PR #35, opened 2017-08-31
Adding mysql-client to Dockerfile so that `docker-compose run web rails db` works

jameshowardwang:


.travis.yml
```diff
@@ -11,7 +11,7 @@ before_install:
   - docker-compose up -d db
   - sleep 10
   - docker-compose run -e RAILS_ENV=test web rake db:create
-script: docker-compose run web rails test
+script: docker-compose run web rails test && echo "quit" | docker-compose run web rails dbconsole
```
yunfangjuan: can you only run one script or you can run multiple scripts? 
also instead of doing "quit", maybe we can do "show tables;"

PR #40, opened 2017-09-01
wiring up airflow webserver connection

yunfangjuan:


config/airflow.yml
```diff
@@ -0,0 +1,37 @@
+default: &default
+  protocol: http
+  host: 127.0.0.1
+  port: 8081
```
ryanking: default is port 8080

config/initializers/airflow.rb
```diff
@@ -0,0 +1,15 @@
+# Module to connect to Airflow webserver
+module Airflow
+  CONFIG = YAML.load_file(Rails.root.join('config/airflow.yml'))[Rails.env]
```
ryanking: this should be `Rails.root.join('config', 'airflow.yml')`

PR #41, opened 2017-09-01
fix bin/setup

ryanking:


ryanking:
@jameshowardwang ah yeah good catch

PR #43, opened 2017-09-01
resque setup

yunfangjuan:


docker-compose.yml
```diff
@@ -14,3 +21,12 @@ services:
       - "3000:3000"
     depends_on:
       - db
```
ryanking: doesn't web need to depend on redis too?

ryanking: oh nevermind I misread it

PR #47, opened 2017-09-07
Improve reports

cdebourcy:


app/controllers/pipeline_outputs_controller.rb
```diff
@@ -29,6 +29,8 @@ def create
 
     respond_to do |format|
       if @pipeline_output.save
+        @pipeline_output.name = ['ID#', @pipeline_output.id, ' (', @pipeline_output.sample.name, ')'].join('')
```
ryanking: why not do this before the first save?

ryanking: note that, as written, if the second save fails we'll not handle the error gracefully

app/views/backgrounds/show.html.erb
```diff
@@ -6,11 +6,18 @@
 </p>
 
 <p>
-  <strong>Used in report(s):</strong>
-  <% @background.reports[0...-1].each do |r| %>
-    <%= link_to r.name, r if r %>, 
+  <% if @background.reports.count == 0 %>
```
ryanking: You can simplify this whole block to:

```
<strong>Used in <%= pluralize(@background.reports.count, report)%>:</strong>
<%= @background.reports.map do |r|
  link_to(r.name, r)
end.join(',')
%>
```

app/views/pipeline_outputs/show.html.erb
```diff
@@ -26,19 +26,29 @@
 </p>
 
 <p>
-  <strong>Report(s):</strong>
-  <% @pipeline_output.reports[0...-1].each do |r| %>
-    <%= link_to r.name, r if r %>, 
+  <% if @pipeline_output.reports.count == 0 %>
```
ryanking: Similiar to above, you can simplify this.

app/views/projects/show.html.erb
```diff
@@ -6,11 +6,18 @@
 </p>
 
 <p>
-  <strong>User(s):</strong>
-  <% @project.users[0...-1].each do |u| %>
-    <%= link_to u.name, u if u%>, 
+  <% if @project.users.count == 0 %>
```
ryanking: simplify

db/migrate/20170906201622_create_backgrounds_pipeline_outputs.rb
```diff
@@ -0,0 +1,8 @@
+class CreateBackgroundsPipelineOutputs < ActiveRecord::Migration[5.1]
+  def change
+    create_table :backgrounds_pipeline_outputs do |t|
+      t.integer :background_id
```
ryanking: these should be bigint

app/controllers/pipeline_outputs_controller.rb
cdebourcy: Can't access pipeline_output.id until it's been saved!

app/views/backgrounds/show.html.erb
cdebourcy: cool!

app/controllers/pipeline_outputs_controller.rb
ryanking: Oh i understand now.

There's no reason to have a `name` column in the db. Just create a method called `#name` in PipelineOutput like so:

```
def name
  return ['ID#', self.id, ' (', self.sample.name, ')'].join('')
end
```

db/migrate/20170906201622_create_backgrounds_pipeline_outputs.rb
cdebourcy: ok

app/views/projects/show.html.erb
cdebourcy: ok

app/views/pipeline_outputs/show.html.erb
cdebourcy: ok

PR #48, opened 2017-09-08
rubocop audit

ryanking:
A pass to configure rubocop in a more reasonable way. Also cleaned up a bunch of low-hanging-fruit errors.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,64 +1,54 @@
 source 'https://rubygems.org'
 
 git_source(:github) do |repo_name|
-  repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
+  repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?('/')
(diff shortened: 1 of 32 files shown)


ryanking:
@cdebourcy i'm not sure we can, bc those are captured in a more general "indentation consistency" rule (http://rubocop.readthedocs.io/en/latest/cops_layout/#layoutindentationconsistency), which we probably don't want to turn off.

I'm pretty sure all of those rules are auto-correctable though, so it shouldn't really have any impact.

cdebourcy:
The reason I thought it might be possible is because of this:
http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/EmptyMethod
It looked like you might be able to choose whether you want to enforce
"compact" or "expanded" style, for example.
Not urgent by any means though!

ryanking:
ah yes, we can follow up with another PR to do that

PR #50, opened 2017-09-08
support for sample uploading

ryanking:
Please help test this by doing this:

1. install the cli by following instructions at https://github.com/chanzuckerberg/idseq-cli
2. start up your idseq-web (`docker-compose up --build -d`)
3. make sure you've migrated the db and run db:seed
3. run something like `idseq -e fake@example.com -t idseq1234 -p 1 -s uganda-`jot -r 1 1 1000000` -u localhost:3000 FILES`, where FILES is whatever files you like


db/migrate/20170908172207_add_status_to_sample.rb
```diff
@@ -0,0 +1,5 @@
+class AddStatusToSample < ActiveRecord::Migration[5.1]
+  def change
+    add_column :samples, :status, :string
```
jameshowardwang: I couldn't figure out from this review -- what is the "status" of a sample used to capture?

bin/upload
```diff
@@ -0,0 +1,55 @@
...
+
+
+data = resp.json()
+
+class tqio(io.BufferedReader):
+    def __init__(self, file_path):
+        super().__init__(open(file_path, "rb"))
+        self.t = tqdm(desc=file_path, unit="bytes", unit_scale=True, total=os.path.getsize(file_path))
+ 
+    def read(self, *args, **kwargs):
```
jameshowardwang: nice! progress bar in a box

jameshowardwang:
hmm, looks like travis is failing due to some dependency changes

ryanking:
@cdebourcy are you sure that you have the right branch running? I ask because that's an error I specifically addressed.

app/controllers/samples_controller.rb
```diff
@@ -17,37 +18,6 @@ def new
     @sample = Sample.new
   end
 
-  # GET /samples/insert
-  def insert
-    # TODO: merge create and insert somehow
-    params.require([:sample_name, :project_name, :s3_input_path])
-    @project = Project.find_by(name: params[:project_name]) ||
```
yunfangjuan: how does def create figure out which project this sample belongs to?

app/models/sample.rb
```diff
@@ -2,4 +2,7 @@ class Sample < ApplicationRecord
   belongs_to :project
   has_many :pipeline_outputs, dependent: :destroy
   has_and_belongs_to_many :backgrounds
+  has_many :input_files, dependent: :destroy
+
```
yunfangjuan: 1. would be nice to have a helper function to figure out the s3 path of the input file.  like s3_path 
2. all our input data are paired fastq.gz, what do you envision the uploading process?  two input_files entries attached to a sample? or one entry corresponding to a s3 directory?

bin/upload
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python
+
+import argparse, io, json, os, requests, subprocess
+from tqdm import tqdm
+
+parser = argparse.ArgumentParser(description='Upload to idseq.')
+
+parser.add_argument('-t', '--token', metavar='token', type=str, required=True)
+parser.add_argument('-p', '--project-id', metavar='id', type=int, required=True)
```
yunfangjuan: it's probably more user friendly to have project-name instead of id here

db/migrate/20170908172207_add_status_to_sample.rb
yunfangjuan: lgtm 
a few comments and questions.

cdebourcy:
Works now! 
It seems to recreate my local folder structure in the S3 bucket though. Can we make it so it takes only the file basename?

app/controllers/samples_controller.rb
ryanking: the payload includes a project_id field

app/models/sample.rb
ryanking: 1. for sure, we can add that

2. I decided not to validate the uploads for now, bc I didn't know if we'd have more diverse input times (no paired reads?). I think we should hold off for now until we can get a better sense for the diversity of inputs.

ryanking: 2.cont: that said, if our pipeline can only handle paired-end reads then we should probably reject anything that doesn't look like paired end read fastq files

ryanking:
What command did you run?

cdebourcy:
idseq -e fake@example.com -t idseq1234 -p 1 -s uganda-`jot -r 1 1 1000000` -u localhost:3000 /Users/charlesdebourcy/Downloads/UGANDA_S30_NP_R1_001.fastq.gz

Resulting in the file being put at s3://czbiohub-idseq-samples-development/samples/Awesome\ Project/uganda-675091/Users/charlesdebourcy/Downloads/UGANDA_S30_NP_R1_001.fastq.gz

app/models/sample.rb
yunfangjuan: we only accept paired reads for now

ryanking:
@cdebourcy ok i can make the change to only use basename over in idseq-cli

PR #52, opened 2017-09-11
Add taxon count_type (and fix user signup)

cdebourcy:


app/views/pipeline_outputs/show.html.erb
```diff
@@ -50,6 +52,14 @@
           <tr>
             <td><%= element.tax_id %></td>
             <td><%= element.name %></td>
+            <td><% case element.tax_level
```
ryanking: This might be a bit cleaner like:

```
<td><%= case element.tax_level
                 when 1; 'species'
                 when 2; 'genus'
                 end %>
</td>
```

or something like that
and similar below

app/views/pipeline_outputs/show.html.erb
```diff
@@ -50,6 +52,16 @@
           <tr>
             <td><%= element.tax_id %></td>
             <td><%= element.name %></td>
+            <td><%= case element.tax_level
+                   when 1; 'species'
```
yunfangjuan: I will probably define
TAX_LEVEL_SPECIES  = 1
TAX_LEVEL_SPECIES = 2 
in the taxid_count model.

app/views/pipeline_outputs/show.html.erb
cdebourcy: cool

app/views/pipeline_outputs/show.html.erb
cdebourcy: cool

PR #54, opened 2017-09-12
Add NT/NR and species/genus distinction

cdebourcy:


app/views/pipeline_outputs/show.html.erb
```diff
@@ -40,27 +40,20 @@
...
+          <th>Species</th>
+          <th>NT Species Count</th>
+          <th>NR Species Count</th>
         </tr>
       </thead>
       <tbody>
-        <% @pipeline_output.taxon_counts.sort_by(&:count).reverse.each do |element| %>
+        <% counts_display = @pipeline_output.taxon_counts %>
+        <% nt_species_counts_display = counts_display.where("count_type = 'NT' and tax_level = #{TaxonCount::TAX_LEVEL_SPECIES}").order(count: :desc) %>
+        <% nt_species_counts_display.sort_by(&:count).reverse.each do |nt_species_element| %>
```
ryanking: you shouldn't have to sort this again, right?

app/views/pipeline_outputs/show.html.erb
```diff
@@ -40,27 +40,20 @@
...
-          <th>Count Type</th>
-          <th>Count</th>
+          <th>Species</th>
+          <th>NT Species Count</th>
+          <th>NR Species Count</th>
         </tr>
       </thead>
       <tbody>
-        <% @pipeline_output.taxon_counts.sort_by(&:count).reverse.each do |element| %>
+        <% counts_display = @pipeline_output.taxon_counts %>
```
ryanking: you might be able to clean this up by implementing it as a ActiveRecord scope: http://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html

app/views/pipeline_outputs/show.html.erb
ryanking: (since you did `.order(count: :desc)` above

app/views/pipeline_outputs/show.html.erb
```diff
@@ -40,27 +40,20 @@
...
-            </td>
-            <td><% case element.count_type
-                   when TaxonCount::COUNT_TYPE_NT; 'nucleotide hit'
-                   when TaxonCount::COUNT_TYPE_NR; 'protein hit'
-                   end %>
-            </td>
-            <td><%= element.count %></td>
+            <td><%= link_to (nt_species_element.name || nt_species_element.tax_id), "https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=#{nt_species_element.tax_id}" %></td>
+           <td><%= nt_species_element.count %></td>
+            <% nr_species_element = counts_display.where("count_type = 'NR' and tax_level = #{TaxonCount::TAX_LEVEL_SPECIES}").find_by(tax_id: nt_species_element.tax_id) %>
```
ryanking: Yeah you could refactor this and the above bit into a scope that takes a param or two. See here [https://stackoverflow.com/questions/16588077/scopes-with-lambda-and-arguments-in-rails-4-style] for some examples of how to do scopes with params.

I can help tomorrow if this doesn't make sense.

cdebourcy:
does refactored version using ActiveRecord scopes look good now?

app/views/pipeline_outputs/show.html.erb
```diff
@@ -38,35 +38,30 @@
...
-    </table>
+  <thead>
+    <tr>
+      <th>Species</th>
+      <th>NT Species Count</th>
+      <th>NR Species Count</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% counts_display = @pipeline_output.taxon_counts %>
```
ryanking: Sorry I didn't say this before, but it didn't occur to me until after posting the previous review–

This logic that deals with finding the right data should really be in either the model or controller (or a combo of both).

Templates should be as free of logic as possible, except for logic related to display/rendering/etc. 

Does that make sense? It might just mean cut/pasting this into the controller.

app/views/reports/show.html.erb
```diff
@@ -35,39 +35,57 @@
...
+      <th>Species</th>
+      <th>NT Species Z</th>
+      <th>NT Species rM</th>
+      <th>NR Species Z</th>
+      <th>NR Species rM</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% zscores = @report.taxon_zscores %>
+    <% if @nt_species_zscore_threshold %>
```
ryanking: same as above for all of these

cdebourcy:
Ryan, I've addressed the requested changes and am ready to merge

PR #55, opened 2017-09-13
Add combined json output

cdebourcy:


aws_batch/pipeline.py
```diff
@@ -301,6 +302,29 @@ def generate_json_from_taxid_counts(sample, rawReadsInputPath, taxidCountsInputP
...
         json.dump(output_dict, outf)
 
+def combine_json(inputPath1, inputPath2, outputPath):
+    with open(inputPath1) as inf1:
+        input1 = json.load(inf1).get("pipeline_output")
+    with open(inputPath2) as inf2:
+        input2 = json.load(inf2).get("pipeline_output")
+    total_reads = max(input1.get("total_reads"),
+                      input2.get("total_reads"))
+    remaining_reads = max(input1.get("remaining_reads"),
```
yunfangjuan: Is this the right accounting I wonder?

aws_batch/pipeline.py
```diff
@@ -83,6 +83,7 @@
 NR_TAXID_COUNTS_TO_JSON_OUT = 'counts.filter.deuterostomes.taxids.rapsearch2.filter.deuterostomes.taxids.gsnapl.unmapped.bowtie2.lzw.cdhitdup.priceseqfilter.unmapped.star.json'
 NR_TAXID_COUNTS_TO_SPECIES_RPM_OUT = 'species.rpm.filter.deuterostomes.taxids.rapsearch2.filter.deuterostomes.taxids.gsnapl.unmapped.bowtie2.lzw.cdhitdup.priceseqfilter.unmapped.star.csv'
 NR_TAXID_COUNTS_TO_GENUS_RPM_OUT = 'genus.rpm.filter.deuterostomes.taxids.rapsearch2.filter.deuterostomes.taxids.gsnapl.unmapped.bowtie2.lzw.cdhitdup.priceseqfilter.unmapped.star.csv'
+COMBINED_JSON_OUT = 'output.json'
```
yunfangjuan: maybe something slightly more informative like idseq_web_sample.json?

aws_batch/pipeline.py
cdebourcy: Total_reads is the same for both files, remaining_reads can differ. 
Really, remaining_reads kind of becomes meaningless: there should be remaining_reads_nt and remaining_reads_nr.
We're planning to change the statistics on the pipeline run and add more of them, so I think I'll keep this preliminary solution until we refactor the whole thing.

aws_batch/pipeline.py
cdebourcy: sounds good

PR #56, opened 2017-09-13
kick off pipeline from sample

yunfangjuan:
Initial commit to integrate idseq-web and pipeline


config/initializers/pipeline.rb
```diff
@@ -0,0 +1,19 @@
+module IdSeqPipeline
+  AWS_ACCESS_KEY_ID = ENV['AWS_ACCESS_KEY_ID']
```
ryanking: We are going to be using instance profiles to configure aws in dev/stage/prod. Locally we should use aws credentials profiles, rather than environment variable.s

config/initializers/pipeline.rb
```diff
@@ -0,0 +1,19 @@
...
+  AWS_SECRET_ACCESS_KEY = ENV['AWS_SECRET_ACCESS_KEY']
+  AWS_DEFAULT_REGION = ENV['AWS_DEFAULT_REGION']
+  S3_SCRIPT_LOC= 's3://czbiohub-infectious-disease/bin/pipeline.py'
```
ryanking: We should put this in this repo, so that we can version them together.

config/initializers/pipeline.rb
```diff
@@ -0,0 +1,19 @@
...
+  BASE_COMMAND = "export AWS_ACCESS_KEY_ID=#{AWS_ACCESS_KEY_ID}" +
+                 " AWS_SECRET_ACCESS_KEY=#{AWS_SECRET_ACCESS_KEY}" +
+                 " AWS_DEFAULT_REGION=#{AWS_DEFAULT_REGION}; "
+  def self.pipeline_command(input_s3_dir, output_s3_dir, db_sample_id)
```
ryanking: this code should not be in initializers

config/initializers/pipeline.rb
yunfangjuan: It's in the repo. I was thinking about aws s3 cp it over during initialization

config/initializers/pipeline.rb
yunfangjuan: where should it go?

config/initializers/pipeline.rb
ryanking: we can worry about this later tho

yunfangjuan: how do we do that inside a docker container?

config/initializers/pipeline.rb
ryanking: Why cp it over at all? Like you're copying it to s3 just to copy it back.

config/initializers/pipeline.rb
ryanking: probably in the job

config/initializers/pipeline.rb
yunfangjuan: because. the pipeline.py is longer than 8192 characters. I couldn't submit the code directly through aegea. I need to put it on s3 and on the batch instance, copy the code to local to run.

ryanking: Oh i get it, nevermind I guess. This feels janky but we can fix it later.

config/initializers/pipeline.rb
ryanking: this is fine for now

config/initializers/pipeline.rb
ryanking: you can forget about the other notes, this is the only part I think we should change now (vs later PRs)

.gitignore
```diff
@@ -6,6 +6,7 @@
 
 # Ignore bundler config.
 /.bundle
+/config/initializers/app_environment_variables.rb
```
ryanking: why is this ignored?

yunfangjuan:
It has the AWS credentials in it ;-)

PR #61, opened 2017-09-18
render nested attributes for pipeline_output/report json 

cdebourcy:
to facilitate integration with React


app/controllers/pipeline_outputs_controller.rb
```diff
@@ -13,6 +13,10 @@ def index
   # GET /pipeline_outputs/1.json
   def show
     @view_level = params[:view_level]
+    respond_to do |format|
+      format.html
+      format.json { render json: @pipeline_output.to_json(include: :taxon_counts) }
```
ryanking: It would be preferable to update the json template (the .json.jbuilder file) rather than do this, but we can merge this for now.

PR #65, opened 2017-09-19
trigger pipeline directly

yunfangjuan:
from status changes.

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -1,21 +1,29 @@
 require 'open3'
 require 'json'
 class Sample < ApplicationRecord
+  STATUS_CREATED  = 'created'.freeze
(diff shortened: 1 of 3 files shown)


app/models/sample.rb
```diff
@@ -30,7 +38,14 @@ def pipeline_command
     command
   end
 
-  def kickoff_pipeline(dry_run = true) # should be triggered when the upload is complete
+  def check_status
+    if [STATUS_UPLOADED, STATUS_RERUN].include?(status)
+      status = STATUS_CHECKED
```
ryanking: rubocop is warning about this, because you might need a `self.` here

yunfangjuan: thanks. i will fix this. sometime rubcop wants me to do self. sometimes doesn't

PR #67, opened 2017-09-19
Add metadata

cdebourcy:


app/models/sample.rb
```diff
@@ -23,7 +23,16 @@ def pipeline_command
     batch_command = "aws s3 cp #{IdSeqPipeline::S3_SCRIPT_LOC} .; chmod 755 #{script_name}; " \
                     "INPUT_BUCKET=#{sample_input_s3_path} " \
                     "OUTPUT_BUCKET=#{sample_output_s3_path} " \
-                    "DB_SAMPLE_ID=#{id} ./#{script_name}"
+                    "DB_SAMPLE_ID=#{id} " \
```
ryanking: why do these need to go to the pipeline?

cdebourcy: In case we want to use them later (e.g. if we want to use SAMPLE_HOST to inform filtering reference)

ryanking: makes sense.

once you resolve the conflicts with master we can approve this

PR #71, opened 2017-09-20
Add deuterostome filtering

cdebourcy:


aws_batch/pipeline.py
```diff
@@ -357,6 +357,40 @@ def generate_taxid_annotated_m8(input_m8, output_m8, accession2taxid_db):
...
     outf.close()
 
+def filter_deuterostomes_from_m8(input_m8, output_m8, deuterostome_file):
+    with open(deuterostome_file) as f:
+        content = f.readlines()
+    taxids_toremove = [x.strip("\n") for x in content]
+    output_f = open(output_m8, 'wb')
+    with open(input_m8, "rb") as input_f:
+        for line in input_f:
+            taxid = (line.split("taxid"))[1].split(":")[0]
```
yunfangjuan: I would add a comment with an example of an example line for m8.

aws_batch/pipeline.py
```diff
@@ -357,6 +357,40 @@ def generate_taxid_annotated_m8(input_m8, output_m8, accession2taxid_db):
...
+            if not taxid in taxids_toremove:
+                output_f.write(line)
+    output_f.close()
+
+def filter_taxids_from_fasta(input_fa, output_fa, annotation_prefix, accession2taxid_path, deuterostome_file):
+    with open(deuterostome_file) as f:
```
yunfangjuan: i would probably extract this out to a function since you repeat it twice,

aws_batch/pipeline.py
```diff
@@ -357,6 +357,40 @@ def generate_taxid_annotated_m8(input_m8, output_m8, accession2taxid_db):
...
+        content = f.readlines()
+    taxids_toremove = [x.strip("\n") for x in content]
+    accession2taxid_dict = shelve.open(accession2taxid_path)
+    input_f = open(input_fa, 'rb')
+    output_f = open(output_fa, 'wb')
+    sequence_name = input_f.readline()
+    sequence_data = input_f.readline()
+    while len(sequence_name) > 0 and len(sequence_data) > 0:
+        read_id = sequence_name.rstrip().lstrip('>')
+        accession_id = (read_id.split(annotation_prefix+":"))[1].split(":")[0] if read_id.startswith(annotation_prefix) else (read_id.split(":"+annotation_prefix+":"))[1].split(":")[0]
```
yunfangjuan: line too long. also i think more explanation will be desirable with an example annotated read_id

aws_batch/pipeline.py
```diff
@@ -457,88 +492,106 @@ def run_sample(sample_s3_input_path, sample_s3_output_path,
...
     # run_generate_taxid_annotated_fasta_from_m8
     run_generate_taxid_annotated_fasta_from_m8(sample_name,
-        result_dir + '/' + GSNAPL_OUT,
-        result_dir + '/' + EXTRACT_UNMAPPED_FROM_SAM_OUT3,
-        result_dir + '/' + FILTER_DEUTEROSTOME_FROM_TAXID_ANNOTATED_FASTA_OUT,
-        'NT',
-        result_dir, sample_s3_output_path, lazy_run=False)
- 
+                    result_dir + '/' + GSNAPL_OUT,
+                    result_dir + '/' + EXTRACT_UNMAPPED_FROM_SAM_OUT3,
```
yunfangjuan: indentation looks a bit off here.

aws_batch/pipeline.py
```diff
@@ -457,88 +492,106 @@ def run_sample(sample_s3_input_path, sample_s3_output_path,
...
-        result_dir + '/' + NT_TAXID_COUNTS_TO_JSON_OUT,
-        result_dir + '/' + NT_TAXID_COUNTS_TO_SPECIES_RPM_OUT,
-        result_dir + '/' + NT_TAXID_COUNTS_TO_GENUS_RPM_OUT,
-        taxid2info_s3_path, 'NT', db_sample_id,
-        sample_host, sample_location, sample_date, sample_tissue,
-        sample_template, sample_library, sample_sequencer, sample_notes,
-        result_dir, sample_s3_output_path, lazy_run=False)
- 
+                    result_dir + '/' + FILTER_DEUTEROSTOMES_FROM_NT_M8_OUT,
+                    fastq_file_1,
```
yunfangjuan: indentation doesn't look right.

aws_batch/pipeline.py
```diff
@@ -457,88 +492,106 @@ def run_sample(sample_s3_input_path, sample_s3_output_path,
...
+                    result_dir + '/' + NR_TAXID_COUNTS_TO_JSON_OUT,
+                    result_dir + '/' + NR_TAXID_COUNTS_TO_SPECIES_RPM_OUT,
+                    result_dir + '/' + NR_TAXID_COUNTS_TO_GENUS_RPM_OUT,
+                    taxid2info_s3_path, 'NR', db_sample_id,
+                    sample_host, sample_location, sample_date, sample_tissue,
+                    sample_template, sample_library, sample_sequencer, sample_notes,
+                    result_dir, sample_s3_output_path, lazy_run=False)
 
     run_combine_json_outputs(sample_name,
-        result_dir + '/' + NT_TAXID_COUNTS_TO_JSON_OUT,
```
yunfangjuan: indentation: either do 4 spaces or align the parameters.

aws_batch/pipeline.py
```diff
@@ -457,88 +492,106 @@ def run_sample(sample_s3_input_path, sample_s3_output_path,
...
-        result_dir + '/' + COMBINED_JSON_OUT,
-        result_dir, sample_s3_output_path, lazy_run=False)
+                    result_dir + '/' + NT_TAXID_COUNTS_TO_JSON_OUT,
+                    result_dir + '/' + NR_TAXID_COUNTS_TO_JSON_OUT,
+                    result_dir + '/' + COMBINED_JSON_OUT,
+                    result_dir, sample_s3_output_path, lazy_run=False)
 
 def run_star(sample_name, fastq_file_1, fastq_file_2, star_genome_s3_path,
              result_dir, scratch_dir, sample_s3_output_path, lazy_run):
     logger = logging.getLogger()
```
yunfangjuan: why did you strip out all the early return for the lazy run? It makes the code a lot harder to read. and it's the coding convention people follow a lot.  Also at some point we need to follow the 80-char length. This would make things really hard.

aws_batch/pipeline.py
```diff
@@ -90,6 +90,18 @@
 LOGS_OUT = 'log.txt'
 
 ### convenience functions
+def count_reads(file_name, file_type):
+    if file_type == "fastq_gz_paired":
+        return 2*sum(1 for line in gzip.open(file_name))
```
yunfangjuan: this is a very inefficient way of counting lines. basically you are creating a temporary array and sum them together. 
If the fastq file has millions of lines, you are using millions of bytes of memory just for this. 
I think you can just start with count=0 and count+=1 

Actually I will redo the function this way. 
count =0 
if file_name[-3:] == '.gz': 
  infile = gz.open(file_name)
else:
  infile = open(filename)

for line in inline:
  if file_type == 'fastq':
    count += 2./4
 elif file_type == 'fasta':
    count += 1
  elif file_type == 'fasta_paired':
    count += 2 if line.startswith(('>')
  else:
     count += 1 
infile.close()
 return count

aws_batch/pipeline.py
```diff
@@ -214,7 +226,7 @@ def generate_tax_counts_from_m8(m8_file, output_file):
     taxid_count_map = {}
     with open(m8_file, 'rb') as m8f:
         for line in m8f:
-            taxid = (line.split("taxid"))[1].split(":")[0]
+            taxid = (line.split("taxid"))[1].split(":")[0] # example line: "taxid9606:NB501961:14:HM7TLBGX2:1:12104:15431:7497/1  KY504764.1  100.0  150  0 [...]"
```
yunfangjuan: move example to a new line. 
keep it short like 
#"taxid9606:NB501961:14:HM7TLBGX2:1:12104:15431: ....

aws_batch/pipeline.py
```diff
@@ -357,6 +359,44 @@ def generate_taxid_annotated_m8(input_m8, output_m8, accession2taxid_db):
             outf.write(new_line)
     outf.close()
 
+def read_columnfile_into_list(file_name):
+    with open(file_name) as f:
+        content = f.readlines()
+    return [x.strip("\n") for x in content]
```
yunfangjuan: rstrip or strip? 
a simpler way would be 
return [x for x in f] 
directly. 

Can you also explain what's read_columnfile_into_list? it's not immediately obvious

aws_batch/pipeline.py
```diff
@@ -357,6 +359,44 @@ def generate_taxid_annotated_m8(input_m8, output_m8, accession2taxid_db):
...
+
+def filter_deuterostomes_from_m8(input_m8, output_m8, deuterostome_file):
+    taxids_toremove = read_columnfile_into_list(deuterostome_file)
```
yunfangjuan: i will just call the function name read_file_into_list?

aws_batch/pipeline.py
```diff
@@ -357,6 +359,44 @@ def generate_taxid_annotated_m8(input_m8, output_m8, accession2taxid_db):
...
+    output_f = open(output_m8, 'wb')
+    with open(input_m8, "rb") as input_f:
+        for line in input_f:
+            taxid = (line.split("taxid"))[1].split(":")[0] # example line: "taxid9606:NB501961:14:HM7TLBGX2:1:12104:15431:7497/1  KY504764.1  100.0  150  0 [...]"
```
yunfangjuan: example line change.

aws_batch/pipeline.py
```diff
@@ -387,27 +427,49 @@ def filter(self, record):
...
+        records_after = count_reads(logparams["after_file_name"], logparams["after_file_type"])
+        percent_removed = (100.0 * (records_before - records_after)) / records_before
+        if "paired" in logparams["before_file_type"] and "paired" in logparams["after_file_type"]:
+            record_description = "read-pairs"
+        else:
+            record_description = "reads"
+        logger.info("%s %% of %s dropped out, %s %s remaining" % (str(percent_removed), record_description,
+                                                                  str(records_after), record_description))
+    # copy log file
+    execute_command("aws s3 cp %s/%s %s/;" % (logparams["result_dir"], LOGS_OUT, logparams["sample_s3_output_path"]))
```
yunfangjuan: Isn't this a lot cleaner? 
Also I would suggest create a different log.txt for each batch job

so it can be like log.txt.{job_id} 

you can access batch job id with 
os.environ['AWS_BATCH_JOB_ID']

aws_batch/pipeline.py
```diff
@@ -387,27 +427,49 @@ def filter(self, record):
...
     sample_dir = DEST_DIR + '/' + sample_name
     fastq_dir = sample_dir + '/fastqs'
     result_dir = sample_dir + '/results'
     scratch_dir = sample_dir + '/scratch'
     execute_command("mkdir -p %s %s %s %s" % (sample_dir, fastq_dir, result_dir, scratch_dir))
     execute_command("mkdir -p %s " % REF_DIR);
-    
+
     # configure logger
+    log_file = result_dir+'/'+LOGS_OUT
```
yunfangjuan: add the batch job id as described.

aws_batch/pipeline.py
```diff
@@ -435,463 +497,530 @@ def run_sample(sample_s3_input_path, sample_s3_output_path,
...
-    logger.info("generated taxon counts from m8")
+    logging.getLogger().info("generated taxon counts from m8")
     generate_json_from_taxid_counts(sample_name, fastq_file_1, taxon_counts_csv_file,
                                     taxoninfo_path, taxon_counts_json_file,
-                                    count_type, db_sample_id, sample_host, sample_location,
-                                    sample_date, sample_tissue, sample_template, sample_library,
-                                    sample_sequencer, sample_notes)
-    logger.info("generated JSON file from taxon counts")
+                                    count_type, db_sample_id)
+    logging.getLogger().info("generated JSON file from taxon counts")
```
yunfangjuan: is logging a global variable? if so, you need to uppercase it.

aws_batch/pipeline.py
yunfangjuan: instead of hardcode the log filename
do 

logger.handlers[0].baseFilename  instead.

aws_batch/pipeline.py
cdebourcy: Okay, just make sure to use file_type "fastq_paired" because the run_and_log function checks if "paired" occurs in file_type to determine messages

aws_batch/pipeline.py
cdebourcy: I thought that was the part you wanted me to extract into a function in one of your previous comments. But since it can be changed into a single line, I think I'll remove this function

aws_batch/pipeline.py
cdebourcy: makes sense

aws_batch/pipeline.py
cdebourcy: no, it's a module

aws_batch/pipeline.py
cdebourcy: ok

aws_batch/pipeline.py
cdebourcy: ok

aws_batch/pipeline.py
cdebourcy: ok

aws_batch/pipeline.py
cdebourcy: ok

aws_batch/pipeline.py
cdebourcy: improved function and left it in as read_file_into_list

aws_batch/pipeline.py
```diff
@@ -87,9 +87,30 @@
...
+        elif file_type == "fastq_paired":
+            count += 2./4 
+        elif file_type == "fasta_paired":
+            if line.startswith('>'):
+                count += 2
+        elif file_type == "fasta":
+            if line.startswith('>'):
+                count += 1
+        else:
+            count += 1
```
yunfangjuan: add f.close()

PR #72, opened 2017-09-20
Feature/react rails setup

sadelokiki:


yunfangjuan:
Do you need to check in all the aws_cli files?

PR #79, opened 2017-09-21
add links to log file and fasta file

cdebourcy:


app/models/sample.rb
```diff
@@ -59,6 +59,16 @@ def sample_output_s3_path
     "s3://#{SAMPLES_BUCKET_NAME}/#{sample_path}/results"
   end
 
+  def sample_annotated_fasta_url
+    file_base_name = "taxids.rapsearch2.filter.deuterostomes.taxids.gsnapl.unmapped.bowtie2.lzw.cdhitdup.priceseqfilter.unmapped.star.fasta"
```
yunfangjuan: Move this to the CONSTANTS section on the top and upper case it.

app/models/sample.rb
```diff
@@ -59,6 +59,16 @@ def sample_output_s3_path
...
+    "https://s3.console.aws.amazon.com/s3/object/#{SAMPLES_BUCKET_NAME}/#{sample_path}/results/#{file_base_name}"
+  end
+
+  def sample_log_url
+    file_base_name = "log.txt"
```
yunfangjuan: same as above.

PR #87, opened 2017-09-26
Default background

cdebourcy:


app/controllers/reports_controller.rb
```diff
@@ -35,6 +35,7 @@ def edit
   # POST /reports.json
   def create
     @report = Report.new(report_params)
+    @report.background = Background.find(1) unless @report.background
```
yunfangjuan: This is a bit dangerous. 
You can implement a function called
Background.default
and where you specify the name for it and find_by. 
harcoding find(1) is not good.

app/controllers/reports_controller.rb
```diff
@@ -43,23 +44,9 @@ def create
...
-        else
-          normalized_bg_count = 0
-        end
-        sum += normalized_bg_count
-        sum_sq += normalized_bg_count**2
-        n += 1
-      end
-      mean = sum.to_f / n
-      stdev = Math.sqrt((sum_sq.to_f - sum**2 / n) / (n - 1))
+      summary = @report.background.summarize.detect {|s| s[:tax_id] == taxon_count.tax_id and s[:count_type] == taxon_count.count_type}
```
yunfangjuan: it's probably a lot faster to join the backgrounds with the taxon_counts table directly.

app/models/background.rb
```diff
@@ -7,4 +7,37 @@ class Background < ApplicationRecord
   def validate_size
     errors.add(:base, "Need to select at least 2 pipeline runs.") if pipeline_outputs.size < 2
   end
+
+  def summarize
```
yunfangjuan: again. i think it's better to do this through database calculation. This would be slow for a direct query.

app/controllers/reports_controller.rb
```diff
@@ -35,38 +35,16 @@ def edit
...
-                   end
-      @report.taxon_zscores.new(a)
+    unless Background.find_by(name: "default") || @report.background
+      raise "No background specified and background 'default' does not exist"
     end
+    @report.background = Background.find_by(name: "default") unless @report.background
+    total_reads = @report.pipeline_output.total_reads
+    summary = @report.background.summarize
+    data = @report.pipeline_output.taxon_counts.map { |h| h.attributes.merge(norm_count: h["count"] / total_reads.to_f) }
+    data_and_background = (data + summary).group_by { |h| [h["tax_id"], h["tax_level"], h["name"], h["count_type"]] }.map { |_k, v| v.reduce(:merge) }
```
yunfangjuan: wouldn't you get a bunch of records without norm_count if there's no count for the pipeline output?  In that case, I believe you don't need to generate the z_score for those.

app/models/background.rb
boris-dimitrov: how does one run a SQL query from this magical place?

cdebourcy: check out for example the latest version of app/models/background.rb!

PR #97, opened 2017-09-28
fix rounding error

cdebourcy:
Report generation was failing due to a rounding error leading to sqrt of a negative number


app/controllers/reports_controller.rb
```diff
@@ -106,6 +106,8 @@ def compute_rpm(count, total_reads)
   def compute_zscore(count, mean, stdev)
     if count && mean
       (count - mean) / stdev
+    elsif count
+      Float::MAX
```
yunfangjuan: I think we should have a discussion with Joe on this when there's no observation on background.

cdebourcy: agree

PR #98, opened 2017-09-28
Add index choice

cdebourcy:


app/models/sample.rb
```diff
@@ -75,6 +75,7 @@ def pipeline_command
     script_name = File.basename(IdSeqPipeline::S3_SCRIPT_LOC)
     batch_command = "aws s3 cp #{IdSeqPipeline::S3_SCRIPT_LOC} .; chmod 755 #{script_name}; " \
       "INPUT_BUCKET=#{sample_input_s3_path} OUTPUT_BUCKET=#{sample_output_s3_path} " \
+      "STAR_GENOME=#{s3_star_index_path} BOWTIE2_GENOME=#{s3_bowtie2_index_path} " \
```
yunfangjuan: this can be a problem if user doesn't specify it. I think you will only add this in when the index paths are set

db/migrate/20170928215837_add_star_index_path_to_sample.rb
```diff
@@ -0,0 +1,5 @@
+class AddStarIndexPathToSample < ActiveRecord::Migration[5.1]
+  def change
+    add_column :samples, :s3_star_index_path, :text
```
yunfangjuan: seems overkill to generate two migrations for this. You can just put both of these lines to one migration.

app/models/sample.rb
cdebourcy: Isn't that problem dealt with in aws_batch/pipeline.py? There's a default in the pipeline. So far we've never set some of the other ENV variables (e.g. SAMPLE_SEQUENCER) and it always worked, so probably won't cause failure?

PR #99, opened 2017-09-28
Put the user creation/update behind login

yunfangjuan:


app/models/user.rb
```diff
@@ -1,8 +1,8 @@
 class User < ApplicationRecord
   acts_as_token_authenticatable
   # Include default devise modules. Others available are:
-  # :confirmable, :lockable, :timeoutable and :omniauthable
-  devise :database_authenticatable, :registerable,
-         :recoverable, :rememberable, :trackable, :validatable
+  # :confirmable, :lockable, :timeoutable and :omniauthable, :registerable
```
boris-dimitrov: Can't find info on the interwebs what "registerable" means.

config/initializers/devise.rb
```diff
@@ -245,7 +245,7 @@
   # config.navigational_formats = ['*/*', :html]
 
   # The default HTTP method used to sign out a resource. Default is :delete.
-  config.sign_out_via = :delete
+  config.sign_out_via = :get
```
boris-dimitrov: Is this where bad requests get redirected somehow?

config/routes.rb
```diff
@@ -5,9 +5,7 @@
   resources :reports
   resources :pipeline_outputs
   devise_for :users
-  resources :samples do
-    get :insert, on: :collection
-  end
```
boris-dimitrov: and similarly i don't know what this does

app/models/user.rb
ryanking: registerable -> https://github.com/plataformatec/devise/blob/master/lib/devise/models/registerable.rb

config/initializers/devise.rb
ryanking: i don't think this is a good idea. GET for anything that mutates seems bad.

yunfangjuan: i guess so. what's the best way to sign out through a link directly on chrome?

config/routes.rb
yunfangjuan: insert no longer exists

app/models/user.rb
yunfangjuan: i disable that so people don't just go to the sign up page and signup for an account.

boris-dimitrov: How do you create user accounts?

In particular, if I check out this branch and pull the dev db, will I be locked out?

boris-dimitrov: I am guessing the way this will work is that we manually insert user entries in the SQL DB and then the user follows the "i forgot my password" link to reset it?

yunfangjuan: we will manually enter one through "rails c" and i will enable inviteable so you can invite people if necessary.

config/initializers/devise.rb
yunfangjuan: ok never mind. there's always simple rest client. I will change it back.

PR #100, opened 2017-09-28
support alpha environment

ryanking:


bin/deploy
```diff
@@ -70,7 +69,7 @@ def run_task(client, cluster, service, task_revision, container, command)
   puts "  logs -> #{log_url(REGION, LOG_GROUP, TASK_NAME, CONTAINER_NAME, task_id)}"
 
   loop do
-    resp2 = client.describe_tasks(cluster: CLUSTER, tasks: [migrate_task_arn])
+    resp2 = client.describe_tasks(cluster: cluster, tasks: [migrate_task_arn])
```
yunfangjuan: where is cluster defined?

ryanking: it is a param to this method

PR #102, opened 2017-09-28
Fix report more

cdebourcy:


app/controllers/reports_controller.rb
```diff
@@ -103,11 +103,11 @@ def compute_rpm(count, total_reads)
     end
   end
 
-  def compute_zscore(count, mean, stdev)
-    if count && mean
-      (count - mean) / stdev
-    elsif count
-      Float::MAX
+  def compute_zscore(rpm, mean, stdev)
+    if rpm && mean
```
yunfangjuan: isn't checking  stdev more intuitive than mean? since you are checking for divide by zero

app/controllers/reports_controller.rb
```diff
@@ -39,9 +39,9 @@ def create
     raise "No background specified and background #{Background::DEFAULT_BACKGROUND_MODEL_NAME} does not exist" unless @report.background
     total_reads = @report.pipeline_output.total_reads
     summary = @report.background.summarize
-    data = @report.pipeline_output.taxon_counts.map { |h| h.attributes.merge(norm_count: h["count"] / total_reads.to_f) }
```
yunfangjuan: I think all the logic here probably should be moved to Report model. But you can do that in the next refactoring.

app/controllers/reports_controller.rb
cdebourcy: agree

app/controllers/reports_controller.rb
cdebourcy: agree

PR #105, opened 2017-09-29
Feature/newclientdevelop

sadelokiki:
Add Pagination to Pipelineoutputs page


app/controllers/home_controller.rb
```diff
@@ -1,11 +1,7 @@
 class HomeController < ApplicationController
   def home
-    @users = User.all
-    @projects = Project.all
-    @samples = Sample.all
     @output_data = {}
-
-    @pipeline_outputs = PipelineOutput.order("created_at DESC")
+    @pipeline_outputs = PipelineOutput.order("created_at DESC").paginate(page: params[:page], per_page: 6)
```
yunfangjuan: maybe we can not hardcode the 6 but put it a constant somewhere? SAMPLES_PER_PAGE = 20? 
Also I think the right logic should be getting the most recent pipeline output for each sample, so you don't display more than one pipeline output per sample. 
Otherwise it looks good.

sadelokiki: oh okay!

PR #106, opened 2017-09-29
add health check gem

ryanking:
Currently the load balancer uses '/' to check health. This appears to be a pretty slow and expensive page to build.

This gem gives us a simple endpoint specifically for checking the health of the running server.


ryanking:
@boris-dimitrov no need, the defaults are fine

PR #108, opened 2017-09-29
add and integrate airbrake

ryanking:


db/schema.rb
```diff
@@ -70,7 +70,7 @@
   end
 
   create_table "projects", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
-    t.string "name", collation: "utf8_general_ci"
+    t.string "name", collation: "latin1_swedish_ci"
```
yunfangjuan: is this utf8 -> latin1 intended?

ryanking: oh that's weird. will revert

PR #110, opened 2017-09-29
add stats json

cdebourcy:


aws_batch/pipeline.py
```diff
@@ -455,15 +459,21 @@ def run_and_log(logparams, func_name, *args):
     if logparams["count_reads"]:
         percent_removed = (100.0 * (records_before - records_after)) / records_before
         logger.info("%s %% of reads dropped out, %s reads remaining" % (str(percent_removed), str(records_after)))
+        STATS[func_name.__name__] = {'reads_before': records_before, 'reads_after': records_after}
```
yunfangjuan: Make STATS an array as discussed and move func_name into the dictionary

aws_batch/pipeline.py
```diff
@@ -455,15 +459,21 @@ def run_and_log(logparams, func_name, *args):
...
+        logger.info("duplicate compression ratio: %s" % str(compression_ratio))
     if func_name.__name__ == "run_priceseqfilter":
-      pass_percentage = (100.0 * records_after) / records_before
-      logger.info("percentage of reads passing QC filter: %s %%" % str(pass_percentage))
+        pass_percentage = (100.0 * records_after) / records_before
+        logger.info("percentage of reads passing QC filter: %s %%" % str(pass_percentage))
     # copy log file
     execute_command("aws s3 cp %s %s/;" % (logger.handlers[0].baseFilename, logparams["sample_s3_output_path"]))
+    # write stats
+    stats_path = os.path.join(result_dir, STATS_OUT)
```
yunfangjuan: add the AWS_BATCH_ID to keep track which job generates which?

aws_batch/pipeline.py
```diff
@@ -489,6 +500,8 @@ def run_sample(sample_s3_input_path, sample_s3_output_path,
     handler.addFilter(TimeFilter())
     handler.setFormatter(formatter)
     logger.addHandler(handler)
+    default_logparams = {"sample_s3_output_path": sample_s3_output_path,
```
yunfangjuan: probably make this a constant and upper cading it.

cdebourcy: ok I uppercased it. Didn't make it global because it needs sample_s3_output_path and result_dir, which are only defined within run_sample

PR #112, opened 2017-09-29
Fix logic to display samples

sadelokiki:


app/controllers/home_controller.rb
```diff
@@ -1,16 +1,12 @@
...
+    @samples = Sample.paginate(page: params[:page]).order('created_at DESC')
 
-    @pipeline_outputs = PipelineOutput.order("created_at DESC")
-
-    @pipeline_outputs.each do |output|
-      sample_info =  output.sample
-      project_info = output.sample.project
-      @output_data['sample_info'] = sample_info
+    @samples.each do |output|
+      pipeline_output_info = output.pipeline_outputs.order('created_at').last
```
yunfangjuan: this will break if there's no pipeline_output for a sample yet.  I think you will still show something on the page just say results are pending as shown currently.

app/controllers/home_controller.rb
```diff
@@ -1,16 +1,12 @@
...
     @output_data = {}
+    @samples = Sample.paginate(page: params[:page]).order('created_at DESC')
 
-    @pipeline_outputs = PipelineOutput.order("created_at DESC")
-
-    @pipeline_outputs.each do |output|
-      sample_info =  output.sample
-      project_info = output.sample.project
-      @output_data['sample_info'] = sample_info
+    @samples.each do |output|
```
yunfangjuan: This will generate a db query for each sample. Maybe an easy way is to query once and analyze the results? 

Or do something like 
pipeline_output_ids = PipelineOutput.select("sample_id, max(id) as maxid").where(sample_id: @samples.pluck(:id)).group(:sample_id).map {|u| u.attributes }.pluck("maxid")
@pipeline_outputs = PipelineOutput.where(id: [pipeline_output_ids]) 
and analyze the results.

yunfangjuan:
checking in idseq_dev.sql probably is not a great idea. It makes the web docker image a lot bigger over time.

boris-dimitrov:
@yunfangjuan @sadelokiki   I am working on a script to fetch a fresh 'idseq_dev.sql' mysql dump from the alpha or dev environment.    Almost done.

app/controllers/home_controller.rb
boris-dimitrov: @yunfangjuan I am new to this query mechanism, and just looked up `pluck`.   Is there a danger here that the output list of `maxid`s would not match the order of the input `samples.pluck(:id)`?

yunfangjuan: you don't really need the right order. You still have to post process them into a hash to do the sample => pipeline_output mapping.

PR #114, opened 2017-10-02
Script to run BASH commands on the WEB container for DEV/ALPHA

boris-dimitrov:
To obtain a SQL dump from environment 'dev', run it like so:

    bin/clam dev 'mysqldump -h $RDS_ADDRESS -u $DB_USERNAME --password=$DB_PASSWORD idseq_dev | gzip -c' | gzip -dc > id_dev.sql

Note how the environment variables are expanded on the remote container,
and how the sql dump is compressed on the remote machine, streamed in
compressed format, and uncompressed on the receiving machine.

This differs from the bin/shell script in that it does not run the command
in a tty but does run it in the correct container environment to enable
the expansion of environment variables and shell pipes.

Make sure to use single-quotes as in the above example.

### DIFF
--- a/bin/clam
+++ b/bin/clam
@@ -1,0 +1,27 @@
+#!/bin/bash
+#
+#  In order for this script to work, you will need PEM credentials for SSH.
+#
+#  Usage:
+#
+#      bin/clam dev bash
+#
+#      bin/clam dev 'echo $HOSTNAME'
+#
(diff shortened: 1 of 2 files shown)


boris-dimitrov:
I am not 100% thrilled with this since it still doesn't correctly forward STDIN, but it's plenty good enough for STDOUT like dumping SQL DB.

PR #115, opened 2017-10-02
web submission

yunfangjuan:


app/models/sample.rb
```diff
@@ -44,7 +44,7 @@ def initiate_s3_cp
     fastq2 = input_files[1].source
     command = "aws s3 cp #{fastq1} #{sample_input_s3_path}/;"
     command += "aws s3 cp #{fastq2} #{sample_input_s3_path}/;"
-    if s3_preload_result_path.present?
+    if s3_preload_result_path.size > 5 && s3_preload_result_path[0..4] == 's3://'
```
ryanking: you can skip the first part of this condition. ruby will not error when you slice longer than the string

app/controllers/samples_controller.rb
```diff
@@ -1,6 +1,7 @@
 class SamplesController < ApplicationController
```
ryanking: This logic is getting significant. Maybe not right now, but it would be great to see some more tests.

app/models/sample.rb
yunfangjuan: ok cool~

app/controllers/samples_controller.rb
yunfangjuan: agree. i have a few more things i want to add in for rerunning pipelines, etc. i will add all the test then.

app/models/sample.rb
```diff
@@ -46,7 +46,7 @@ def initiate_s3_cp
     fastq2 = input_files[1].source
     command = "aws s3 cp #{fastq1} #{sample_input_s3_path}/;"
     command += "aws s3 cp #{fastq2} #{sample_input_s3_path}/;"
-    if s3_preload_result_path.present?
+    if s3_preload_result_path[0..4] == 's3://'
```
boris-dimitrov: Don't you need to keep the presence check, before the prefix check?   In my samples db (fresh import from DEV), there are lots of legit-looking rows with NULL s3_preload_result_path, and in the rails console calling [0..4] on a nil value raises an exception.

PR #119, opened 2017-10-03
Add taxon db

cdebourcy:


app/models/taxon_category.rb
```diff
@@ -0,0 +1,3 @@
+class TaxonCategory < ApplicationRecord
```
yunfangjuan: maybe add some documentation for each model. they are not self explanatory.

db/migrate/20171002234500_create_taxon_descriptions.rb
```diff
@@ -0,0 +1,9 @@
+class CreateTaxonDescriptions < ActiveRecord::Migration[5.1]
+  def change
+    create_table :taxon_descriptions do |t|
+      t.bigint :version
```
yunfangjuan: maybe you can actually add a description field here?  which will have descriptions like "download from 10/2/2017 ncbi"

db/migrate/20171002234843_create_taxon_child_parents.rb
```diff
@@ -0,0 +1,12 @@
+class CreateTaxonChildParents < ActiveRecord::Migration[5.1]
+  def change
+    create_table :taxon_child_parents do |t|
+      t.integer :taxid
+      t.integer :parent_taxid
+      t.string :rank
+      t.references :taxon_description, foreign_key: true
```
yunfangjuan: i thought we discussed yesterday we don't want to add the version for now. because every time we run this script, a lot of new records would be generated and overtime, it would slow down the database. If we want to keep version, we should probably have another table which archived the old version and move the records over. The actualy TaxonChildParents datababase will only keep the most recent K versions.

yunfangjuan: Also, it's probably a lot more efficient to just do (taxid => lineage) table instead of the child/parent as it has a lot more records.

lib/tasks/taxondb.rake
```diff
@@ -0,0 +1,22 @@
+require 'open3'
+require 'csv'
+task load_taxon_db: :environment do
+  t = TaxonDescription.new(version: 1)
```
yunfangjuan: use create  instead of new/save abstract out version and descriptions as arguments.

lib/tasks/taxondb.rake
```diff
@@ -0,0 +1,22 @@
...
+  nodes_s3_file = "nodes.dmp"
+  categories_s3_file = "categories.dmp"
+  s3_taxonomy_path = "s3://czbiohub-infectious-disease/taxonomy"
+  local_taxonomy_path = "/app/tmp/taxonomy"
+  command = "mkdir -p #{local_taxonomy_path};"
+  command += "aws s3 cp #{s3_taxonomy_path} #{local_taxonomy_path} --recursive;"
+  Open3.capture3(command)
+  nodes_file = "#{local_taxonomy_path}/#{nodes_s3_file}"
+  CSV.open(nodes_file, "r", { :col_sep => "\t|\t" }).each do |n|
+    t.taxon_child_parents.create(taxid: n[0].to_i, parent_taxid: n[1].to_i, rank: n[2])
```
yunfangjuan: calculate lineage first and load into database instead of loading child parents directly?

app/models/taxon_category.rb
cdebourcy: ok

db/migrate/20171002234843_create_taxon_child_parents.rb
cdebourcy: ok, I removed TaxonDescription. Will do full lineage table in a later PR

boris-dimitrov:
This may be a silly question, but why are we using a relational database here?   What types of queries are we likely to run against it?   How many rows will there be?

I suspect this was discussed already and I am misusing the pull request as a design review discussion.  Please feel free to just point me to the right doc.

cdebourcy:
We'll use this to find the correct taxids when the user decided to have
reports displayed at a different level of taxonomic granularity. E.g see
family-level scores instead of species-level scores. Or the user may want
to see only genus-level scores, but then expand one of the genuses to see
the detailed species counts within that genus. There will be ~1.6 million
rows in TaxonChildParent

ryanking:
@boris-dimitrov the reason for a relational db is simplicity. We could use something else but that'd take longer and this is good enough.

PR #120, opened 2017-10-03
document rails db console, aws setup, exec vs run

boris-dimitrov:


boris-dimitrov:
@jameshowardwang I took some care to exclude site-specific instructions, lmk if I missed some.

PR #123, opened 2017-10-03
Feature/fixpagination

sadelokiki:
Fix pagination.


app/assets/javascripts/components/PipelineSampleReads.jsx
```diff
@@ -69,14 +69,14 @@ class PipelineSampleReads extends React.Component {
             </div>
 
             <div className="sub-title">
-              <a href="/"> {this.projectInfo.name} </a> > { this.sampleInfo.name }
+              <a href="/"> {!this.projectInfo ? 'No project to display yet' : this.projectInfo.name} </a> > { !this.sampleInfo ? 'No sample to display' : this.sampleInfo.name}
```
boris-dimitrov: Forgive me for being a complete front-end n00b, but two questions:

1.  The language inside {} -- is it Java or Ruby?

2.  Is it possible to take care of the uninitialized data case at the model level?   Where if there is no data, we crate a stub with empty fields, so that in this template we don't have to check for existence of every field.

sadelokiki: @boris-dimitrov, the lang inside {} is javascript
the uninitialized data case is checked for to take care of cases where null is being returned as a value from the API. 
Also, if an object with empty fields is returned, there might still need to be a way to check and show that there is "no data" returned for such fields.

boris-dimitrov: I think it would be better for the logic that decodes data (including NIL or missing data) to be in one place, rather than scattered like this.   Perhaps someone more familiar with webapp development can suggest what that place might be.   As an onbserver and occasional fixer of frontend bugs in several past projects, I can attest that scattering this type of logic tends to make it harder to assure correctness and make changes in future.

ryanking:
@sadelokiki In order to make PRs easy to read and understand for the team, they should only have commits listed that are part of the change.

You may not have intended this, but this PR lists a bunch of irrelevant commits. I'm not sure how to fix it because I don't know how you created this branch.

boris-dimitrov:
My guess is the number of commits shown might shrink considerably after rebasing on master, but unfortunately the rebase cannot be done automatically due to conflicts.

I think it may be best to recreate the branch and just copy the changed files over to create a single commit.   There are very few changes relative to master, so this will be quick.   Then don't forget to delete this branch.

Am I correct that our recommended process for using git is as follows?
```
start:
git fetch
git checkout master
git checkout -b susan/my_new_feature origin/master
<change some files>
git commit
git push origin susan/my_new_feature
<create pull request>
<when approved, squash and merge to master via ui>
<go to start>
```

Specifically for this branch, please remember to exclude 'idseq_dev.sql' from the commit.

ryanking:
@boris-dimitrov yes and you can use `bin/pr` for creating the pull request (it will rebase and run tests first)

PR #124, opened 2017-10-03
add taxon description tables

cdebourcy:


cdebourcy:
this is a redo of add-taxon-db with Yun-Fang's comments addressed. It was easier to just start over as far as the migrations are concerned

lib/tasks/taxondb.rake
```diff
@@ -0,0 +1,23 @@
...
+  f.each do |n|
+    TaxonChildParent.find_or_create_by(taxid: n[0].to_i).update(parent_taxid: n[1].to_i, rank: n[2])
+  end
+  f.close
+  categories_file = "#{local_taxonomy_path}/#{categories_s3_file}"
+  f = CSV.open(categories_file, "r", col_sep: "\t")
+  f.each do |c|
+    TaxonCategory.find_or_create_by(taxid: c[2].to_i).update(category: c[0])
+  end
+  f.close
```
yunfangjuan: you might want to remove the files in the end.

cdebourcy: agreed

PR #130, opened 2017-10-04
make sure we are not on master before pushing branch

boris-dimitrov:


boris-dimitrov:
Because the branch was rebased and pushed to github with 'bin/pr', we can also be sure that all tests passed.

PR #133, opened 2017-10-04
Fix pagination

sadelokiki:


boris-dimitrov:
Wow this diff cleaned up nicely.

PR #135, opened 2017-10-04
Add lineage db

cdebourcy:


db/migrate/20171004210243_create_taxon_lineages.rb
```diff
@@ -0,0 +1,16 @@
+class CreateTaxonLineages < ActiveRecord::Migration[5.1]
+  def change
+    create_table :taxon_lineages do |t|
+      t.integer :taxid
```
ryanking: this should probably be not null, yeah?

db/migrate/20171004210243_create_taxon_lineages.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.integer :superkingdom_taxid
```
ryanking: probably this too?

db/migrate/20171004210452_create_taxon_names.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTaxonNames < ActiveRecord::Migration[5.1]
+  def change
+    create_table :taxon_names do |t|
+      t.integer :taxid
```
ryanking: probably should be not null

db/migrate/20171004210243_create_taxon_lineages.rb
cdebourcy: agree

db/migrate/20171004210243_create_taxon_lineages.rb
cdebourcy: sure

db/migrate/20171004210452_create_taxon_names.rb
cdebourcy: ok

ncbitax2lin/README.md
```diff
@@ -0,0 +1,89 @@
+# NCBItax2lin
```
yunfangjuan: should we put all the code here?  could you just put ncbitax2lin into Dockerfile? and maybe write a quick script to run the lineage?

cdebourcy: ok, I forked ncbitax2lin and made it a separate repo that gets cloned inside the rake task

db/schema.rb
```diff
@@ -150,6 +150,26 @@
     t.index ["pipeline_output_id"], name: "index_taxon_counts_on_pipeline_output_id"
   end
 
+  create_table "taxon_lineages", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
+    t.integer "taxid", null: false
+    t.integer "superkingdom_taxid", null: false
+    t.integer "phylum_taxid"
```
yunfangjuan: maybe you can make them all not null?

cdebourcy: yep!

PR #136, opened 2017-10-05
speed up taxon db import

cdebourcy:


lib/tasks/taxondb.rake
```diff
@@ -9,16 +9,13 @@ task load_taxon_db: :environment do
...
-    TaxonCategory.find_or_create_by(taxid: c[2].to_i).update!(category: c[0])
-  end
-  f.close
+  `
+    cd #{local_taxonomy_path};
+    awk -F "\t|\t" '{print x+=1,","$1","$3","$5",2017-10-04,2017-10-04"}' #{nodes_file} | sed 's= ,=,=' > taxon_child_parents;
+    mysqlimport --local --user=$DB_USERNAME --host=$RDS_ADDRESS --password=$DB_PASSWORD --fields-terminated-by=',' idseq_#{Rails.env} taxon_child_parents;
+    awk -F "\t" '{print x+=1,","$3","$1",2017-10-04,2017-10-04"}' #{categories_file} | sed 's= ,=,=' > taxon_categories;
+    mysqlimport --local --user=$DB_USERNAME --host=$RDS_ADDRESS --password=$DB_PASSWORD --fields-terminated-by=',' idseq_#{Rails.env} taxon_categories;
+  `
```
boris-dimitrov: Cool.   Two quick questions.   If anything fails here, does the rake task fail?   Will the table have been created in the DB by the time this script runs, and by what entity/code?

boris-dimitrov: Also may be worth noting in comment the number of rows ~200k and time it takes otherwise. :)

cdebourcy: Yes, the rake task should fail if the backtick block fails, but I have now added an explicit failure message for that case. Yes, this rake task is run after the table is already created. It doesn't run automatically, we're executing it manually for now. 
I've added comments about the numbers of rows and the durations of the two approaches we tried.

PR #137, opened 2017-10-05
add prometheus client

ryanking:
* set up client
* use rack exporter and collector

collector will get us some baseline throughput + latency stats

exporter is how we will get the data out


boris-dimitrov:
What are good examples for using this tool?

ryanking:
For examples or specifics on how prometheus works, google is probably the best way to answer questions.

Regarding using other tools - yes, of course. We can use other things. If you're interested, do the research and figure it out and do it.

PR #138, opened 2017-10-05
enable restart from web

yunfangjuan:


yunfangjuan:
Installment #1. for rerun from web interface. More user management cleanup to come.

PR #142, opened 2017-10-06
honeycomb logging

boris-dimitrov:
Honeycomb is a super basic analytics DB that lets you filter and group timestamped log events and chart sum/avg/percentile of any event field over time.

The rails code here emits stats that are already being gathered by rails.  For example, db_runtime, the total latency of DB queries in the request.  We can extend this event listener pattern as needed to emit more information.   The idea is to keep all logging here.

To use this, you need to set IDSEQ_HONEYCOMB_WRITE_KEY to a special value and IDSEQ_HONEYCOMB_DATA_SET to idseq_web.   Ryan can invite you to our team account on honeycomb so you can issue queries against the events.   Events show up pretty instantaneously.

TODO:  Add event fields for environment, host,  logged in user id, more per-request stats.

### DIFF
--- a/config/initializers/honeycomb.rb
+++ b/config/initializers/honeycomb.rb
@@ -1,0 +1,26 @@
+# See https://honeycomb.io/docs/guides/rails/
+require 'libhoney'
+
+if ENV["IDSEQ_HONEYCOMB_WRITE_KEY"] && ENV["IDSEQ_HONEYCOMB_DATA_SET"]
+
+  HONEYCOMB = Libhoney::Client.new(writekey: ENV["IDSEQ_HONEYCOMB_WRITE_KEY"],
+                                   dataset: ENV["IDSEQ_HONEYCOMB_DATA_SET"])
+
+  # See http://guides.rubyonrails.org/active_support_instrumentation.html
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -27,6 +27,10 @@ services:
       - redis
     environment:
       - SAMPLES_BUCKET_NAME=czbiohub-idseq-samples-development
+      - IDSEQ_AIRBRAKE_PROJECT_ID
```
boris-dimitrov: We can make some changes to idseq_infra and airbrake.rb to take advantage of this.

config/initializers/honeycomb.rb
```diff
@@ -0,0 +1,24 @@
+require 'libhoney'
```
boris-dimitrov: The code in this file was dumped verbatim from the "getting started" example.

ryanking: maybe drop a link to that example in a comment here?

boris-dimitrov: Great idea.   I also found every logging/profiling tool ever made for Rails by googling for this line from the example:
```
 ActiveSupport::Notifications.subscribe(/process_action.action_controller/) do
```
They all use it :)

PR #143, opened 2017-10-06
Cdebourcy patch 1

cdebourcy:


aws_batch/pipeline.py
```diff
@@ -460,15 +460,15 @@ def run_and_log(logparams, func_name, *args):
...
+        records_before = int(count_reads(logparams["before_file_name"], logparams["before_file_type"]))
+        records_after = int(count_reads(logparams["after_file_name"], logparams["after_file_type"]))
     if logparams["count_reads"]:
         percent_removed = (100.0 * (records_before - records_after)) / records_before
         logger.info("%s %% of reads dropped out, %s reads remaining" % (str(percent_removed), str(records_after)))
         STATS.append({'task': func_name.__name__, 'reads_before': records_before, 'reads_after': records_after})
     # function-specific logs
     if func_name.__name__ == "run_cdhitdup":
-        compression_ratio = records_before / records_after
+        compression_ratio = (1.0 * records_before) / records_after
```
yunfangjuan: handle division by zero?

aws_batch/pipeline.py
```diff
@@ -460,15 +460,15 @@ def run_and_log(logparams, func_name, *args):
     # count records
     required_params = ["before_file_name", "before_file_type", "after_file_name", "after_file_type"]
     if all(param in logparams for param in required_params):
-        records_before = count_reads(logparams["before_file_name"], logparams["before_file_type"])
-        records_after = count_reads(logparams["after_file_name"], logparams["after_file_type"])
+        records_before = int(count_reads(logparams["before_file_name"], logparams["before_file_type"]))
```
yunfangjuan: why don't you just int() inside the count_reads function?

aws_batch/pipeline.py
cdebourcy: due to the nature of run_cdhitdup (removing duplicates, keeping one copy), records_after can't be zero here

PR #146, opened 2017-10-06
disable SSL redirect for health check so the load balancer can fetch it

boris-dimitrov:


config/environments/alpha.rb
```diff
@@ -46,6 +46,7 @@
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
+  config.ssl_options = { redirect: { exclude: ->(request) { request.path =~ /healthcheck/ } } }
```
ryanking: path is /health_check I think?

PR #147, opened 2017-10-06
Adding Coveralls

ryanking:
Coveralls is a service for tracking test coverage and quality -> coveralls.io

It will take a while of collecting baseline data before there's a ton to see there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ rvm:
 services:
   - docker
 before_install:
-  - env
+  - env | sort
   - mkdir -p ~/.aws && printf "[default]\nregion = $AWS_REGION\naws_access_key_id = $AWS_ACCESS_KEY_ID\naws_secret_access_key = $AWS_SECRET_ACCESS_KEY\n" > ~/.aws/credentials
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -12,7 +12,7 @@ before_install:
   - docker-compose up -d db
   - docker-compose up -d redis
   - sleep 10
-  - docker-compose run -e RAILS_ENV=test web rake db:create
+  - docker-compose run web rake db:create
```
boris-dimitrov: Why no longer just 'test'?

test/test_helper.rb
```diff
@@ -1,6 +1,9 @@
 require File.expand_path('../../config/environment', __FILE__)
 require 'rails/test_help'
 
+require 'coveralls'
+Coveralls.wear!
```
boris-dimitrov: cute

.travis.yml
ryanking: travis-ci set RAILS_ENV=test and bc the docker-compose now passes RAILS_ENV through, this is redundant.

PR #150, opened 2017-10-07
add indices

cdebourcy:


db/migrate/20171007001015_add_indices.rb
```diff
@@ -0,0 +1,16 @@
+class AddIndices < ActiveRecord::Migration[5.1]
+  def change
+    add_index :taxon_zscores, [:report_id, :hit_type, :tax_level, :tax_id], unique: true, name: "index_taxon_zscores_detailed"
```
boris-dimitrov: I was curious if the order in which these columns are listed matters, and apparently it does.
```
MySQL can use multiple-column indexes for queries that test all the columns in the index, or queries that test just the first column, the first two columns, the first three columns, and so on. If you specify the columns in the right order in the index definition, a single composite index can speed up several kinds of queries on the same table.
```

boris-dimitrov: https://dev.mysql.com/doc/refman/5.7/en/multiple-column-indexes.html

db/schema.rb
```diff
@@ -180,6 +191,7 @@
     t.string "name"
     t.float "rpm", limit: 24
     t.string "hit_type"
+    t.index ["report_id", "hit_type", "tax_level", "tax_id"], name: "index_taxon_zscores_detailed", unique: true
```
boris-dimitrov: I meant to place my earlier comments on this line, sorry.

db/migrate/20171007001015_add_indices.rb
cdebourcy: good point. I think I'll want to change the order to [:report_id, :tax_level, :hit_type, :tax_id] actually

PR #153, opened 2017-10-09
update readme

cdebourcy:


ryanking:
Do we know why this needs to change?

cdebourcy:
"run" works and "exec" does not (but we don't yet know why)

boris-dimitrov:
The behavior of docker-compose changed last week, possibly after Docker update, or may be we changed something.

ryanking:
👍

PR #157, opened 2017-10-09
generate reports on the fly, without storing in DB

boris-dimitrov:
now faster with direct SQL query, and added back sort and filter

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -12,8 +12,9 @@ class ReportsController < ApplicationController
   def show
     @report_details = report_details(@report)
     @view_level = params[:view_level] ? params[:view_level].downcase : 'genus'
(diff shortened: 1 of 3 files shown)


boris-dimitrov:
A test is failing for this change.   Not sure why, yet.

```
# Running:

....................E

Error:
PipelineOutputsControllerTest#test_should_show_pipeline_output:
NoMethodError: undefined method `pipeline_output' for nil:NilClass
    app/lib/report_helper.rb:25:in `compute_taxon_zscores'
    app/lib/report_helper.rb:4:in `external_report_info'
    app/controllers/pipeline_outputs_controller.rb:20:in `show'
    test/controllers/pipeline_outputs_controller_test.rb:29:in `block in <class:PipelineOutputsControllerTest>'

```

boris-dimitrov:
Tests now pass.   There were some spurious '@' characters.

PR #158, opened 2017-10-10
add lineage/name table

cdebourcy:
We haven't fully decided whether it is best to have:
(1) a lineage table (expressed as taxids) plus a taxid-to-name conversion table, or
(2) a table that gives the lineage for every taxid in terms of the names of the parents.
@cyrielo, it sounds like you would find (2) useful as it allows you to look up the info you need for report displays directly and easily -- so that's what this pull request does

### DIFF
--- a/app/models/taxon_lineage_name.rb
+++ b/app/models/taxon_lineage_name.rb
@@ -1,0 +1,2 @@
+class TaxonLineageName < ApplicationRecord
+end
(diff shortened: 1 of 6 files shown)


cyrielo:
I believe option 2 could be useful in looking up the names, but if we can prevent a lookup query while iterating the lineage that could be more efficient but for now this would do.

PR #159, opened 2017-10-10
fix typo that affects report filtering (highest z score computation)

boris-dimitrov:


boris-dimitrov:
@yunfangjuan I am adding view_level to the permitted parameters, but I am puzzled that nt_rpm_threshold parameters work yet they are not permitted.  What am I missing?

The external_report_info() info function appears to be called from several controllers, should all be changed?

boris-dimitrov:
I think the problem is again with Ruby symbols vs strings.

Here it is expected to be a symbol
```
 wanted_level = level == :species ? TaxonCount::TAX_LEVEL_SPECIES : TaxonCount::TAX_LEVEL_GENUS
```
but here it is expected to be a string
```
    @view_level = params[:view_level] ? params[:view_level].downcase : 'genus'

```
Is there a convention or something about this?

yunfangjuan:
I believe if you are doing  ==, you should not use symbols but use single quoted strings instead. symbols are basically like enums.

yunfangjuan:
does permit flag no longer applies? if not, we should remove it but from the controller level, it seems we still use it.

boris-dimitrov:
I fixed this and other bugs, but am now failing a test...

```
bin/rails test test/controllers/reports_controller_test.rb:30

..............................E

Error:
PipelineOutputsControllerTest#test_should_show_pipeline_output:
NoMethodError: undefined method `map' for nil:NilClass
    app/lib/report_helper.rb:78:in `highest_tax_counts'
    app/lib/report_helper.rb:39:in `taxonomy_details'
    app/lib/report_helper.rb:6:in `external_report_info'
    app/controllers/pipeline_outputs_controller.rb:20:in `show'
    test/controllers/pipeline_outputs_controller_test.rb:15:in `block in <class:PipelineOutputsControllerTest>'
```

boris-dimitrov:
This passes all tests now.

app/lib/report_helper.rb
```diff
@@ -1,11 +1,11 @@
 # rubocop:disable ModuleLength
```
jameshowardwang: should we just disable it globally?

app/lib/report_helper.rb
```diff
@@ -33,53 +33,65 @@ def compute_taxon_zscores(report)
     zscore_array
   end
 
-  def taxonomy_details(view_level, _report, params, taxon_zscores)
-    sz = select_zscore(view_level, taxon_zscores)
+  # rubocop:disable Metrics/CyclomaticComplexity
```
jameshowardwang: ditto with these -- do we actually want them applied elsewhere?

jameshowardwang: my personal preference is to not have a bunch of rubocop exception comments littering our code

boris-dimitrov: Yes I think these warnings are less useful.   They tend to steer us toward making the code more compact, but sometimes that makes it more difficult to understand.

PR #162, opened 2017-10-10
Fix Filter component to handle invalid range filter

cyrielo:
This updated PR implement manual range filtering, disable the range slider when range values are invalid example when the min and max are equal or min is less than max.

### DIFF
--- a/app/assets/javascripts/components/ReportFilter.jsx
+++ b/app/assets/javascripts/components/ReportFilter.jsx
@@ -2,7 +2,6 @@ class ReportFilter extends React.Component {
   constructor(props) {
     super(props);
     this.background_model = props.background_model || 'N/A';
-    this.report_title = props.report_title || 'Not Set';
 
     this.highest_nt_zscore = this.props.highest_tax_counts.highest_nt_zscore;
     this.highest_nt_rpm = this.props.highest_tax_counts.highest_nt_rpm;
--- a/app/assets/javascripts/components/ReportFilter.jsx
+++ b/app/assets/javascripts/components/ReportFilter.jsx
@@ -1,11 +1,24 @@
 class ReportFilter extends React.Component {
   constructor(props) {
     super(props);
-    const view_level = props.view_level || 'Genus';
     this.background_model = props.background_model || 'N/A';
     this.report_title = props.report_title || 'Not Set';
-    const default_nt_zscore_threshold = `0,${this.props.highest_tax_counts.highest_nt_zscore}`;
-    const default_nt_rpm_threshold = `0,${this.props.highest_tax_counts.highest_nt_rpm}`;
+
+    this.highest_nt_zscore = this.props.highest_tax_counts.highest_nt_zscore;
+    this.highest_nt_rpm = this.props.highest_tax_counts.highest_nt_rpm;
+
+    this.lowest_nt_zscore = this.props.highest_tax_counts.lowest_nt_zscore;
+    this.lowest_nt_rpm = this.props.highest_tax_counts.lowest_nt_rpm;
+
+    this.has_valid_nt_zscore_range =
+    (this.highest_nt_zscore && (this.lowest_nt_zscore !== this.highest_nt_zscore));
+
+    this.has_valid_nt_rpm_range =
+    (this.highest_nt_rpm && (this.lowest_nt_rpm !== this.highest_nt_rpm));
+
+
+    const default_nt_zscore_threshold = `${this.lowest_nt_zscore},${this.highest_nt_zscore}`;
+    const default_nt_rpm_threshold = `${this.lowest_nt_rpm},${this.highest_nt_rpm}`;
 
     const nt_zscore_threshold = ReportFilter.getFilter('nt_zscore_threshold')
       ? ReportFilter.getFilter('nt_zscore_threshold') : default_nt_zscore_threshold;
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/ReportFilter.jsx
```diff
@@ -73,7 +75,7 @@ class ReportFilter extends React.Component {
       behaviour: 'tap-drag',
       range: {
         'min': 0,
-        'max': this.props.highest_tax_counts.highest_nt_zscore
+        'max':this.highest_nt_zscore
```
jameshowardwang: missing a space

app/assets/javascripts/components/ReportFilter.jsx
```diff
@@ -4,8 +4,10 @@ class ReportFilter extends React.Component {
     const view_level = props.view_level || 'Genus';
     this.background_model = props.background_model || 'N/A';
     this.report_title = props.report_title || 'Not Set';
-    const default_nt_zscore_threshold = `0-${this.props.highest_tax_counts.highest_nt_zscore}`;
-    const default_nt_rpm_threshold = `0-${this.props.highest_tax_counts.highest_nt_rpm}`;
+    this.highest_nt_zscore = this.props.highest_tax_counts.highest_nt_zscore || 100;
+    this.highest_nt_rpm = this.props.highest_tax_counts.highest_nt_rpm || 100;
```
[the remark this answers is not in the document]
boris-dimitrov: Valid points, @cdebourcy    @cyrielo @eggerdesigns and I had a chat about filter ranges, and we decided that 1) the default range should be the min ... max from the report data  2) if min=max, the sliders will be turned off in UI.   This will address the issue that prompted this pull request, namely that a javascript exception was being thrown by the slider UI component when min=max.

app/assets/javascripts/components/ReportFilter.jsx
```diff
@@ -14,22 +30,23 @@ class ReportFilter extends React.Component {
...
 
     const nt_rpm_start =
-      (nt_rpm_threshold.split(',').length > 0) ? parseInt(nt_rpm_threshold.split(',')[0], 10) : 0;
+      (nt_rpm_threshold.split(',').length > 0) ? parseInt(nt_rpm_threshold.split(',')[0], 10) :
+      this.lowest_nt_rpm;
     const nt_rpm_end =
       (nt_rpm_threshold.split(',').length > 1) ? parseInt(nt_rpm_threshold.split(',')[1], 10) :
-        this.props.highest_tax_counts.highest_nt_rpm;
-
+      this.highest_nt_rpm;
```
boris-dimitrov: There is a lot of repeated code here, when we go to 8 columns with 2 bounds each, it will likely be unmanageable.  Probably should generalize/parameterize so you don't have to repeat each line 16 times:  for each {highest, lowest} x {genus, species} x {nt, nr} x {zscore, rpm}.

app/assets/javascripts/components/ReportFilter.jsx
```diff
@@ -55,65 +72,82 @@ class ReportFilter extends React.Component {
...
+        },
+        format: wNumb({
+          decimals: 0
+        })
+      });
+
+      nt_zscore.noUiSlider.on('update', (values, handle) => {
+        if (handle === 0) {
+          this.setState({
+            nt_zscore_start: Math.round(values[handle])
```
boris-dimitrov: start and end shouldn't be rounded, because data points may end up outside the rounded range

to convert `start` and `end` to integers, use the function `Math.floor` for `start`, and `Math.ceil` for end
(i am assuming JS has those, and I am also assuming that `Math.floor(-1.1)` is `-2.0` and not `-1.0` -- could you please confirm that).

But to the broader question, why are you converting to integers?   If the actual data range is [-4.35 to -1.21], shouldn't users be permitted to select values in that range?

My suggestion is for now if you need to convert to integers, to use floor and ceil, and also research whether converting to integers is wise.

app/assets/javascripts/components/ReportFilter.jsx
```diff
@@ -55,65 +72,82 @@ class ReportFilter extends React.Component {
...
+          this.setState({
+            nt_rpm_start: Math.round(values[handle])
+          });
+        } else {
+          this.setState({
+            nt_rpm_end: Math.round(values[handle])
+          });
+        }
+      });
+    }
```
boris-dimitrov: Is it typical to duplicate code like that in JS?   For example, when you figure out the correct rounding math, if the code is duplicated, you will need to fix in two places.

app/assets/javascripts/components/ReportFilter.jsx
```diff
@@ -174,54 +208,50 @@ class ReportFilter extends React.Component {
...
                       </div>
-                      <div className="slider-values">
-                        <div className="nt_rpm start">{ this.state.nt_rpm_start }</div>
-                        <div className="nt_rpm end">{ this.state.nt_rpm_end }</div>
-                      </div>
-                      <div id="nt_rpm"></div>
-                    </div>
+                    ) : ''}
+                     { (!this.has_valid_nt_rpm_range && !this.has_valid_nt_zscore_range) ?
+                    <div className="center"><small>Cannot set thresholds</small></div> : '' }
```
boris-dimitrov: Let's show this to Rebecca when it's pushed, she may prefer some different way to indicate the min=max condition.   Personally I don't think of that as an error condition.

PR #163, opened 2017-10-11
relax rubocop warnings

boris-dimitrov:


boris-dimitrov:
Disabled ClassLength warning as suggested by Charles.

PR #166, opened 2017-10-11
Slack deploy notifications

cdebourcy:
This will notify your Slack channel of deploys if you have the environment variable SLACK_HOOK set to an appropriate incoming webhook.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -149,12 +149,20 @@ def notify_slack(env, image, stage)
       puts `curl -X POST --data-urlencode 'payload={"icon_emoji": ":ghost:", "text": "starting to deploy #{image} to #{env}"}' #{slack_hook}`
     elsif stage=="end"
       puts `curl -X POST --data-urlencode 'payload={"icon_emoji": ":sharkdance:", "text": "finished deploying #{image} to #{env}"}' #{slack_hook}`
+    elsif stage=="failure"
+      puts `curl -X POST --data-urlencode 'payload={"icon_emoji": ":sob:", "text": "deploy of #{image} to #{env} failed"}' #{slack_hook}`
     end
+  else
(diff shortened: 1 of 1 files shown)


bin/deploy
```diff
@@ -142,6 +142,19 @@ def deploy(client, cluster, service, repo, image, run_migrations = false)
...
 
+def notify_slack(env, image, stage)
+  slack_hook = ENV["SLACK_HOOK"]
+  if slack_hook
+    if stage=="start"
+      puts `curl -X POST --data-urlencode 'payload={"icon_emoji": ":ghost:", "text": "starting to deploy #{image} to #{env}"}' #{slack_hook}`
+    elsif stage=="end"
+      puts `curl -X POST --data-urlencode 'payload={"icon_emoji": ":sharkdance:", "text": "finished deploying #{image} to #{env}"}' #{slack_hook}`
+    end
+  end
```
ryanking: Put an else here where you print a message like "not notifying slack because blah, blah blah" and explain how to set it up

bin/deploy
```diff
@@ -142,6 +142,19 @@ def deploy(client, cluster, service, repo, image, run_migrations = false)
...
+end 
+
+notify_slack(env, image, "start")
 deploy(client, env, RAILS_SERVICE, REPO, image, true)
 deploy(client, env, RESQUE_WORKER_SERVICE, REPO, image, false)
 deploy(client, env, RESQUE_SCHEDULER_SERVICE, REPO, image, false)
+notify_slack(env, image, "end")
```
ryanking: put a begin/rescue around this then do a notify of failure

jameshowardwang:
nice!

bin/deploy
cdebourcy: ok done

bin/deploy
cdebourcy: ok done

PR #167, opened 2017-10-11
Cyril/display genus and species

cyrielo:
This PR adds the ability to display the genus and the species in the table, and also changes the delimiter for the range filtering to use comma(,)

### DIFF
--- a/app/assets/javascripts/components/PipelineSampleReport.jsx
+++ b/app/assets/javascripts/components/PipelineSampleReport.jsx
@@ -14,7 +14,6 @@ class PipelineSampleReport extends React.Component {
         ? current_sort.sort_query  : 'sort_by=highest_zscore'
     };
     this.applySort = this.applySort.bind(this);
-    console.log(props)
   }
 
   uppCaseFirst(name) {
--- a/app/assets/javascripts/components/PipelineSampleReport.jsx
+++ b/app/assets/javascripts/components/PipelineSampleReport.jsx
@@ -91,16 +91,16 @@ class PipelineSampleReport extends React.Component {
     return (
       <div>
         <div id="reports" className="reports-screen tab-screen col s12">
-          <div className="container tab-screen-content">
+          <div className="tab-screen-content">
             <div className="row">
-              <div className="col s3">
+              <div className="col s2">
                 <ReportFilter
                   background_model = {this.report_details.background_model.name}
                   report_title = { this.report_details.report_info.name }
                   view_level={this.uppCaseFirst(this.view_level)}
                   highest_tax_counts={this.highest_tax_counts}/>
               </div>
-              <div className="col s9 reports-main ">
+              <div className="col s10 reports-main ">
                 <div className="report-sort right">
 
                   <span className="">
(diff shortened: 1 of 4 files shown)


app/lib/report_helper.rb
```diff
@@ -50,11 +55,25 @@ def taxonomy_details(view_level, report, params)
       )
     end
     sort_report!(nt, params[:sort_by])
+
     nr_zscore_by_taxon = nr.group_by { |h| h[:tax_id] }
+    genus_nr_zscore_by_taxon = nr_genus.group_by { |h| h[:tax_id] }
+
     tax_details = []
     nt.each do |h|
```
cdebourcy: This takes all the NT hits and displays them with their corresponding NR scores, if any. It does not show hits that were only present in NR but not in NT. However, those hits should also be displayed. Could you add that in, by refactoring so that and nt and nr are treated equivalently?

app/assets/javascripts/components/ReportFilter.jsx
```diff
@@ -14,15 +14,15 @@ class ReportFilter extends React.Component {
       ? ReportFilter.getFilter('nt_rpm_threshold') : default_nt_rpm_threshold;
 
     const nt_zscore_start =
-      (nt_zscore_threshold.split('-').length > 0) ? parseInt(nt_zscore_threshold.split('-')[0], 10) : 0;
+      (nt_zscore_threshold.split(',').length > 0) ? parseInt(nt_zscore_threshold.split('-')[0], 10) : 0;
```
boris-dimitrov: I think you forgot to convert some '-' to ','.

Would you mind approving my pull request that specifically does just that, so we can have an unbroken master?   Since the work here needs one more iteration.

app/lib/report_helper.rb
```diff
@@ -32,14 +32,19 @@ def compute_taxon_zscores(report)
...
 
-  def taxonomy_details(view_level, report, params)
+  def taxonomy_details(report, params)
     taxon_zscores = compute_taxon_zscores(report)
-    wanted_level = view_level == 'species' ? TaxonCount::TAX_LEVEL_SPECIES : TaxonCount::TAX_LEVEL_GENUS
-    data = taxon_zscores.group_by { |h| h[:tax_level] == wanted_level && h[:hit_type] }
+    genus_level = TaxonCount::TAX_LEVEL_GENUS
+    species_level = TaxonCount::TAX_LEVEL_SPECIES
+    data = taxon_zscores.group_by { |h| h[:tax_level] == species_level && h[:hit_type] }
+    data2 = taxon_zscores.group_by { |h| h[:tax_level] == genus_level && h[:hit_type] }
```
boris-dimitrov: Can you do this with one group by?
```
data = taxon_zscores.group_by { |h| [h[:tax_level], h[:hit_type]] }
```
Then you can refer to, e.g.
```
data[[genus_level, 'NT']]
```

app/lib/report_helper.rb
```diff
@@ -50,11 +55,25 @@ def taxonomy_details(view_level, report, params)
...
+    genus_nr_zscore_by_taxon = nr_genus.group_by { |h| h[:tax_id] }
+
     tax_details = []
     nt.each do |h|
+      lineage_info = TaxonLineage.find_by(taxid: h[:tax_id])
+      category_taxid = lineage_info.superkingdom_taxid
+
+      genus_taxid = lineage_info.genus_taxid
+      found_genus = nt_genus.select { |genus| genus[:tax_id] == genus_taxid }
+      matched_genus = found_genus.length == 1 ? found_genus[0] : nil
```
boris-dimitrov: I think Ruby does exactly this if you just write
```
    matched_genus = found_genus[0]
```

app/lib/report_helper.rb
```diff
@@ -50,11 +55,25 @@ def taxonomy_details(view_level, report, params)
...
+
+      category = TaxonName.find_by(taxid: category_taxid)
```
boris-dimitrov: Are we doing queries inside a loop here?   That's not good.

app/lib/report_helper.rb
```diff
@@ -103,15 +122,15 @@ def sort_report!(nt_zscores, sort_by)
 
   def compute_rpm(count, total_reads)
     if count
-      count * 1e6 / total_reads.to_f
+      (count * 1e6 / total_reads.to_f).round(3)
```
boris-dimitrov: Should rounding happen here or in JS?   I see advantages to both.  Note, however, this isn't the only place that computes zscores, so you can't rely on the other zscores merged from background summary to be rounded the way you like.   This would be an argument to push rounding into JS.

app/lib/report_helper.rb
```diff
@@ -32,29 +32,52 @@ def compute_taxon_zscores(report)
...
       )
     end
-    sort_report!(nt, params[:sort_by])
-    nr_zscore_by_taxon = nr.group_by { |h| h[:tax_id] }
+    sort_report!(nt_nr, params[:sort_by])
     tax_details = []
-    nt.each do |h|
-      x = nr_zscore_by_taxon[h[:tax_id]]
-      tax_details.push(nt_ele: h, nr_ele: x && x[0])
+    nt_nr.each do |h|
```
boris-dimitrov: outside the loop
```
all_tax_ids = nt_nr.map {|h| h[:tax_id]}
tax_ids_str = ""
all_tax_ids.each tax_id do
  tax_ids_str += tax_id.to_s
  tax_ids_str += ","
end
lineage_arr = TaxonLineage.connection.select_all("select tax_id, superkingdom_taxid,  genus_taxid from taxon_lineage where tax_id in ({#tax_ids_str})").to_hash
taxon_name_arr = TaxonLineage.connection.select_all("select tax_id, name from taxon_name where tax_id in ({#tax_ids_str})").to_hash
lineage_info = lineage_arr.group_by {|h| h[:tax_id]}
taxon_name_info = taxon_name_arr.group_by {|h| h[:tax_id]}
```

inside the loop
```
   category_taxid = lineage_info[my_tax_id] && lineage_info[my_tax_id][0]["superkingdom_taxid"]
   genus_taxid = lineage_info[my_tax_id] && lineage_info[my_tax_id][0]["genus_taxid"]
   taxon_name = taxon_name_info[my_tax_id][0]["name"]
```

still not the best way

best to pull this data at the start when you are doing the earlier SQL query using a join

boris-dimitrov: Above in the bigger SQL query
```
TaxonCount.connection.select_all("select tax_id, pipeline_output_id, tax_level, count, name, count_type from taxon_counts where pipeline_output_id = #{pipeline_output_id}").to_hash
```
you may pull the additional fields you need with a JOIN, like so:
```
TaxonCount.connection.select_all("select tax_id, pipeline_output_id, tax_level, count, name, count_type, `taxon_lineage`.`genus_tax_id` as `genus_tax_id` from taxon_counts INNER JOIN `taxon_lineage`  ON `taxon_lineage`.`tax_id` = `taxon_counts`.`tax_id` where pipeline_output_id = #{pipeline_output_id}").to_hash
```

boris-dimitrov: There is also a RAILS-native way of doing this.

app/lib/report_helper.rb
```diff
@@ -32,29 +32,75 @@ def compute_taxon_zscores(report)
...
+      tax_ids_str += taxon[:tax_id].to_s
+      tax_ids_str += ','
+    end
+    tax_ids_str = tax_ids_str.chomp(',')
+    lineage_arr = TaxonLineage.connection.select_all("SELECT taxid, superkingdom_taxid,  genus_taxid FROM taxon_lineages WHERE taxid IN (#{tax_ids_str})").to_hash
+    lineage_info = lineage_arr.group_by { |h| h['taxid'] }
+
+    category_taxids = ''
+    lineage_info.each do |taxid|
+      category_taxids += taxid[1][0]['superkingdom_taxid'].to_s
```
boris-dimitrov: I think this code doesn't do what you intend.

Don't you mean
```
  lineage_info.each do |taxid, taxons|
    category_taxids += taxons[0]['superkindgdom_tax_id'].to_s
```
and may be only if the superkingdom_tax_id is not null?

app/lib/report_helper.rb
```diff
@@ -32,29 +32,75 @@ def compute_taxon_zscores(report)
...
+
+    nt_genus = data[[genus_level, 'NT']] || []
+    nr_genus = data[[genus_level, 'NT']] || []
+    genus_nt_nr = nt_genus.concat nr_genus
+
     # filter and sort the nt_scores
+    htc = highest_tax_counts(nt, nr)
     rp = resolve_params(params, htc)
-    nt.keep_if do |h|
+    nt_nr = nt.concat nr
```
boris-dimitrov: Isn't `nt_nr` basically the same data as `taxon_zscores`?

app/lib/report_helper.rb
```diff
@@ -32,29 +32,75 @@ def compute_taxon_zscores(report)
...
+    data = taxon_zscores.group_by { |h| [h[:tax_level], h[:hit_type]] }
+    nt = data[[species_level, 'NT']] || []
+    nr = data[[species_level, 'NR']] || []
+
+    nt_genus = data[[genus_level, 'NT']] || []
+    nr_genus = data[[genus_level, 'NT']] || []
+    genus_nt_nr = nt_genus.concat nr_genus
+
     # filter and sort the nt_scores
+    htc = highest_tax_counts(nt, nr)
```
boris-dimitrov: The min/max are only computed for the species scores.   Is that the design?

app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -14,7 +14,7 @@ class PipelineSampleReport extends React.Component {
         ? current_sort.sort_query  : 'sort_by=highest_zscore'
     };
     this.applySort = this.applySort.bind(this);
-
+    console.log(props)
```
boris-dimitrov: Do you mean to commit this?

boris-dimitrov:
The API code changes are not production quality and contain bugs, but since the frontend changes look fine, I think we should get this in asap, and then one of us can fix the API code quickly in a follow up request.

PR #168, opened 2017-10-11
Susan/useradminui

sadelokiki:
Create UI Layout and validations for Login


config/routes.rb
```diff
@@ -4,7 +4,14 @@
   resources :backgrounds
   resources :reports
   resources :pipeline_outputs, only: [:index, :show]
-  devise_for :users
+  devise_for :users, controllers: {
+    sessions: 'sessions',
+    registrations: 'registrations'
+  }
+  devise_scope :user do
```
yunfangjuan: why do we need this? 
why not just use /users/sign_in, /users/sign_out?

app/assets/javascripts/components/login.jsx
```diff
@@ -0,0 +1,130 @@
...
+    this.setState.isChecked = !this.setState.isChecked;  
+  }
+
+  renderLogin() {
+    return (
+        <div className="form-wrapper">
+          <div className="row">
+            <form ref="form" className="new_user" id="new_user" onSubmit={ this.handleSubmit }>
+              <div className="row title">
+                <p className="col s6 verify">Not a user? Sign up</p>
```
yunfangjuan: We don't allow people to sign up directly so remove this line for now.

app/assets/javascripts/components/login.jsx
```diff
@@ -0,0 +1,130 @@
...
+                  <i className="fa fa-key" aria-hidden="true"></i>
+                  <input ref= "password" type="password" className="" onFocus={ this.clearError }   />
+                  <label htmlFor="user_password">Password</label>
+                </div>
+                <div className="">  
+                  <input ref="remember_me" type="checkbox" name="switch" className="filled-in" id="remember_me" onChange={ this.toggleCheckBox } value={this.setState.isChecked ? 1 : 0} />
+                  <label htmlFor="remember_me">Remember me</label>
+                </div>
+              </div>
+                <div className="forgot-password"><span onClick={ this.gotToForgotPassword }>Forgot your password?</span><br /></div>
```
yunfangjuan: the forget password flow seems a bit odd. probably not for this pr, but we should fix it soon.

config/routes.rb
sadelokiki: I figured it is easier go to login, instead of the longer route? @yunfangjuan. It renders the same view.

yunfangjuan: but it's redudant and makes code management more complicated.

PR #170, opened 2017-10-11
fix home directory and username for GSNAPL instance

cdebourcy:


aws_batch/pipeline.py
```diff
@@ -886,28 +886,30 @@ def run_gsnapl_remotely(sample, input_fa_1, input_fa_2,
     execute_command("aws s3 cp %s %s/" % (gsnap_ssh_key_s3_path, REF_DIR))
     key_path = REF_DIR +'/' + key_name
     execute_command("chmod 400 %s" % key_path)
-    commands =  "mkdir -p /home/ec2-user/batch-pipeline-workdir/%s;" % sample
-    commands += "aws s3 cp %s/%s /home/ec2-user/batch-pipeline-workdir/%s/ ; " % \
-                 (sample_s3_output_path, input_fa_1, sample)
-    commands += "aws s3 cp %s/%s /home/ec2-user/batch-pipeline-workdir/%s/ ; " % \
-                 (sample_s3_output_path, input_fa_2, sample)
-    commands += " ".join(['/home/ec2-user/bin/gsnapl',
+    remote_home_dir = "/home/ubuntu"
```
yunfangjuan: why don't you just do 
remote_work_dir = ""/home/ubuntu/batch-pipeline-workdir/%s" % sample  ?

cdebourcy: need it in several different paths, see below

cdebourcy: need it in several different paths, see below

yunfangjuan: can probably do both 
remote_home_dir 
remote_work_dir 
to make the code more succint though.

aws_batch/pipeline.py
```diff
@@ -16,7 +16,7 @@
 
 INPUT_BUCKET = 's3://czbiohub-infectious-disease/UGANDA'
 OUTPUT_BUCKET = 's3://yunfang-workdir/id-uganda'
-KEY_S3_PATH = 's3://cdebourcy-test/cdebourcy_7-19-17.pem'
+KEY_S3_PATH = 's3://cdebourcy-test/idseq-alpha.pem'
```
yunfangjuan: since you are changing it, why not put is into a s3 directory which is not cdebourcy-test?

PR #173, opened 2017-10-12
container name changed from chanzuckerberg/idseq-web to ecs-idseq-web

boris-dimitrov:


ryanking:
Note that this was looking for the image name before, and now the container name. Should be fine, just FYI.

PR #174, opened 2017-10-12
Susan/useradminui

sadelokiki:


app/assets/javascripts/components/newUser.jsx
```diff
@@ -0,0 +1,53 @@
...
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.clearError = this.clearError.bind(this)
+    this.state = {
+      isChecked: false,
+      showFailed: false,
+      errorMessage: ''
+    }
+  }
+
+  renderSignUp() {
```
yunfangjuan: Where are you calling this? Also, we don't let users sign up now. So it's more of a pure create user.

sadelokiki: It is for creating users. Will take it off in this PR and include it in the next for modularity.

app/assets/javascripts/components/Header.jsx
```diff
@@ -1,19 +1,29 @@
...
     return (
       <div>
         <div className="site-header">
           {/* Dropdown menu */}
           <ul id="dropdown1" className="dropdown-content">
-            <li><a href="#!">New Project</a></li>
-            <li><a href="#!">Account</a></li>
-            <li className="divider"></li>
-            <li><a href="#!">Sign out</a></li>
+            {<li><a href="#!">New Project</a></li>}
```
yunfangjuan: can probably remove it for now or do 'New Sample' instead.

PR #175, opened 2017-10-12
Add host genome

yunfangjuan:
add host genome to the samples page and prepare the wiring for default report generation.

### DIFF
--- a/db/migrate/20171011225553_create_host_genomes.rb
+++ b/db/migrate/20171011225553_create_host_genomes.rb
@@ -1,0 +1,14 @@
+class CreateHostGenomes < ActiveRecord::Migration[5.1]
+  def change
+    create_table :host_genomes do |t|
+      t.string :name, null:false, unique: true
+      t.text :s3_star_index_path
+      t.text :s3_bowtie2_index_path
+      t.bigint :default_background_id
+      t.timestamps
+    end
+
+    add_column :samples, :host_genome_id, :bigint
+
+  end
+end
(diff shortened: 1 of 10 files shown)


db/migrate/20171011225553_create_host_genomes.rb
```diff
@@ -0,0 +1,14 @@
+class CreateHostGenomes < ActiveRecord::Migration[5.1]
+  def change
+    create_table :host_genomes do |t|
+      t.string :name, null:false, unique: true
+      t.text :s3_star_index_path
+      t.text :s3_bowtie2_index_path
+      t.bigint :default_background_id
```
boris-dimitrov: is this a circular dependency?

  -- in order to create background, you need several runs

  -- for each run, you need a host genome

  -- for a host genome, you need a background?

lib/tasks/host_genomes.rake
```diff
@@ -0,0 +1,10 @@
+require 'open3'
```
boris-dimitrov: what runs this task?

db/migrate/20171011225553_create_host_genomes.rb
yunfangjuan: well that's why it's called default_background_id. 

For each host genome, there would be a background that consists of runs that are run based on the host genome, which has the indexes that areapecified in the beginning. 

Once those are done, a default could be created but it's not there in the beginning.

lib/tasks/host_genomes.rake
yunfangjuan: guess we don't need it here.

yunfangjuan:
Run 

docker-compose run web "rake create_host_genomes" 

To initialize host genomes, (human/mosquito)

PR #179, opened 2017-10-12
store background summaries

cdebourcy:


app/controllers/backgrounds_controller.rb
```diff
@@ -25,7 +25,18 @@ def edit
   # POST /backgrounds
   # POST /backgrounds.json
   def create
-    @background = Background.new(background_params)
+    @background = Background.create(background_params)
+
+    data = @background.summarize.map { |h| h.slice('tax_id', 'count_type', 'tax_level', 'name', :background_id, :created_at, :updated_at, :mean, :stdev) }
+    columns = data.first.keys
+    values_list = data.map do |hash|
+      hash.values.map do |value|
```
boris-dimitrov: Are the values here always in the same order for each of the hashes?

app/controllers/backgrounds_controller.rb
```diff
@@ -25,7 +25,18 @@ def edit
...
+        ActiveRecord::Base.connection.quote(value)
+      end
+    end
+    ActiveRecord::Base.connection.execute <<-SQL
+    INSERT INTO taxon_summaries (#{columns.join(",")}) VALUES #{values_list.map { |values| "(#{values.join(",")})" }.join(", ")}
+    SQL
```
boris-dimitrov: Beautiful.

db/schema.rb
```diff
@@ -248,5 +261,6 @@
   add_foreign_key "reports", "backgrounds"
   add_foreign_key "reports", "pipeline_outputs"
   add_foreign_key "taxon_counts", "pipeline_outputs"
+  add_foreign_key "taxon_summaries", "backgrounds"
```
boris-dimitrov: The taxon_zscores table is now unused, right?   No action suggested, just making sure I understand.

cdebourcy: No, taxon_summaries could really be called "background_taxon_summaries" -- it is used together with taxon_counts to generate taxon_zscores

cdebourcy:
to do: 
- add validity flag to background (did large taxon_summaries insert succeed?)
- check if hash value order is guaranteed

db/schema.rb
boris-dimitrov: Got it thanks.

app/controllers/backgrounds_controller.rb
cdebourcy: yes: https://stackoverflow.com/questions/31418673/is-order-of-a-ruby-hash-literal-guaranteed

db/schema.rb
yunfangjuan: should we call the table background_taxon_summaries? if not, we should at least document in in taxon_summary.rb it's exclusively for background.

app/assets/javascripts/components/PipelineSampleReads.jsx
```diff
@@ -91,9 +91,6 @@ class PipelineSampleReads extends React.Component {
         <a className="custom-button" href= { this.sampleInfo.sample_output_folder_url }>
           <i className="fa fa-cloud-download left"></i> GO TO RESULTS FOLDER
         </a>
-        <a className="custom-button" href= { '/reports/new?report[pipeline_output_id]=' + this.pipelineOutput.id }>
-          <i className="fa fa-file-text-o left"></i> GENERATE REPORT
```
yunfangjuan: please merge master!

app/controllers/backgrounds_controller.rb
```diff
@@ -69,6 +72,24 @@ def set_background
...
+        ActiveRecord::Base.connection.quote(value)
+      end
+    end
+    begin
+      ActiveRecord::Base.connection.execute <<-SQL
+      INSERT INTO taxon_summaries (#{columns.join(',')}) VALUES #{values_list.map { |values| "(#{values.join(',')})" }.join(', ')}
+      SQL
+      @background.status = Background::STATUS_SUCCESS
+    rescue
+      @background.status = Background::STATUS_FAILED
```
boris-dimitrov: what is the initial/default background status?

boris-dimitrov: does this no longer require chunking?

app/lib/report_helper.rb
```diff
@@ -20,7 +20,12 @@ def report_details(report)
   end
 
   def compute_taxon_zscores(report)
-    summary = report.background.summarize
+    bg = report.background
```
boris-dimitrov: You'll need to `pull --rebase origin master` and do a manual merge/resolve here.    Otherwise, as is, this pull request would revert some recent changes :)

boris-dimitrov: Hmmm, not exactly.   It's very puzzling why github is showing the diff against an older master.

boris-dimitrov:
Finally, we need to update all queries that fetch backgrounds to actually exclude backgrounds without SUCCESS status.    And we need a migration job that will decorate good existing backgrounds with a SUCCESS status.   This seems like adding some complexity, but not sure what the alternatives are, if background creation can fail ---  @yunfangjuan what is the RAILS typical approach in these cases?   We have to create a background id before we can insert background taxon summaries, but if those inserts fails, the background id remains, but we don't want to use it because it is incomplete.

db/schema.rb
cdebourcy: ok I'll add a comment

cdebourcy:
There are only 3 backgrounds in alpha right now so we can just hit the update button manually on alpha.idseq.net/backgrounds/1/edit etc. All new ones that get created will have status set natively. Also, there is no default background status. I thought it was quite neat that if status is FAILED for whatever reason, the user can still generate a report from the background by doing it on the fly without using the (non-existent) stored summary?

cdebourcy:
moved to store-bg-summary-2

PR #182, opened 2017-10-13
Boris/bugfixes 5

boris-dimitrov:


boris-dimitrov:
thanks for reviewing, Charles!

PR #185, opened 2017-10-13
Store bg summary 2

cdebourcy:
I re-did store-bg-summary due to the rebase issues and . Think this one should be good to go

### DIFF
--- a/app/controllers/backgrounds_controller.rb
+++ b/app/controllers/backgrounds_controller.rb
@@ -27,7 +27,7 @@ class BackgroundsController < ApplicationController
   def create
     ActiveRecord::Base.transaction do
       @background = Background.create(background_params)
-      store_summary
+      @background.store_summary
     end
 
     respond_to do |format|
@@ -45,7 +45,7 @@ class BackgroundsController < ApplicationController
   # PATCH/PUT /backgrounds/1.json
   def update
     ActiveRecord::Base.transaction do
-      store_summary
+      @background.store_summary
     end
 
     respond_to do |format|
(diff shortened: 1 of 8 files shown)


app/controllers/backgrounds_controller.rb
```diff
@@ -25,7 +25,10 @@ def edit
   # POST /backgrounds
   # POST /backgrounds.json
   def create
-    @background = Background.new(background_params)
+    @background = Background.create(background_params)
```
yunfangjuan: shouldn't this be inside the transaction if we want the whole thing to rollback if the creation failed?

cdebourcy: absolutely, thanks

app/controllers/backgrounds_controller.rb
```diff
@@ -25,7 +25,10 @@ def edit
   # POST /backgrounds
   # POST /backgrounds.json
   def create
-    @background = Background.new(background_params)
+    ActiveRecord::Base.transaction do
```
yunfangjuan: Great!!

app/controllers/backgrounds_controller.rb
```diff
@@ -69,6 +76,22 @@ def set_background
     @background = Background.find(params[:id])
   end
 
+  def store_summary
```
yunfangjuan: I think this function should really go into model/background.rb

cdebourcy:
After merging and closing this PR, go to alpha.idseq.net/backgrounds/{id}/edit for each existing background id and click "Update" to create the relevant taxon_summaries.

PR #186, opened 2017-10-13
Display all categories

cyrielo:


app/assets/javascripts/components/ReportFilter.jsx
```diff
@@ -157,18 +160,15 @@ class ReportFilter extends React.Component {
...
-                      <label htmlFor="fungi">Fungi</label>
-                    </p>
-                    <p>
-                      <input type="checkbox" className="filled-in" id="virus" defaultChecked="checked" />
-                      <label htmlFor="virus">Virus</label>
-                    </p>
+                    { this.all_categories.map((category, i) => {
+                      return (
+                        <p key={i}>
+                          <input type="checkbox" className="filled-in" id={category.name} defaultChecked="checked" />
```
yunfangjuan: i think the box shouldn't be checked by default. It should reflect what people selected.

app/lib/report_helper.rb
```diff
@@ -175,6 +185,20 @@ def resolve_params(params, data_ranges)
...
 
+  def all_categories
+    cat_ids = TaxonLineage.distinct.pluck(:superkingdom_taxid).join(', ')
+    taxon_name_arr = []
+    taxon_name_arr = TaxonName.connection.select_all("select taxid, name from taxon_names where taxid in (#{cat_ids})").to_hash unless cat_ids.empty?
+    taxon_name_arr
+  end
+
+  def all_genus
+    genus_taxids = TaxonLineage.distinct.pluck(:genus_taxid).join(', ')
```
yunfangjuan: Isn't this a very long list?

app/lib/report_helper.rb
```diff
@@ -193,7 +193,7 @@ def all_categories
   end
 
   def all_genus
-    genus_taxids = TaxonLineage.distinct.pluck(:genus_taxid).join(', ')
+    genus_taxids = TaxonLineage.distinct.limit(20).pluck(:genus_taxid).join(', ')
```
yunfangjuan: this seems a bit arbitrary. the right way to do it should be a remote typeahead. just getting the arbitrary 20 genus ids doesn't seem to be right.

app/assets/javascripts/components/PipelineSampleReads.jsx
```diff
@@ -8,6 +8,7 @@ class PipelineSampleReads extends React.Component {
     this.reportInfo =  (Object.keys(this.props.reportInfo).length > 0) ? this.props.reportInfo : null;
     this.jobStatistics = this.props.jobStatistics;
     this.gotoReport = this.gotoReport.bind(this);
+    console.log(props);
```
boris-dimitrov: do you mean to commit this logging?

app/lib/report_helper.rb
```diff
@@ -6,30 +6,39 @@ def external_report_info(report, view_level, params)
...
+      pipeline_output_id = report.pipeline_output.id
+      data = TaxonCount.connection.select_all("select tax_id, pipeline_output_id, tax_level, count, name, count_type from taxon_counts where pipeline_output_id = #{pipeline_output_id}").to_hash
+      data.each do |h|
+        h[:rpm] = compute_rpm(h["count"], total_reads)
+      end
+      data_and_background = (data + summary).group_by { |h| [h["tax_id"], h["tax_level"], h["name"], h["count_type"]] }.map { |_k, v| v.reduce(:merge) }.select { |h| h["count"] }
+      zscore_array = data_and_background.map { |h| { tax_id: h["tax_id"], tax_level: h["tax_level"], name: h["name"], rpm: h[:rpm], hit_type: h["count_type"], zscore: compute_zscore(h[:rpm], h[:mean], h[:stdev]) } }
+      zscore_array
+    else
+      []
```
boris-dimitrov: When you have a long function like that it's acceptable to put a `return [] unless report` at the top.  In this case that would be better than indenting the entire long function.

Note it is still considered poor form to return from the middle of a long function.

app/lib/report_helper.rb
```diff
@@ -175,6 +184,13 @@ def resolve_params(params, data_ranges)
     new_params
   end
 
+  def all_categories
+    cat_ids = TaxonLineage.distinct.pluck(:superkingdom_taxid).join(', ')
+    taxon_name_arr = []
+    taxon_name_arr = TaxonName.connection.select_all("select taxid, name from taxon_names where taxid in (#{cat_ids})").to_hash unless cat_ids.empty?
+    taxon_name_arr
```
boris-dimitrov: What are the SQL queries that you see in your log as a result of adding this code, and how long do they take to run?

Hey @yunfangjuan these are big static tables that are only going to change when we pull in new NCBI info.   I've been thinking reading these from SQL on each request isn't really great.   Caching either in the RAILS server memory (like APC at FB) or memcache may be good for this.  Does RAILS have a caching pattern based on version increment?   I imagine when we refresh that data from NCBI, we increment a version number in SQL;  when RAILS request want to load this data, they first check the version against what they last loaded;   this saves the big query.

yunfangjuan:
I just checked out this branch. I was not able to switch to see the results based on species. How do you do that right now?

PR #188, opened 2017-10-16
Charles/bug fix

cdebourcy:
Fixed a bug where in some cases species names were being displayed twice. Fixed bug in genus display.

### DIFF
--- a/app/lib/report_helper.rb
+++ b/app/lib/report_helper.rb
@@ -113,50 +113,50 @@ module ReportHelper
     unsortable = []
 
     tax_id_set.each do |tax_id|
-      linfo = lineage_info[tax_id] || [{}]
-      linfo = linfo.first
-
-      category_taxid = linfo["superkingdom_taxid"]
-      category_info = cat_name_info[category_taxid]
-      category_name = category_info ? category_info[0]['name'] : 'Other'
-
-      taxon_nts = tax_hit[[tax_id, 'NT']] || [nil]
-      taxon_nrs = tax_hit[[tax_id, 'NR']] || [nil]
-
-      if level[tax_id] == species_level
-        species_nts = taxon_nts
-        species_nrs = taxon_nrs
-        genus_taxid = linfo["genus_taxid"]
-        genus_nts = tax_hit[[genus_taxid, 'NT']] || [nil]
-        genus_nrs = tax_hit[[genus_taxid, 'NR']] || [nil]
-      else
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171013213906) do
+ActiveRecord::Schema.define(version: 20171016172945) do
```
boris-dimitrov: You probably shouldn't commit this file.

yunfangjuan: actually he has to because there's actual schema change.

app/lib/report_helper.rb
```diff
@@ -19,6 +19,16 @@ def report_details(report)
     }
   end
 
+  def view_level_name2int(view_level)
+    case view_level.downcase
```
yunfangjuan: should all the functions really be here or it should be in app/helpers ?

db/schema.rb
boris-dimitrov: hah good point

app/lib/report_helper.rb
boris-dimitrov: Should this specific function use TaxonCount::TAX_LEVEL_SPECIES and TaxonCount::TAX_LEVEL_GENUS constants?   I assume those are available without SQL queries?

app/lib/report_helper.rb
```diff
@@ -103,53 +119,55 @@ def taxonomy_details(report, params)
...
 
-      if details[:sort_key]
-        sortable.push(details)
-      else
-        unsortable.push(details)
+        if details[:sort_key]
+          sortable.push(details)
+        else
+          unsortable.push(details)
+        end
```
boris-dimitrov: Please move the filtering code from `taxon_zscores.keep_if` to down here.   We don't want the taxon_zscores to be filtered before this loop, because that would potentially throw away some of the genus taxons before they are accessed from species taxons here.

cdebourcy: Good point. I will leave as is so you can fix it in your upcoming PR as discussed.

app/lib/report_helper.rb
cdebourcy: Yes, I'll change to use TaxonCount::TAX_LEVEL_SPECIES etc

PR #190, opened 2017-10-16
Add user role

yunfangjuan:
add admin role


app/models/user.rb
```diff
@@ -5,4 +5,13 @@ class User < ApplicationRecord
   devise :database_authenticatable, :recoverable,
          :rememberable, :trackable, :validatable
   has_and_belongs_to_many :projects
+  ROLE_ADMIN = 1
+
+  def as_json(_options = {})
+    super(methods: [:admin])
+  end
+
+  def admin
```
boris-dimitrov: Isn't ruby convention to end predicates with '?'

yunfangjuan: Cant because i am exporting it to json

yunfangjuan:
Test plan: 
- create a user with admin on/off
- edit a user with admin on/off
See user.admin change accordingly

PR #206, opened 2017-10-18
change number displays and fix bugs

cdebourcy:


app/assets/javascripts/helpers/strings.jsx
```diff
@@ -0,0 +1,7 @@
+export function numberWithCommas(x) {
```
cyrielo: Remove export: The helper function would get imported automatically, so no need for the `export` keyword

PR #208, opened 2017-10-18
Sampleupload

sadelokiki:
- Complete sample upload
- Automatically populate added project
- Make path copyable
- Submit or click on button to ADD project
- Include host genome field
- Format host genomes structure
- Enter default values for job queue and memory
- Show sample page on successful sample creation


app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -0,0 +1,278 @@
...
+    axios.post('/samples', {
+      sample: {
+        name: this.refs.name.value.trim(),
+        project_name: this.state.project.trim(),
+        input_files_attributes: [{source_type: 's3', source: this.refs.first_file_source.value.trim() },
+        {source_type: 's3', source: this.refs.second_file_source.value.trim() }],
+        s3_preload_result_path: this.refs.s3_preload_result_path.value.trim(),
+        job_queue: this.refs.job_queue.value.trim(),
+        memory: this.refs.memory.value.trim(),
+        status: 'created'
```
cdebourcy: I don't think you should specify ```status: 'created'``` here, that should be set when the post actually succeeds -- @yunfangjuan ?

yunfangjuan: you do need that actually. the status needs to be set to created to trigger all the following actions

yunfangjuan:
- the project part looks confusing. I think they should be in one line. 
- the s3 preload path got truncated
- the host genome needs to be added
- the queue should have the default value: 'aegea_batch' 
- the memory should have default value 64000

app/assets/javascripts/components/sampleUpload.jsx
cdebourcy: ok thanks for clarifying

yunfangjuan:
- After submitting it redirects to home page. It should redirect to the sample page instead. 
-  (bug), when you don't select a project and type in the project name, the sample submit fails. Supposedly, it would create a new project and submits the sample. 
- If some select and project and type in the new project name, it only uses the selected project. It's confusing. I think if someone type in a project name, once they are done, their should be an ajax response to add that into the project list and select that new project automatically.

app/assets/javascripts/components/createUser.jsx
```diff
@@ -1,7 +1,7 @@
 class CreateUser extends React.Component {
   constructor(props, context) {
     super(props, context);
-    this.csrf = this.props.csrf;
+    this.csrf = props.csrf;
```
yunfangjuan: why is this file in this diff?

app/assets/javascripts/components/createUser.jsx
```diff
@@ -89,7 +90,7 @@ class CreateUser extends React.Component {
 
   renderCreateUser() {
     return (
-      <div className="form-wrapper">
+      <div className="user-form">
           <div className="row">
```
yunfangjuan: you will need to add the admin flag in the create user form here.

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -0,0 +1,336 @@
...
+    .catch(function (error) {
+     that.setState({
+      invalid: true,
+       errorMessage: 'Failed to create sample'
+     })
+    });
+  }
+
+  filePathValid(str) {
+    var regexPrefix = /^s3:\/\//;
```
yunfangjuan: can you trim the white space on both end of the input. It's a common mistake people make.

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -0,0 +1,550 @@
...
+      hostName: this.hostName,
+      hostId: this.hostId,
+      invalid: false,
+      errorMessage: '',
+      success: false,
+      successMessage: '',
+      project: 'Select a Project',
+      projectId: null,
+      job_queue: 'aegea_batch',
+      memory: 64000,
```
yunfangjuan: I think the reason why the memory is not set correctly is that this field should be called sample_memory rather than memory.

sadelokiki: @yunfangjuan, the sample memory is pre-filled if there is sample_memory data, if there isn't, the field is empty and open to edit.

sadelokiki: @yunfangjuan, You could check a sample that contains sample_memory data, u'll see that the field is pre-filled

yunfangjuan: you can try to create a new sample and see the data in the rails console, the sample_memory is not set there although it was set to 64000

PR #212, opened 2017-10-19
Yf/genus aggregation

yunfangjuan:
doing genus aggregation on the db load

Test Plan
pr = PipelineRun.find(31)
pr.pipeline_output.destroy!
pr.load_results_from_s3 

and check results!!

### DIFF
--- a/app/models/pipeline_output.rb
+++ b/app/models/pipeline_output.rb
@@ -19,4 +19,20 @@ class PipelineOutput < ApplicationRecord
 
     end
   end
+
+  def generate_aggregate_counts(tax_level_name)
+    current_date = Time.now.strftime("%Y-%m-%d")
+    tax_level_id = TaxonCount::NAME_2_LEVEL[tax_level_name]
+    TaxonCount.connection.execute(
+      "INSERT INTO taxon_counts(pipeline_output_id, tax_id, name,
+                                tax_level, count_type, count, created_at, updated_at)
+       SELECT #{id}, taxon_lineages.#{tax_level_name}_taxid, taxon_lineages.#{tax_level_name}_name,
+              #{tax_level_id}, taxon_counts.count_type,
+              sum(taxon_counts.count), '#{current_date}', '#{current_date}'
+       FROM  taxon_lineages, taxon_counts
+       WHERE taxon_lineages.taxid = taxon_counts.tax_id AND
+             taxon_counts.pipeline_output_id = #{id} AND
+             taxon_counts.tax_level = #{TaxonCount::TAX_LEVEL_SPECIES}
+      GROUP BY 1,2,3,4,5")
+  end
 end
--- a/app/models/pipeline_output.rb
+++ b/app/models/pipeline_output.rb
@@ -21,7 +21,7 @@ class PipelineOutput < ApplicationRecord
   end
 
   def generate_aggregate_counts(tax_level_name)
-    current_date = Time.now.strftime("%Y-%m-%d")
+    current_date = Time.zone.now.strftime("%Y-%m-%d")
     tax_level_id = TaxonCount::NAME_2_LEVEL[tax_level_name]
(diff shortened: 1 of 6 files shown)


app/models/taxon_count.rb
```diff
@@ -2,8 +2,12 @@ class TaxonCount < ApplicationRecord
   belongs_to :pipeline_output
   TAX_LEVEL_SPECIES = 1
   TAX_LEVEL_GENUS = 2
+  TAX_LEVEL_FAMILY = 3
   COUNT_TYPE_NT = 'NT'.freeze
   COUNT_TYPE_NR = 'NR'.freeze
+  NAME_2_LEVEL = { 'specifies' => TAX_LEVEL_SPECIES,
```
cdebourcy: 'species' instead of 'specifies'

boris-dimitrov:
cool, i want to read more carefully when i get to a big screen, particularly as to what happens for species that have no genus

yunfangjuan:
@boris-dimitrov in that case, no genus counts for it.

boris-dimitrov:
I think it would be useful to create a pseudo-genus for those taxons, so that we can get count and z score for all of them as a category, on the report.   The reason this will be useful is that we will have expand/collapse function showing species under their genuses, with sorting and filtering working on the genus metrics.  Sort of an officially supported “Other” category could be useful.  Don’t we currently assign -999 genus id for those guys?

cdebourcy:
Agree with Boris and yes when we generate the taxon lineage table we do fill all missing values with -999.

yunfangjuan:
I think anything map to other we actually dont show on report.

app/models/pipeline_output.rb
```diff
@@ -19,4 +19,21 @@ def generate_report
...
+       SELECT #{id}, taxon_lineages.#{tax_level_name}_taxid, taxon_lineages.#{tax_level_name}_name,
+              #{tax_level_id}, taxon_counts.count_type,
+              sum(taxon_counts.count), '#{current_date}', '#{current_date}'
+       FROM  taxon_lineages, taxon_counts
+       WHERE taxon_lineages.taxid = taxon_counts.tax_id AND
+             taxon_counts.pipeline_output_id = #{id} AND
+             taxon_counts.tax_level = #{TaxonCount::TAX_LEVEL_SPECIES}
+      GROUP BY 1,2,3,4,5"
+    )
+  end
```
boris-dimitrov: How does this deal with reads that map to a genus without mapping to any particular species?

app/models/pipeline_run.rb
```diff
@@ -77,11 +77,24 @@ def load_results_from_s3
...
+
+    pipeline_output_dict['taxon_counts_attributes'] = taxon_counts_attributes_filtered
     pipeline_output_dict['job_stats_attributes'] = stats_array
     po = PipelineOutput.new(pipeline_output_dict)
     po.sample = sample
     po.pipeline_run = self
     po.save
+    # aggregate the data at genus level
+    po.generate_aggregate_counts('genus')
+
```
boris-dimitrov: If the RAILS process crashes, are these rolled back?

app/models/pipeline_output.rb
cdebourcy: I don't think our pipeline currently allows for such a case, but it should in the future

boris-dimitrov: Ok, then all I request is a small comment here to that effect.

yunfangjuan: I think it already does that for most part because we do have all the taxon ids and some of them do map to 999. I will add a to-do and make the change after we finalize the taxon_leages not found code.

app/models/pipeline_run.rb
yunfangjuan: good point. i will put it in a transaction.

PR #215, opened 2017-10-19
Charles/bg unique index

cdebourcy:


app/models/background.rb
```diff
@@ -37,7 +37,7 @@ def store_summary
         end
       end
       ActiveRecord::Base.connection.execute <<-SQL
-      INSERT INTO taxon_summaries (#{columns.join(',')}) VALUES #{values_list.map { |values| "(#{values.join(',')})" }.join(', ')}
+      REPLACE INTO taxon_summaries (#{columns.join(',')}) VALUES #{values_list.map { |values| "(#{values.join(',')})" }.join(', ')}
```
yunfangjuan: this can have a potential problem when you remove some samples from the background. Old data will hang around. I think doing a delete and insert is cleaner. You can wrap a transaction around it.

cdebourcy: Ok done. Transaction is wrapped around calls to store_summary so I didn't put one inside the function.

PR #217, opened 2017-10-19
fix bug that was dropping rows with zero counts

boris-dimitrov:


app/lib/report_helper.rb
```diff
@@ -251,18 +258,20 @@ def resolve_params(params, view_level_str, data_ranges)
     new_params
   end
 
-  def highest_tax_counts(taxon_zscores, view_level_str)
+  def highest_tax_counts(taxon_counts_2d, view_level_str)
     # compute min/max extents for 8 columns
     bounds = {}
```
yunfangjuan: where is view_level_str used in this function?

boris-dimitrov:
```
      if level_type_metric.split("_").length < 3
        level_type_metric = "#{view_level_str}_#{level_type_metric}".to_sym
      end
```
Needless to say this is getting simplified very soon.

PR #218, opened 2017-10-19
move delete to correct place

cdebourcy:


app/models/background.rb
```diff
@@ -27,6 +27,9 @@ def summarize
   end
 
   def store_summary
+    ActiveRecord::Base.connection.execute <<-SQL
```
yunfangjuan: wrap this around ActiveRecord::Base.transaction ?

PR #223, opened 2017-10-20
Yf/category filter

yunfangjuan:
Test Plan: 
go to 
localhost:3000/samples/4 
click the filters and make sure it filters the right thing.


app/lib/report_helper.rb
```diff
@@ -79,7 +82,13 @@ def get_raw_taxon_counts(report)
     ").where(
       pipeline_output_id: pipeline_output_id,
       count_type: %w[NT NR]
-    ).to_a.map(&:attributes)
+    )
+
+    if categories.present?
+      query_results = query_results.where("taxon_lineages.superkingdom_taxid in (#{categories})")
```
boris-dimitrov: Rows for taxids that have no entry in the taxon_lineages table will be killed by this where clause.   Move it to the left outer join ON clause so that we still get them?   There should be no such rows right now, but I am becoming weary of the (lack of) cleanliness in our data.  Who knows when we'll have some taxon ids added to some DB but not to taxon_lineages...

yunfangjuan: well but if you are filtering for category. They won't have any category anyway so they won't be shown.

yunfangjuan:
I think we should totally filter by query to save some network bandwith. esp for infinite scroll in the future. I don't thin fetching the whole results and filter is the right approach for infinite scroll.

PR #224, opened 2017-10-20
Charles/alignment scores

cdebourcy:
Test plan:
- in rails console: Sample.find(1).update(status: 'need_rerun')
- when sample has finished processing, average alignment quality scores for each taxid should show up in report


app/views/samples/_samples.html.erb
```diff
@@ -34,9 +34,7 @@
             <% if pipeline_run %>
               <%= pipeline_run.job_status %>
               &nbsp;
-              <% if pipeline_run.job_status == 'FAILED' %>
-                <%= button_to 'rerun', kickoff_pipeline_sample_path(sample), method: :put %>
-              <% end %>
+              <%= button_to 'rerun', kickoff_pipeline_sample_path(sample), method: :put %>
```
cdebourcy: would love to be able to rerun any sample -- so we can easily generate these additional scores for the existing samples

app/lib/report_helper.rb
```diff
@@ -61,6 +61,9 @@ def get_raw_taxon_counts(report)
       taxon_counts.count               AS  count,
       (count / #{total_reads}.0
         * 1000000.0)                   AS  rpm,
+      taxon_counts.percent_identity    AS  percent_identity,
```
boris-dimitrov: What are the default values for these columns?   If they can be NULL, then I'd prefer suitable defaults be set here in the query --- unless it can be done in the DB schema.   In the query you can use IF syntax like for IF(stddev IS NULL, ...) below.

cdebourcy: If they are not set, then no value should be shown at all (as is currently the case for all the samples that haven't been reprocessed with the modified pipeline). There is really no suitable default value for the case where you have a hit but no percent-identity/alignment-length/e-value

yunfangjuan: it should be fine. NULL shouldn't really happen but if it does, showing blank should be ok. 
Maybe add a TODO: comment if we plan to filter on this field in the future.

cdebourcy: Got it, we'll need to set defaults when we implement filtering on these fields. I've added a "to do" comment.

PR #228, opened 2017-10-20
further simply report gen

boris-dimitrov:
Note:   This is far from ready, so not intended to be merged yet;  Just want to keep @yunfangjuan and @cdebourcy current with my work and gather feedback earlier rather than later.


boris-dimitrov:
A more advanced version of this code change has already been merged into master.

PR #237, opened 2017-10-24
Userpagefixes

sadelokiki:
- Import rails error responses to React views
- Go to users page after creating or updating a user

### DIFF
--- a/app/assets/javascripts/components/createUser.jsx
+++ b/app/assets/javascripts/components/createUser.jsx
@@ -216,11 +216,7 @@ class CreateUser extends React.Component {
                 <i className="fa fa-success"></i>
                  <span>{this.state.successMessage}</span>
                 </div> : null }
-                <div className={this.state.showFailed ? 'error-info' : ''} >{ this.state.serverErrors.length && this.state.showFailed ? this.state.serverErrors.map((err, i) => {
-                  return <p className="error center-align" key={i}> {err} </p>
-                    }) : this.state.showFailed ? 
-                  <span>{this.state.errorMessage}</span>
-              : null }</div>
+                <div className={this.state.showFailed ? 'error-info' : ''} >{this.displayError(this.state.showFailed, this.state.serverErrors, this.state.errorMessage)}</div>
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/components/createUser.jsx
```diff
@@ -217,10 +216,11 @@ class CreateUser extends React.Component {
                 <i className="fa fa-success"></i>
                  <span>{this.state.successMessage}</span>
                 </div> : null }
-              { this.state.showFailed ? <div className="error-info" >
-                  <i className="fa fa-error"></i>
+                <div className={this.state.showFailed ? 'error-info' : ''} >{ this.state.serverErrors.length && this.state.showFailed ? this.state.serverErrors.map((err, i) => {
```
yunfangjuan: For readability reasons, I will do 
{this.state.showFailed ? <div className="error-info" > {this.serverErrors.length? this.state.serverErrors.map((err, i) => { return <p className="error center-align" key={i}> {err} </p> }): <span>{this.state.errorMessage}</span></div> : null }

app/assets/javascripts/components/createUser.jsx
```diff
@@ -267,10 +267,11 @@ class CreateUser extends React.Component {
                 <i className="fa fa-success"></i>
                  <span>{this.state.successMessage}</span>
                 </div> : null }
-              { this.state.showFailed ? <div className="error-info" >
-                  <i className="fa fa-error"></i>
+              <div className={this.state.showFailed ? 'error-info' : ''} >{ this.state.serverErrors.length && this.state.showFailed ? this.state.serverErrors.map((err, i) => {
```
yunfangjuan: same as above. maybe you should make this code block into a function.

app/assets/javascripts/components/createUser.jsx
```diff
@@ -255,9 +251,19 @@ class CreateUser extends React.Component {
     )
   }
 
+  displayError(failedStatus, serverError, formattedError) {
```
yunfangjuan: great!

PR #238, opened 2017-10-24
Updated samples page

sadelokiki:
- Fix valid path bug

### DIFF
--- a/app/assets/javascripts/components/sampleUpload.jsx
+++ b/app/assets/javascripts/components/sampleUpload.jsx
@@ -217,7 +217,7 @@ class SampleUpload extends React.Component {
 
   filePathValid(str) {
     var regexPrefix = /^s3:\/\//;
-    var regexSuffix = /(\A[^\s\/]+\.fastq.gz)/igm;
+    var regexSuffix = /([^\s\/]+\.fastq.gz)/igm;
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -217,7 +217,7 @@ class SampleUpload extends React.Component {
 
   filePathValid(str) {
     var regexPrefix = /^s3:\/\//;
```
yunfangjuan: i think you can allow the white space in the beginning because the backend automatically trims it.

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -217,7 +217,7 @@ class SampleUpload extends React.Component {
...
-    var regexSuffix = /(\A[^\s\/]+\.fastq.gz)/igm;
+    var regexSuffix = /([^\s\/]+\.fastq.gz)/igm;
```
yunfangjuan: Also the bug Rebecca reported is really from the preload s3 path, you only do the check when the preload s3 path is not empty.

sadelokiki: hey @yunfangjuan, are u referring to the preload result path, cos that field is optional

PR #241, opened 2017-10-24
Genus search

yunfangjuan:
Test Plan: 

* go to http://localhost:3000/samples/6 and click around
* type and select things in the genus search box


yunfangjuan:
Yeah. The autocomplete also interfaces with the thresholding a little bit but we can have susan/cyril figure out the HTML element for us soon.

PR #242, opened 2017-10-24
Charles/taxon fasta 3

cdebourcy:


app/models/postprocess_run.rb
```diff
@@ -0,0 +1,98 @@
+# to kick off postprocessing for existing pipeline_outputs,
```
yunfangjuan: Instead of creating a new postprocess_run. I think we should keep pipeline_run and add a 'type' to it. We probably will have a few different pipeline_run types down the road. creating one model each time we do it is not scalable.

PR #243, opened 2017-10-25
Charles/taxon fasta 4

cdebourcy:
Test plan:
1. Rerun a sample
2. When rerun has finished, click on any of the numbers in the 'NT Species r' column and you'll see the corresponding reads!

### DIFF
--- a/app/assets/javascripts/components/PipelineSampleReport.jsx
+++ b/app/assets/javascripts/components/PipelineSampleReport.jsx
@@ -6,6 +6,7 @@ class PipelineSampleReport extends React.Component {
     this.taxonomy_details = props.taxonomy_details;
     this.all_categories = props.all_categories || [];
     this.checked_categories = props.checked_categories || props.all_categories
+    this.pipeline_output_id = props.report_details.pipeline_info.id;
 
     this.view_level = ReportFilter.getFilter('view_level') || 'species';
     this.highest_tax_counts = props.highest_tax_counts;
@@ -274,9 +275,7 @@ class PipelineSampleReport extends React.Component {
                         <td>
                           { (this.view_level==='species' && taxon.nt_ele) ?
                             <span className="link">
-                              <a href={`https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=${
-                                taxon.nt_ele.tax_id}`}>{ taxon.nt_ele.name }
-                              </a>
+                              <a href={`/sequence_locators/${this.pipeline_output_id}/NT/${taxon.nt_ele.tax_id}`}>{ taxon.nt_ele.name }</a>
                             </span> : ''
                           }
                         </td>
--- a/app/assets/javascripts/components/PipelineSampleReport.jsx
+++ b/app/assets/javascripts/components/PipelineSampleReport.jsx
@@ -275,7 +275,8 @@ class PipelineSampleReport extends React.Component {
                         <td>
                           { (this.view_level==='species' && taxon.nt_ele) ?
                             <span className="link">
-                              <a href={`/pipeline_outputs/${this.pipeline_output_id}/${taxon.nt_ele.tax_id}/fasta`}>{ taxon.nt_ele.name }</a>
+                              <a href={`https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=${
+                                 taxon.nt_ele.tax_id}`}>{ taxon.nt_ele.name }</a>
                             </span> : ''
                           }
                         </td>
@@ -303,7 +304,10 @@ class PipelineSampleReport extends React.Component {
                         </td>
 
                         <td>
-                          { (this.view_level==='species' && (taxon.nt_ele && taxon.nt_ele.hasOwnProperty('count')) ) ? numberWithCommas(taxon.nt_ele.count) : '' }
+                          { (this.view_level==='species' && (taxon.nt_ele && taxon.nt_ele.hasOwnProperty('count')) ) ? 
+                            <span className="link">
+                              <a href={`/pipeline_outputs/${this.pipeline_output_id}/${taxon.nt_ele.tax_id}/fasta`}>{ numberWithCommas(taxon.nt_ele.count) }</a>
+                            </span> : '' }
                         </td>
 
                         <td>
--- a/app/assets/javascripts/components/PipelineSampleReport.jsx
+++ b/app/assets/javascripts/components/PipelineSampleReport.jsx
@@ -275,7 +275,7 @@ class PipelineSampleReport extends React.Component {
                         <td>
                           { (this.view_level==='species' && taxon.nt_ele) ?
                             <span className="link">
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -304,7 +304,10 @@ class PipelineSampleReport extends React.Component {
                         </td>
 
                         <td>
-                          { (this.view_level==='species' && (taxon.nt_ele && taxon.nt_ele.hasOwnProperty('count')) ) ? numberWithCommas(taxon.nt_ele.count) : '' }
+                          { (this.view_level==='species' && (taxon.nt_ele && taxon.nt_ele.hasOwnProperty('count')) ) ? 
+                            <span className="link">
+                              <a href={`/pipeline_outputs/${this.pipeline_output_id}/${taxon.nt_ele.tax_id}/fasta`}>{ numberWithCommas(taxon.nt_ele.count) }</a>
+                            </span> : '' }
```
boris-dimitrov: I hope I don't lose this detail in my refactor of this page :)   Watch out when you review my diff tomorrow.

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -1,2 +1,12 @@
 module PipelineOutputsHelper
+  Client = Aws::S3::Client.new
+
+  def get_taxid_fasta(pipeline_output, taxid)
+    uri_parts = pipeline_output.sample.sorted_taxid_annotated_fasta_s3_path.split("/", 4)
+    bucket = uri_parts[2]
+    key = uri_parts[3]
+    taxon_location = pipeline_output.taxon_byteranges.find_by(taxid: taxid)
+    resp = Client.get_object(bucket: bucket, key: key, range: "bytes=#{taxon_location.first_byte}-#{taxon_location.last_byte}")
```
boris-dimitrov: Cool.

app/models/pipeline_run.rb
```diff
@@ -65,15 +66,18 @@ def local_json_path
   end
 
   def load_results_from_s3
-    return if pipeline_output
+    return if pipeline_output && !pipeline_output.taxon_byteranges.empty?
```
boris-dimitrov: Do you have an extra `!` here?   If the byte ranges have not been computed, don't you want to compute them?

aws_batch/postprocess.py
```diff
@@ -0,0 +1,244 @@
...
+def return_merged_dict(dict1, dict2):
+    result = dict1.copy()
+    result.update(dict2)
+    return result
+
+# processing functions
+def accession2taxid(read_id, accession2taxid_dict, hit_type, lineage_map):
+    accid_short = ((read_id.split(hit_type+':'))[1].split(":")[0]).split(".")[0]
+    taxid = accession2taxid_dict.get(accid_short, "NA")
+    species_taxid, genus_taxid, family_taxid = lineage_map.get(taxid, ("-1", "-2", "-3"))
```
boris-dimitrov: Isn't this supposed to be -100, -200, -300?

aws_batch/postprocess.py
```diff
@@ -0,0 +1,244 @@
...
+            taxon_sequence_locations.append({'taxid': taxid, 'first_byte': first_byte,
+                                             'last_byte': new_first_byte - 1})
+            taxid = new_taxid
+            first_byte = new_first_byte
+            new_first_byte = first_byte + len(sequence_name) + len(sequence_data)
+        else:
+            new_first_byte += len(sequence_name) + len(sequence_data)
+    f.close()
+    with open(output_json, 'wb') as f:
+       json.dump(taxon_sequence_locations, f)
```
boris-dimitrov: How big do these files tend to be again?

db/schema.rb
```diff
@@ -140,6 +140,16 @@
...
   end
 
+  create_table "taxon_byteranges", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
+    t.integer "taxid"
+    t.bigint "first_byte"
+    t.bigint "last_byte"
+    t.bigint "pipeline_output_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["pipeline_output_id"], name: "index_taxon_byteranges_on_pipeline_output_id"
```
boris-dimitrov: no index on (pipeline_output_id, taxid)?

boris-dimitrov:
How long does it take to fetch the reads?

Otherwise looks pretty good.

Fine with me if @yunfangjuan approves (since she is more familiar with the code being changed).

app/models/pipeline_run.rb
cdebourcy: I think the logic is correct. If taxon_byteranges already exist ("not empty"), the function returns early. If taxon_byteranges are empty, the function loads them in. BUT typing up this answer I realize I don't need to check taxon_byteranges at all; checking pipeline_output is enough because a new pipeline_run (even for a sample that has been run before) will always have no pipeline_output at first.

aws_batch/postprocess.py
cdebourcy: 10-100 KB

aws_batch/postprocess.py
```diff
@@ -0,0 +1,244 @@
...
+        new_taxid = get_taxid(sequence_name, taxid_field)
+        if new_taxid != taxid:
+            taxon_sequence_locations.append({'taxid': taxid, 'first_byte': first_byte,
+                                             'last_byte': new_first_byte - 1})
+            taxid = new_taxid
+            first_byte = new_first_byte
+            new_first_byte = first_byte + len(sequence_name) + len(sequence_data)
+        else:
+            new_first_byte += len(sequence_name) + len(sequence_data)
+    f.close()
```
yunfangjuan: you didn't handle the boundary condition when processing the very last taxid in the file.

cdebourcy: I thought I did. After reading the last record, the while loop will be executed one more time. In that execution, sequence_name will be '', new_taxid will be 'none', which is different from taxid, and so first_byte and last_byte of the last record will be written to the taxon_sequence_locations output. After that the loop won't be executed again. Where does it fail?

yunfangjuan: ok. Could you document the boundary condition when sequnce name is nil? Also new_first_byte is really unintuitive to read? Can you use last_byte instead?

aws_batch/postprocess.py
cdebourcy: yes

aws_batch/postprocess.py
cdebourcy: Ok I added that documentation. Changed `new_first_byte` to `end_byte`, hope that's ok :) Since `last_byte` = `end_byte - 1` in the output file

yunfangjuan: much better

cdebourcy:
@boris-dimitrov It was instantaneous on the example I tried :)

db/schema.rb
cdebourcy: Added index!

PR #247, opened 2017-10-26
Susan/newhomepage

sadelokiki:
- New HomePage layout
- Search samples functionality


app/models/sample.rb
```diff
@@ -41,6 +41,14 @@ def input_files_checks
     # TODO: for s3 input types, test permissions before saving, by making a HEAD request
   end
 
+  def self.search(search)
+    if search
+      where('name LIKE ?', "%#{search}%")
```
yunfangjuan: make sure sample.name is indexed on the db side

PR #254, opened 2017-10-26
Charles/taxon fasta nr

cdebourcy:
This extends taxon fasta displays to higher taxonomic levels and different hit types.


config/routes.rb
```diff
@@ -21,5 +21,5 @@
   mount Resque::Server.new, at: '/resque'
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
   root to: 'home#home'
-  get 'pipeline_outputs/:id/:taxid/fasta', to: 'pipeline_outputs#show_taxid_fasta'
+  get 'pipeline_outputs/:id/:taxid/fasta/:tax_level/:hit_type', to: 'pipeline_outputs#show_taxid_fasta'
```
yunfangjuan: maybe pipeline_outputs/:id/fasta/:taxid/:tax_level/:hit_type instead?

app/models/sample.rb
```diff
@@ -79,8 +85,14 @@ def sample_postprocess_s3_path
     "s3://#{SAMPLES_BUCKET_NAME}/#{sample_path}/postprocess"
   end
 
-  def sorted_taxid_annotated_fasta_s3_path
-    "#{sample_postprocess_s3_path}/#{SORTED_TAXID_ANNOTATED_FASTA}"
+  def s3_paths_for_taxon_byteranges
+    # by tax_level and hit_type
+    { 1 => { 'NT' => "#{sample_postprocess_s3_path}/#{SORTED_TAXID_ANNOTATED_FASTA}",
```
yunfangjuan: instead of 1, 2, 3 use TaxonCount::TAX_LEVEL_SPECIES, etc

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -1,12 +1,17 @@
...
+  def get_taxid_fasta(pipeline_output, taxid, tax_level, hit_type)
+    uri = pipeline_output.sample.s3_paths_for_taxon_byteranges[tax_level][hit_type]
+    uri_parts = uri.split("/", 4)
     bucket = uri_parts[2]
     key = uri_parts[3]
-    taxon_location = pipeline_output.taxon_byteranges.find_by(taxid: taxid)
-    return 'No records' if taxon_location.nil?
+    taxon_location = pipeline_output.taxon_byteranges.find_by(taxid: taxid, hit_type: hit_type)
+    if taxon_location.nil? && hit_type == 'NT' && tax_level == TaxonCount::TAX_LEVEL_SPECIES
+      taxon_location = pipeline_output.taxon_byteranges.find_by(taxid: taxid, hit_type: nil, tax_level: nil)
```
yunfangjuan: can you just run a migration instead of having the code the code here?

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -1,12 +1,17 @@
 module PipelineOutputsHelper
   Client = Aws::S3::Client.new
 
-  def get_taxid_fasta(pipeline_output, taxid)
-    uri_parts = pipeline_output.sample.sorted_taxid_annotated_fasta_s3_path.split("/", 4)
+  def get_taxid_fasta(pipeline_output, taxid, tax_level, hit_type)
+    uri = pipeline_output.sample.s3_paths_for_taxon_byteranges[tax_level][hit_type]
```
yunfangjuan: Maybe add some documentation of what the uri looks like?

app/controllers/pipeline_outputs_controller.rb
```diff
@@ -28,7 +28,13 @@ def show
   end
 
   def show_taxid_fasta
-    @taxid_fasta = get_taxid_fasta(@pipeline_output, params[:taxid])
+    if params[:hit_type] == "NT_or_NR"
+      nt_array = get_taxid_fasta(@pipeline_output, params[:taxid], params[:tax_level].to_i, 'NT').split("^>")
+      nr_array = get_taxid_fasta(@pipeline_output, params[:taxid], params[:tax_level].to_i, 'NR').split("^>")
+      @taxid_fasta = ((nt_array | nr_array) - ['']).join(">")
```
yunfangjuan: cool. but since you split by '>', the first read of each nt_ nr_ array might lose '>'

cdebourcy: fixed it!

app/helpers/pipeline_outputs_helper.rb
cdebourcy: Removed it. There are no taxon_byteranges on production yet, so no need to handle the old case.

app/helpers/pipeline_outputs_helper.rb
cdebourcy: ok

app/models/sample.rb
cdebourcy: ok

config/routes.rb

PR #257, opened 2017-10-27
Susan/formsand sample page updates

sadelokiki:
- Loader to show Sample processing

- Take off authentication fields from User create and edit forms

- Remove sample memory and job queue field for non-admins

- Rerun sample if Pipeline run failed

### DIFF
--- a/app/assets/javascripts/components/PipelineSampleReads.jsx
+++ b/app/assets/javascripts/components/PipelineSampleReads.jsx
@@ -25,7 +25,7 @@ class PipelineSampleReads extends React.Component {
     PipelineSampleReads.setTab('pipeline_display','reports');
   }
 
-  shouldProcessPipeline(status) {
+  pipelineInProgress(status) {
     if (status === null || this.state.rerun === true ) {
       return true;
     } else if ( status === 'ERROR' || status === 'CHECKED' || status === 'FAILED') {
@@ -120,7 +120,7 @@ class PipelineSampleReads extends React.Component {
         sample_id = {this.sampleId}
       />;
     } else {
-      d_report = <div className="center-align text-grey text-lighten-2 no-report">{ this.shouldProcessPipeline(this.pipelineStatus) ? <div>Processing Sample...<p><i className='fa fa-spinner fa-spin fa-3x'></i></p></div> : 
+      d_report = <div className="center-align text-grey text-lighten-2 no-report">{ this.pipelineInProgress(this.pipelineStatus) ? <div>Processing Sample...<p><i className='fa fa-spinner fa-spin fa-3x'></i></p></div> : 
         <div>
           <h6 className="failed"><i className="fa fa-frown-o"></i>  {this.state.failureText}  </h6>
           <p><a onClick={ this.rerunPipeline }className="custom-button small"><i className="fa fa-repeat left"></i>RERUN PIPELINE</a></p>
--- a/app/assets/javascripts/components/PipelineSampleReads.jsx
+++ b/app/assets/javascripts/components/PipelineSampleReads.jsx
@@ -2,19 +2,52 @@ class PipelineSampleReads extends React.Component {
 
   constructor(props) {
     super(props);
-    this.pipelineOutput = this.props.pipelineOutput;
-    this.sampleInfo = this.props.sampleInfo;
-    this.projectInfo = this.props.projectInfo;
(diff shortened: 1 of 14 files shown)


app/assets/javascripts/components/PipelineSampleReads.jsx
```diff
@@ -2,21 +2,54 @@ class PipelineSampleReads extends React.Component {
...
+    }
   }
 
   gotoReport() {
     $('ul.tabs').tabs('select_tab', 'reports');
     PipelineSampleReads.setTab('pipeline_display','reports');
   }
 
+  shouldProcessPipeline(status) {
+    if (status === null || this.state.rerun === true ) {
```
yunfangjuan: where do you set the status? the status would be null for a few minutes after the run started? 

The rerun button should only show when status === 'FAILED' or 'SUCCESS'

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -41,8 +42,9 @@ class SampleUpload extends React.Component {
       successMessage: '',
       project: 'Select a Project',
       projectId: null,
-      job_queue: null,
-      memory: null,
+      job_queue: 'aegea_batch',
```
yunfangjuan: You don't have to set the default values for job_queue and memory any more.  It would pick up from host_genomes if it's not set. so keep line 44 and 45

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -211,7 +215,8 @@ class SampleUpload extends React.Component {
     .catch(function (error) {
      that.setState({
       invalid: true,
-       errorMessage: 'Failed to update sample'
+      serverErrors: error.response.data,
+      errorMessage: 'Failed to update sample'
```
yunfangjuan: same above.

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -175,10 +178,11 @@ class SampleUpload extends React.Component {
...
     })
     .catch(function (error) {
-     that.setState({
-      invalid: true,
-       errorMessage: JSON.stringify(error.response.data)
-     })
+      that.setState({
+        invalid: true,
+        serverErrors: error.response.data,
+        errorMessage: 'Failed to create sample'
```
yunfangjuan: do you still need line 184?

yunfangjuan:
In general, looks good. See comments about rerun. I don't know where the rerun button is enabled but the logic needs some work. 

Also, after sample creation, it redirects to /samples, which is bad. It should go to / instead. /samples is kinda backdoor path for developers to debug. 

thanks,

app/assets/javascripts/components/PipelineSampleReads.jsx
sadelokiki: yea, this actually checks for status === null, which means the sample is either uploading, initalizing or in progress, it is in the else condition that the rerun button is rendered @yunfangjuan

app/assets/javascripts/components/sampleUpload.jsx
sadelokiki: No, taking it off

app/assets/javascripts/components/PipelineSampleReads.jsx
yunfangjuan: this function should be called "pipelineInProgress then

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -199,7 +203,7 @@ class SampleUpload extends React.Component {
       },
```
yunfangjuan: status: 'created' needs to be removed.

yunfangjuan: where is this.state.id set?

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -164,7 +164,7 @@ class SampleUpload extends React.Component {
         sample_memory: this.state.memory,
         host_genome_id: this.state.hostId,
         host_genome_name: this.state.hostName,
-        status: 'created'
+        status: ''
```
yunfangjuan: can you remove this status completely.

app/assets/javascripts/components/sampleUpload.jsx
sadelokiki: it is set from props

PR #260, opened 2017-10-27
handle data absence

cdebourcy:
Tested on my local dev. Temporary message to handle data absence for samples that have not been re-processed with the new pipeline. Eventually we'll change this to check for presence of data files in S3 and, if not, pass a flag to the front-end that prevents download links from appearing.

### DIFF
--- a/app/controllers/pipeline_outputs_controller.rb
+++ b/app/controllers/pipeline_outputs_controller.rb
@@ -32,6 +32,7 @@ class PipelineOutputsController < ApplicationController
       nt_array = get_taxid_fasta(@pipeline_output, params[:taxid], params[:tax_level].to_i, 'NT').split(">")
       nr_array = get_taxid_fasta(@pipeline_output, params[:taxid], params[:tax_level].to_i, 'NR').split(">")
       @taxid_fasta = ">" + ((nt_array | nr_array) - ['']).join(">")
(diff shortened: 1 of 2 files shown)


app/controllers/pipeline_outputs_controller.rb
```diff
@@ -32,6 +32,7 @@ def show_taxid_fasta
       nt_array = get_taxid_fasta(@pipeline_output, params[:taxid], params[:tax_level].to_i, 'NT').split(">")
       nr_array = get_taxid_fasta(@pipeline_output, params[:taxid], params[:tax_level].to_i, 'NR').split(">")
       @taxid_fasta = ">" + ((nt_array | nr_array) - ['']).join(">")
+      @taxid_fasta = "Coming soon" if @taxid_fasta == ">" # Temporary fix
```
yunfangjuan: Maybe you dont need this line if you do get ‘’. It just means it has no data? If it’s a key error, you will actually get coming soon.

cdebourcy: Ok, I moved 'coming soon' code to pipeline_outputs_helper in order not to have these temporary fixes in too many different places. @taxid_fasta should never be empty though, because if the taxid shows up in the report, it should have at least one fasta record (we're displaying the union of NT and NR now)

PR #261, opened 2017-10-28
refactor report page -- fewer columns and smarter sorting

boris-dimitrov:
for now category filter and genus filter are disabled,
and the links to reads are not being generated (but
should be very easy to add)

### DIFF
--- a/app/assets/javascripts/components/PipelineSampleReport.jsx
+++ b/app/assets/javascripts/components/PipelineSampleReport.jsx
@@ -3,358 +3,177 @@ class PipelineSampleReport extends React.Component {
   constructor(props) {
     super(props);
     this.report_details = props.report_details;
-    this.taxonomy_details = props.taxonomy_details;
-    this.all_categories = props.all_categories || [];
-    this.checked_categories = props.checked_categories || props.all_categories
-    this.pipeline_output_id = props.report_details.pipeline_info.id;
-    this.genus_info = props.genus_info
-    this.sample_id = props.sample_id
-
-    this.view_level = ReportFilter.getFilter('view_level') || 'species';
-    this.highest_tax_counts = props.highest_tax_counts;
-    this.defaultSortBy = 'highest_species_nt_zscore';
-    const current_sort = PipelineSampleReport.currentSort();
-    this.state = {
-      sort_query: current_sort.sort_query
-        ? current_sort.sort_query  : `sort_by=${this.defaultSortBy}`
-    };
-    this.applySort = this.applySort.bind(this);
-    this.columnSorting = this.columnSorting.bind(this);
+    this.real_length = props.taxonomy_details[0];
+    this.taxonomy_details = props.taxonomy_details[1];
+    this.all_categories = props.all_categories;
+    this.applyViewLevel = this.applyViewLevel.bind(this);
+    this.applyNewFilterThresholds = this.applyNewFilterThresholds.bind(this);
   }
 
-  static uppCaseFirst(name) {
-    return (name)? name.charAt(0).toUpperCase() + name.slice(1) : name;
+  refreshPage(overrides) {
+    new_params = Object.assign({}, this.props.report_page_params, overrides);
+    window.location = location.protocol + '//' + location.host + location.pathname + '?' + jQuery.param(new_params);
   }
 
-  static currentSort() {
-   const sort_by = ReportFilter.getFilter('sort_by');
-   let current_sort = {};
-   if(sort_by) {
-     current_sort = {
-       sort_query: `sort_by=${sort_by}`
-     }
-   }
-   return current_sort;
+  applyViewLevel(view_level) {
+    this.refreshPage({view_level});
   }
 
-  applySort(sort_query) {
-    this.setState({ sort_query });
-    const url = PipelineSampleReport.deleteUrlParam(window.location.href, 'sort_by');
-    window.location = (PipelineSampleReport.hasQuery(url))
-      ? `${url}&${sort_query}` : `${url}?${sort_query}`;
+  // applySort needs to be bound at time of use, not in constructor above
+  applySort(sort_by) {
+    this.refreshPage({sort_by});
   }
 
-  static deleteUrlParam(url, parameter) {
-    const queryString = url.split('?');
-    if (queryString.length >= 2) {
-      const prefix = encodeURIComponent(parameter)+'=';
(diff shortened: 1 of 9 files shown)


app/lib/report_helper.rb
```diff
@@ -38,312 +156,355 @@ def all_categories
...
+  def cleanup_names!(taxon_counts_2d)
+    # There are still taxons without names
+    missing_names = Set.new
+    taxon_counts_2d.each do |tax_id, tax_info|
+      name_from_lineages = tax_info.delete('name_from_lineages')
+      name_from_counts = tax_info.delete('name_from_counts')
+      if tax_id < 0
+        # Usually -1 means accession number did not resolve to species.
+        # TODO: Can we keep the accession numbers to show in these cases?
+        level_str = tax_info['tax_level'] == TaxonCount::TAX_LEVEL_SPECIES ? 'species' : 'genuses'
```
cdebourcy: prefer 'genera' over 'genuses'?

app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -3,358 +3,176 @@ class PipelineSampleReport extends React.Component {
...
-                    </th> : '' }
-
+                    <th>Taxonomy</th>
+                    { this.render_column_header('NT+NR', 'ZZRPM',  'nt_aggregatescore') }
+                    { this.render_column_header('NT', 'Z',   'nt_zscore') }
+                    { this.render_column_header('NT', 'rPM', 'nt_rpm')    }
+                    { this.render_column_header('NT', 'r',   'nt_r')      }
+                    { this.render_column_header('NT', '%id', 'nt_percentidentity')    }
+                    { this.render_column_header('NT', 'AL',   'nt_alignmentlength')    }
+                    { this.render_column_header('NT', 'Lg1/E',  'nt_neglogevalue')    }
```
cdebourcy: log(1/E) [less cryptic?]

app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -3,358 +3,176 @@ class PipelineSampleReport extends React.Component {
...
+        </span>
+      );
+    }
+    if (tax_info.tax_level == 1) {
+      // indent species rows
+      foo = <span>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;{foo}</span>
     } else {
-      return url;
+      // emphasize genus, soften category and species count
+      foo = <span><b>{foo}</b>&nbsp;&nbsp;&nbsp;&nbsp;<span style={{'color':'#A0A0A0'}}><i>({tax_info.species_count}&nbsp;{tax_info.category_name}&nbsp;species)</i></span></span>
```
cdebourcy: This display feels slightly confusing to me. Would prefer for category name and species-count info to be separated rather than intertwined. Maybe just move category name outside the parenthesis next to the genus name?

app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -3,358 +3,176 @@ class PipelineSampleReport extends React.Component {
...
-        }
-      }
-      url = queryString[0] + (pars.length > 0 ? '?' + pars.join('&') : '');
-      return url;
+  render_name(tax_info) {
+    foo = <i>{tax_info.name}</i>;
+    if (tax_info.tax_id > 0) {
+      foo = (
+        <span className="link">
+          <a href={`https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=${tax_info.tax_id}`}>
```
cdebourcy: <a href={`/pipeline_outputs/${this.pipeline_output_id}/fasta/tax_info.tax_level/${tax_info.tax_id}/NT_or_NR}`>
(or wherever you can get pipeline_output_id from)

app/lib/report_helper.rb
```diff
@@ -1,23 +1,141 @@
...
+    metric = parts[2]
+    return nil unless METRICS.include? metric
+    {
+      direction:    direction,
+      count_type:   count_type,
+      metric:       metric
+    }
+  end
+
+  def number_or_nil(string)
```
yunfangjuan: what's this function for? can you just do string.to_f?

app/lib/report_helper.rb
```diff
@@ -1,23 +1,141 @@
...
     data = {}
+    data[:report_page_params] = params
     data[:report_details] = report_details(report)
-    htc, td = taxonomy_details(report, params, view_level)
-    data[:highest_tax_counts] = htc
-    data[:taxonomy_details] = td
-    data[:view_level] = view_level
+    data[:taxonomy_details] = taxonomy_details(report, params)
     data[:all_categories] = all_categories
-    if params[:categories].present?
```
yunfangjuan: If it's not too hard, can you add the following lines back? (even commented this fine)

yunfangjuan:
I couldn't load the page when checking out your branch. see slack channel for more details.

app/lib/report_helper.rb
boris-dimitrov: thank you

app/assets/javascripts/components/PipelineSampleReport.jsx
boris-dimitrov: Interesting.   I thought `log` meant natural logarithm (base e), same as `ln`, whereas `lg` meant log base 10.   That definition, apparently, is not as widely accepted as I had thought.  http://mathworld.wolfram.com/Lg.html

app/assets/javascripts/components/PipelineSampleReport.jsx
boris-dimitrov: Let's try that.

app/assets/javascripts/components/PipelineSampleReport.jsx
boris-dimitrov: That's very helpful, thank you, Charles.

app/lib/report_helper.rb
boris-dimitrov: string.to_f raises an exception when its argument is malformed;   i wanted it to return nil in that case

my bigger goal here was to simplify the JS and move argument validation & default values to rails

app/lib/report_helper.rb
boris-dimitrov: Yeah as part of reinstating category filter.   Thanks.

PR #263, opened 2017-10-30
Clean up date filter

sadelokiki:


db/schema.rb
```diff
@@ -237,6 +237,15 @@
     t.index ["taxid"], name: "index_taxon_names_on_taxid", unique: true
   end
 
+  create_table "taxon_sequence_files", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
```
boris-dimitrov: Seems like this change wasn't meant to be part of the pull request.   Could you remove?

PR #265, opened 2017-10-31
Charles/csv download

cdebourcy:


app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -119,6 +125,7 @@ class PipelineSampleReport extends React.Component {
               <div className="col s10 reports-main ">
                 <table id="report-table" className='bordered report-table'>
                   <thead>
+                  { download_button }
```
boris-dimitrov: I didn't realize HTML permitted elements inside thead but not inside a tr.   Does it?

cdebourcy: It's working for me, so... ¯\_(ツ)_/¯
I expect that this button will be subject to future UI improvements though.

PR #266, opened 2017-10-31
Susan/hotfixesonsamplepage

sadelokiki:
Show number of samples found during search in real time
Swop qc and compression ratio values on the table
Refactor header styling


cdebourcy:
Could you make it so that the values in the "Total Reads | Final Reads | Percentage Reads | QC | Compression Ratio | Pipeline run status" columns move around correctly when User searches for a sample and the list of displayed samples changes? Right now, they do not change, i.e. they display the wrong values

PR #267, opened 2017-10-31
Yf/bulkupload

yunfangjuan:
bulk loader works but probably needs more error handling.

### DIFF
--- a/app/assets/javascripts/components/bulkUploadImport.jsx
+++ b/app/assets/javascripts/components/bulkUploadImport.jsx
@@ -30,7 +30,8 @@ class BulkUploadImport extends React.Component {
       projectId: null,
       selectedBulkPath: '',
       samples: [],
-      selectedSampleIndices: []
+      selectedSampleIndices: [],
+      createdSampleIds: []
     }
   }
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -0,0 +1,399 @@
...
+        that.initializeSelectTag();
+      });
+    })
+    .catch((error) => {
+      that.setState({
+        invalid: true,
+        errorMessage: 'Project exists already or is invalid'
+      })
+    });
+  }
```
boris-dimitrov: very nice & ajaxy thanks

app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -0,0 +1,399 @@
...
+
+  isProjectInvalid() {
+    if (this.refs.new_project.value === '' && this.state.project === 'Select a project') {
+      this.setState({
+        invalid: true,
+        errorMessage: 'Please enter valid project name'
+      })
+      return true;
+    } else {
+      return false;
```
boris-dimitrov: it just seems a bit unexpected for a predicate function to modify state, especially only in one of the cases

app/assets/javascripts/components/bulkUploadImport.jsx
yunfangjuan: actually it's susan's code.

boris-dimitrov: great @sadelokiki

PR #270, opened 2017-10-31
fix bugs

cdebourcy:
fix bad behavior in CSV generation and remove the download button for now

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -75,6 +75,7 @@ class ReportsController < ApplicationController
 
   def send_report_csv
     # override DEFAULT_PARAMS of report_helper.rb to get raw data
+    # TO DO: add flag to turn off thresholding rather than editing params
     params[:threshold_zscore] = 0
     params[:threshold_rpm] = 0
     params[:threshold_r] = 0
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -74,6 +74,16 @@ class ReportsController < ApplicationController
   end
 
   def send_report_csv
+    # override DEFAULT_PARAMS of report_helper.rb to get raw data
+    params[:threshold_zscore] = 0
+    params[:threshold_rpm] = 0
+    params[:threshold_r] = 0
+    params[:threshold_percentidentity] = 0
+    params[:threshold_alignmentlength] = 0
+    params[:threshold_neglogevalue] = 0
+    params[:threshold_aggregatescore] = 0
+    params[:excluded_categories] = 'None'
+    # serve CSV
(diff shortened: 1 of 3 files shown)


app/controllers/reports_controller.rb
```diff
@@ -74,6 +74,16 @@ def destroy
   end
 
   def send_report_csv
+    # override DEFAULT_PARAMS of report_helper.rb to get raw data
```
yunfangjuan: a bit hacky. probably should set a flag in geneate_report_csv to turn off the thresholdings. genus_search would need it too. @boris-dimitrov. At least at a TODO here.

app/lib/report_helper.rb
```diff
@@ -563,10 +563,12 @@ def get_tax_detail(tax_info, column_name)
 
   def generate_report_csv(report_info)
     rows = report_info[:taxonomy_details][1]
+    view_level_int = TaxonCount::NAME_2_LEVEL[report_info[:report_page_params][:view_level].downcase]
     attributes = %w[category_name tax_id name NT.zscore NT.rpm NT.r NR.zscore NR.rpm NR.r]
     CSV.generate(headers: true) do |csv|
       csv << attributes
       rows.each do |tax_info|
+        next unless tax_info['tax_level'] == view_level_int
```
boris-dimitrov: This removes genus entries from species results.   That's what you want?

db/schema.rb
```diff
@@ -94,7 +94,7 @@
   end
 
   create_table "projects", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
-    t.string "name", collation: "latin1_swedish_ci"
+    t.string "name", collation: "utf8_general_ci"
```
boris-dimitrov: remove this file from pull request?

app/controllers/reports_controller.rb
```diff
@@ -74,6 +74,16 @@ def destroy
...
+    params[:threshold_zscore] = 0
+    params[:threshold_rpm] = 0
+    params[:threshold_r] = 0
+    params[:threshold_percentidentity] = 0
+    params[:threshold_alignmentlength] = 0
+    params[:threshold_neglogevalue] = 0
+    params[:threshold_aggregatescore] = 0
+    params[:excluded_categories] = 'None'
+    # serve CSV
```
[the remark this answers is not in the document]
cdebourcy: Ok

app/controllers/reports_controller.rb
cdebourcy: ok

app/lib/report_helper.rb
cdebourcy: yes

PR #272, opened 2017-10-31
Susan/searchfix

sadelokiki:
Update search response with correct information
Search results are not paginated, will need to create custom pagination for them in another PR

### DIFF
--- a/app/assets/javascripts/components/samples.jsx
+++ b/app/assets/javascripts/components/samples.jsx
@@ -214,7 +214,7 @@ class Samples extends React.Component {
             </div>
           </div>
         </div>
-          {!this.state.displayedSamples && !this.state.displayedOutputData && this.state.displayedPipelineRunInfo ? <div className="no-data"><i className="fa fa-frown-o" aria-hidden="true"> No data to display</i></div> : this.renderTable(this.state.displayedSamples, this.state.displayedOutputData, this.state.displayedPipelineRunInfo)}
+          {!this.state.displayedSamples && !this.state.displayedOutputData && !this.state.displayedPipelineRunInfo ? <div className="no-data"><i className="fa fa-frown-o" aria-hidden="true"> No data to display</i></div> : this.renderTable(this.state.displayedSamples, this.state.displayedOutputData, this.state.displayedPipelineRunInfo)}
       </div>
     )
   }
--- a/app/assets/javascripts/components/samples.jsx
+++ b/app/assets/javascripts/components/samples.jsx
@@ -1,17 +1,20 @@
 class Samples extends React.Component {
   constructor(props, context) {
     super(props, context);
-    this.project = props.project;
+    this.project = props.project || null;
+    this.projectId = this.project ? this.project.id : null;
     this.samples = props.samples;
     this.samplesAmount = props.samples_count;
-    this.outputData = props.outputData
-    this.pipeline_run_info = props.pipeline_run_info
+    this.outputData = props.outputData || [];
+    this.pipelineRunInfo = props.pipeline_run_info || [];
     this.all_project = props.all_project|| [];
     this.defaultSortBy = 'newest';
     const currentSort = SortHelper.currentSort();
     this.state = {
       displayedSamples: this.samples || [],
-      samplesCount: this.samplesAmount || [],
+      displayedOutputData: this.outputData || [],
+      displayedPipelineRunInfo: this.pipelineRunInfo || [],
+      samplesCount: this.samplesAmount,
       sort_query: currentSort.sort_query
       ? currentSort.sort_query  : `sort_by=${this.defaultSortBy}`,
     };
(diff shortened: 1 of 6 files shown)


app/helpers/samples_helper.rb
```diff
@@ -63,7 +63,47 @@ def parsed_samples_for_s3_path(s3_path, project_id, host_genome_id)
...
-
     sample_list
   end
+
+  def samples_output_data(samples)
+    final_result = []
+    samples.each do |output|
+      output_data = {}
+      pipeline_info = output.pipeline_runs.first ? output.pipeline_runs.first.pipeline_output : nil
+      job_stats = output.pipeline_outputs.first ? output.pipeline_outputs.first.job_stats : nil
```
yunfangjuan: I think this is a bug. This should be pipelin_info? Pipelin_info. Job_stats?
Also, why do you call it pipeline_info? Pipeline_output will be more informative.

app/helpers/samples_helper.rb
```diff
@@ -63,7 +63,47 @@ def parsed_samples_for_s3_path(s3_path, project_id, host_genome_id)
...
+      pipeline_run_status = output.pipeline_runs.first ? output.pipeline_runs.first.job_status : nil
+      pipeline_run_entry[:job_status_description] =
+        if %w[CHECKED SUCCEEDED].include?(pipeline_run_status)
+          'COMPLETE'
+        elsif %w[FAILED ERROR].include?(pipeline_run_status)
+          'FAILED'
+        elsif %w[RUNNING LOADED].include?(pipeline_run_status)
+          'IN PROGRESS'
+        else
+          'UPLOADING'
```
yunfangjuan: I think you can add a ‘runnable’ and call it ‘initializing’

app/assets/javascripts/components/samples.jsx
```diff
@@ -1,17 +1,20 @@
 class Samples extends React.Component {
   constructor(props, context) {
     super(props, context);
-    this.project = props.project;
+    this.project = props.project || null;
+    this.projectId = this.project ? this.project.id : null;
     this.samples = props.samples;
     this.samplesAmount = props.samples_count;
```
yunfangjuan: this.sampleCount is a much better name.

app/helpers/samples_helper.rb
cdebourcy: Agreed. Also, when you change `job_stats` to `job_stats = pipeline_info ? pipeline_info.job_stats : nil`, please make the same change in the corresponding line in `home_controller.rb`. (It was me who put the buggy line there in the first place...)

app/controllers/home_controller.rb
```diff
@@ -20,30 +18,29 @@ def home
       @samples = sort_by(Sample.includes(:pipeline_runs, :pipeline_outputs).paginate(page: params[:page]), sort)
       @samples_count = Sample.all.size
     end
+    @final_result = samples_info(@samples) ? samples_info(@samples)[:final_result] : nil
```
yunfangjuan: instead of calling samples_info 4 times, I think you should just call int one time. 

basically do this

samples_info_output = samples_info(@samples) || {}
@final_results = samples_info_output[:final_result]
@pipeline_run_info = samples_info_output[:pipeline_run_info]

app/controllers/home_controller.rb
```diff
@@ -20,30 +18,29 @@ def home
...
-                                                    else
-                                                      'UPLOADING'
-                                                    end
-      @pipeline_run_info.push(pipeline_run_entry)
+    @samples = if project_id.present?
+                 Sample.includes(:pipeline_runs, :pipeline_outputs).search(search_query).where(project_id: project_id).paginate(page: params[:page])
+               else
+                 Sample.includes(:pipeline_runs, :pipeline_outputs).search(search_query).paginate(page: params[:page])
+               end
+    @final_result = samples_info(@samples)[:final_result]
```
yunfangjuan: same comment as above.

app/helpers/samples_helper.rb
```diff
@@ -63,7 +63,47 @@ def parsed_samples_for_s3_path(s3_path, project_id, host_genome_id)
...
+    end
+    final_result
+  end
+
+  def samples_pipeline_run_info(samples)
+    pipeline_run_info = []
+    pipeline_run_entry = {}
+    samples.each do |output|
+      pipeline_run_status = output.pipeline_runs.first ? output.pipeline_runs.first.job_status : nil
+      pipeline_run_entry[:job_status_description] =
```
yunfangjuan: pipeline_run_entry should be initialized inside each iteration.

app/helpers/samples_helper.rb
sadelokiki: okay, would "initializing" be a new status or replacing one of these @yunfangjuan?

app/helpers/samples_helper.rb
```diff
@@ -63,7 +63,47 @@ def parsed_samples_for_s3_path(s3_path, project_id, host_genome_id)
...
+        if %w[CHECKED SUCCEEDED].include?(pipeline_run_status)
+          'COMPLETE'
+        elsif %w[FAILED ERROR].include?(pipeline_run_status)
+          'FAILED'
+        elsif %w[RUNNING LOADED].include?(pipeline_run_status)
```
yunfangjuan: Can you also add RUNNABLE here

app/helpers/samples_helper.rb
cdebourcy: You can add
```
elsif pipeline_run_status == 'RUNNABLE'
  'INITIALIZING'
```

app/helpers/samples_helper.rb
sadelokiki: sure @cdebourcy

app/helpers/samples_helper.rb
cdebourcy: before the 
```
else
  'UPLOADING'
```

app/helpers/samples_helper.rb
cdebourcy: Perhaps related to this, notice that job status was still not displaying correctly when you test it on a few partially typed sample names

app/helpers/samples_helper.rb
sadelokiki: I actually moved the method into the sample helper so home controller does not have that line of code anymore @cdebourcy

app/controllers/home_controller.rb
sadelokiki: Great, thanks!

yunfangjuan:
There's still a type ahead bug as discussed but it works better than now.

PR #273, opened 2017-10-31
restore genus search

boris-dimitrov:


app/lib/report_helper.rb
```diff
@@ -34,7 +34,9 @@ module ReportHelper
     threshold_alignmentlength: 0.0,
     threshold_neglogevalue:    0.0,
     threshold_aggregatescore:  0.0,
-    excluded_categories: 'None'
+    excluded_categories: 'None',
+    selected_genus: 'None',
+    disable_filters: 0
```
boris-dimitrov: @cdebourcy You just need to add `&disable_filters=1` for your CSV download.

boris-dimitrov:
I will merge Cyril’s pull request after this. One merge at a time.

cdebourcy:
Sorry, I meant just the styles related to the genus search, from a week ago. (Not the new expand/collapse.) If that's clear and it's still on the agenda, then feel free to go ahead :-)

boris-dimitrov:
Oh I didnt realize I was missing those, good catch, Charles.   Let me look
for where they got lost!

app/lib/report_helper.rb
cdebourcy: thanks!

app/models/pipeline_output.rb
```diff
@@ -23,15 +23,6 @@ def generate_report
     end
   end
 
-  def distinct_genuses
```
yunfangjuan: I don't understand the choice of removing this and adding more complexity to  taxononomy_details function

app/lib/report_helper.rb
```diff
@@ -497,9 +509,13 @@ def taxonomy_details(report, params)
 
     count_species_per_genus!(tax_2d)
 
-    # Compute aggregate scores
+    # Compute aggregate scores and all genera
+    all_genera = Set.new
```
yunfangjuan: It's hard for me to determine if all_genera including all genera (including non-filtered) in the sample.

boris-dimitrov:
Happy to replace None with a word of your choice. We need that option so
the user has a way to clear the genus search.

The filters are not rendered on the page when the genus search is applied.
I think you are asking why still send the filter parameters in the URL. I
made that decision for a couple reasons:


— First, I didn’t want the user to lose any threshold values they may have
entered prior to activating genus search.

— Second, I wanted to greatly simplify the JS state management. You should
compare to what it was before the refactor. We can chat more with higher
bandwidth if unclear how this simplifies JS in general. For now note that
absence of parameters implies default values coded in rails, not JS.

The disable_filters parameter is not used in genus search. When set to 1,
it disables all filters including genus search. It is needed for Charles’
CSV download.

The logic to disable other filters when genus search is active is in the
rails code, because that minimizes the additional parameter changes that
would need to be tracked when switching between states in JS.

I thought genus names rather than ids was more user friendly but I may be
wrong. Happy to flip back to IDs if you prefer.

boris-dimitrov:
all_genera include all genera in the sample (before filtering)

the reason you need to compute all genera in the taxonomy details function
rather than by simple SQL query is that the all genera computation must
happen after data cleanup

data cleanup provides the display name for genus -200 and for species -1
and also deals with messy data (species with no genus in taxonomy_count get
singleton genera created on the fly — this no longer happens since we fixed
the aggregation rake task but it can happen again in future and so I have
kept the logic to handle it;   the rationale is that doing data cleanup
helps the presentation look less like a bug in the tool and also simplifies
the rest of the code)

PR #274, opened 2017-11-01
check if taxon fasta data is present

cdebourcy:


app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -41,14 +41,17 @@ class PipelineSampleReport extends React.Component {
     this.refreshPage({excluded_categories});
   }
 
-  render_name(tax_info, pipeline_output_id) {
+  render_name(tax_info, report_details) {
     foo = <i>{tax_info.name}</i>;
+    if (report_details.taxon_fasta_flag) {
+      link_url = `/pipeline_outputs/${report_details.pipeline_info.id}/fasta/${tax_info.tax_level}/${tax_info.tax_id}/NT_or_NR`
+    } else {
+      link_url = `https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=${tax_info.tax_id}`
```
boris-dimitrov: I'd replace this link with `<span>{tax_info.name}</span>`, OR, i'd always link to the report details page and there on that page show this link to NCBI.

boris-dimitrov: Rationale, as we discussed, is to make the behavior more consistent for users.

cdebourcy: agreed

PR #276, opened 2017-11-01
environments for pipeline

cdebourcy:
Lets us use development/alpha/production versions of the pipeline


config/initializers/pipeline.rb
```diff
@@ -1,4 +1,4 @@
 module IdSeqPipeline
-  S3_SCRIPT_LOC = ENV['IDSEQ_S3_SCRIPT_LOC'] || 's3://czbiohub-infectious-disease/bin/pipeline.py'
-  S3_POSTPROCESS_SCRIPT_LOC = ENV['IDSEQ_POSTPROCESS_S3_SCRIPT_LOC'] || 's3://czbiohub-infectious-disease/bin/postprocess.py'
+  S3_SCRIPT_LOC = ENV['IDSEQ_S3_SCRIPT_LOC'] || "s3://czbiohub-infectious-disease/bin/pipeline-#{Rails.env}.py"
```
yunfangjuan: I think a better way to structure this is to do 
s3://czbiohub-infectious-disease/bin/#{Rails.env}/pipeline.py 
and etc

cdebourcy: ok

PR #280, opened 2017-11-02
Add black list

yunfangjuan:
- get black list from the vector.gz file 
- change the genus_tax_id to -201 for all the black listed species ids.


config/taxon_blacklist.txt
```diff
@@ -0,0 +1,597 @@
+-1
```
boris-dimitrov: is -1 always blacklisted?  what if that's something genuinely never seen before? :P)

boris-dimitrov: i think right now -1 goes under genus -200 rather than -201 but anyway we will need to reaggregate

boris-dimitrov:
```
mysql> select * from taxon_lineages where taxid < 0;
Empty set (0.00 sec)
```
So this means -1 will always be put under -200, might as well exclude it from -201.

yunfangjuan:
can you disable the submit button once i clicked on it. It should prolly say submitting when doing it.

PR #282, opened 2017-11-02
Show spinner when submitting a form

sadelokiki:
Indicate when a submission process is ongoing.


app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -186,12 +193,16 @@ class BulkUploadImport extends React.Component {
         createdSampleIds: response.data.sample_ids
       });
       setTimeout(() => {
+        that.setState({
+          submitting: false
+        });
         that.gotoPage(`/?ids=${that.state.createdSampleIds.join(',')}`);
       }, 5000)
```
yunfangjuan: maybe we should make the delay just 200ms instead. otherwise people will click on the button again

PR #285, opened 2017-11-02
Disable s3 check for bulk upload

yunfangjuan:
disable s3 check for bulk upload to speed things up.

### DIFF
--- a/app/models/input_file.rb
+++ b/app/models/input_file.rb
@@ -16,10 +16,12 @@ class InputFile < ApplicationRecord
       if source[0..4] != 's3://'
         errors.add(:input_files, "file source doesn't start with s3:// for s3 input")
       end
-      command = "aws s3 ls #{source}"
-      _stdout, _stderr, status = Open3.capture3(command)
-      unless status.exitstatus.zero?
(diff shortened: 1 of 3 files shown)


app/models/input_file.rb
```diff
@@ -16,10 +16,12 @@ def s3_source_check
       if source[0..4] != 's3://'
         errors.add(:input_files, "file source doesn't start with s3:// for s3 input")
       end
-      command = "aws s3 ls #{source}"
-      _stdout, _stderr, status = Open3.capture3(command)
-      unless status.exitstatus.zero?
-        errors.add(:input_files, "file source #{source} doesn't exist")
+      unless sample.bulk_mode # skip the check for bulk mode
```
boris-dimitrov: Is it a better solution to "bulk ls" instead of skip?

yunfangjuan: can't really do that because each active record object is atomic ... . the thing about bulk mode is the data is kinda pre-validated unless someone is hacking our javascript.

yunfangjuan:
thanks!

PR #287, opened 2017-11-03
Charles/fasta input

cdebourcy:
Allow input files to be fasta. We'll need this to process all the new background samples we've just received

### DIFF
--- a/app/models/input_file.rb
+++ b/app/models/input_file.rb
@@ -7,13 +7,13 @@ class InputFile < ApplicationRecord
 
   FILE_REGEX = %r{\A[^\s\/]+\.(fastq|fastq.gz|fasta|fasta.gz)\z}
   validates :name, presence: true, format: { with: FILE_REGEX, message: "file must match format '#{FILE_REGEX}'" },
-    unless: :optional_input?
+                   unless: :optional_input?
   validates :source_type, presence: true, inclusion: { in: %w[local s3] }
   validate :s3_source_check, unless: :optional_input?
   validate :file_type_consistency_check, unless: :optional_input?
 
   def optional_input?
-    # If host filtering is skipped, read2 is optional. If a file name is given, it should nevertheless be validated. 
+    # If host filtering is skipped, read2 is optional. If a file name is given, it should nevertheless be validated.
     # TO DO: determine if we're indeed dealing with read2 (right now implicit, since frontend won't allow empty read1 name)
     sample.host_genome.name == HostGenome::NO_HOST_NAME && name == ''
   end
--- a/app/models/input_file.rb
+++ b/app/models/input_file.rb
@@ -14,13 +14,14 @@ class InputFile < ApplicationRecord
 
   def optional_input?
     # If host filtering is skipped, read2 is optional. If a file name is given, it should nevertheless be validated. 
-    # TO DO: less hacky way of determining that we're dealing with read2
-    sample.host_genome.name == HostGenome::NO_HOST_NAME && id > sample.input_files.first.id && name == ''
+    # TO DO: determine if we're indeed dealing with read2 (right now implicit, since frontend won't allow empty read1 name)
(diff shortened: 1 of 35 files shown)


db/schema.rb
```diff
@@ -10,293 +10,293 @@
 #
 # It's strongly recommended that you check this file into your version control system.
```
cdebourcy: Huge diff because the indentations on every line differ. How strange for an autogenerated file

cdebourcy: Actually this made me realize we don't even need to add type column to input_files. Will change it to just a method and keep schema out of the diff

aws_batch/pipeline.py
```diff
@@ -555,16 +554,16 @@ def run_sample(sample_s3_input_path, sample_s3_output_path,
                          "stats_file": os.path.join(result_dir, STATS_OUT)}
 
     # Download fastqs
-    command = "aws s3 ls %s/ |grep fastq.gz" % (sample_s3_input_path)
+    command = "aws s3 ls %s/ | grep %s" % (sample_s3_input_path, file_type)
```
boris-dimitrov: this is ok if filetype is one of (fasta, fastq.gz) but won't discern between fastq and fastq.gz

boris-dimitrov: also... ```fastanimation.fastq.gz```

may be just `grep '\.%s$'`

cdebourcy: Thanks!

app/models/input_file.rb
```diff
@@ -36,4 +36,8 @@ def s3_source_check
   def file_path
     File.join(sample.sample_path, 'fastqs', name)
   end
+
+  def type
```
yunfangjuan: type seems to be a sensitive reserve word. Use file_type instead?

app/models/input_file.rb
```diff
@@ -5,7 +5,7 @@ class InputFile < ApplicationRecord
   SOURCE_TYPE_LOCAL = 'local'.freeze
   SOURCE_TYPE_S3 = 's3'.freeze
 
-  FILE_REGEX = %r{\A[^\s\/]+\.fastq.gz}
+  FILE_REGEX = %r{\A[^\s\/]+\.(fastq.gz|fasta)\z}
```
yunfangjuan: and |fastq|fasta.gz?

aws_batch/pipeline.py
```diff
@@ -1152,6 +1177,7 @@ def main():
     global STAR_GENOME
     global BOWTIE2_GENOME
     INPUT_BUCKET = os.environ.get('INPUT_BUCKET', INPUT_BUCKET)
+    FILE_TYPE = os.environ.get('FILE_TYPE', FILE_TYPE)
```
yunfangjuan: I think you need to add 
global FILE_TYPE to be able to change the FILE_TYPE

cdebourcy: thanks

app/models/input_file.rb
cdebourcy: thanks!

aws_batch/pipeline.py
cdebourcy: thanks!

app/models/input_file.rb
cdebourcy: ok

app/models/sample.rb
```diff
@@ -48,11 +48,16 @@ def sample_path
   validates_associated :input_files
 
   def input_files_checks
-    # validate that we have exactly 2 input files
-    errors.add(:input_files, "file_size !=2 for sample") unless input_files.size == 2
-    # validate that both input files have the same source_type
+    # validate that we have the correct number of input files
+    if host_genome && host_genome.name == HostGenome::NO_HOST_NAME
+      errors.add(:input_files, "no input files") unless input_files.size >= 1
```
yunfangjuan: have to be between 1 and 2.

PR #288, opened 2017-11-03
Indication for 'running' and 'runnable' status

sadelokiki:


app/assets/javascripts/components/PipelineSampleReads.jsx
```diff
@@ -26,9 +26,9 @@ class PipelineSampleReads extends React.Component {
   }
 
   pipelineInProgress(status) {
-    if (status === null || this.state.rerun === true ) {
+    if (status === null || this.state.rerun === true || status === 'RUNNABLE' || status === 'RUNNING') {
       return true;
-    } else if ( status === 'ERROR' || status === 'CHECKED' || status === 'FAILED') {
```
yunfangjuan: you should keep the checked

sadelokiki: Yea, i intend to add it back when I discuss where to fix the button on the reports page with Rebecca. @yunfangjuan

PR #292, opened 2017-11-04
remove view_level selector

boris-dimitrov:
also fix recently introduced bug that was causing all categories
to show as "Uncategorized"

### DIFF
--- a/app/assets/javascripts/components/PipelineSampleReport.jsx
+++ b/app/assets/javascripts/components/PipelineSampleReport.jsx
@@ -8,11 +8,12 @@ class PipelineSampleReport extends React.Component {
     this.taxonomy_details = props.taxonomy_details[2];
     this.all_genera_in_sample = props.all_genera_in_sample;
     this.all_categories = props.all_categories;
-    this.applyViewLevel = this.applyViewLevel.bind(this);
     this.applyNewFilterThresholds = this.applyNewFilterThresholds.bind(this);
     this.applyExcludedCategories = this.applyExcludedCategories.bind(this);
     this.applyGenusFilter = this.applyGenusFilter.bind(this);
     this.expandOrCollapseGenus = this.expandOrCollapseGenus.bind(this);
+    this.expandTable = this.expandTable.bind(this);
+    this.collapseTable = this.collapseTable.bind(this);
     this.disableFilters = this.disableFilters.bind(this);
     this.enableFilters = this.enableFilters.bind(this);
   }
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -161,13 +160,23 @@ class PipelineSampleReport extends React.Component {
...
-    if (attr[1] == 'fa-angle-right') {
-      e.target.attributes.class.nodeValue = `fa fa-angle-down ${taxId}`;
-    } else {
-      e.target.attributes.class.nodeValue = `fa fa-angle-right ${taxId}`;
-    }
-    e.nativeEvent.stopImmediatePropagation();
+    $(`.report-arrow.${taxId}`).toggleClass('hidden');
+  }
+
+  expandTable(e) {
```
boris-dimitrov: Note that the negative genera ("all taxa without genera" and "all artificial constructs") will not be expanded by the Table expander.    I thought this is how users would want it.   Let me know if I am wrong.

app/lib/report_helper.rb
```diff
@@ -358,14 +354,14 @@ def cleanup_names!(taxon_counts_2d)
...
-        elsif !(TaxonLineage::MISSING_LINEAGE_ID.values.include? tax_id)
+        elsif !(TaxonLineage::MISSING_LINEAGE_ID.values.include? tax_id) && tax_id != TaxonLineage::MISSING_SPECIES_ID_ALT
           tax_info['name'] += " #{tax_id}"
         end
       elsif !tax_info['name']
         missing_names.add(tax_id)
         tax_info['name'] = "Unnamed taxon #{tax_id}"
       end
-      category_id = tax_info.delete('superkindom_taxid')
+      category_id = tax_info.delete('superkingdom_taxid')
```
boris-dimitrov: this is the bugfix

app/lib/report_helper.rb
```diff
@@ -358,14 +354,14 @@ def cleanup_names!(taxon_counts_2d)
         tax_info['name'] = "All taxa without #{level_str} classification"
         if tax_id == TaxonLineage::BLACKLIST_GENUS_ID
           tax_info['name'] = "All artificial constructs"
-        elsif !(TaxonLineage::MISSING_LINEAGE_ID.values.include? tax_id)
+        elsif !(TaxonLineage::MISSING_LINEAGE_ID.values.include? tax_id) && tax_id != TaxonLineage::MISSING_SPECIES_ID_ALT
```
boris-dimitrov: this is also a bugfix

PR #294, opened 2017-11-06
Charles/verbose logs

cdebourcy:


boris-dimitrov:
I find in these situations it's good to flush on most python print statements.   One way to do it is replacing "print" with your own function "tsprint" where "ts" that roughly does this:
```
tsprint_lock = threading.RLock()

# Thread-safe and timestamped prints.
def tsprint(msg):
    ts = msstr(time.time()) + " "
    msg = ts + msg.replace("\n", "\n" + ts)
    with tsprint_lock:
        print msg

def msstr(t):
    # We do not use "{:.3f}".format(time.time()) because its result may be
    # up to 0.5 ms in the future due to rounding.  We want flooring here.
    s = str(int(t * 1000))
    return s[:-3] + "." + s[-3:]
```
Among the several advantages of this approach, you can add "flush" right at the end of the print function, so it doesn't need to be sprinkled everywhere.  It also works under both python 2.7 and 3.6+.

An alternative approach to using flush() is to unbuffer stdout.  Often that's done together with merging stdout and stderr.
```
    # Unbuffer stdout and redirect stderr into stdout.  This helps observe logged events in realtime.
    sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
    os.dup2(sys.stdout.fileno(), sys.stderr.fileno())
```
The only disadvantage of that approach is that it does not work at all on python 3+.

boris-dimitrov:
For examples of tsprint and unbuffer see https://github.com/boris-dimitrov/eversnap/blob/master/eversnap.py

cdebourcy:
Thanks @boris-dimitrov ! I've added stdout unbuffering but not tsprint since stdout will have cloudwatch time stamps and the logfile will have time stamps produced by TimeFilter

aws_batch/pipeline.py
```diff
@@ -468,7 +468,10 @@ def return_merged_dict(dict1, dict2):
 
 def execute_command(command):
     print command
+    sys.stdout.flush()
     output = subprocess.check_output(command, shell=True)
+    print output
+    sys.stdout.flush()
     return output
 
 def wait_for_server(service_name, command, max_concurrent):
```
boris-dimitrov: 1) I find it helpful to log messages whenever a wait for a resource can happen.

2) May be the AWS queuing system can handle that wait?

3) Otherwise it's also possible to package the wait like so:

```
    with SharedResource(resource_type="gsnap instance", requestor=job, priority="high", max_wait=3600) as gsnap_instance:
        gsnap_instance.run(job)
```

where the `__enter__` and `__exit__` functions implementing this block functionality will do the appropriate things, including raise an exception if the max_wait is exceeded.

4) The running of the job too can be encapsulated in such a block, in a way that enforces timeouts, requires keepalives, and allows you to filter the subprocess output.   Usually this is done using  a few more python threads. 

An individual log listener could be as simple as a thread that just gets lines from the subprocess stdout and inspects those lines and/or puts some of them into a queue from which another thread consumes and tsprints into the main process output log, possibly adding annotations that indicate which subprocess they belong to.

boris-dimitrov:
You may need tsprint as soon as you start using multiple python threads (without the timestamps, I guess, though it wouldn't hurt to have the tsprint timestamps, because the Cloudwatch ones can't possibly be of a useful time resolution -- the infrastructure that generates them doesn't even pick up the line until some time after they are flushed, does it?)

aws_batch/pipeline.py
```diff
@@ -468,9 +468,24 @@ def return_merged_dict(dict1, dict2):
...
+    print command
+    sys.stdout.flush()
+    process2 = subprocess.Popen("touch %s ; tail -f %s" % (progress_file, progress_file), shell=True)
+    process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
+    while True:
+        line = process.stdout.readline().rstrip()
+        if not line: break
+        print line
+    process.wait()
+    process2.kill()
```
boris-dimitrov: Pretty cool, tho as we discussed the lines can get broken and interleaved in this technique.   Also, for output to be emitted in realtime you'd need the unbuffering trick.  As is, your output will not flush after each line.

Of course, if you do unbuffer, then you could just
```
tail = subprocess.Popen("touch %s ; tail -f %s" % (progress_file, progress_file), shell=True)
try:
  subprocess.check_call(command, shell=True)
finally:
  subprocess.kill(tail)
```

cdebourcy: Thank you @boris-dimitrov ! Changed it in charles/verbose-logs-2

PR #296, opened 2017-11-07
Star partitions

yunfangjuan:
STAR Indexing partitioning
Test Plan
- Run indexing.py on science.czi-e.com
- Run pipeline.py on sciece.czi-e.com 

Ensure everything works with the right output.

### DIFF
--- a/aws_batch/indexing.py
+++ b/aws_batch/indexing.py
@@ -13,6 +13,8 @@ STAR="STAR"
 BOWTIE2_BUILD="bowtie2-build"
 
 STAR_INDEX_OUT = 'STAR_genome.tar.gz'
+MAX_STAR_PART_SIZE = 3252010122
+
 BOWTIE2_INDEX_OUT = 'bowtie2_genome.tar.gz'
 
 ### Example usage of the present script:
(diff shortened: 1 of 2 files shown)


aws_batch/pipeline.py
```diff
@@ -784,6 +784,23 @@ def run_sample(sample_s3_input_path, sample_s3_output_path,
...
+                           '--outFilterMatchNminOverLread', '0.5',
+                           '--outReadsUnmapped', 'Fastx',
+                           '--outFilterMismatchNmax', '999',
+                           '--outSAMmode', 'None',
+                           '--clip3pNbases', '0',
+                           '--runThreadN', str(multiprocessing.cpu_count()),
+                           '--genomeDir', genome_dir,
+                           '--readFilesIn', fastq_file_1, fastq_file_2]
+    if fastq_file_1[-3:] == '.gz':
+        ['--readFilesCommand', 'zcat']
```
cdebourcy: star_command_params.extend(['--readFilesCommand', 'zcat']) ?

aws_batch/pipeline.py
```diff
@@ -798,23 +815,25 @@ def run_star(sample_name, fastq_file_1, fastq_file_2, star_genome_s3_path,
...
+        num_parts = int(open("%s/STAR_genome/parts.txt" % REF_DIR, 'rb').read())
+        part_idx = 0
+        tmp_result_dir = "%s/star-part-%d" % (scratch_dir, part_idx)
+        run_star_part(tmp_result_dir, REF_DIR + "/STAR_genome/part-%d" % part_idx, fastq_file_1, fastq_file_2)
+        for i in range(1, num_parts):
+            fastq_1 = "%s/Unmapped.out.mate1" % tmp_result_dir
+            fastq_2 = "%s/Unmapped.out.mate2" % tmp_result_dir
+            tmp_result_dir = "%s/star-part-%d" % (scratch_dir, i)
+            run_star_part(tmp_result_dir, REF_DIR + "/STAR_genome/part-%d" % i, fastq_1, fastq_2)
+        # extract out unmapped files
```
cdebourcy: nice...

aws_batch/pipeline.py
yunfangjuan: good catch. meant to do +=

aws_batch/indexing.py
```diff
@@ -25,20 +27,62 @@ def execute_command(command):
...
+
+            if line[0] == '>': # got a new read
+                current_output_file.write(current_read)
+                current_size += len(current_read)
+                current_read = line
+            else:
+                current_read += line
+        current_output_file.write(current_read)
+        current_output_file.close()
+    return fasta_file_list
```
boris-dimitrov: I see a couple of bugs here.

First, current_size is not initialized when it is first accessed.

Second, the last read in the input file is lost (never output).

aws_batch/indexing.py
```diff
@@ -25,20 +27,62 @@ def execute_command(command):
...
 
+def split_fasta(fasta_file, max_fasta_part_size):
+    fasta_file_list = []
+    part_idx = 0; current_size = 0
+    current_output_file_name = "%s.%d" % (fasta_file, part_idx)
+    current_output_file = open(current_output_file_name, 'wb')
+    fasta_file_list.append(current_output_file_name)
+    with open(fasta_file, 'rb') as input_f:
+        current_read = input_f.readline()
+        for line in input_f:
```
boris-dimitrov: i believe in python 2.7 this "for line in input_f" reads the entire file in memory.

aws_batch/indexing.py
yunfangjuan: line 32: part_idx = 0; current_size = 0 
line 53: current_output_file.write(current_read)

aws_batch/indexing.py
yunfangjuan: https://stackoverflow.com/questions/8009882/how-to-read-large-file-line-by-line-in-python

boris-dimitrov: there is a small probability the last fragment would be very short

aws_batch/indexing.py
```diff
@@ -25,20 +27,62 @@ def execute_command(command):
     output = subprocess.check_output(command, shell=True)
     return output
 
+def split_fasta(fasta_file, max_fasta_part_size):
+    fasta_file_list = []
+    part_idx = 0; current_size = 0
```
boris-dimitrov: I totally missed this semicolon, could you put on second line :P)

aws_batch/indexing.py
boris-dimitrov: yeah i just found that stackoverflow question myself, the answer says this is good for python versions >= 2.6

i think if this doesn't blow up for our 4GB file we are good :)

yunfangjuan: yeah. i tested the indexing part. it works.

aws_batch/indexing.py
```diff
@@ -25,20 +27,62 @@ def execute_command(command):
...
+    for i in range(len(fasta_file_list)):
+        print "start making STAR index part %d" % i
+        star_genome_part_dir = "%s/part-%d" % (star_genome_dir_name, i)
+        star_command_params = ['cd', scratch_dir, ';',
+                               'mkdir -p ', star_genome_part_dir, ';',
+                               STAR, '--runThreadN', str(multiprocessing.cpu_count()),
+                               '--runMode', 'genomeGenerate',
+                               '--genomeDir', star_genome_part_dir,
+                               '--genomeFastaFiles', fasta_file_list[i]]
+        execute_command(" ".join(star_command_params))
```
boris-dimitrov: I've heard that Python 3.6 has interpolated strings.   :)

aws_batch/pipeline.py
```diff
@@ -798,23 +815,25 @@ def run_star(sample_name, fastq_file_1, fastq_file_2, star_genome_s3_path,
...
-                           '--genomeDir', REF_DIR + '/STAR_genome',
-                           '--readFilesIn', fastq_file_1, fastq_file_2]
-    execute_command(" ".join(star_command_params))
-    logging.getLogger().info("finished job")
-    # extract out unmapped files
-    execute_command("cp %s/%s %s/%s;" % (scratch_dir, 'Unmapped.out.mate1', result_dir, STAR_OUT1))
-    execute_command("cp %s/%s %s/%s;" % (scratch_dir, 'Unmapped.out.mate2', result_dir, STAR_OUT2))
+    # Check if parts.txt file exists, if so use the new version of (partitioned indices). Otherwise, stay put
+    if os.path.isfile("%s/STAR_genome/parts.txt" % REF_DIR):
+        num_parts = int(open("%s/STAR_genome/parts.txt" % REF_DIR, 'rb').read())
```
boris-dimitrov: are we worried about leaking file descriptors?   i guess this is not a long running process?

yunfangjuan: sure. i can close it.

aws_batch/pipeline.py
boris-dimitrov: nice optimization

should we delete the unmapped files after we are done with them?

curious if for the very big jobs (4GB input) this will increase time due to writing out all the unmapped files... if we split the input into N parts, then the amount of intermediate output will be input_size*(N-1)/2

also for the case when we run this in parallel, may be useful to just turn off the unmapped output?  or send it to /dev/null

boris-dimitrov: oh i see we actually need the unmapped ones for the overall result

yunfangjuan: yes

PR #297, opened 2017-11-07
fix genome selection

cyrielo:
This PR fixes the genus selection bug in the sample create form


sadelokiki:
Thanks! Please change initial value of this.state.hostName to this.state.host

yunfangjuan:
I just checked out the branch and it's not working if you test on samples/new

cdebourcy:
same here

cyrielo:
You're right @yunfangjuan, it would need to send the updated host id. @sadelokiki would look into it to update the PR

yunfangjuan:
i have a fix for this. you can close the pr now. I will submit it shortly.

PR #298, opened 2017-11-07
UI Fixes

sadelokiki:
- Overflow on All Projects Dropdown
- Color code for In Progress job status

### DIFF
--- a/app/assets/javascripts/components/samples.jsx
+++ b/app/assets/javascripts/components/samples.jsx
@@ -61,7 +61,7 @@ class Samples extends React.Component {
   applyClass(status) {
     if(status === 'COMPLETE') {
       return 'complete';
-    } else if (status === 'UPLOADING') {
+    } else if (status === 'UPLOADING' || status === 'IN PROGRESS') {
       return 'uploading';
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/samples.jsx
```diff
@@ -61,7 +61,7 @@ class Samples extends React.Component {
   applyClass(status) {
     if(status === 'COMPLETE') {
       return 'complete';
-    } else if (status === 'UPLOADING') {
+    } else if (status === 'UPLOADING' || status === 'IN PROGRESS') {
```
yunfangjuan: why is this uploading? 
if 'IN PROGRESS', status should just be 'IN PROGRESS' or 'RUNNING'?

sadelokiki: I think job_status_descriptions UPLOADING, IN PROGRESS, FAILED AND COMPLETE, which is what is being displayed on the samples page. Rebecca suggested we have the same colors for uploading and in progress @yunfangjuan

yunfangjuan: ok got it.

PR #301, opened 2017-11-08
logging improvements

boris-dimitrov:
flush python log prior to copying to S3

echo python log info statements to stdout so they make it into cloudwatch

merge stdout and stderr of most subprocesses into the main python process
stdout and stderr so all that makes it into cloudwatch

for long-running subproceses, print a timer tick every minute, so we can
see in cloudwatch that the python driver process is alive and just waiting

quiet bowtie2 because it always completes very quickly and its output
is rather voluminous;  TODO: sample the output instead of suppressing

### DIFF
--- a/aws_batch/pipeline.py
+++ b/aws_batch/pipeline.py
@@ -14,6 +14,7 @@ import datetime
 import gzip
 import logging
 import math
+import threading
 
 ENVIRONMENT = 'alpha'
 INPUT_BUCKET = 's3://czbiohub-infectious-disease/UGANDA' # default to be overwritten by environment variable
@@ -117,8 +118,8 @@ def lzw_fraction(sequence):
     dictionary = {}
     # Initialize dictionary with single char
     for c in sequence:
-       dict_size += 1
-       dictionary[c] = dict_size
+        dict_size += 1
+        dictionary[c] = dict_size
 
     word = ""
     results = []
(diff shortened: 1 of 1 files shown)


boris-dimitrov:
still need to test this but thought i'd put it up for review first

boris-dimitrov:
updated after testing the new code in a standalone environment

aws_batch/pipeline.py
```diff
@@ -506,17 +564,31 @@ def filter(self, record):
...
+        return "%3.1f" % percent
+    except:
+        return str(percent)
+
+def logger_echo(level, msg)
+    print level + ": " + msg
+
+def logger_info(msg)
+    logger_echo("INFO", msg)
+    logger_info(msg)
```
cdebourcy: did you mean `logger.info(msg)` inside `logger_info`? (I think we should still produce the logfile in S3 as well)

boris-dimitrov: yes that is the result of a careless find-and-replace!   thank you for catching it

boris-dimitrov: as it is, it would have crashed on bottomless recursion right away :P)

boris-dimitrov:
@cdebourcy stackoverflow had a better way of echoing logger.info(...) messages to stdout, so i switched to that scheme

PR #302, opened 2017-11-08
Samples page improvement

yunfangjuan:
Fix a bug and a config change 
- vcpus from 16 to 4
- check the sources are different from read 1 and read
- make samples pages more informative

### DIFF
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -33,6 +33,25 @@ module SamplesHelper
     (100.0 * bowtie2_stats.reads_after) / star_stats.reads_before unless bowtie2_stats.nil? || star_stats.nil?
   end
 
+  def sample_status_display(sample)
+    if sample.status == Sample::STATUS_CREATED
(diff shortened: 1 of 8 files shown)


app/helpers/samples_helper.rb
```diff
@@ -33,6 +33,25 @@ def compute_percentage_reads(jobstats)
     (100.0 * bowtie2_stats.reads_after) / star_stats.reads_before unless bowtie2_stats.nil? || star_stats.nil?
   end
 
+  def sample_status_display(sample)
+    if sample.status == Sample::STATUS_CREATED
+      return 'uploading'
+    elsif sample.status == Sample::STATUS_CHECKED
+      pipeline_run = sample.pipeline_runs.first
+      return '' unless pipeline_run
```
boris-dimitrov: what is the meaning of empty string?

yunfangjuan: basically it's a transitional state

PR #303, opened 2017-11-08
Update report_helper.rb

cdebourcy:


cdebourcy:
ok will do! will merge this for now so cyril can get CSVs

PR #304, opened 2017-11-08
Susan/general bug fixes

sadelokiki:
Form validation bulk upload
Format error message for Bulk import
Add IDSeq mail access link
Update search UI fonts and layout
Add formatted error responses for unhandled exceptions from by db upgrade

### DIFF
--- a/app/assets/javascripts/components/bulkUploadImport.jsx
+++ b/app/assets/javascripts/components/bulkUploadImport.jsx
@@ -244,6 +244,19 @@ class BulkUploadImport extends React.Component {
   }
 
   isImportFormInvalid() {
+    if (this.state.project === 'Select a project' && this.refs.bulk_path === '') {
+      this.setState({
+        invalid: true,
+        errorMessage: 'Please fill in all required fields'
+      })
+      return true;
+    } else if (this.state.project === 'Select a project') {
+      this.setState({
+        invalid: true,
+        errorMessage: 'Please select a project'
+      })
+      return true;
+    }
     if (this.refs.bulk_path === '') {
         this.setState({
           invalid: true,
@@ -256,7 +269,7 @@ class BulkUploadImport extends React.Component {
           errorMessage: 'Please fill in a valid bulk_path '
         })
         return true;
-    }
+    } 
     return false;
   }
 
--- a/app/assets/javascripts/components/bulkUploadImport.jsx
+++ b/app/assets/javascripts/components/bulkUploadImport.jsx
@@ -204,7 +204,7 @@ class BulkUploadImport extends React.Component {
      that.setState({
       submitting: false,
       invalid: true,
-      errorMessage: error.response.data.status || 'Failed to upload bulk'
+      errorMessage: error.response.data.status || 'Something went wrong'
      })
     });
   }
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -192,28 +199,26 @@ class BulkUploadImport extends React.Component {
...
       that.initializeSelectTag()
       that.initializeSelectAll()
-    }).catch(function (error) {
+    })
+    .catch((error) => {
      that.setState({
       submitting: false,
       invalid: true,
-      errorMessage: JSON.stringify(error.response)
+      errorMessage: error.response.data.status || 'Something went wrong'
```
yunfangjuan: what does the status tell us from the server side?

app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -272,7 +300,8 @@ class BulkUploadImport extends React.Component {
     const samples = this.state.samples
     samples[e.target.id].project_id = this.state.allProjects[e.target.selectedIndex].id
     this.setState({
-      samples: samples
+      samples: samples,
+      project: e.target.value.trim()
```
yunfangjuan: what is this line for? why change project when you just change project for one sample.

app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -327,7 +364,7 @@ class BulkUploadImport extends React.Component {
                     <label htmlFor={i}> {sample.name}</label>
                   </p>
                   <div className="col s4">
-                    <select className="" id={i} onChange={ this.handleProjectChangeForSample } value={sample.project_id}>
+                    <select className="" id={i} onChange={ this.handleProjectChangeForSample } value={this.state.project}>
```
yunfangjuan: why do you set the value to this.state.project? individual samples can map to different project id.

app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -424,7 +463,8 @@ class BulkUploadImport extends React.Component {
   render() {
     return (
       <div>
-        { this.state.imported ? this.renderBulkUploadSubmitForm(): this.renderBulkUploadImportForm() }
+        { this.state.imported ? this.renderBulkUploadSubmitForm() : null }
```
yunfangjuan: what's the reason of separating them into different lines?

app/assets/javascripts/components/bulkUploadImport.jsx
sadelokiki: it returns the specific error object returned from the server - 'No samples imported under bulk_path'.  @yunfangjuan

app/assets/javascripts/components/bulkUploadImport.jsx
sadelokiki: Just noticed that all sample project ids are being updated on project change, so no need to change then individually, i would revert it, thanks.

app/assets/javascripts/components/bulkUploadImport.jsx
sadelokiki: having them on both sides of the ternary operator was throwing a DOM error poiting to hiding and not completely removing the component(after rendering), so I decided to either show or return null for both components as opposed to hiding. @yunfangjuan

PR #305, opened 2017-11-08
WIP - Project visualization

cyrielo:
**Progress 73%**
TODO
Still working on fixing the color scale
Then updating the colors when the count type is switched

Add a button to sort the pathogens so that the colors distribute based on the value of the count type.

The labels on x and y axis are overflowing on long pathogen names, maybe use a legend to map the x axis to display sample initials, then display full name on hover.

Add a loading spinner when fetching the pathogens from the server.

Right now clicking the chevron icon changes the dataset, ideally we should be visualizing the selected pathogen. (the size of the rectangle gets really big when showing few pathogen because of the scaling).


cyrielo:
New additions:
Added ability to sort the pathogens based on higest zscores or rpm
Ability to switch the count type and updated the color scale
visualizing only the selected pathogens

cdebourcy:
Cool. For the absent data, gray squares moving into blank spots looks a bit confusing. It might look more natural to the scientists it the heatmap just changed instantly upon selection update rather than having a continuous motion

cyrielo:
Cool, we can definitely try out different ideas

PR #306, opened 2017-11-08
Full csv

cdebourcy:
remove filters and truncation for csv outputs

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -75,7 +75,6 @@ class ReportsController < ApplicationController
   end
 
   def send_report_csv
-    params[:disable_filters] = 1
     params[:is_csv] = 1
     @report_csv = generate_report_csv(external_report_info(@report, params))
     send_data @report_csv, filename: @report.pipeline_output.sample.name + '_report.csv'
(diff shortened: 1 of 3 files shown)


app/controllers/reports_controller.rb
```diff
@@ -75,18 +75,10 @@ def destroy
...
-    params[:threshold_r] = 0
-    params[:threshold_percentidentity] = 0
-    params[:threshold_alignmentlength] = 0
-    params[:threshold_neglogevalue] = 0
-    params[:threshold_aggregatescore] = 0
-    params[:excluded_categories] = 'None'
-    # serve CSV
-    @report_csv = generate_report_csv(external_report_info(@report, params))
+    params[:is_csv] = 1
+    params[:sort_by] = "highest_nt_aggregatescore"
```
boris-dimitrov: oh i see... i think it may be fine to put the disable_filters here then, so we don't need to have the check for filters in taxonomy_details

the check for truncating results we still need, or we can make max_rows a URI param and make this perfectly uniform

cdebourcy: Makes sense. I think we'll redesign the report CSV functionality once the button is back in the frontend and we get some user feedback on whether they mostly want to download raw data or whether they want to be able to filter/truncate/sort... I'll postpone reworking this until then.

PR #307, opened 2017-11-09
Update pipeline.py

cdebourcy:


aws_batch/pipeline.py
```diff
@@ -34,7 +34,7 @@
 
 LZW_FRACTION_CUTOFF = 0.45
 
-GSNAPL_MAX_CONCURRENT = 5
+GSNAPL_MAX_CONCURRENT = 15
```
boris-dimitrov: Go to 20 at least. :)

PR #310, opened 2017-11-09
Update pipeline.py

cdebourcy:


aws_batch/pipeline.py
```diff
@@ -600,8 +601,10 @@ def get_server_ips(service_name, environment):
     return server_ips
 
 def wait_for_server_ip(service_name, key_path, remote_username, environment, max_concurrent):
+    instance_ips = get_server_ips(service_name, environment)
+    i = 1
     while True:
-        instance_ips = get_server_ips(service_name, environment)
+        if i % 10 == 0: instance_ips = get_server_ips(service_name, environment)
         ip_nproc_dict = {}
         for ip in instance_ips:
```
boris-dimitrov: may be randomize order here like

```
  import random
  for ip in sorted(instance_ips, key=lambda _: random.randint(1,1000))
```

cdebourcy: we're actually choosing the machine with lowest number of running processes a few lines later. You want to randomize so that we don't always choose the same machine if there are several with equal utilization?

PR #313, opened 2017-11-10
Revert "change 'INITIALIZING' to 'QUEUED'"

cdebourcy:
Reverts chanzuckerberg/idseq-web#308


cdebourcy:
Reverting because "queued" shows as red instead of gray like "initializing" did

PR #318, opened 2017-11-14
Check results status

yunfangjuan:
Test plan: 
- Downloaded alpha. make sample 13 pipeline_output nil. 
- Make Rails.env = 'production'
- pr = Sample.find(13).pipeline_runs.first # a mosquito sample 
- pr.check_job_status

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -24,11 +24,11 @@ class PipelineRun < ApplicationRecord
   end
 
   def check_job_status
-    return if pipeline_output
-    if job_status == STATUS_SUCCESS || job_status == STATUS_CHECKED
+    if pipeline_output
       self.job_status = STATUS_CHECKED
-      Resque.enqueue(LoadResultsFromS3, id)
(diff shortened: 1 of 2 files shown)


app/models/pipeline_run.rb
```diff
@@ -139,4 +150,17 @@ def download_file(s3_path, destination_dir)
     return nil unless status.exitstatus.zero?
     "#{destination_dir}/#{File.basename(s3_path)}"
   end
+
+  def file_generated_since_run(s3_path)
+    command = "aws s3 ls #{s3_path}"
+    stdout, _stderr, status = Open3.capture3(command)
+    return false unless status.exitstatus.zero?
+    s3_file_time = DateTime.strptime(stdout[0..18], "%Y-%m-%d %H:%M:%S")
+    (s3_file_time > created_at)
```
cdebourcy: Are we guaranteeing that the timezone in S3 and in the app are the same?

yunfangjuan: They are all UTC

PR #320, opened 2017-11-14
Susan/resultrowreport

sadelokiki:
Add Download Reads, NCBI links and Download report for Reports Page.


app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -17,6 +17,20 @@ class PipelineSampleReport extends React.Component {
     this.disableFilters = this.disableFilters.bind(this);
     this.enableFilters = this.enableFilters.bind(this);
   }
+  
+  componentDidMount() {
```
boris-dimitrov: Please find a way to do this without coponentDidMount().   It has been a major source of performance problems in the past.

app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -63,19 +77,42 @@ class PipelineSampleReport extends React.Component {
...
+
+  //download Fasta  
+  downloadFastaUrl(reportDetails, taxInfo) {
+    location.href = `/pipeline_outputs/${reportDetails.pipeline_info.id}/fasta/${taxInfo.tax_level}/${taxInfo.tax_id}/NT_or_NR`;
+  }
+
+  displayTags(taxInfo, reportDetails) {
+    return (
+      <span className="link-tag">
+        <i onClick={this.gotoNCBI.bind(this, taxInfo.tax_id)} className="fa fa-link cloud" aria-hidden="true"></i>
```
boris-dimitrov: Will this create a closure for every row in the table?   That's too many closures, which will cause the browser to pause for long reports.  Find a way to extract the information differently, e.g. from the class name.

app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -180,23 +218,29 @@ class PipelineSampleReport extends React.Component {
...
+  downloadReport(id) {
+    location.href = `/reports/${id}/csv`
+  }
+
   render() {
     const parts = this.props.report_page_params.sort_by.split("_")
     const sort_column = parts[1] + "_" + parts[2];
     var t0 = Date.now();
-    filter_stats = this.rows_passing_filters + ' rows passing filters, out of ' + this.rows_total + ' total rows.';
+    filter_stats = this.rows_passing_filters + ' Results matching filter criteria';
```
boris-dimitrov: It's really important to know how many rows did *not* match the criteria.   Is that information being removed?

app/assets/javascripts/components/PipelineSampleReport.jsx
sadelokiki: @boris-dimitrov, componentDidMount() is required to invoke a method after a component is mounted. the only thing that happens after the component is mounted is that the tags are hidden until the names are hovered on. Also when hovered on, it takes a maximum of 1millisecond to show and hide the tags. The other way to do this would be using DOM manipulation with jQuery which is not the React way.

boris-dimitrov: Please find a big report (>3000 rows) and measure the browser layout time before and after this change.   If there is a way to *not* introduced componentDidMount() please find it.   Literally every time it has been introduced on this page, performance has suffered, and I have had to delete it.

sadelokiki: hey @boris-dimitrov, I just did a test with huge data, it actually doesn't affect the performance cos it is not walking through the entire DOM, and is only called if a name is hovered on. in which case, the response is instantaneous.

app/assets/javascripts/components/PipelineSampleReport.jsx
sadelokiki: Thanks.

app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -17,6 +17,20 @@ class PipelineSampleReport extends React.Component {
...
   }
+  
+  componentDidMount() {
+    this.hideTags();
+    //only display download links when genus name is hovered on
+    $('.hover-wrapper').hover(function() {
+      $(this).find(".link-tag").show(); 
+    }, function() {
+      $(this).find(".link-tag").hide(); 
+    });
```
boris-dimitrov: This may be just me not understanding how browsers work, but it really looks like the browser is going to have to find all the elements with class .hover_wrapper and then for each one create a closure (a function) on the fly.

Why can't all the tags be hidden initially, with a hove function that shows them?   Wouldn't that let you eliminate the componentDidMount() function?

I am concerned about adding this function here because it represents a style of programming which doesn't generally scale for large pages, and will invite future additions of other code that is even less scalable.   All things equal, for this particular page, I think we would be well served to try and find a way of avoiding this.

app/assets/javascripts/components/PipelineSampleReport.jsx
sadelokiki: thanks, the new design had only the matching filters, but would revert back to show all as well.

yunfangjuan:
It works OK but when the taxid is <0, it goes to a dead link to NCBI. So I think the link to NCBI shouldn't be shown if the taxid is negative

app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -63,19 +63,49 @@ class PipelineSampleReport extends React.Component {
...
+    const taxLevel = e.target.getAttribute('data-tax-level');
+    const taxId = e.target.getAttribute('data-tax-id');
+    console.log(pipelineId, taxLevel, taxId, 'fasta params');
+    location.href = `/pipeline_outputs/${pipelineId}/fasta/${taxLevel}/${taxId}/NT_or_NR`;
+  }
+
+  displayTags(taxInfo, reportDetails) {
+    return (
+      <span className="link-tag">
+        <i data-tax-id={taxInfo.tax_id} onClick={this.gotoNCBI} className="fa fa-link cloud" aria-hidden="true"></i>
```
yunfangjuan: if taxInfo.tax_id is negative, we shouldn't show the link. The negative taxid is internal for our use, which usually means genus/family doesn't exist

app/assets/javascripts/components/PipelineSampleReport.jsx
```diff
@@ -63,19 +63,49 @@ class PipelineSampleReport extends React.Component {
...
+    const taxId = e.target.getAttribute('data-tax-id');
+    console.log(pipelineId, taxLevel, taxId, 'fasta params');
+    location.href = `/pipeline_outputs/${pipelineId}/fasta/${taxLevel}/${taxId}/NT_or_NR`;
+  }
+
+  displayTags(taxInfo, reportDetails) {
+    return (
+      <span className="link-tag">
+        { taxInfo.tax_id > 0 ? <i data-tax-id={taxInfo.tax_id} onClick={this.gotoNCBI} className="fa fa-link cloud" aria-hidden="true"></i> : null }
+        { taxInfo.tax_id > 0 && reportDetails.taxon_fasta_flag ? <i data-pipeline-id={reportDetails.pipeline_info.id}
```
yunfangjuan: you might want to check with @cdebourcy on if the download link should be displayed when tax_id is less than 0.

cdebourcy: yes, link should be displayed even when tax_id < 0 -- that way, the user can inspect these "uncategorized" reads

PR #322, opened 2017-11-14
fix bug

cdebourcy:


aws_batch/pipeline.py
```diff
@@ -785,6 +797,9 @@ def run_sample(sample_s3_input_path, file_type, filter_host_flag, sample_s3_outp
         result_dir, sample_s3_output_path, False)
 
     # run_generate_taxid_annotated_fasta_from_m8
+    if not filter_host_flag:
```
yunfangjuan: how about instead of checking the flag. run this step when the merged fasta doesn't exist.

PR #325, opened 2017-11-14
Charles/pipeline separation and input chunking

cdebourcy:
# Description
- Breaks the pipeline into 3 separate stages (3 scripts: host_filtering.py, non_host_alignment.py, postprocess.py; all 3 use functions from common.py).
- For gsnap/rapsearch steps, breaks input files into small chunks and processes chunks sequentially -- each chunk will have to wait for resources independently once the previous chunk is finished. That way smaller jobs can get into gsnap/rapsearch processing even before an earlier longer job is completely done.
- Pipeline kickoff commands in the rails app will need to be changed to use these new scripts. The folder structure in S3 will need to be changed: the "results" folder needs to be split into a "stage 1 results" and a "stage 2 results" folder for lazy-run to work efficiently (avoid copying large files from stage 1 when executing stage 2).

## Type of change
- New feature (non-breaking change which adds functionality)
- Breaking change (fix a feature that knowingly causes existing functionality to not work as expected)

# Test and Reproduce
For each one of the three scripts :
1. Inspect the main() function of the script for the environment variables it uses.
2. Run the script with these environment variables set as needed for an example sample of your choice. E.g.:
```
BASE_PATH=s3://czbiohub-idseq-samples-alpha/samples/11/199;
sudo INPUT_BUCKET=$BASE_PATH/results OUTPUT_BUCKET=$BASE_PATH/postprocess python postprocess.py
```
3. Check if script runs without errors and produces the desired outputs.

Data processing between upload and report display.

# To be tested extensively


yunfangjuan:
Maybe it's time for us to create a python package for all the pipeline code. 
And we can just 
pip install python <package_name> in the container.

aws_batch/non_host_alignment.py
```diff
@@ -554,514 +399,455 @@ def wait_for_server_ip(service_name, key_path, remote_username, environment, max
...
-    STATS.append({'total_reads': get_total_initial_reads(fastq_files[0], initial_file_type_for_log, stats_file)})
-
-    # run gsnap remotely
-    if filter_host_flag:
-        gsnapl_input_files = [EXTRACT_UNMAPPED_FROM_SAM_OUT1, EXTRACT_UNMAPPED_FROM_SAM_OUT2]
-        before_file_name_for_log = os.path.join(result_dir, EXTRACT_UNMAPPED_FROM_SAM_OUT1)
-        before_file_type_for_log = "fasta_paired"
+def check_s3_file_presence(s3_path):
+    command = "aws s3 ls %s | wc -l" % s3_path
+    return int(execute_command_with_output(command).rstrip())
```
yunfangjuan: I think this is a bug.  
when you aws s3 ls $non_existence file, it will return 1. you will have to catch it, otherwise the script will fail

aws_batch/non_host_alignment.py
```diff
@@ -554,514 +399,455 @@ def wait_for_server_ip(service_name, key_path, remote_username, environment, max
...
+        input_file_full_local_path = os.path.join(RESULT_DIR, input_file)
+        out_prefix = input_file_full_local_path + part_suffix
+        out_prefix_base = input_file + part_suffix
+        execute_command("split --numeric-suffixes -l %d %s %s" % (chunk_nlines, input_file_full_local_path, out_prefix))
+        execute_command("aws s3 cp %s/ %s/ --recursive --exclude '*' --include '%s*'" % (RESULT_DIR, SAMPLE_S3_OUTPUT_CHUNKS_PATH, out_prefix_base))
+        partial_files = [os.path.basename(partial_file) for partial_file in execute_command_with_output("ls %s*" % out_prefix).rstrip().split("\n")]
+        part_lists.append(partial_files)
+    input_chunks = [list(part) for part in zip(*part_lists)]
+    # e.g. [["input_R1.fasta-part-1", "input_R2.fasta-part-1"],["input_R1.fasta-part-2", "input_R2.fasta-part-2"],["input_R1.fasta-part-3", "input_R2.fasta-part-3"],...]
+    return input_chunks
```
yunfangjuan: cool

PR #334, opened 2017-11-17
Susan/restructuresamplesdata

sadelokiki:
# Description
Restructure/Format samples data to contain pipeline run info and statistics
Refactor search and paginate
Search within a project

## Type of change

# Test and Reproduce

1. 

Which pages?
- [x ] All Projects Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###All Project Page
* Switch between projects 
* Search for a sample

# Checklist:


app/helpers/samples_helper.rb
```diff
@@ -122,8 +122,30 @@ def samples_pipeline_run_info(samples)
...
+    filtered
+  end
+
+  def format_samples(samples)
+    formatted_samples = []
+    samples.each_with_index do |_sample, i|
+      job_info = {}
+      final_result = samples_output_data(samples)
+      pipeline_run_info = samples_pipeline_run_info(samples)
+      job_info[:job] = samples[i]
```
yunfangjuan: as discussed, :job is a very confusing name. Can you change it to db_sample

app/helpers/samples_helper.rb
```diff
@@ -122,8 +122,30 @@ def samples_pipeline_run_info(samples)
...
+      job_info[:statistics] = final_result[i]
```
yunfangjuan: this is also confusing, because there's job_stats,  let's call  it something  like derived_sample_output

app/helpers/samples_helper.rb
```diff
@@ -88,7 +88,7 @@ def samples_output_data(samples)
     final_result = []
     samples.each do |output|
       output_data = {}
-      pipeline_output = output.pipeline_runs.first ? output.pipeline_runs.first.pipeline_output : nil
+      pipeline_output = output.pipeline_runs && output.pipeline_runs.first ? output.pipeline_runs.first.pipeline_output : nil
```
yunfangjuan: i don't think you need to change this line. ouput.pipeline_runs will be [] if it's empty

app/controllers/home_controller.rb
```diff
@@ -1,48 +1,45 @@
+require 'will_paginate/array'
 class HomeController < ApplicationController
   before_action :login_required
   include SamplesHelper
+
   def home
     @all_project = Project.all
     project_id = params[:project_id]
+    search_query = params[:search]
```
yunfangjuan: call it name_search_query to be more informative?

app/controllers/home_controller.rb
```diff
@@ -1,48 +1,45 @@
...
+    filter_query = params[:filter]
     sort = params[:sort_by]
     @project_info = nil
```
yunfangjuan: status_filter_query to be more informative.

app/controllers/home_controller.rb
```diff
@@ -1,48 +1,45 @@
+require 'will_paginate/array'
 class HomeController < ApplicationController
   before_action :login_required
   include SamplesHelper
+
   def home
```
yunfangjuan: i think the logic here is awfully complicated. 
You should be able to stack your query by doing. 
results = Sample.includes(:pipeline_runs, :pipeline_outputs) 
if (params[:project_id]) 
  results = results.where(project_id: 9) 
end

if (params[:search]) 
  results = results.where(name like "%%") 
end 
.....

app/controllers/home_controller.rb
```diff
@@ -1,48 +1,45 @@
...
-    samples_info_output = samples_info(@samples) || {}
-    @final_result = samples_info_output[:final_result] || []
-    @pipeline_run_info = samples_info_output[:pipeline_run_info] || []
-  end
-
-  def search
-    project_id = params[:project_id]
-    search_query = params[:search]
-
-    @samples = if project_id.present?
```
yunfangjuan: similar comment as the home function.

app/controllers/home_controller.rb
sadelokiki: okay

app/controllers/home_controller.rb
sadelokiki: there is no longer a search method.

app/helpers/samples_helper.rb
```diff
@@ -122,8 +122,30 @@ def samples_pipeline_run_info(samples)
     pipeline_run_info
   end
 
-  def samples_info(samples)
-    { final_result: samples_output_data(samples),
-      pipeline_run_info: samples_pipeline_run_info(samples) }
+  def filter_samples(samples, query)
+    filtered = []
```
yunfangjuan: You can try something like
if query == 'UPLOADING' 
  samples = samples.where(pipeline_run_id: nil)
else
samples = samples..where(“pipeline_id in (select max(pipeline_id) from pipeline_runs group by sample_id”).where(job_status: filter_qury) 
end

app/helpers/samples_helper.rb
```diff
@@ -122,8 +122,30 @@ def samples_pipeline_run_info(samples)
...
+    samples.each do |output|
+      pipeline_run_status = output.pipeline_runs.first ? output.pipeline_runs.first.job_status : nil
+      if query == 'UPLOADING'
+        filtered.push(output) unless pipeline_run_status
```
yunfangjuan: change to 
samples = samples.where(status: 'created')

app/helpers/samples_helper.rb
```diff
@@ -122,8 +122,30 @@ def samples_pipeline_run_info(samples)
...
+      elsif pipeline_run_status == query
+        filtered.push(output)
```
yunfangjuan: change to 

samples = samples.joins("INNER JOIN pipeline_runs ON pipeline_runs.sample_id and samples.id").where(status: 'checked').where("pipeline_runs.id in (select max(id)  from pipeline_runs group by sample_id)").where("pipeline_runs.job_status = #{query}")

sadelokiki: returns inaccurate samples

PR #339, opened 2017-11-21
Susan/removeduplicate

sadelokiki:
Remove duplicate line

### DIFF
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -125,7 +125,6 @@ module SamplesHelper
   def filter_samples(samples, query)
     if query == 'UPLOADING'
       samples = samples.where(status: 'created')
-      samples = samples.where(status: 'created')
     else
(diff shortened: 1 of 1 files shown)


app/helpers/samples_helper.rb
```diff
@@ -123,12 +123,11 @@ def samples_pipeline_run_info(samples)
...
-    if query == 'UPLOADING'
-      samples = samples.where(status: 'created')
-      samples = samples.where(status: 'created')
-    else
-      samples = samples.joins("INNER JOIN pipeline_runs ON pipeline_runs.sample_id = samples.id").where(status: 'checked').where("pipeline_runs.id in (select max(id) from pipeline_runs group by sample_id)").where("pipeline_runs.job_status = '#{query}'")
-    end
+    samples = if query == 'UPLOADING'
+                samples.where(status: 'created')
+              else
+                samples.joins("INNER JOIN pipeline_runs ON pipeline_runs.sample_id = samples.id").where(status: 'checked').where("pipeline_runs.id in (select max(id) from pipeline_runs group by sample_id)").where("pipeline_runs.job_status = '#{query}'")
```
cyrielo: You probably want to escape the query param, as oppose to passing the value directly into the sql, to avoid sql injection

PR #347, opened 2017-11-28
Susan/multistage ui

sadelokiki:
- Display chunked pipeline run stages in UI


app/assets/javascripts/components/samples.jsx
```diff
@@ -54,6 +54,17 @@ class Samples extends React.Component {
       let dbSample = sample.db_sample;
       let derivedOutput = sample.derived_sample_output;
       let runInfo = sample.run_info
+      rowWithChuckStatus = (
```
yunfangjuan: you mean Chunk?

app/assets/javascripts/components/samples.jsx
```diff
@@ -74,12 +85,38 @@ class Samples extends React.Component {
...
     } else if (status === 'INITIALIZING') {
       return 'initializing';
-    } else {
+    } else if (status === 'FAILED') {
+      return 'failed';
+    }
+  }
+
+  applyChunkStatusClass(postProcess, gsnapAlignment, hostFiltering) {
+    if (postProcess === 'LOADED' || gsnapAlignment ===   'LOADED' || hostFiltering === 'LOADED') {
```
yunfangjuan: it should just be called Alignment because there're two types of alignments: gsnapl/rapsearch2.

app/assets/javascripts/components/samples.jsx
```diff
@@ -74,12 +85,38 @@ class Samples extends React.Component {
...
+    } else {
+      return 'initializing';
+    }
+  }
+
+  getChunckedStatus(postProcess, gsnapAlignment, hostFiltering) {
+    if (postProcess && gsnapAlignment === 'LOADED' && hostFiltering === 'LOADED') {
+      return  `Post Processing: ${postProcess}`;
+    } else if (postProcess === null && gsnapAlignment && hostFiltering === 'LOADED') {
+      return `Gsnap Alignment: ${gsnapAlignment}`;
```
yunfangjuan: Just alignment. please.

app/assets/javascripts/components/samples.jsx
```diff
@@ -74,12 +85,38 @@ class Samples extends React.Component {
...
-    } else {
+    } else if (status === 'FAILED') {
+      return 'failed';
+    }
+  }
+
+  applyChunkStatusClass(postProcess, gsnapAlignment, hostFiltering) {
+    if (postProcess === 'LOADED' || gsnapAlignment ===   'LOADED' || hostFiltering === 'LOADED') {
+      return 'complete';
+    } else if (postProcess === 'CHECKED' || gsnapAlignment === 'CHECKED' || hostFiltering === 'CHECKED') {
```
yunfangjuan: There are also 'RUNNING' and 'RUNNABLE' stata, we should just show them.

app/assets/javascripts/components/samples.jsx
```diff
@@ -74,15 +85,37 @@ class Samples extends React.Component {
...
+
+  getChunckedStatus(postProcess, alignment, hostFiltering) {
+    if (postProcess && alignment === 'LOADED' && hostFiltering === 'LOADED') {
+      return  'Post Processing';
+    } else if (postProcess === null && alignment && hostFiltering === 'LOADED') {
+      return 'Alignment';
+    } else if (postProcess === null && alignment === null && hostFiltering) {
+      return 'Host Filtering';
+    } else {
+      return 'WAITING';
```
yunfangjuan: where is FAILED status?

app/helpers/samples_helper.rb
```diff
@@ -100,38 +99,61 @@ def samples_output_data(samples)
...
+    elsif query == 'UPLOADING'
+      db_query = [query, '1.Host Filtering-STARTED', '1.Host Filtering-CHECKED', '1.Host Filtering-LOADED',
+                  '2.GSNAPL/RAPSEARCH alignment-STARTED', '2.GSNAPL/RAPSEARCH alignment-CHECKED', '2.GSNAPL/RAPSEARCH alignment-LOADED', '3.Post Processing-STARTED', '3.Post Processing-CHECKED']
+      samples = samples.joins("INNER JOIN pipeline_runs ON pipeline_runs.sample_id = samples.id").where(status: 'checked').where("pipeline_runs.id in (select max(id) from pipeline_runs group by sample_id)").where("pipeline_runs.job_status IN (?)", db_query)
+    elsif query == 'CHECKED'
+      db_query = [query, '3.Post Processing-CHECKED']
+      samples = samples.joins("INNER JOIN pipeline_runs ON pipeline_runs.sample_id = samples.id").where(status: 'checked').where("pipeline_runs.id in (select max(id) from pipeline_runs group by sample_id)").where("pipeline_runs.job_status IN (?)", db_query)
+    else samples = samples
```
yunfangjuan: new line after else

app/helpers/samples_helper.rb
```diff
@@ -100,38 +99,61 @@ def samples_output_data(samples)
...
   end
 
+  def filter_samples(samples, query)
+    if query == 'WAITING'
+      samples = samples.joins("LEFT OUTER JOIN pipeline_runs ON pipeline_runs.sample_id = samples.id").where("samples.status = ?  or pipeline_runs.job_status is NULL", 'created')
+    elsif query == 'FAILED'
+      db_query = [query, '1.Host Filtering-FAILED', '2.GSNAPL/RAPSEARCH alignment-FAILED', '3.Post Processing-FAILED', 'ERROR']
+      samples = samples.joins("INNER JOIN pipeline_runs ON pipeline_runs.sample_id = samples.id").where(status: 'checked').where("pipeline_runs.id in (select max(id) from pipeline_runs group by sample_id)").where("pipeline_runs.job_status IN (?)", db_query)
+    elsif query == 'UPLOADING'
+      db_query = [query, '1.Host Filtering-STARTED', '1.Host Filtering-CHECKED', '1.Host Filtering-LOADED',
```
yunfangjuan: this is problematic because there are also status like RUNNING and RUNNABLE. 
To make it simple you can just do 

where("pipeline_runs.job_status NOT IN (?) and pipeline_runs.finalized != 1",  ['CHECKED',  'FAILED'])

app/helpers/samples_helper.rb
```diff
@@ -100,38 +99,61 @@ def samples_output_data(samples)
...
-          'COMPLETE'
-        elsif %w[FAILED ERROR].include?(pipeline_run_status)
-          'FAILED'
-        elsif %w[RUNNING LOADED].include?(pipeline_run_status)
-          'IN PROGRESS'
-        elsif pipeline_run_status == 'RUNNABLE'
-          'INITIALIZING'
+      if output.pipeline_runs.first
+        recent_pipeline_run = output.pipeline_runs.first
+        pipeline_run_entry[:job_status_description] = 'WAITING' if recent_pipeline_run.job_status.nil?
```
yunfangjuan: next

app/helpers/samples_helper.rb
```diff
@@ -100,38 +99,61 @@ def samples_output_data(samples)
...
+        if recent_pipeline_run.pipeline_run_stages.present?
+          run_stages = recent_pipeline_run.pipeline_run_stages || []
+          host_filtering = run_stages ? run_stages.first.job_status : nil
```
yunfangjuan: this is not true. It's possible we skip the host filtering for same samples. 
The right way should be 

run_stages.each do |rs|
  pipeline_run_entry[rs[:name]] = rs
end

app/helpers/samples_helper.rb
```diff
@@ -100,38 +99,61 @@ def samples_output_data(samples)
...
+          pipeline_run_entry[:host_filtering] = host_filtering
         else
-          'UPLOADING'
+          pipeline_run_status = recent_pipeline_run.job_status
+          pipeline_run_entry[:job_status_description] =
+            if %w[CHECKED SUCCEEDED].include?(pipeline_run_status)
+              'COMPLETE'
+            elsif %w[FAILED ERROR].include?(pipeline_run_status)
+              'FAILED'
+            elsif %w[RUNNING LOADED].include?(pipeline_run_status)
```
yunfangjuan: Hm LOADED shouldn't be here.

app/assets/javascripts/components/samples.jsx
sadelokiki: this method only shows what stages are presently running for that sample, not the status.
the name will be changed to getChunkedStage @yunfangjuan

app/helpers/samples_helper.rb
```diff
@@ -100,38 +99,55 @@ def samples_output_data(samples)
     final_result
   end
 
+  def filter_samples(samples, query)
+    samples = if query == 'WAITING'
+                samples.joins("LEFT OUTER JOIN pipeline_runs ON pipeline_runs.sample_id = samples.id").where("samples.status = ?  or pipeline_runs.job_status is NULL and pipeline_runs.finalized != 1", 'created')
+              elsif query == 'FAILED'
+                samples.joins("INNER JOIN pipeline_runs ON pipeline_runs.sample_id = samples.id").where(status: 'checked').where("pipeline_runs.id in (select max(id) from pipeline_runs group by sample_id)").where("pipeline_runs.job_status IN (?) and pipeline_runs.finalized = 1", [query || 'ERROR'])
```
yunfangjuan: this is problematic. "ERROR" is also a transient status and the job status will say: "step-name FAILED" 
It should be changed to 
samples.joins("INNER JOIN pipeline_runs ON pipeline_runs.sample_id = samples.id").where(status: 'checked').where("pipeline_runs.id in (select max(id) from pipeline_runs group by sample_id)").where("pipeline_runs.job_status like '%FAILED' ")
and in this case you don't have to change finalized at all.

PR #370, opened 2017-12-04
Cyril/refine single sample upload design

cyrielo:
# Description
This PR incorporates the new design layout for uploading a single sample

## Type of change

Which pages?
- [x] Single Upload Page

Specific components 

* SampleUpload

# Testing Script:

### Single Upload

# Checklist:


app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -51,18 +51,20 @@ class SampleUpload extends React.Component {
       selectedJobQueue: this.selected.jobQueue || '',
```
jameshowardwang: i know this wasn't part of this diff review but in line 23 of this file, there is a reference to hostGenomes[0] which breaks if there aren't any hostGenomes uploaded yet (e.g. when I do a ./bin/setup from scratch). could you please fix that, either as part of this diff, or separately?

cyrielo: Sure would reproduce this error and apply a fix

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -267,106 +263,145 @@ class SampleUpload extends React.Component {
...
-        errorMessage: 'Please select a host genome'
-      })
-      return true;
+    const errors = {};
+    if(this.state.selectedProject) {
+      if(this.state.selectedProject.toLowerCase() === 'select a project') {
+        errors.selectedProject = 'Please select a project';
+      }
+    } else {
+      errors.selectedProject = 'Please select a project';
```
cdebourcy: Could you make text like 'select a project' and 'Please select a project' constants so that we only have to change them in the code once if we want to modify them?

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -127,7 +131,7 @@ class SampleUpload extends React.Component {
     .catch((error) => {
       that.setState({
         invalid: true,
-        errorMessage: 'Project exists already or is invalid',
+        errors: { 'selectedProject': 'Project already exists or is in-valid' },
```
cdebourcy: I think it should be `invalid` rather than `in-valid`

cdebourcy:
It doesn't seem to quite work yet:
screen shot 2017-12-04 at 10 48 13 am

app/assets/javascripts/components/sampleUpload.jsx
cdebourcy: I think the solution would be to seed the database with a host genome as part of bin/setup. I can add that

cyrielo: Yeah I have also at least prevented the page from breaking if there aren't any host genomes

PR #367, opened 2017-11-30
Susan/check status in dropdown

sadelokiki:
# Description
Display checkmarks on filter when selected in dropdown

## Type of change
- [x] New feature (non-breaking change which adds functionality)

# Test and Reproduce
Outline the steps to test or reproduce the PR here.

1. 

## Impacted Areas in Application
List general components of the application that this PR will affect:

Which pages?
- [ ] All Projects Page

Specific components 

* 

# Testing Script:

###All Project Page
* Switch between projects 
* Search for a sample
* Navigate through different pages of samples
* Select Upload new sample button
* Select sample to see generated report

# Checklist:

- [x] I have run through the testing script to make sure current functionality is unchanged
- [x] I have done relevant tests that prove my fix is effective or that my feature works
- [x] I have spent time testing out edge cases for my feature
- [x] I have updated the test script or pull request template if necessary
- [x] New and existing unit tests pass locally with my changes

PR #371, opened 2017-12-04
Susan/lazyloading

sadelokiki:
# Description

Infinite scroll on projects page
Remove page reloads

## Type of change

# Test and Reproduce

1. 

Which pages?
- [ ] All Projects Page

Specific components 

* 

# Testing Script:

###All Project Page
* Switch between projects 
* Search for a sample

# Checklist:


yunfangjuan:
In general, this looks good but since the URL no longer changes, it's no longer obvious if there's a pipeline_run status filter. I think it needs to be clearer there. Otherwise, when you change projects, it's not clear it's showing all samples or samples that have certain status.

app/controllers/project_samples_controller.rb
```diff
@@ -0,0 +1,34 @@
+class ProjectSamplesController < ApplicationController
```
yunfangjuan: why do you need a separate controller for this? why don't you just put this under samples? 
I think you can move the current samples#index controller to some where else and just use samples#index instead of creating a new one.

sadelokiki:
hey @yunfangjuan, there is an existing PR that shows a checkmark when a filter is chosen here https://github.com/chanzuckerberg/idseq-web/pull/367. If this is not sufficient, maybe a design tweak to illustrate this would help @eggerdesigns. Thanks

yunfangjuan:
As discussed, I think when a user go to an url like 

https://idseq.net/?filter=CHECKED&project_id=12 

we should still honor the project_id and filter selection on the first load. 

Currently it ignores them but we do have links go to https://idseq.net/?project_id=X on each sample report page. Please change that.

app/assets/javascripts/components/samples.jsx
```diff
@@ -148,6 +156,113 @@ class Samples extends React.Component {
...
+    this.setState({ loading: true })
+    this.fetchProjects();
+    this.fetchSamples();
+  }
+
+  //load more paginated samples
+  loadMore() {
+    const params = this.getParams();
+    this.setState({ isRequesting: true })
+    axios.get(`/samples${params}`).then((res) => {
```
yunfangjuan: this seems more intuitive to put ? in here instead of in params

app/assets/javascripts/components/samples.jsx
```diff
@@ -148,6 +156,113 @@ class Samples extends React.Component {
...
+        pageEnd: prevState.pageEnd
+      }))
+    })
+  }
+
+  //fetch project, filter and search params
+  getParams() {
+    const projectId = parseInt(this.state.selectedProjectId);
+    
+    const paramsWithProject = `?filter=${this.state.filterParams}&page=${this.state.pagesLoaded+1}&project_id=${projectId}&search=${this.state.searchParams}`;
```
yunfangjuan: simplify this with 
let params = `?filter=${this.state.filterParams}&page=${this.state.pagesLoaded+1}&search=${this.state.searchParams}`
if project_id !== null {
  params += `&project_id=${project_id}` 
}
return params

PR #386, opened 2017-12-06
update automatic sample name generation mechanism

cyrielo:
# Description
This PR generates a sample name from the file path and gives the user ability to edit the name

## Type of change

Which pages?

- [x] Single Upload Page

Specific components 

* SampleUpload

# Testing Script:

###Single Upload

# Checklist:

### DIFF
--- a/app/assets/javascripts/components/sampleUpload.jsx
+++ b/app/assets/javascripts/components/sampleUpload.jsx
@@ -18,7 +18,7 @@ class SampleUpload extends React.Component {
     this.hostGenomes = props.host_genomes || [];
     this.sample = props.selectedSample || '';
     this.userDetails = props.loggedin_user;
-    this.updateGeneratedSampleName = this.updateGeneratedSampleName.bind(this);
+    this.updateSampleName = this.updateSampleName.bind(this);
     const selectedHostGenomeName = (this.hostGenomes[0] && this.hostGenomes[0].name) ? this.hostGenomes[0].name : '';
     const selectedHostGenomeId = (this.hostGenomes[0] && this.hostGenomes[0].id) ? this.hostGenomes[0].id : '';
     const adminGenomes = this.hostGenomes.filter((g) => {
@@ -59,7 +59,7 @@ class SampleUpload extends React.Component {
       id: this.selected.id,
       errors: {},
       adminGenomes,
-      generatedSampleName: ''
+      sampleName: ''
     };
   }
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -778,6 +803,32 @@ class SampleUpload extends React.Component {
...
+                      <div className='field-title'>
+                        <div className='read-count-label'>
+                          Sample name
+                        </div>
+                      </div>
+                    </div>
+                  </div>
+                  <div className='row input-row'>
+                    <div className='col no-padding s12'>
+                      <input type='text' ref='sample_name' className='browser-default' value={this.state.generatedSampleName} onChange={ (e) => {
```
yunfangjuan: I think sample name should be above the admin options.

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -778,6 +803,32 @@ class SampleUpload extends React.Component {
...
+                        this.setState({
+                            generatedSampleName: e.target.value.trim()
```
yunfangjuan: I think you should also set a state so if people change Read path, the sample name  entered by the user won't be  overwritten. Also I think the function should be extracted out.

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -57,7 +58,8 @@ class SampleUpload extends React.Component {
       selectedMemory: this.selected.memory || '',
       id: this.selected.id,
       errors: {},
-      adminGenomes
+      adminGenomes,
+      generatedSampleName: ''
```
yunfangjuan: this should just be called sample name.

app/assets/javascripts/components/sampleUpload.jsx
```diff
@@ -412,6 +425,17 @@ class SampleUpload extends React.Component {
     }
   }
 
+  updateGeneratedSampleName(e) {
+    let value = e.target.value.trim();
```
yunfangjuan: as mentioned, only do this when a user not editing the name.

app/assets/javascripts/components/sampleUpload.jsx
cyrielo: Yeah we didn't want to have the user manually type in a sample name, so we put it at the bottom so we can generate it and have the user edit it

yunfangjuan: but it's below the admin options. users should be able to edit it. 
it's odd to have <user options> <admin options> and then <user options> again.

yunfangjuan:
Can you please also move the sample name field to be above the admin options?

PR #398, opened 2017-12-12
Charles/move thresholds

cdebourcy:
# Description
Moves threshold filters to column headers. Hit enter to apply.

## Type of change

- [X] Report Page


app/lib/report_helper.rb
```diff
@@ -453,20 +463,19 @@ def sort_key(tax_2d, tax_info, sort_by)
...
     rows.each do |tax_info|
       should_delete = false
-      # if any metric is below threshold in every type, delete
-      METRICS.each do |metric|
-        should_delete = COUNT_TYPES.all? do |count_type|
+      # if any metric is below threshold in the specified type, delete
+      METRICS.any? do |metric|
+        COUNT_TYPES.any? do |count_type|
+          should_delete = !(tax_info[count_type][metric]) || (thresholds[count_type][metric] && tax_info[count_type][metric] < thresholds[count_type][metric])
           # aggregatescore is null for genera, and thus not considered in filtering genera
```
boris-dimitrov: Is this comment still true?

cdebourcy: I don't believe I made any changes that affect its truth

PR #410, opened 2017-12-15
metadata upload bug fix

cdebourcy:
# Description
Bug fix: sample name is only unique in combination with project

### DIFF
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -2,21 +2,42 @@ require 'open3'
 require 'csv'
 
 module SamplesHelper
+  include ApplicationHelper
   include PipelineOutputsHelper
 
   def populate_metadata_bulk(csv_s3_path)
+    # Get list of available samples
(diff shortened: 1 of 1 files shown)


app/helpers/samples_helper.rb
```diff
@@ -5,14 +5,17 @@ module SamplesHelper
   include PipelineOutputsHelper
 
   def populate_metadata_bulk(csv_s3_path)
-    # CSV should have columns "sample_name" and any desired columns from Sample::METADATA_FIELDS
+    # CSV should have columns "sample_name", "project_name", and any desired columns from Sample::METADATA_FIELDS
     csv = get_s3_file(csv_s3_path)
     csv.delete!("\uFEFF") # remove BOM if present (file likely comes from Excel)
     CSV.parse(csv, headers: true) do |row|
       h = row.to_h
+      sample_project = Project.find_by(name: h['project_name'])
```
cyrielo: We can simplify the SQL queries using JOINs to be performant optimal instead of having them in the iterations

PR #413, opened 2017-12-16
Cyril/batch upload design update

cyrielo:
# Description
Pushing the updates for bulk upload

## Type of change

Which pages?

- [x] Single Upload Page
- [x] Batch Upload Page

Specific components 

### Single Upload

### Batch Upload

# Checklist:

### DIFF
--- a/app/assets/javascripts/components/bulkUploadImport.jsx
+++ b/app/assets/javascripts/components/bulkUploadImport.jsx
@@ -15,6 +15,11 @@ class BulkUploadImport extends React.Component {
     this.handleHostChangeForSample = this.handleHostChangeForSample.bind(this);
     this.handleBulkPathChange = this.handleBulkPathChange.bind(this);
     this.selectSample = this.selectSample.bind(this);
+    this.toggleNewProjectInput = this.toggleNewProjectInput.bind(this);
+    this.adminGenomes = this.hostGenomes.map((g) => {
+      return g.name.toLowerCase().indexOf('test') >= 0 ? g.name : '';
+    });
+    this.userDetails = props.loggedin_user;
     this.state = {
       submitting: false,
       allProjects: props.projects || [],
@@ -33,7 +38,15 @@ class BulkUploadImport extends React.Component {
       selectedSampleIndices: [],
       createdSampleIds: [],
       allChecked: false,
-    }
+      disableProjectSelect: false,
+      errors: {},
+      serverErrors: null
+    };
+  }
+  componentDidUpdate() {
+    $('.custom-select-dropdown').dropdown({
+      belowOrigin: true
+    });
   }
 
   componentDidMount() {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -248,32 +278,29 @@ class BulkUploadImport extends React.Component {
...
     }
-    return false;
+    if(this.refs.bulk_path) {
+      if (this.refs.bulk_path.value === '') {
+        errors.bulk_path = 'Please fill in the S3 bulk_path path';
+      } else if(!this.filePathValid(this.refs.bulk_path.value)) {
+        errors.bulk_path = 'S3 bulk path is invalid';
+      }
+    } else {
+      errors.bulk_path = 'Please fill in the S3 bulk_path path';
```
cdebourcy: Should be `errors.bulk_path = 'Please fill in the S3 bulk path';`?

app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -15,6 +15,11 @@ class BulkUploadImport extends React.Component {
     this.handleHostChangeForSample = this.handleHostChangeForSample.bind(this);
     this.handleBulkPathChange = this.handleBulkPathChange.bind(this);
     this.selectSample = this.selectSample.bind(this);
+    this.toggleNewProjectInput = this.toggleNewProjectInput.bind(this);
+    this.adminGenomes = this.hostGenomes.map((g) => {
+      return g.name.toLowerCase().indexOf('test') >= 0 ? g.name : '';
```
cdebourcy: nice!

app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -333,131 +380,310 @@ class BulkUploadImport extends React.Component {
...
+                          </div> : null
+                      }
+                    </div>
+                  </div>
+                </div>
+                <div className='field'>
+                  <div className='row'>
+                    <div className='col field-title no-padding s5 tooltipped'
+                         data-position="top" data-delay="50"
+                         data-tooltip='This would be subtracted by the pipeline'>
```
cdebourcy: This tooltip is not showing up for me. 
And text should be "This will be subtracted by the pipeline" ?

app/assets/javascripts/components/bulkUploadImport.jsx
```diff
@@ -333,131 +380,310 @@ class BulkUploadImport extends React.Component {
...
+                                     id={g.name} onClick={() => this.handleHostChange(g.id, g.name)}>
+                                       { g.name }
+                                   </div>
+                                 );
+                              }
+                            })
+                          }
+                        </div> : null
+                    }
+                    </div>
```
cdebourcy: I think it would be nice to extract the block from line 581 to 598 into a function display_admin_genomes so that it is (a) easier to keep an overview of all the different elements being displayed on the page and (b) easier to figure out which lines to modify if we ever want to change the way admin genomes are displayed going back to this code months from now. Similar comment for lines 603-622 or 535-539

cyrielo: right, but it gets harder to keep track of the HTML elements structure if the blocks extracted is pretty long. you won't be able to tell the children element without having to scroll all the way up and down.

cdebourcy: Hmm ok. Well, maybe it's not too big a deal here, but in general I think it would be a good habit to avoid having too much inline logic scattered around the code, especially repeated logic (e.g. lines 437 and 586). See for example how much good that practice did for the readability of the report page, even with the small disadvantage of not having all html tags in one place

PR #417, opened 2017-12-18
Charles/column selection

cdebourcy:
# Description
Add or remove columns on the sample list page. 

## Type of change

# Test and Reproduce

1. Go to front page and play with the new dropdown.

Which pages?
- [X] All Projects Page


cdebourcy:
Ah. It seems like that problem arises when some of the fields have really long entries.

PR #422, opened 2017-12-19
apply threshold on focus-out

cdebourcy:
# Description
Apply thresholds when you click outside of threshold field.


cdebourcy:
yes, I think we should -- I'd expect to be able to submit the threshold both ways (enter / focus-out)

PR #436, opened 2017-12-21
Susan/favoriteprojects

sadelokiki:
# Description

This PR implements the ability to add and remove favorite projects from a list of projects. 

## Type of change

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [x] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

### DIFF
--- a/app/assets/javascripts/components/samples.jsx
+++ b/app/assets/javascripts/components/samples.jsx
@@ -14,7 +14,9 @@ class Samples extends React.Component {
     this.setUrlLocation = this.setUrlLocation.bind(this);
     this.sortSamples = this.sortSamples.bind(this);
     this.switchColumn = this.switchColumn.bind(this);
-    this.pageSize = props.pageSize || 30;
+    this.uploadSample = this.uploadSample.bind(this);
+    this.toggleDisplayProjects = this.toggleDisplayProjects.bind(this);
+    this.pageSize = props.pageSize || 30
     this.state = {
       project: null,
       totalNumber: null,
@@ -22,6 +24,7 @@ class Samples extends React.Component {
       filterParams: this.fetchParams('filter') || '',
       searchParams: this.fetchParams('search') || '',
       sampleIdsParams: this.fetchParams('ids') || [],
+      favouriteProjects: [],
       allSamples: [],
       allProjects: [],
       sort_by: this.fetchParams('sort_by') || 'id,desc',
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/components/samples.jsx
```diff
@@ -434,6 +447,103 @@ class Samples extends React.Component {
...
+    }
+  }
+
+
+  // add Projects to favorites list
+  addProjectToFavorites(id, projects) {
+    let updatedProject = projects.filter(project => project.id == id);
+
+    this.setState({
+        formattedFavProjectList: [...this.state.formattedFavProjectList, ...updatedProject],
```
yunfangjuan: do you need the second ... here?

app/controllers/projects_controller.rb
```diff
@@ -65,6 +61,14 @@ def visuals
     end
   end
 
+  def favorite
+    if current_user.favorites.include? @project
```
yunfangjuan: this is a bit dangerous. if the call is sent twice advertently, it will undo what it is supposed to do.  can you do add_favorite and remove_favorite instead?

app/assets/javascripts/components/samples.jsx
sadelokiki: yes, the second is also an array, we join them using the spread syntax.

yunfangjuan:
The functionality works well but a few quick feedback on the user experience
1. There's a lot of white space when my favorite project list is empty and the list of projects got pushed down below the fold. Can we not show the white space when favorite project list is short? 
2. When clicking on the project name, it doesn't update the samples unless I click on the actual link. Can you make the whole cell clickable? It's a bit confusing.

app/controllers/projects_controller.rb
sadelokiki: yes @yunfangjuan, this method is meant to do exactly that, toggle favorite/unfavorite, currently on the client side, the user will only be allowed to make the call for unfavorite if the project is favorited already and vice versa.

yunfangjuan: well but I think from the backend standpoint, it's not a deterministic action. sometimes, it's delete. sometimes it's add. It can make the future debugging hard to do. I think we should still break it do add_favorite and remove_favorite. if the frontend accidentally call the api twice (which could happen in some brower), it won't do anything.

PR #439, opened 2017-12-22
add logging for corrupt m8 files

cdebourcy:
# Description
A number of samples fail because the m8 file seems incompletely written (truncated line). This PR adds some logging to diagnose those cases better. Eventually we should switch to s3mi and have proper file integrity controls.


boris-dimitrov:
this looks good
may be logging the line number would be good too

PR #443, opened 2018-01-03
Charles/re org pipeline run

cdebourcy:
# Description
When a new pipeline_run is kicked off, the info on older pipeline_runs for the sample should get archived in a file in S3. 
There is a rake task for one-time use to archive all existing pipeline_runs, except the most recent one for each sample.

## Type of change

# Test and Reproduce

1. Call the function archive_old_pipeline_runs on a sample with multiple pipeline_runs. All but the most recent one should get archived in "s3://czbiohub-idseq-samples-development/pipeline_runs/#{id}" and deleted from the mysql database.

2. Rerun a sample and verify that a new pipeline_run gets created and older ones get archived as above.

3. Run the rake task to see if archiving works in bulk for all samples.

Which pages?
- samples/all page
- sample upload

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -25,7 +25,7 @@ class PipelineRun < ApplicationRecord
   after_create :kickoff_job
 
   def archive_s3_path
-    's3://#{SAMPLES_BUCKET_NAME}/pipeline_runs/#{id}'
+    "s3://#{SAMPLES_BUCKET_NAME}/pipeline_runs/#{id}"
   end
 
   def self.in_progress
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -24,6 +24,10 @@ class PipelineRun < ApplicationRecord
   before_save :check_job_status
   after_create :kickoff_job
 
+  def archive_s3_path
+    's3://#{SAMPLES_BUCKET_NAME}/pipeline_runs/#{id}'
+  end
+
   def self.in_progress
     where("job_status != '#{STATUS_FAILED}' OR job_status IS NULL")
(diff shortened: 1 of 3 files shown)


app/models/pipeline_run.rb
```diff
@@ -24,6 +24,10 @@ class PipelineRun < ApplicationRecord
   before_save :check_job_status
   after_create :kickoff_job
 
+  def archive_s3_path
+    "s3://#{SAMPLES_BUCKET_NAME}/pipeline_runs/#{id}"
```
yunfangjuan: can we embed sample_id in there too?

app/models/sample.rb
```diff
@@ -209,12 +210,34 @@ def check_status
...
+    old_pipeline_runs.each do |pr|
+      # Write pipeline_run data to file
+      file = Tempfile.new
+      file.write(pr.to_json(include: :pipeline_run_stages))
+      file.close
+      # Copy file to S3
+      pr_s3_file_name = "pipeline_run_#{pr.id}.json"
+      command = "aws s3 cp #{file.path} #{pr.archive_s3_path}/#{pr_s3_file_name}"
+      _stdout, _stderr, status = Open3.capture3(command)
+      # Delete pipeline_run
```
yunfangjuan: Hm I think we should keep the database records but remove the taxon_counts, taxon_byte_ranges. because if we delete all the records, we won't know what's going on and we can't link back to the cloud watch logs.

app/models/pipeline_run.rb

app/models/sample.rb
cdebourcy: Oh I see

PR #446, opened 2018-01-04
require admin authentication to mass assign role

cdebourcy:
# Description
Only permit mass assignment of role if user is properly authenticated as admin.


yunfangjuan:
great

PR #449, opened 2018-01-05
tweak gsnap options for i3.16xlarge according to perf experimenrts

boris-dimitrov:
pasting from experiment report published in slack proj-biohub
on Nov 21, 2017:

bdimitrov [7:16 PM]
My recommendation is to split paired-end reads jobs into chunks of 31,000 reads.
A chunk running alone on a 64 vCPU machine should take 5 minutes.
About 90% of max throughput is achieved running 4 chunks concurrently
with -t 32 each, in batch mode 0.   Command:

nohup time /mnt/nvme_raid8/gmap-gsnap/src/gsnapl.avx2 -A m8 --batch=0 --use-shared-memory=0 --gmap-mode=none --npaths=1 --ordered -t 32 --maxsearch=5  -D /mnt/nvme_raid8/chunked/nt_k16 -d nt_k16 r1.fasta.3 r2.fasta.3 &> chunk_3.m8 &

I've tried a variety of other scenarios, but this is overall the best.
Each chunk slows down from 5min to 8m 20s, but overall throughput
is nearly doubled.

The number of concurrent chunks can be increased quite a bit (I tried 3,4,5,6,7,8)
and nothing bad happens;  this is very convenient when scaling up/down based on load.
Throughput peaks at 5 whereas latency is best at 2.   For bigger jobs, even the
overall job latency is more determined by throughput, so my recommendation is 4,
skewed toward higher throughput .   If we found that the majority of jobs were tiny,
we could set it to 3.   Note each chunk gets 32 threads so it can use the entire
CPU if no other chunks are runnig.

### DIFF
--- a/aws_batch/non_host_alignment.py
+++ b/aws_batch/non_host_alignment.py
@@ -40,7 +40,7 @@ FILE_TYPE = 'fastq.gz'
 ENVIRONMENT = 'production'
 
 # compute capacity
-GSNAPL_MAX_CONCURRENT = 20 # number of gsnapl jobs allowed to run concurrently on 1 machine
+GSNAPL_MAX_CONCURRENT = 5 # number of gsnapl jobs allowed to run concurrently on 1 machine
 RAPSEARCH2_MAX_CONCURRENT = 3
 GSNAPL_CHUNK_SIZE = 30000 # number of fasta records in a chunk
 RAPSEARCH_CHUNK_SIZE = 10000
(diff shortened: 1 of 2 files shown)


boris-dimitrov:
Please note that deploying this needs to be coordinated with deploying the terraform change that copies the index fast and skips the loading of the index in shared memory.   Shared memory is no longer used.  Also after that change we can autoscale more aggressively since the time to spin up a gsnap instance is greatly reduced.

boris-dimitrov:
On i3.16xlarge I don’t expect OOM with even 8 concurrent jobs.  My recommendation was 4-5 and this change sets the limit to 5.

boris-dimitrov:
But I tested 8 with no problems.

boris-dimitrov:
4 is the best compromise between throughout and latency

boris-dimitrov:
Fundamentally, not using shared memory means each chunk loads only the parts of the index it wants.  Even with 8 chunks there is still room in the filesystem cache for the entire index.

PR #450, opened 2018-01-05
add brakeman to travis

cdebourcy:
Aims to append brakeman output to Travis report


cdebourcy:
I don't want it to have any effect on whether the build passes or not. But I do want the stdout of the command for informative purposes. Won't that work?

PR #452, opened 2018-01-06
check gsnap output format, retry if corrupt

cdebourcy:
# Description
When gsnap is done, check if all lines have the correct number of columns. If not, something went wrong so the chunk should be reprocessed.

### DIFF
--- a/aws_batch/non_host_alignment.py
+++ b/aws_batch/non_host_alignment.py
@@ -549,7 +549,7 @@ def run_gsnapl_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work
             min_column_number = 0
             max_tries = 2
             try_number = 1
-            while (min_column_number < correct_number_of_output_columns and try_number <= max_tries):
+            while (min_column_number != correct_number_of_output_columns and try_number <= max_tries):
                 write_to_log("waiting for server")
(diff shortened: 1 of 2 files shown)


aws_batch/non_host_alignment.py
```diff
@@ -544,16 +544,27 @@ def run_gsnapl_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work
...
-            gsnapl_instance_ip = wait_for_server_ip('gsnap', key_path, remote_username, ENVIRONMENT, GSNAPL_MAX_CONCURRENT)
-            write_to_log("starting alignment for chunk %s on machine %s" % (chunk_id, gsnapl_instance_ip))
-            remote_command = 'ssh -o "StrictHostKeyChecking no" -i %s %s@%s "%s"' % (key_path, remote_username, gsnapl_instance_ip, commands)
-            execute_command(remote_command)
-        # move gsnapl output back to local
+            correct_number_of_output_columns = 12
+            min_column_number = 0
+            max_tries = 2
+            try_number = 1
+            while (min_column_number < correct_number_of_output_columns and try_number <= max_tries):
```
boris-dimitrov: probably `!=` would be better than `<` here, to match the assert below

PR #454, opened 2018-01-09
pipeline migration

cdebourcy:
# Description
Switches to installing the pipeline package (https://github.com/chanzuckerberg/idseq-pipeline) instead of downloading individual scripts from S3. 

# Testing
1. Create a new sample and verify that
    - it gets processed correctly through all stages
    - logging works as desired
    - a file containing the sha of the pipeline commit used for processing is added to the result folder. 
2. Also rerun an old sample to test lazy run.


aws_batch/blacklist.py
```diff
@@ -1,34 +0,0 @@
-#!/usr/bin/env python
```
yunfangjuan: why remove this unrelated script?

aws_batch/indexing.py
```diff
@@ -1,148 +0,0 @@
-#!/usr/bin/env python
```
yunfangjuan: we might still need this?

aws_batch/Dockerfile
```diff
@@ -61,6 +61,9 @@ RUN wget http://launchpadlibrarian.net/162265652/fastx-toolkit_0.0.14-1_amd64.de
 RUN gdebi --non-interactive libgtextutils0_0.7-1_amd64.deb
 RUN gdebi --non-interactive fastx-toolkit_0.0.14-1_amd64.deb
 
+# Install IDseq pipeline package
+RUN pip install git+https://github.com/chanzuckerberg/idseq-pipeline.git
```
yunfangjuan: I actually don't think we should put this into the Docker image. We should install the package as part of the batch command. Updating dockerimage is kinda painful.

app/models/pipeline_run_stage.rb
```diff
@@ -139,9 +138,7 @@ def host_filtering_command
     if sample.s3_bowtie2_index_path.present?
       batch_command_env_variables += " BOWTIE2_GENOME=#{sample.s3_bowtie2_index_path} "
     end
-    batch_command = "aws s3 cp #{IdSeqPipeline::S3_HOST_FILTER_SCRIPT_LOC} .; chmod 755 #{script_name}; " \
-      "aws s3 cp #{IdSeqPipeline::S3_COMMON_SCRIPT_LOC} .; chmod 755 #{common_script_name}; " +
-                    batch_command_env_variables + " ./#{script_name}"
+    batch_command = batch_command_env_variables + " idseq_pipeline host_filtering"
```
yunfangjuan: nice, but maybe for the next diff we should move from environment variables to command arguments.

aws_batch/Dockerfile
ryanking: painful how? we can get it automated

aws_batch/indexing.py
cdebourcy: Ok I've put it back in this repo while also leaving it in idseq-pipeline. I do think it will be good to just make it a part of idseq-pipeline because eventually we want to update index regularly and automatically. Then the indexing script and the pipeline scripts should have identical versions (e.g. say, we switch to a new release of GSNAP). So it makes sense to have them in the same package

app/models/pipeline_run_stage.rb
```diff
@@ -2,6 +2,9 @@ class PipelineRunStage < ApplicationRecord
   include ApplicationHelper
   include PipelineOutputsHelper
   belongs_to :pipeline_run
+
+  PIPELINE_INSTALL_COMMAND = "pip install git+https://github.com/chanzuckerberg/idseq-pipeline.git".freeze
```
boris-dimitrov: You probably want `pip install --upgrade` here?   If ever reinstalling on same instance.

boris-dimitrov: This is fine assuming it only runs once on the instance/docker container, and there are no concurrent invocations.

cdebourcy:
Yes, idseq-pipeline is open-sourced

app/models/pipeline_run_stage.rb
cdebourcy: Thanks!

PR #458, opened 2018-01-10
add update build script to add yarn and node dependency for frontend bundling

cyrielo:
# Description
This PR updates the build script to add yarn and node for the npm dependency

## Type of change

Which pages?
 - None

# Testing Script:
  - 
# Checklist:

### DIFF
--- a/bin/reset
+++ b/bin/reset
@@ -1,0 +1,29 @@
+#!/usr/bin/env ruby
+require 'pathname'
+require 'fileutils'
+include FileUtils
+
+# path to your application root.
+APP_ROOT = Pathname.new File.expand_path('../../', __FILE__)
+
+def system!(*args)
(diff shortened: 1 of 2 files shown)


bin/reset
```diff
@@ -0,0 +1,29 @@
...
+end
+
+chdir APP_ROOT do
+  # This script is a way to update your development environment automatically.
+  # Add necessary update steps to this file.
+
+  puts "\n== Removing old logs and tempfiles =="
+  system! 'bin/rails log:clear tmp:clear'
+  puts '== Installing dependencies =='
+  system! 'gem install bundler --conservative'
```
ryanking: I don't think this is needed. It will get run again inside the container on line 23. Running it outside the container doesn't really do anything.

cyrielo: You're right, updated.

PR #461, opened 2018-01-10
change column type of archived background data from json to text

boris-dimitrov:
json is new in MySQL 5.7.8 and thus not yet supported on Aurora

### DIFF
--- a/app/controllers/backgrounds_controller.rb
+++ b/app/controllers/backgrounds_controller.rb
@@ -41,7 +41,7 @@ class BackgroundsController < ApplicationController
   # PATCH/PUT /backgrounds/1
   # PATCH/PUT /backgrounds/1.json
   def update
-    current_data = @background.as_json(include: [{ pipeline_outputs: { only: [:id, :pipeline_run_id] } },
+    current_data = @background.to_json(include: [{ pipeline_outputs: { only: [:id, :pipeline_run_id] } },
                                                  { samples: { only: :id } }])
     current_data_full_string = @background.to_json(include: [{ pipeline_outputs: { only: [:id, :pipeline_run_id] } },
                                                              { samples: { only: :id } },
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -156,6 +156,16 @@
     t.index ["user_id"], name: "index_projects_users_on_user_id"
   end
 
+  create_table "reports", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
```
boris-dimitrov: Wow what happened here...  I didn't expect to get this table back, sorry!    I think it was left over in my local DB.

Have to redo this.

boris-dimitrov: I think what happened here is that the old tables still exist on alpha.   So alpha doesn't have the latest code or somehow hasn't run the migrations that delete those tables.

cdebourcy:
I think you also need to change from "as_json" to "to_json" in line 44 of background_controller so that the object is actually a string, not json

app/controllers/backgrounds_controller.rb
```diff
@@ -41,7 +41,7 @@ def create
   # PATCH/PUT /backgrounds/1
   # PATCH/PUT /backgrounds/1.json
   def update
-    current_data = @background.as_json(include: [{ pipeline_outputs: { only: [:id, :pipeline_run_id] } },
+    current_data = @background.to_json(include: [{ pipeline_outputs: { only: [:id, :pipeline_run_id] } },
```
boris-dimitrov: ok, changed that

how do i test it?   does it run on every background modification?

cdebourcy: Yes - you can test it by modifying any background (go to localhost:3000/backgrounds/:id/edit) and checking that an ArchivedBackground record gets created correctly

boris-dimitrov: thanks, confirmed it works

PR #466, opened 2018-01-11
display brakeman output -- fixed

cdebourcy:
Move brakeman to testing script to see if its output will be displayed in the build report for informative purposes. Not supposed to have any influence on whether the build passes.


.travis.yml
```diff
@@ -19,6 +19,7 @@ script:
   - echo "quit" | docker-compose run web rails dbconsole
   - docker-compose run web rubocop -R --config .rubocop_todo.yml
   - docker-compose run web rake db:drop db:create db:migrate db:seed
+  - gem install brakeman && brakeman --no-pager || true
```
ryanking: the `gem install` should do in the install section. I think that improves the caching, but I'm not 100% sure of that.

PR #468, opened 2018-01-11
downgrade development mysql to 5.6 for compatibility with aws aurora

boris-dimitrov:
per @ryanking's suggestion, seems to work


db/migrate/20180104234247_create_archived_backgrounds.rb
```diff
@@ -2,7 +2,7 @@ class CreateArchivedBackgrounds < ActiveRecord::Migration[5.1]
   def change
     create_table :archived_backgrounds do |t|
       t.bigint :archive_of
-      t.json :data
```
yunfangjuan: I thought this was an old change.

ryanking: Yeah but there's no real danger in changing it. It will never be re-run in alpha or production. But it means that it can be run in the development env where the json type is not supported.

PR #473, opened 2018-01-13
Charles/version

cdebourcy:
# Description
Load version info into database. Goes with changes to idseq-pipeline: https://github.com/chanzuckerberg/idseq-pipeline/pull/16


app/models/pipeline_run_stage.rb
```diff
@@ -180,6 +180,9 @@ def db_load_host_filtering
     pr.total_reads = (stats_array[0] || {})['total_reads'] || 0
     stats_array = stats_array.select { |entry| entry.key?("task") }
 
+    version_s3_path = "#{sample_output_s3_path}/#{PipelineRun::VERSION_JSON_NAME}"
+    pr.version = `aws s3 cp #{version_s3_path} -`
```
yunfangjuan: does this a actually work?

cdebourcy: yes :D

PR #485, opened 2018-01-18
Charles/subsampling

cdebourcy:
# Description
Allows us to set a subsample flag. If set, the inputs will be subsampled to the specified number of read pairs after host filtering and before gsnap/rapsearch. Note, rpm values and zscores must then be computed with respect to an adjusted total read number to make sense. This PR goes with https://github.com/chanzuckerberg/idseq-pipeline/pull/19

## Type of change


app/models/pipeline_run_stage.rb
```diff
@@ -19,7 +19,9 @@ class PipelineRunStage < ApplicationRecord
   def install_pipeline
     "cd /mnt; " \
     "git clone https://github.com/chanzuckerberg/idseq-pipeline.git; " \
-    "cd idseq-pipeline; git rev-parse master > #{COMMIT_SHA_FILE_ON_WORKER}; " \
+    "cd idseq-pipeline; " \
+    "git checkout charles/subsampling; " \
```
cdebourcy: this line to be removed once charles/subsampling is merged in idseq-pipeline

db/migrate/20180119200528_add_subsample_to_sample.rb
```diff
@@ -0,0 +1,5 @@
+class AddSubsampleToSample < ActiveRecord::Migration[5.1]
+  def change
+    add_column :samples, :subsample, :integer
```
boris-dimitrov: what sql data type is integer?   I hope it can hold the number 1,000,000.

cdebourcy: Maximum value is 2,147,483,647 according to http://sqlserverplanet.com/tsql/sql-server-max-int-value

PR #486, opened 2018-01-18
Cyril/migrate single upload

cyrielo:
# Description
This PR migrates the singleUpload page to use the new rendering style. It however breaks the bulkupload which should be updated in the next PR

## Type of change
- [x] Breaking change (fix a feature that knowingly causes existing functionality to not work as expected)

# Test and Reproduce

1. 

Which pages?

- [x] Single Upload Page
- [x] Batch Upload Page

Specific components 

* Sample Upload
* IconComponent

# Testing Script:

### Login Page

### Single Upload

### Batch Upload

# Checklist:


app/views/samples/new.html.erb
```diff
@@ -1,3 +1,11 @@
-<%= react_component 'SampleUpload',
-  projects: @projects, csrf: form_authenticity_token, sample: @sample,
-  host_genomes: @host_genomes, loggedin_user: current_user %>
+<div id="sample_upload_page">
+    <%= javascript_tag do %>
+      react_component('SampleUpload', {
+        projects: JSON.parse('<%= raw escape_json(@projects) %>'),
```
cdebourcy: We still need to find a better solution for this. I know "escape_json" was my suggestion initially :) but it's now clear we need something more native and secure. @trevorb has been busy over the past week, but hopefully we can get some input from him soon

cdebourcy:
LGTM, except for `JSON.parse('<%= raw escape_json(...) %>')`. But let's not merge until bulk sample upload is fixed as well

trevorb:
Yep, including raw json leaves you vulnerable to script injection since it doesn't properly escape html tags which could cause the browser to break out of the string. 

There's an ERB helper that will properly escape json for this purpose:
http://api.rubyonrails.org/classes/ERB/Util.html#method-c-json_escape

cdebourcy:
Thank you @trevor! This should make it secure. By the way, the escape_json in the current diff was actually a method we wrote ourselves to make sure quotes are properly escaped in strings that came from ruby. It turned out when passing to JSON.parse, quotes in the payload had not been escaped properly by ruby! How do we deal with that?

trevorb:
Interesting, can you provide an example of a string that ruby is not properly escaped?  In my tests quotes seem to be escaped as expected.

cdebourcy:
Cool, @cyrielo would you mind pasting the example error you got before we added the custom escaping? Also, does it still occur if you use the json_escape function Trevor pointed us to?

@trevorb, the error happened in an example like this: 
User enters a note into a text field, for example: `A String With "Quotes"`
In ruby/rails it ends up looking like this:
`@projects == { notes: "A string with \"quotes\"", something_else: 123 }`
The argument to JSON.parse (`'<%= raw @projects %>'`) ends up looking like this in the browser console, IRC:
`{ "notes": "A string with "quotes"", "something_else": "123"}`,
breaking JSON.parse. 

The problem goes away if you pass `'<%= raw @projects.gsub!("\\", "\\\\\\") %>'` as an argument to JSON.parse instead, but it doesn't seem like a good solution -- there must be a built-in way to solve the problem?

trevorb:
If you use json_escape the way the document recommends, it should work fine with quotes as in the example provided above.

aka
```
@projects = { notes: "A string with \"quotes\"", something_else: 123 }


<script>
  var projects = <%= raw json_escape(@projects.to_json) %>;
</script>
```

cdebourcy:
Thanks!!

cyrielo:
Thanks @trevorb this is very helpful, I should implement this in the next changes to revert this method.

cyrielo:
This PR was meant to test the changes, and give an idea of the changeset necessary for the migration to the entire team. This PR is no longer relevant as there's another PR for the actual work

PR #502, opened 2018-01-23
Ajaxify bulk report downloads

cdebourcy:
To be re-styled by Susan/Cyril. Currently allows only one download at a time, which is fine for now.


app/controllers/projects_controller.rb
```diff
@@ -30,10 +30,24 @@ def send_project_csv
     send_data project_csv, filename: project_name + '_sample-table.csv'
   end
 
-  def project_reports_csv
-    output_file = bulk_report_csvs_from_params(@project, params)
-    return unless output_file
-    send_file output_file
+  def make_project_reports_csv
+    `aws s3 rm #{@project.report_tar_s3}`
+    Resque.enqueue(GenerateProjectReportsCsv, params)
```
yunfangjuan: we need to make sure the output file is id'ed by user because each user might see different content once we put in the access control.

cdebourcy: Ok!

app/assets/javascripts/components/samples.jsx
```diff
@@ -98,6 +102,39 @@ class Samples extends React.Component {
...
   }
 
+  displayReportProgress(res) {
+      $('.download-progress')
+      .html(`${res.data.status_display}`)
+      .css('display', 'block');
+
+      setTimeout(() => {
+        this.checkReportDownload()
+      }, 2000)
```
boris-dimitrov: does the polling stop after the report generation reaches a terminal state?

every 2 seconds seems real aggressive;   perhaps start at 5 seconds, and add 5 on each wakeup

cdebourcy: Yes, it stops after it actually gets the file. Ok, will increase

app/controllers/projects_controller.rb
boris-dimitrov: Good catch, @yunfangjuan.   I think it would be slightly more optimal if id'ed on visibility rather than user.   I.e. if the user's access is "public" then they can get the "report_public.tar", and if the user's access is "contributor" then they can get the full "report.tar".   That way when a public link to a report is shared, the public doesn't overwhel our servers with CSV generation -- but instead each public member immediately downloads the existing "report_public.tar".

boris-dimitrov:
Perhaps a small further improvement would be not to delete the report file when a new report generation is requested, but rather to include timestamp along with visibility in the report filename, so that a new user deciding they want to regenerate the report doesn't break a download.   Similarly, the status polling can show all the reports that exist so far in s3, basically the result of "aws s3 ls".

boris-dimitrov:
One larger improvement needed here is to prevent a new report generation from being started while a report generation is in progress, and also to make it possible for other users to know that a report is being generated.   In other words, when you go to start this, you should see status.

boris-dimitrov:
Reflecting further on this, generating a different file in S3 for every user who wants to download the report seems wasteful, and similarly keeping too many old versions would be wasteful.   At least for the privacy thing it's easy to workaround by visibility typing.

PR #505, opened 2018-01-23
Cyril/frontend migration

cyrielo:
# Description
This PR handles the general migration for all components

## Type of change

# Test and Reproduce

Which pages?

- [x] Login Screen
- [x] Single Upload Page
- [x] Batch Upload Page
- [x] All Projects Page
- [x] Sample Details Page
- [x] Report Page

Specific components 

* 

# Testing Script:

###Login Page

### Single Upload

### Batch Upload

### All Project Page
* Switch between projects 
* Search for a sample

### Sample Details Page
* Can navigate to report page 

### Report Page
* Data loads successfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:


app/views/users/new.html.erb
```diff
@@ -1,7 +1,14 @@
-<div>
+<div id='user_create_page'>
   <% if current_user && current_user.admin %>
-    <%= react_component 'CreateUser', { csrf: form_authenticity_token, user: @user } %>
+    <%= javascript_tag do %>
+      react_component('CreateUser', {
+        user: JSON.parse('<%= raw escape_json(@user) %>'),
```
cdebourcy: Have you tried the json_escape function Trevor suggested? Did it work? Then you shouldn't need the custom function escape_json anymore

cyrielo: Oh I have tried it, it escaped the json using a single backslash, so I update the custom method we created to use the json_escape method trevor suggested, but to replace single backslash with double instead

app/helpers/application_helper.rb
```diff
@@ -13,4 +13,11 @@ def hash_array_json2csv(input_file, output_file, keys)
       end
     end
   end
+
+  def escape_json(hash)
+    # using json_escape to prevent XSS vulnerability
+    str = json_escape(hash.to_json) unless hash.class == 'String'
+    str = str.gsub!("\\", "\\\\\\") if str.include? "\\"
```
trevorb: Strings need to be json_escaped as well, I'd recommend that you use json_escape like the docs recommend rather than JSON.parse on the output.

https://apidock.com/rails/ERB/Util/json_escape

PR #508, opened 2018-01-24
"Can-see" functions

cdebourcy:
# Description
Utility functions to 
(a) check whether a sample/project is visible to a user
(b) generate a list of all visible projects/samples for a user


cdebourcy:
Closing because we're going to use consul for our access/permission management.

PR #528, opened 2018-01-25
Make project public, Add user to project

cdebourcy:
Two new features.

(1) Checkbox "Make project public" when creating a new project (sample upload page, bulk upload page)

(2) Ability to add user to project by email on the project page. If the email address does not exist, a new user is created (with a randomly generated password).


app/assets/src/components/SampleUpload.jsx
```diff
@@ -129,7 +130,7 @@ class SampleUpload extends React.Component {
 
   toggleCheckBox(e) {
     this.setState({
-      omit_subsampling_checked: e.target.value == "true" ? false : true,
+      [e.target.id]: e.target.value == "true" ? false : true,
```
yunfangjuan: what is the array [] state?  they should at least belong to a hash entry, not expose direct ids.

PR #538, opened 2018-01-29
Closes #495 Cyril/show member info

### LINKED ISSUE #495: Add member information for Projects
After selecting a project and seeing all samples in the project the user should see three things in the right-hand corner: 
1. If the project is private (hidden) or public (visible)
2. How many members are part of a project
3. A button to add project members

cyrielo:
# Description
This PR adds a project menu list to the subheader to view total members in a project, a button to add a user to the project and project visibility.

## Type of change

Which pages?

- [x] All Projects Page

Specific components 

* Samples

# Testing Script:

### All Project Page
* Switch between projects 
* Search for a sample

# Checklist:

### DIFF
--- a/app/assets/src/components/Samples.jsx
+++ b/app/assets/src/components/Samples.jsx
@@ -228,7 +228,7 @@ class Samples extends React.Component {
       let email_to_add = this.refs.add_user.value;
       let project_id = this.state.selectedProjectId;
       if (all_user_emails.includes(email_to_add)) {
-        axios.post(`/projects/${project_id}/add_user_to_project`, 
+        axios.post(`/projects/${project_id}/add_user_to_project`,
                    { user_emails_to_add: [email_to_add],
                      authenticity_token: this.csrf })
         .then((res) => {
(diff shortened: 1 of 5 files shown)


app/controllers/projects_controller.rb
```diff
@@ -15,6 +15,20 @@ def index
   # GET /projects/1
   # GET /projects/1.json
   def show
+    # all exisiting project are null, we ensure private projects are explicitly set to 0
+    public_access = @project.public_access.nil? ? 1 : @project.public_access
```
cdebourcy: I think it's a better default to set the variable to 0 when it is nil

cyrielo: I assume 0 means private right? so that means all projects are private by default?

cdebourcy: Yes. Until we set them to public explicitly

cyrielo: Okay, updating ...

app/assets/src/components/Samples.jsx
```diff
@@ -740,22 +736,61 @@ class Samples extends React.Component {
...
+               <span>
+                 <i className="tiny material-icons">lock</i> Private Project
+               </span>
+            ) : null }
+          </li>
+          <li>
+              { this.state.project ? (
+                this.state.project.total_members ?
+                <span>
+                  <i className="tiny material-icons">people</i> {this.state.project.total_members} Member(s)
```
cdebourcy: In this PR or the next, please get rid of the "(s)" and check the plural to make it correct. There must be a "pluralize" function out there that does it automatically

PR #559, opened 2018-01-30
Charles/protect emails

cdebourcy:
# Description
Remove route that showed all emails


config/routes.rb
```diff
@@ -35,10 +35,10 @@
     put :remove_favorite, on: :member
     post :add_user_to_project, on: :member
```
yunfangjuan: remove this.

PR #560, opened 2018-01-30
Closes #555: Add pipeline retries

### LINKED ISSUE #555: Retry if sample fails due to spot instance termination

cdebourcy:
Already a large fraction of the newly submitted samples failed just because of spot instance termination.
Also deduplicated some of the batch submission code.

### DIFF
--- a/app/models/pipeline_run_stage.rb
+++ b/app/models/pipeline_run_stage.rb
@@ -3,7 +3,7 @@ class PipelineRunStage < ApplicationRecord
   include PipelineOutputsHelper
   belongs_to :pipeline_run
   DEFAULT_MEMORY_IN_MB = 4000
-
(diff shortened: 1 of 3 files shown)


app/models/pipeline_run_stage.rb
```diff
@@ -93,6 +106,7 @@ def update_job_status
       if job_hash['container'] && job_hash['container']['logStreamName']
         self.job_log_id = job_hash['container']['logStreamName']
       end
+      run_job if instance_terminated?(job_hash) # retry if necessary
```
yunfangjuan: if instance_terminated?(job_hash)
run_job
return
end 

so you don't save twice

cdebourcy: Ok! Thanks

PR #562, opened 2018-01-31
fix cli functionality

cdebourcy:
# Description
Files were uploaded to wrong S3 location because presigned URL was generated before sample ID existed. Sample ID is embedded in the S3 path.


app/models/input_file.rb
```diff
@@ -25,10 +25,10 @@ def s3_source_check
     end
   end
 
-  after_validation(on: :create) do
+  def set_presigned_url_for_local_upload
```
yunfangjuan: can you just make this an  after_save hook?

cdebourcy: yes, thanks

PR #600, opened 2018-02-05
pipeline optimizations:  tweak vCPU and memory requirements

boris-dimitrov:

### DIFF
--- a/app/models/pipeline_run_stage.rb
+++ b/app/models/pipeline_run_stage.rb
@@ -2,9 +2,7 @@ class PipelineRunStage < ApplicationRecord
   include ApplicationHelper
   include PipelineOutputsHelper
   belongs_to :pipeline_run
-  DEFAULT_MEMORY_IN_MB = 30_000
 
-  DEFAULT_STORAGE_IN_GB = 500
   JOB_TYPE_BATCH = 1
   COMMIT_SHA_FILE_ON_WORKER = "/mnt/idseq-pipeline/commit-sha.txt".freeze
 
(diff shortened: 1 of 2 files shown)


app/models/pipeline_run_stage.rb
```diff
@@ -25,10 +23,19 @@ def install_pipeline
...
+      vcpus = Sample::DEFAULT_VCPUS
+      queue = Sample::DEFAULT_QUEUE
+    else
+      vcpus = Sample::DEFAULT_VCPUS_HIMEM
+      queue = Sample::DEFAULT_QUEUE_HIMEM
+    end
+    if pipeline_run.sample.job_queue.present?
+      queue = pipeline_run.sample.job_queue
+    end
+    command += " --storage /mnt=#{Sample::DEFAULT_STORAGE_IN_GB} --ecr-image idseq --memory #{memory} --queue #{queue} --vcpus #{vcpus}"
```
boris-dimitrov: I think the latest changes cleaned this up a bit better.   Does it still look good, @cdebourcy?

boris-dimitrov:
Testing new `idseq` queue in https://alpha.idseq.net/samples/321

boris-dimitrov:
This passes tests but I am presently struggling to run it in localhost env (because my env is broken).

app/models/pipeline_run_stage.rb
cdebourcy: It looks good to me!

boris-dimitrov:
As soon as this is deployed on production, we need to `update host_genomes SET job_queue=NULL`.

PR #633, opened 2018-02-08
Show all output files, even nested ones

cdebourcy:


cdebourcy:
Yes, just added that

app/models/sample.rb
```diff
@@ -105,11 +105,25 @@ def self.get_signed_url(s3key)
     nil
   end
 
-  def results_folder_files
+  def path_end(key, n = 1)
+    parts = key.split('/')
+    return "#{parts[-2]}/#{parts[-1]}" if n == 2
```
yunfangjuan: n == 2 ? "#{parts[-2]}/#{parts[-1]}" : parts[-1]

app/models/sample.rb
```diff
@@ -105,11 +105,25 @@ def self.get_signed_url(s3key)
...
+  end
+
+  def results_folder_files
+    prs = pipeline_runs.first.pipeline_run_stages.first
+    stage1_files = list_outputs(prs.sample_output_s3_path)
+    stage2_files = list_outputs(prs.alignment_output_s3_path, 2)
+    stage1_files + stage2_files
   end
 
   def fastqs_folder_files
```
yunfangjuan: for consistency reasons, you should change the format here too.

app/models/sample.rb
```diff
@@ -105,11 +105,25 @@ def self.get_signed_url(s3key)
     nil
   end
 
-  def results_folder_files
+  def path_end(key, n = 1)
```
yunfangjuan: I think end_path might be a better name.

app/models/sample.rb
cdebourcy: ok!

PR #639, opened 2018-02-09
Charles/skip deutero filter

cdebourcy:
# Description
Goes with https://github.com/chanzuckerberg/idseq-pipeline/pull/52/files
When ready, need to deploy this to production, set skip_deutero_filter to 1 for "mosquito" host and "no host subtraction", and rerun recent samples.

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -166,6 +166,10 @@ class Sample < ApplicationRecord
     host_genome && host_genome.name == HostGenome::NO_HOST_NAME ? 0 : 1
   end
 
+  def skip_deutero_filter_flag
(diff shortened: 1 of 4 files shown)


app/models/sample.rb
```diff
@@ -166,6 +166,10 @@ def filter_host_flag
     host_genome && host_genome.name == HostGenome::NO_HOST_NAME ? 0 : 1
   end
 
+  def skip_deutero_filter_flag
+    !host_genome || host_genome.skip_deutero_filter == 1 ? 1 : 0
```
boris-dimitrov: Did you mean to skip deutero filtering if there is no host genome?

cdebourcy: Yes: deutero filtering is really a part of host subtraction

PR #663, opened 2018-02-14
speed up pipeline state transitions

boris-dimitrov:
Detect completed jobs much faster by checking S3 every 20 seconds.   Detect failed jobs by checking with aegea batch every 100 seconds, just slightly more frequent than before.   If the checks become slower, then they will run less frequently.   Metrics for scaling are updated within 20 seconds of when they change, and the autoscaling groups respond immediately if the request is from prod, or with up to 5 minute delay if responding to a change in one fo the other environments.

### DIFF
--- a/app/jobs/check_pipeline_runs.rb
+++ b/app/jobs/check_pipeline_runs.rb
@@ -1,19 +1,67 @@
 # Jos to check the status of pipeline runs
 require 'logger'
 require 'resque/plugins/lock'
+
 class CheckPipelineRuns
   extend Resque::Plugins::Lock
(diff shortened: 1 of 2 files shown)


yunfangjuan:
cool

boris-dimitrov:
Proof it works on alpha:  https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logEventViewer:group=ecs-logs-alpha;stream=idseq-resque/idseq-resque/23528072-d63a-41e1-bc31-5cdcc9a1c301

PR #667, opened 2018-02-14
Closes #671: Charles/display version

### LINKED ISSUE #671: Display version next to pipeline

cdebourcy:
# Description
First pass at displaying version information

### DIFF
--- a/app/assets/src/components/PipelineSampleReads.jsx
+++ b/app/assets/src/components/PipelineSampleReads.jsx
@@ -406,7 +406,7 @@ class PipelineSampleReads extends React.Component {
         <SubHeader>
           <div className="sub-header">
             <div className="title">
-              PIPELINE
+              PIPELINE v{this.pipelineRun.version.pipeline}, NT v{this.pipelineRun.version.nt}, NR v{this.pipelineRun.version.nr}
             </div>
 
             <div className="sub-title">
--- a/app/assets/src/components/PipelineSampleReads.jsx
+++ b/app/assets/src/components/PipelineSampleReads.jsx
@@ -400,13 +400,23 @@ class PipelineSampleReads extends React.Component {
     sample_dropdown = <span>{ this.sampleInfo.name }</span>
     }
 
+    let version_display = !this.pipelineRun ? '' :
+                            !this.pipelineRun.version ? '' :
+                              !this.pipelineRun.version.pipeline ? '' :
+                                'v' + this.pipelineRun.version.pipeline
+    if (version_display != '' && this.pipelineRun.version.nt) {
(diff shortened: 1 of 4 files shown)


app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -400,13 +400,23 @@ class PipelineSampleReads extends React.Component {
     sample_dropdown = <span>{ this.sampleInfo.name }</span>
     }
 
+    let version_display = !this.pipelineRun ? '' :
+                            !this.pipelineRun.version ? '' :
+                              !this.pipelineRun.version.pipeline ? '' :
+                                'v' + this.pipelineRun.version.pipeline
```
boris-dimitrov: Pretty.

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -1,6 +1,24 @@
...
+    pipeline_run_display["version"] = { pipeline: select_version_aspect(pipeline_run, "idseq-pipeline"),
+                                        nt: select_version_aspect(pipeline_run, "nt_k16"),
+                                        nr: select_version_aspect(pipeline_run, "nr_rapsearch") }
+    pipeline_run_display
+  end
+
+  def select_version_aspect(pipeline_run, aspect)
+    return nil unless pipeline_run.version
+    version_hash = JSON.parse(pipeline_run.version)
+    aspect_hash = version_hash.select { |item| item["name"] == aspect }[0]
```
boris-dimitrov: Is the `[0]` guaranteed not to go out-of-bounds?

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -1,6 +1,24 @@
...
+    return nil unless aspect_hash
+    version_key = %w[nt_k16 nr_rapsearch].include?(aspect) ? "source_version" : "version"
+    aspect_hash.key?(version_key) ? aspect_hash[version_key] : nil
```
boris-dimitrov: To my reading of the Ruby spec, you can just write the last line simply as follows
```
aspect_hash.key(version_key)
```

app/helpers/pipeline_outputs_helper.rb
cdebourcy: `[0]` will just give nil if it is out-of-bounds and that case is then handled in the next line.
Ok, I'm adding an example of version_hash. It should actually be called version_array.

app/helpers/pipeline_outputs_helper.rb
cdebourcy: Hmm that doesn't actually work though

PR #674, opened 2018-02-14
Closes #654: escape single quote

### LINKED ISSUE #654: Apostrophe in project name breaks site
Tried creating a project with apostrophe in the name and uploading sample to it. Subsequently pages cannot be rendered:
screen shot 2018-02-12 at 6 58 06 pm
Pretty serious.

cdebourcy:
No idea why this should still be necessary, suggestions on deeper fix welcome.
Without this fix, creating a project named `Charles's project` will break the "all samples" page, for example.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -18,6 +18,7 @@ module ApplicationHelper
     # using json_escape to prevent XSS vulnerability
     str = json_escape(hash.to_json) unless hash.class == 'String'
     str = str.gsub!("\\", "\\\\\\") if str.include? "\\"
+    str = str.gsub("'", "\\\\'")
     str
   end
 end


app/helpers/application_helper.rb
```diff
@@ -18,6 +18,7 @@ def escape_json(hash)
     # using json_escape to prevent XSS vulnerability
     str = json_escape(hash.to_json) unless hash.class == 'String'
     str = str.gsub!("\\", "\\\\\\") if str.include? "\\"
+    str = str.gsub("'", "\\\\'")
```
boris-dimitrov: it appears this isn't the first time we've ran into it,  i would git blame line 20 and see who added that :)

also seems like this truly belongs inside json_escape;   may be worth creating a security issue

cdebourcy: Yes, line 20 was added during the big frontend migration. We couldn't figure out the right solution at the time and still haven't...

PR #676, opened 2018-02-15
Charles/report details cleanup

cdebourcy:
# Description
Cleanup after ajaxification of report page. Reduce confusion by removing unused code and adding comments.
Make explicit that default background ID is same for report page and report csv download.

### DIFF
--- a/app/lib/report_helper.rb
+++ b/app/lib/report_helper.rb
@@ -176,10 +176,11 @@ module ReportHelper
     data
   end
 
-  def report_details(pipeline_run, _background)
-    # Gives some auxiliary information on pipeline_run.
-    # Not actually anything to do with report-specific data and does not use background argument.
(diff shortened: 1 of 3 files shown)


app/lib/report_helper.rb
```diff
@@ -176,13 +176,13 @@ def external_report_info(pipeline_run_id, background_id, params)
     data
   end
 
-  def report_details(pipeline_run, background)
+  def report_details(pipeline_run, _background = nil)
+    # Provides some auxiliary information on pipeline_run.
+    # Not actually anything to do with report-specific data and does not use _background argument.
```
boris-dimitrov: there is always room for more cleanup, i suppose -- like getting rid of unused argument and renaming to closer match the meaning

but an improvement already

PR #680, opened 2018-02-15
679 susan/metadat filtering samples page

sadelokiki:
# Description

This PR creates a metadata filter at the top of the samples table.

## Type of change

# Test and Reproduce
1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -17,3 +17,4 @@ class HomeController < ApplicationController
     dir ||= default_dir
     dir == 'newest' ? samples.order(id: :desc) : samples.order(id: :asc)
   end
+end
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -8,6 +8,7 @@ class HomeController < ApplicationController
     @favorite_projects = current_user.favorites
     @projects = current_power.projects
     @editable_project_ids = current_power.updatable_projects.pluck(:id)
+    @host_genomes = HostGenome.all
     render 'home'
   end
 
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -8,7 +8,7 @@ class HomeController < ApplicationController
     @favorite_projects = current_user.favorites
     @projects = current_power.projects
     @editable_project_ids = current_power.updatable_projects.pluck(:id)
-    @host_genomes = HostGenome.all
+    @host_genomes = HostGenome.all.reject {|hg| hg.name.downcase.include?("__test__")}
     render 'home'
   end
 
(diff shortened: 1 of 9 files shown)


cdebourcy:
Looks nice. I think it would be good if there were an option to reset filters to None. In fact, I think the filters should be checkboxes instead of radio buttons, so that you can check/uncheck which ever ones you like? Then the dropdown also shouldn't close as soon as you click

app/views/layouts/_page_header.html.erb
```diff
@@ -6,7 +6,7 @@
         userDetails: JSON.parse('<%=raw escape_json(current_user)%>'),
         user_auth_token: '<%= current_user.authentication_token %>',
         demoUser: <%= current_user.demo_user? ? 1: 0 %>,
-        host_genome_names: JSON.parse('<%= raw escape_json(HostGenome.all.map(&:name)) %>'),
+        host_genome_names: JSON.parse('<%= raw escape_json(HostGenome.all.map(&:name).reject {|n| n.downcase.include?("__test__")}) %>'),
         signoutEndpoint: '<%=destroy_user_session_path%>',
```
cdebourcy: diff seems to be comparing to an old master

app/controllers/home_controller.rb
```diff
@@ -8,6 +8,7 @@ def index
     @favorite_projects = current_user.favorites
     @projects = current_power.projects
     @editable_project_ids = current_power.updatable_projects.pluck(:id)
+    @host_genomes = HostGenome.all
```
cdebourcy: There are some host genomes with "__test__" in the name which we just use for test samples. Probably shouldn't show those in front-end. You can replace `HostGenome.all` with something like `HostGenome.all.reject {|hg| hg.name.downcase.include?("__test__")}`

cdebourcy:
Awesome! Maybe also remove the existing tissue-type filter then? I remember in the beginning you had implemented it for when the user clicks on the tissue-type column. We don't need that anymore.

app/helpers/samples_helper.rb
```diff
@@ -185,6 +185,11 @@ def filter_by_tissue_type(samples, query)
     samples
   end
 
+  def filter_by_host(samples, query)
+    samples = samples.where(host_genome_id: query)
+    samples
```
cdebourcy: could just write 
`def filter_by_host(samples, query)
     samples.where(host_genome_id: query)
 end`
no?

PR #685, opened 2018-02-15
Closes #618: hide test hosts from CLI modal

### LINKED ISSUE #618: hide "test" hosts from the options in CLI modal

cdebourcy:


app/views/layouts/_page_header.html.erb
```diff
@@ -6,7 +6,7 @@
         userDetails: JSON.parse('<%=raw escape_json(current_user)%>'),
         user_auth_token: '<%= current_user.authentication_token %>',
         demoUser: <%= current_user.demo_user? ? 1: 0 %>,
-        host_genome_names: JSON.parse('<%= raw escape_json(HostGenome.all.map(&:name)) %>'),
+        host_genome_names: JSON.parse('<%= raw escape_json(HostGenome.all.map(&:name).reject {|n| n.downcase.include?("test")}) %>'),
```
boris-dimitrov: Isn't it likely that a user would choose a name that contains the text "test"?   In that case, that user, would report it as a bug if they can't see this.   I recommend changing this from "test" to something like "__test__" and renaming the existing test genomes accordingly.

PR #687, opened 2018-02-15
WIP 655 cyril implement report top filter

cyrielo:
# Description
This PR implements advanced filtering on the report page and moves sidebars filters to the top 

## Type of change

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads successfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

### DIFF
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -758,27 +758,8 @@ class PipelineSampleReport extends React.Component {
         </span>
       </div>
     );
-<<<<<<< HEAD
-    const report_filter =
-      (<ReportFilter
-        ref="report_filter"
-        all_categories={this.all_categories}
-        all_backgrounds={this.all_backgrounds}
-        search_keys_in_sample={this.state.search_keys_in_sample}
-        default_background={this.report_details.default_background}
-        report_title={this.report_details.sample_info.name}
-        report_page_params={this.report_page_params}
-        applyNameType={this.applyNameType}
-        applyExcludedCategories={this.applyExcludedCategories}
-        applySearchFilter={this.applySearchFilter}
-        flash={this.flash}
-        filter_row_stats={filter_row_stats}
-        enableFilters={this.enableFilters}
-        resetAllFilters={this.resetAllFilters}
-=======
     const report_filter = (<ReportFilter ref="report_filter"
         filter_row_stats={ filter_row_stats }
->>>>>>> restructure top filters
       />);
     let param_background_id = this.fetchParams("background_id")
     let cookie_background_id = Cookies.get('background_id')
@@ -797,10 +778,10 @@ class PipelineSampleReport extends React.Component {
         <div id="reports" className="reports-screen tab-screen col s12">
           <div className="tab-screen-content">
             <div className="row reports-container">
-              {/* <div className="col s2 reports-sidebar">
+              <div className="col s2 reports-sidebar">
                 {report_filter}
-              </div> */}
-              <div className="col s12 reports-section">
+              </div>
+              <div className="col s10 reports-section">
                 <div className="reports-count">
                   { filter_row_stats }
                   <div className='report-top-filters'>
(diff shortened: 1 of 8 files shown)


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -27,6 +28,9 @@ class PipelineSampleReport extends React.Component {
 
     this.state = {
       taxonomy_details: [],
+      backgroundName: Cookies.get('background_name') || this.report_details.background_model.name,
```
cdebourcy: New master doesn't have a background_model field in report_details anymore so you may need to refactor a bit

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -394,104 +464,130 @@ class PipelineSampleReport extends React.Component {
...
 
-    this.applySearchFilter(0, this.state.excluded_categories, thresholded_taxons);
+  getSavedThresholdFilters() {
+    const activeThresholds = window.localStorage.getItem('activeThresholds');
+    return (activeThresholds) ? JSON.parse(activeThresholds) : [];
+  }
 
-    if (play_animation) {
-      this.flash();
+  applyThresholdFilters(candidateTaxons, rules, animate = true) {
```
yunfangjuan: I think there are way too many changes than just move the filters to the top.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -55,24 +65,60 @@ class PipelineSampleReport extends React.Component {
...
+        }, {
+          name: 'R log(1/e)',
+          value: 'NR_neglogevalue'
+        }, {
+          name: 'NR %conc',
+          value: 'NR_percentconcordant'
+        }]
+      };
     this.applySearchFilter = this.applySearchFilter.bind(this);
-    this.applyThresholdFilters = this.applyThresholdFilters.bind(this);
```
yunfangjuan: why do you remove this? This still exists.

cyrielo: We were binding the `this` context to the method, but we have too many binds already and  we could avoid that by using the arrow functions like this 
`<button onClick={ () => this.methodName() }></button>` then  we don't have to call the bind method again

app/assets/src/components/PipelineSampleReport.jsx
cyrielo: Yeah  because we also had to remove the sidebar and re-style form inputs ui and implement the advanced filtering

yunfangjuan:
A few things observed from the new behavior:
1.  Category filtering: it automatically closes after I click on one of the checkboxes. I think it should let people clicking multiple check boxes without closing. It would only close the category filtering box when the user clicks out of the box. 
2. Scrolling: it doesn't load more results because of the content length. If you download the alpha data, 
http://localhost:3000/samples/18 gives you a good idea. We should make the scrolling work or have a load more link on the bottom. 
3. Advance Filter: it doesn't close when the user clicks outside of the box. Also, we should just use '>=', '<=', nobody is going to use the '==' 
4. Advance filtering doesn't work: if you pick a filter, first time it works. But once you X out the filter, you don't get the results back. The # of results do change but the report section is not rendered. 
5. The search box and advance filtering interacts in an odd way. When a user search something, it shows the filtered search result. But when you open the advance filtering box and pick something, it ignores the search query. It should filter down the search results.  Not showing the results just based on the advance filters. 
6. When you add an advance filter and doesn't click the operand, it should show an error. Or you can make the operand default to '>='

PR #710, opened 2018-02-21
Charles/version date

cdebourcy:
# Description
Display the generation date of NT/NR indexes rather than opaque version number.

### DIFF
--- a/app/helpers/pipeline_outputs_helper.rb
+++ b/app/helpers/pipeline_outputs_helper.rb
@@ -11,22 +11,20 @@ module PipelineOutputsHelper
   end
 
   def select_version_aspect(pipeline_run, aspect)
-    begin
-      version_hashes = JSON.parse(pipeline_run.version)
-      # example for version_hashes:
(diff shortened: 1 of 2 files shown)


app/helpers/pipeline_outputs_helper.rb
```diff
@@ -11,17 +11,20 @@ def curate_pipeline_run_display(pipeline_run)
...
-    return nil unless aspect_hash
-    version_key = %w[nt_k16 nr_rapsearch].include?(aspect) ? "source_version" : "version"
-    aspect_hash.key?(version_key) ? aspect_hash[version_key] : nil
+    if %w[nt_k16 nr_rapsearch].include?(aspect)
+      return DateTime.parse(aspect_hash["generation_date"]).utc.strftime("%m-%Y")
+    else
+      return aspect_hash["version"]
+    end
+  rescue
+    return nil
```
boris-dimitrov: I hope we will get some sort of error message if this crashes so we can fix the format parser :)

PR #721, opened 2018-02-21
try to release the lock while shutting down

boris-dimitrov:


boris-dimitrov:
Hey @yunfangjuan I can't get the signal handler to catch SIGTERM when running within Resque.   Works fine from a Ruby script but not from Resque.   From my reading of the internets, Resque isn't known for correctly propagating signals to workers.

https://stackoverflow.com/questions/15074551/recovering-cleanly-from-resquetermexception-or-sigterm-on-heroku

I am going to try a few more things and then probably just turn off this lock completely and look for some other way to run a daemon.

boris-dimitrov:
I am also a bit unclear on how safe it is to simply override the TERM handle in some RAILS source file;   what if that file gets loaded into the running webapp or some other important process?    Normally there would be a chain of handlers, similar to the exit chain in C...  So this really looks like the wrong way to do it, and unfortunately I don't know what the right way is.

boris-dimitrov:
This new way of doing it -- rake daemon instead of resque scheduler, and versioned locks for the workers --- seems to work fine in local development.    In order to test more fully on alpha, I've created a branch merging Charlse' changes to the alpha DB with this work:  https://github.com/chanzuckerberg/idseq-web/pull/723

boris-dimitrov:
Testing on alpha confirmed this works correctly.   Yay.

PR #722, opened 2018-02-21
Charles/lineage versions

cdebourcy:
# Description
Adds date range of validity to taxon_lineages. Uses dates to update taxon_lineages without destroying old records. Uses dates to determine which entries of taxon_lineages should be used for a given taxon_count.

# Testing
Tested on alpha:
- [X] Temporarily copied s3://cdebourcy-test/names_test-2-21-2018.csv.gz to s3://czbiohub-infectious-disease/references/names.csv.gz. Had edited this file to have a new, made-up name for the Klebsiella genus and the Salmonella enterica species. Ran "rake update_lineage_db". Exited fine and affected records of TaxonLineage were only Klebsiella/Salmonella ones.
- [X] Existing reports look same as before, even after running pipeline_run.update_names and pipeline_run.update_genera.
- [X] Ran new sample and checked the new names showed up, both species and genus level.
- [X] Made another update, this time to both taxid-lineages.csv and names.csv, putting one species into a new genus taxid with a made-up name. Apart from that, all names were reverted to the NCBI values. The report correctly showed reverted names and the new genus for the species that got moved.


cdebourcy:
- I only dump records with the maximum ended_at date from the current table. So the way the records_to_retire and records_to_insert get calculated is correct - only "current" entries get considered.
- Good point though - I want to  go back and check that the correct records get updated with records_to_retire when multiple entries are already present. I am not sure of it now.
- When a record gets edited back to an old date, I think it is still the most sensible thing to do to make a new record identical to the old one but for the validity range. That way we avoid complexity on our side and we captured the back-and-forth that happened in NCBI (this should be very rare, no?).

lib/tasks/update_lineagedb.rake
```diff
@@ -0,0 +1,64 @@
...
+   comm -23 old_taxon_lineages_sorted.csv new_taxon_lineages_sorted.csv > records_to_retire.csv
+   comm -13 old_taxon_lineages_sorted.csv new_taxon_lineages_sorted.csv > records_to_insert.csv
+   # Add ended_at column for retired records, started_at column for new records
+   sed -e 's/$/,#{current_date}/' -i records_to_retire.csv
+   sed -e 's/$/,#{current_date}/' -i records_to_insert.csv
+
+   ## Import changes to taxon_lineages
+   # retired records:
+   mv records_to_retire.csv taxon_lineages
+   mysqlimport --replace --local --user=$DB_USERNAME --host=#{host} --password=$DB_PASSWORD --columns=#{column_names},ended_at --fields-terminated-by=',' idseq_#{Rails.env} taxon_lineages;
```
cdebourcy: to do: ensure this affects the correct records

PR #727, opened 2018-02-23
fix 725

boris-dimitrov:
when LoadResults fails the job becomes indefinitely stuck 'in progress'

while there also make only jobs in stage 1 and 2 trigger scaling up
of gsnap and rapsearch tiers


boris-dimitrov:
This pasts tests through my local environment.

PR #786, opened 2018-02-27
remove some race conditions and simplify state

boris-dimitrov:


app/models/pipeline_run_stage.rb
```diff
@@ -110,7 +98,8 @@ def run_load_db
       update(job_status: STATUS_FAILED)
       raise
     ensure
-      pipeline_run.update_job_status
+      pipeline_run.check_job_status
```
yunfangjuan: don't you need pipeline_run.save?

PR #794, opened 2018-02-27
dedup check_job_status and update_job_status

boris-dimitrov:


boris-dimitrov:
Btw this works fine on my development environment so I will push it to alpha for testing there.

boris-dimitrov:
Confirmed works on alpha.   Merging into master.

PR #821, opened 2018-03-03
run alignmnet viz on himem queue so it gets more vcpu

boris-dimitrov:
for now this is the only way to give more vCPUs to the alignment viz code, which seems to really need them


boris-dimitrov:
This per se doesn't help!   (2201.68s elapsed) https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logEventViewer:group=/aws/batch/job;stream=aegea_batch/default/bd60fbc3-2291-4707-9265-5eaf53926b9b

Utilizing those vCPUs requires python multiprocessing.

Issue #827, opened 2018-03-06
After successful bulk upload, I get redirected to an URL that says 'No samples found'


cdebourcy:
But if I just go to localhost:3000, the samples are there

cdebourcy:
adding fix as part of wider bulk-upload update

PR #829, opened 2018-03-06
Closes #777 and #827: Charles/bulk upload

### LINKED ISSUE #777: Make sure bulk upload can handle edge cases
Bulk upload not flexible enough. There were a bunch of fasta / fastq / fa / fq files in s3://cdebourcy-test but it didn’t pick them up

cdebourcy:
# Description
Make bulk upload slightly more flexible, display file format in case of error, and fix bugs.

### DIFF
--- a/app/assets/src/components/BulkUploadImport.jsx
+++ b/app/assets/src/components/BulkUploadImport.jsx
@@ -124,16 +124,18 @@ class BulkUploadImport extends React.Component {
   }
 
   handleUploadSubmit(e) {
+    console.log('handleUploadSubmit')
     e.preventDefault();
     $('html, body').stop().animate({ scrollTop: 0 }, 200, 'swing', () => {
+      console.log('inside');
       this.clearError();
-      if(!this.isUploadFormInvalid()) {
-        this.setState({
-          submitting: true
-        });
-        this.bulkUploadSubmit();
-      }
     });
+    if(!this.isUploadFormInvalid()) {
+      this.setState({
+        submitting: true
+      });
+      this.bulkUploadSubmit();
+    }
   }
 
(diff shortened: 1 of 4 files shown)


app/helpers/samples_helper.rb
```diff
@@ -126,14 +126,22 @@ def parsed_samples_for_s3_path(s3_path, project_id, host_genome_id)
...
+      matched_paired = InputFile::BULK_FILE_PAIRED_REGEX.match(file_name)
+      matched_single = InputFile::BULK_FILE_SINGLE_REGEX.match(file_name)
+      if matched_paired
+        matched = matched_paired
+        read_idx = matched[2].to_i - 1
+      elsif matched_single
+        matched = matched_single
+        read_idx = 0
+      else
+        next
```
boris-dimitrov: what does this mean?   malformed input is skipped?

app/assets/src/components/BulkUploadImport.jsx
```diff
@@ -271,7 +272,7 @@ class BulkUploadImport extends React.Component {
         that.setState({
           submitting: false
         });
-        that.gotoPage(`/?ids=${that.state.createdSampleIds.join(',')}`);
+        that.gotoPage(`/?project_id=${that.state.projectId}`);
       }, 2000)
```
boris-dimitrov: was the problem perhaps that this timeout is only 2000 milliseconds, i.e., enough to submit a single sample but not enough to submit a batch of samples?

app/helpers/samples_helper.rb
cdebourcy: files in the s3 folder that are not fastq/fasta files are skipped (e.g. metadata files)

PR #836, opened 2018-03-07
fix read downloads

boris-dimitrov:


boris-dimitrov:
charles' fix is better

PR #845, opened 2018-03-08
multipart upload through CLI

cdebourcy:
goes with https://github.com/chanzuckerberg/idseq-cli/pull/18

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -274,6 +274,7 @@ class Sample < ApplicationRecord
   end
 
   def concatenate_input_parts
+    return unless [STATUS_CREATED, STATUS_UPLOADED].include?(status)
     input_files.each do |f|
       next unless f.source_type == 'local'
       parts = f.parts.split(", ")
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -274,7 +274,7 @@ class Sample < ApplicationRecord
   end
 
   def concatenate_input_parts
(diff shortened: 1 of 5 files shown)


jshoe:
looks fine after you fix that file.parts reference?

cdebourcy:
Good catch! Haven’t added the schema migration

app/models/sample.rb
```diff
@@ -83,7 +84,12 @@ def set_presigned_url_for_local_upload
     input_files.each do |f|
       if f.source_type == 'local'
         # TODO: investigate the content-md5 stuff https://github.com/aws/aws-sdk-js/issues/151 https://gist.github.com/algorist/385616
-        f.update(presigned_url: S3_PRESIGNER.presigned_url(:put_object, bucket: SAMPLES_BUCKET_NAME, key: f.file_path))
+        parts = f.parts.split(", ")
```
yunfangjuan: split by comma and space?

app/models/sample.rb
```diff
@@ -267,6 +273,27 @@ def check_host_genome
     s3_preload_result_path.strip!
   end
 
+  def concatenate_input_parts
```
yunfangjuan: you are doing this for every sample.save?

cdebourcy: thanks, added a condition

app/models/sample.rb
cdebourcy: yes, is that ok?

Issue #796, opened 2018-02-28
problems with new lineage data and its start and end times


boris-dimitrov:
1) Missing genus names.  Bug in the scripts?

2) The BETWEEN operator is not the correct operator to select rows whose start/end time contains the taxon_count row time, because "t BETWEEN start AND end" is equivalent to "start <= t <= end" and what we actually need here is "start <= t < end" (the interval must be half-open).

3) Renaming a genus is highly problematic.  The current implementation permits rows with the old genus_name and the new genus_name to both be in effect -- with overlapping start/end times.  This means at the time we run genus_aggregation there is no uniquely defined name for the genus.  In fact, this causes every job to fail after host alignment, because the query in genus_aggregation fails with duplicate key update because of this. 

This approach of storing incremental version history in the table is not really feasible unless we normalize the schema and split out the name table again.

My recommendation is to revert this change until it better thought through.

cc @yunfangjuan 

```
mysql> select distinct(genus_name), genus_taxid, started_at, ended_at from taxon_lineages where genus_taxid in (65013, 81736, 518515, 337498, -200)  order by genus_taxid;
+-------------+-------------+---------------------+---------------------+
| genus_name  | genus_taxid | started_at          | ended_at            |
+-------------+-------------+---------------------+---------------------+
| NULL        |        -200 | 2000-01-01 00:00:00 | 3000-01-01 00:00:00 |
|             |        -200 | 2000-01-01 00:00:00 | 2018-02-23 00:58:30 |
|             |        -200 | 2018-02-23 00:58:30 | 2018-02-23 01:17:29 |
|             |        -200 | 2018-02-23 01:17:29 | 3000-01-01 00:00:00 |
|             |       65013 | 2000-01-01 00:00:00 | 2018-02-23 00:58:30 |
|             |       65013 | 2018-02-23 01:17:29 | 3000-01-01 00:00:00 |
| Marquesia   |       65013 | 2018-02-23 00:58:30 | 2018-02-23 01:17:29 |
|             |       81736 | 2000-01-01 00:00:00 | 2018-02-23 00:58:30 |
|             |       81736 | 2018-02-23 01:17:29 | 3000-01-01 00:00:00 |
| Barnardia   |       81736 | 2018-02-23 00:58:30 | 2018-02-23 01:17:29 |
|             |      337498 | 2000-01-01 00:00:00 | 2018-02-23 00:58:30 |
[...]

PR #861, opened 2018-03-09
replace null by empty string in taxon_lineages

cdebourcy:
See my post in https://github.com/chanzuckerberg/idseq-web/issues/796


jshoe:
LGTM

jshoe:
@cdebourcy this is good to merge right?

PR #887, opened 2018-03-12
Move out onClick listener so whole project row is clickable

### LINKED ISSUE #771: Make whole box clickable for project names
hovering over project box (but not over name) results in the box highlighted but unclickable - you have to click on the name to navigate. we should make the entire box clickable

jshoe:
Closes #771 . Moves out the onClick listener so the whole project row incl. the text title is clickable.

screen shot 2018-03-12 at 3 45 17 pm

### DIFF
--- a/app/assets/src/components/ProjectSelection.jsx
+++ b/app/assets/src/components/ProjectSelection.jsx
@@ -50,6 +50,7 @@ import Samples from './Samples';
   }
 
   toggleFavorite(e) {
+    e.stopPropagation();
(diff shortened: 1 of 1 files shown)


cdebourcy:
Almost. When you click only the star to "favorite" the project, the project should not load. Upon click anywhere else in the box, the project should load.

jshoe:
Ah right right

jshoe:
@cdebourcy ok should be fixed now!

PR #904, opened 2018-03-14
Add filter tags to samples page

cdebourcy:


cdebourcy:
screen shot 2018-03-14 at 10 28 08 am

app/assets/src/components/Samples.jsx
```diff
@@ -917,7 +917,42 @@ class Samples extends React.Component {
     location.href = `/projects/${id}/csv`;
   }
 
+  applyExcluded(e, type, state_var) {
+    let id = e.target.getAttribute('data-exclude')
+    if (type === "int") {
+      id = +id
+    }
+    let list = this.state[state_var].splice(0);
```
yunfangjuan: use Object.assign instead if you intend to copy the content?

app/assets/src/components/Samples.jsx
```diff
@@ -917,7 +917,42 @@ class Samples extends React.Component {
...
+    let index;
+    index = list.indexOf(id);
+    list.splice(index, 1);
```
yunfangjuan: check index >= 0 in case of contention

app/assets/src/components/Samples.jsx

app/assets/src/components/Samples.jsx

PR #907, opened 2018-03-14
ratio of gsnap to rapsearch servers needs to increase

boris-dimitrov:
based on recent observations with the reduced gsnap chunk size


boris-dimitrov:
This is super low risk, and I tested locally.    Feel free to merge into master at any point.    The impact on cost is considerable as it reduces the time the entire rapsearch tier sits idle while waiting for gsnap to finish.

PR #916, opened 2018-03-14
Make style tweaks to report page top filters

jshoe:
- So I managed to swap out the "name type" and "background model" dropdowns for Semantic UI dropdowns since that was straightforward.
- The search box, category box, and advanced filters box had a lot of custom JS/HTML so I didn't want to change all of that right now for the sake of time, so I just standardized CSS tweaks for those.
- So all of the JS hooks and actions should be the same / functionally the same as before basically.
- Might not fix all the visual bugs people mentioned but good enough to get out now.

PART ONE:
part 1

PART TWO:
part 2

### DIFF
--- a/app/assets/src/components/PipelineSampleReads.jsx
+++ b/app/assets/src/components/PipelineSampleReads.jsx
@@ -508,7 +508,7 @@ class PipelineSampleReads extends React.Component {
                   <li>
                     <a href={this.getDownloadLink()}>
                       <Button icon labelPosition="left" className="icon link download-btn">
-                        <Icon name="cloud download alternate" />
+                        <Icon name="cloud download" />
                         Download
                       </Button>
                     </a>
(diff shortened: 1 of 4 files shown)


yunfangjuan:
Hi,
I tested the branch. It looks a lot better. A few comments. 
1. When you have multiple advanced filters, the save button will be pushed out of the modal, which is not desirable. 
2.  when the advanced filters wrapped, it's not aligned with other drop downs. 
3. The "Categories" and "Advanced Filters" text are not vertically centered in the dropdown.

jshoe:
@yunfangjuan Okay I made those 3 changes!

part 3 3

yunfangjuan:
awesome let me check

jshoe:
@eggerdesigns advanced filters look a little better now

screen shot 2018-03-14 at 2 27 13 pm

jshoe:
Also fixes "cloud download alternate" references now. Semantic UI React calls the icon 'cloud download' but Semantic UI regular calls it 'cloud download alternate'

PR #935, opened 2018-03-15
Allow project creation through CLI, allow deleting empty projects in UI

cdebourcy:


app/controllers/projects_controller.rb
```diff
@@ -164,10 +164,6 @@ def update
   # DELETE /projects/1.json
   def destroy
     @project.destroy
```
yunfangjuan: I think you need to put in some safe guard here. only destroy the project wheren there are no samples under it.

app/controllers/home_controller.rb
```diff
@@ -6,7 +6,14 @@ class HomeController < ApplicationController
 
   def index
     @favorite_projects = current_user.favorites
-    @projects = current_power.projects
+    project_records = current_power.projects
+    @projects = []
+    project_records.each do |p_r|
+      p = p_r.as_json
+      p["can_delete"] = p_r.can_delete?(current_user) ? 1 : 0
```
yunfangjuan: instead of doing this. I think it would be easier to read by doing 
updatable_projects_ids = current_power.updatable_projects.pluck(:id)
  project_records.each do |prj| 
    p = prj.attributes
    p['can_edit'] = updatable_projects_ids.include?(p['id'))
  end

You can calculate if the user can delete the project in javascript. (user with edit privilege should be able to see all the samples.)

app/controllers/projects_controller.rb
cdebourcy: ok

PR #988, opened 2018-03-19
allow deleting project if empty

cdebourcy:
Ready to merge

### DIFF
--- a/app/assets/src/components/Samples.jsx
+++ b/app/assets/src/components/Samples.jsx
@@ -55,6 +55,7 @@ class Samples extends React.Component {
     this.startReportGeneration = this.startReportGeneration.bind(this);
     this.checkReportDownload = this.checkReportDownload.bind(this);
     this.displayReportProgress = this.displayReportProgress.bind(this);
+    this.deleteProject = this.deleteProject.bind(this);
 
     this.state = {
       invite_status: null,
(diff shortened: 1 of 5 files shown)


app/controllers/projects_controller.rb
```diff
@@ -163,10 +163,16 @@ def update
   # DELETE /projects/1
   # DELETE /projects/1.json
   def destroy
-    @project.destroy
+    deletable = @project.samples.to_a.empty?
```
yunfangjuan: I disabled project.samples so this might always be true. 
Please do Sample.where(project_id: @project.id).count == 0

app/assets/src/components/Samples.jsx
```diff
@@ -937,6 +938,23 @@ class Samples extends React.Component {
     });
   }
 
+  deleteProject(e) {
+    console.log(this.state.selectedProjectId);
```
yunfangjuan: //comment.

app/assets/src/components/Samples.jsx
```diff
@@ -937,6 +938,23 @@ class Samples extends React.Component {
...
+    let projectId = this.state.selectedProjectId;
+    this.nanobar.go(30);
+    console.log(`/projects/${projectId}`)
```
yunfangjuan: //comment

cdebourcy:
@yunfangjuan Thanks, addressed the comments!

PR #1003, opened 2018-03-20
Closes #918: Email user when all samples in project have completed

### LINKED ISSUE #918: Create email alert when all samples in a project have finished running

cdebourcy:
Need to do an end-to-end test on alpha.


app/models/pipeline_run_stage.rb
```diff
@@ -310,6 +310,35 @@ def db_load_postprocess
...
     `
     _stdout, _stderr, _status = Open3.capture3("rm -f #{downloaded_byteranges_path}")
+    notify_users if notify?
+  end
+
+  def project_id
+    pipeline_run.sample.project_id
+  end
+
+  def notify?
```
yunfangjuan: I think 320-341 should be in the model pipeline_run.rb not in pipeline_run_stage.rb  and it will trigger this when all the pipeline_runs are complete.

app/models/pipeline_run_stage.rb
```diff
@@ -310,6 +310,35 @@ def db_load_postprocess
...
+    statuses_in_project = PipelineRun.connection.select_all("
+      SELECT
+        pipeline_runs.job_status
```
yunfangjuan: this is problematic when there are other previous runs that failed. 

This will be my query 

x = PipelineRun.where("id in (select max(id) from pipeline_runs group by sample_id) and sample_id in (select id from samples where project_id = #{project_id.to_i})").where("status != ?",   PipelineRun::STATUS_CHECKED)
x.count > 0

cdebourcy: thanks!!

app/models/pipeline_run_stage.rb
cdebourcy: Ok, moved. Did you want me to attach the action to a different event? I thought it would make the most sense to check whenever a run completes (i.e. when db_load_postprocess happens)

app/models/pipeline_run.rb
```diff
@@ -137,19 +137,20 @@ def update_job_status
     if all_stages_succeeded
       self.finalized = 1
       self.job_status = STATUS_CHECKED
+      save
+      notify_users if notify?
+      return
```
boris-dimitrov: Let's not have returns in the middle of functions.    I refactored this code earlier to get rid of them, because they prove error prone over time, particularly in environments with exceptions and hooks.

app/models/pipeline_run.rb
```diff
@@ -301,4 +302,25 @@ def count_unmapped_reads
     unidentified_fasta = get_s3_file(sample.unidentified_fasta_s3_path)
     unidentified_fasta.lines.select { |line| line.start_with? '>' }.count if unidentified_fasta
   end
+
+  delegate :project_id, to: :sample
+
+  def notify?
+    incomplete_runs = PipelineRun.where("id in (select max(id) from pipeline_runs group by sample_id) and sample_id in (select id from samples where project_id = #{project_id.to_i})").where("job_status != ?", PipelineRun::STATUS_CHECKED)
```
boris-dimitrov: I am not sure what this query does ;)

cdebourcy: For all samples in a project, check whether its most recent pipeline_run is still in progress. (If so, we won't want to notify the user that the project is complete yet.)

yunfangjuan: I think incomplete_runs.count == 0 will probably make this query faster with less data transfered.

PR #1011, opened 2018-03-20
Refactor sample selection

yunfangjuan:
# Description

Refactor sample selection. Get rid of jquery logic. 
## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

# Test and Reproduce

1. Go to samples page. Select samples

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [X ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

### DIFF
--- a/app/assets/src/components/Samples.jsx
+++ b/app/assets/src/components/Samples.jsx
@@ -997,11 +997,6 @@ class Samples extends React.Component {
      })
   }
 
-  downloadTable(id) {
-    _satellite.track('downloadtable');
-    location.href = `/projects/${id}/csv`;
-  }
-
   applyExcluded(e, type, state_var) {
     let id = e.target.getAttribute("data-exclude");
     if (type === "int") {
--- a/app/assets/src/components/Samples.jsx
+++ b/app/assets/src/components/Samples.jsx
@@ -927,8 +927,7 @@ class Samples extends React.Component {
     let sampleList = this.state.selectedSampleIds;
     const checked = e.target.checked;
     const allSamples = this.state.allSamples;
-    for (let idx in allSamples) {
-      let sample = allSamples[idx];
+    for (let sample of allSamples) {
       if (sample.run_info.report_ready != 1) {
         continue;
       }
@@ -1106,7 +1105,6 @@ class Samples extends React.Component {
             id="checkAll"
             className="filled-in checkAll"
             checked = {this.state.allChecked}
-            onChange = {(e) => {}}
             onClick={this.fetchAllSelectedIds}
             />
           <label htmlFor="checkAll"></label>
--- a/app/assets/src/components/Samples.jsx
+++ b/app/assets/src/components/Samples.jsx
@@ -881,7 +881,8 @@ class Samples extends React.Component {
   viewSample(id, e) {
     e.preventDefault();
     // _satellite.track('viewsample')
-    location.href = `/samples/${id}`;
+
+    window.open(`/samples/${id}`);
   }
 
   renderEmptyTable() {
--- a/app/assets/src/components/Samples.jsx
+++ b/app/assets/src/components/Samples.jsx
@@ -591,7 +591,7 @@ class Samples extends React.Component {
               <div className='flex-container'>
                 <ul className='flex-items'>
                   <li className='check-box-container'>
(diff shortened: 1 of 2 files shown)


app/assets/src/components/Samples.jsx
```diff
@@ -588,29 +587,14 @@ class Samples extends React.Component {
...
-                        <label htmlFor={i}>{sample_name_info}</label>
-                      </div>
-                    ) : (
-                      sample_name_info
-                    )}
+            <div className='samples-card white'>
+              <div className='flex-container'>
+                <ul className='flex-items'>
+                  <li className='check-box-container'>
+                    { this.state.displaySelectSamples ? <div><input type="checkbox" id={i} onClick = { this.selectSample } key= {`sample_${dbSample.id}`} data-sample-id ={dbSample.id}
```
trevorb: If statements of this complexity are generally discouraged to do inline.

app/assets/src/components/Samples.jsx
```diff
@@ -938,53 +923,31 @@ class Samples extends React.Component {
...
-    $(".checkbox:enabled").each((id, element) => {
-      let sample_id = element.getAttribute("data-sample-id");
-      const sampleList = that.state.selectedSampleIds;
-      if (!sample_id) {
-        return;
+  fetchAllSelectedIds(e) {
+    let sampleList = this.state.selectedSampleIds;
+    const checked = e.target.checked;
+    const allSamples = this.state.allSamples;
+    for (let idx in allSamples) {
```
trevorb: You can use of in the for statement to get the actual values instead of just the idx. 

```
for (let sample of allSamples)
```

app/assets/src/components/Samples.jsx
```diff
@@ -938,53 +923,31 @@ class Samples extends React.Component {
...
-    var that = this;
-    let sampleList = that.state.selectedSampleIds;
-    $(".checkbox:enabled").each((id, element) => {
-      let sample_id = parseInt(element.getAttribute("data-sample-id"));
-      let indx = sampleList.indexOf(sample_id);
+      let sample_id = sample.db_sample.id;
       if (checked) {
-        if (indx === -1) {
-          sampleList.push(+sample_id);
+        if (sampleList.indexOf(sample_id) === -1) {
```
trevorb: You might want to consider making sampleList a Set object instead so you can use functions like has and delete instead of indexOf != -1 and indexOf/splice
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Babel provides a polyfill to provide full compatibility with older browsers.

app/assets/src/components/Samples.jsx
```diff
@@ -1143,8 +1102,14 @@ class Samples extends React.Component {
 
     let check_all = (
       <div className="check-all">
-        <input type="checkbox" id="checkAll" className="filled-in checkAll" />
-        <label htmlFor="checkAll" />
+          <input type="checkbox"
+            id="checkAll"
+            className="filled-in checkAll"
+            checked = {this.state.allChecked}
+            onChange = {(e) => {}}
```
trevorb: Defining a function in a component prop is generally bad practice as it will create a new function each time the component is rendered which means it will cause unnecessary re-rendering as react will always think that the onChange property changed.

yunfangjuan: OK. I think I got a warning without having it but I will remove it and try again.

app/assets/src/components/Samples.jsx
yunfangjuan: Thanks for the tip.

app/assets/src/components/Samples.jsx
yunfangjuan: ah thx

app/assets/src/components/Samples.jsx
yunfangjuan: yep. we have tons of these.

yunfangjuan:
Thanks!

PR #1016, opened 2018-03-20
Update the species number counts

jshoe:
- Adds feature to update the counts in searches.
- Not sure if there's a better way to reset the counts?

mar-20-2018 16-24-08

### DIFF
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -253,8 +253,6 @@ class PipelineSampleReport extends React.Component {
         const taxon = this.state.taxonomy_details[i];
         if (taxon.genus_taxid == taxon.tax_id) {
           if (matched_taxons.length > 0) {
-            // Update species count on page after a search
-            genus_taxon.species_count = matched_taxons.length;
             selected_taxons.push(genus_taxon);
             selected_taxons = selected_taxons.concat(matched_taxons);
           }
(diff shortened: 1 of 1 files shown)


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -253,6 +253,8 @@ class PipelineSampleReport extends React.Component {
         const taxon = this.state.taxonomy_details[i];
         if (taxon.genus_taxid == taxon.tax_id) {
           if (matched_taxons.length > 0) {
+            // Update species count on page after a search
```
yunfangjuan: line 270 needs the same condition.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -297,15 +299,27 @@ class PipelineSampleReport extends React.Component {
         }
       }
     } else {
-      selected_taxons = thresholded_taxons;
+      // Called for All search
+      let included = thresholded_taxons;
+      for (let i = 0; i < included.length; i++) {
```
yunfangjuan: hm this doesn't apply to excludeCategory filters?

jshoe:
Ah OK I was a little unclear on the exact logic

jshoe:
@yunfangjuan Updated. Makes much more sense now

jshoe:
@cdebourcy

PR #1017, opened 2018-03-20
Charles/phage filter

cdebourcy:
Adds an ad-hoc checkbox for phage to category filter.
Also changes the styles of the tags to match those on the sample page.

First, you need to run
```
PipelineRun.all.map(&:update_genera)
PipelineRun.all.map(&:update_is_phage)
```
Until you do that, everything is classified as "not a phage" by default.


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1045,6 +1077,17 @@ class PipelineSampleReport extends React.Component {
                                 </li>
                               )
                             })}
+                              <br /><div className="divider" /><br />
+                              <li key='phage'>
+                                <input type="checkbox"
+                                className="filled-in cat-filter"
+                                id='phage'
+                                value='phage'
+                                onChange={(e) => {}}
```
yunfangjuan: I think you can remove this empty onChange function.

app/models/pipeline_run.rb
```diff
@@ -264,6 +264,20 @@ def update_genera
     ")
   end
 
+  def update_is_phage
+    TaxonCount.connection.execute("
+      UPDATE taxon_counts
+      INNER JOIN taxon_lineages ON taxon_counts.tax_id = taxon_lineages.taxid
+      SET taxon_counts.is_phage = IF(taxon_lineages.family_name IN ('Myoviridae', 'Siphoviridae', 'Podoviridae', 'Lipothrixviridae',
```
yunfangjuan: I think these names should be extracted out as an array constants.

yunfangjuan: Also, by looking at update_genera function, I think you can actually put family_taxid into taxon_counts in the same query and just update the query in the report_helper, it's cleaner that way.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -226,6 +230,7 @@ class PipelineSampleReport extends React.Component {
     this.setState({
       activeThresholds: [Object.assign({}, this.defaultThreshold)],
       excluded_categories: [],
+      exclude_phage: false,
```
yunfangjuan: this is fine for now but I think since you are going to introduce fungi, you probably should put the list into sub_categories and excluding them will become generic. 

Also, right now when you uncheck virus, the phage filter doesn't get unchecked. I think the relationship should be speficied and when people click on exclude virus, the phage filter will be excluded by default. same for fungi/euracryotes in the future.

app/models/pipeline_run.rb
yunfangjuan: Also, family_taxid is in general pretty useful that we can put into each row in the report soon.

cdebourcy: Ok, set array constant and added family_taxid to taxon_counts. Do we still want to explicitly set the column is_phage? I think yes, so that we don't need to check the entire list of phage families on the fly each time we display a report?

app/assets/src/components/PipelineSampleReport.jsx
cdebourcy: Turns out it's a way to avoid the following warning:
```
Warning: Failed prop type: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`
```

cdebourcy: Using `defaultChecked` doesn't give the desired behavior.

cdebourcy:
Ok, generalized it. It should now be easy to add Fungi filter without changing much frontend code.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -459,6 +473,18 @@ class PipelineSampleReport extends React.Component {
     return displayed_categories;
   }
 
+  displayedSubcats(excludeSubcats) {
+    let displayed_subcats = [];
+    let all_subcats = Object.keys(this.category_child_parent);
+    for (let i = 0; i < all_subcats.length; i++) {
```
yunfangjuan: i think you can do for(let subcat of all_subcats)

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -472,13 +498,50 @@ class PipelineSampleReport extends React.Component {
     } else {
       excluded_categories.push(e.target.value);
     }
+    // Also update subcategory to match category
+    let new_exclude_subcats = this.state.exclude_subcats;
+    let subcats = Object.keys(this.category_child_parent);
+    for (let i = 0; i < subcats.length; i++) {
+      let subcat = subcats[i];
+      let parent = this.category_child_parent[subcat]
+      let i_excluded_categories = excluded_categories.indexOf(parent)
```
yunfangjuan: I will rewrite this whole logic as: 

let parent_excluded = (excluded_categories.indexOf(parent) >= 0); 

if (parent_excluded) {
    //subcat should be excluded 
   if (new_exclude_subcats.indexOf(subcat) <= 0) {
      new_exclude_subcats.push(subcat);
  }
} else {
  let idx = new_exclude_subcats.indexOf(subcat); 
  if (idx >= 0) {
    new_exclude_subcats.splice(idx, 1);
}
}

PR #1018, opened 2018-03-21
Pull out more functions piece by piece

jshoe:


jshoe:
agreed :upside_down_face:

PR #1058, opened 2018-03-26
Closes #1028: add pipeline run date

### LINKED ISSUE #1028: show date of pipeline run
next to "PIPELINE v1.0.10" on report page. E.g. "PIPELINE v1.0.10 (last run on [date])"
Here 'date' should be the 'created_at' date of the sample's latest pipeline_run.

cdebourcy:


app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -473,6 +473,10 @@ class PipelineSampleReads extends React.Component {
       version_display = version_display + ', NR ' + this.pipelineRun.version.nr
     }
 
+    let run_date_display = !this.pipelineRun ? '' :
+                             !this.pipelineRun.created_at ? '' :
```
jshoe: could split into a few lines for readability?

boris-dimitrov: presenting a date right next to the version makes me think this is when the version was cut

PR #1073, opened 2018-03-28
Add retries to Docker push

jshoe:
- Retry up to 5 times total if Docker push fails.

### DIFF
--- a/bin/push-docker
+++ b/bin/push-docker
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -e
 
 export REPO=chanzuckerberg/idseq-web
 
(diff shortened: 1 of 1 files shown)


jshoe:
Oops didn't work on Travis.. looking into this

bin/push-docker
```diff
@@ -12,5 +11,19 @@ echo $COMMIT
...
+
+status=1
+count=1
+# Retry up to 5 times total if failed Docker push
+until [ ${status} -eq 0 ] || [ $count -gt 5 ]; do
+    echo "Trying Docker push..."
+    docker push $REPO
+    status=$?
+    count=$((count+1))
+    sleep 15
```
boris-dimitrov: This is fine for this script, but the pattern in general is flawed, because after the last failed retry it will needlessly delay for 15 seconds and then exit with success code.

boris-dimitrov: In fact it even sleeps after a successful try, which is unnecessary.

jshoe:
@boris-dimitrov better logic

PR #1084, opened 2018-03-29
Load ERCC counts from S3 and fix RPM

cdebourcy:


app/lib/report_helper.rb
```diff
@@ -195,7 +195,8 @@ def all_categories
 
   def fetch_taxon_counts(pipeline_run_id, background_id)
     pipeline_run = PipelineRun.find(pipeline_run_id)
-    adjusted_total_reads = pipeline_run.total_reads * pipeline_run.subsample_fraction
+    total_ercc_reads = pipeline_run.ercc_counts.map(&:count).sum # will be zero if no ercc_counts
```
yunfangjuan: we should just put the total in a new pipeline_run field so we can avoid querying database every time for this.

yunfangjuan:
you probably want to write a quick rake script to load the RR007 reads_per_gene.star.tab into the database.

app/models/pipeline_run_stage.rb
```diff
@@ -234,20 +234,33 @@ def postprocess_command
...
+    # Load version
     version_s3_path = "#{pr.sample_output_s3_path}/#{PipelineRun::VERSION_JSON_NAME}"
     pr.version = `aws s3 cp #{version_s3_path} -`
 
-    # TODO(yf): remove the following line
-    pr.job_stats_attributes = stats_array
-
-    # rm the json
-    _stdout, _stderr, _status = Open3.capture3("rm -f #{downloaded_stats_path}")
+    # Load ERCC counts
```
yunfangjuan: 251-263 should go into its own function.

app/lib/report_helper.rb
cdebourcy: ok

PR #1093, opened 2018-03-29
make index on ERCC name unique

cdebourcy:


db/migrate/20180329221958_add_index_to_ercc_counts.rb
```diff
@@ -0,0 +1,6 @@
+class AddIndexToErccCounts < ActiveRecord::Migration[5.1]
+  def change
+    remove_index :ercc_counts, :name
+    add_index :ercc_counts, :name, unique: true
```
yunfangjuan: Unique on name+ pipeline run id?

cdebourcy: corrected

PR #1095, opened 2018-03-30
Fix RPM in heatmap and backgrounds

cdebourcy:


app/models/background.rb
```diff
@@ -17,7 +17,19 @@ def validate_size
   end
 
   def summarize
-    results = TaxonCount.connection.select_all("SELECT tax_id, count_type, tax_level, sum((1.0*1e6*count)/total_reads) as sum_rpm, sum((1.0*1e6*count*1e6*count)/(total_reads*total_reads)) as sum_rpm2 FROM `taxon_counts` INNER JOIN `pipeline_runs` ON `pipeline_runs`.`id` = `taxon_counts`.`pipeline_run_id` WHERE (pipeline_run_id in (select pipeline_run_id from backgrounds_pipeline_runs where background_id = #{id}))  GROUP BY tax_id, count_type, tax_level").to_hash
+    results = TaxonCount.connection.select_all("
+      SELECT
+        tax_id,
+        count_type,
+        tax_level,
+        sum((1.0*1e6*count)/(total_reads-IFNULL(total_ercc_reads,0))) AS sum_rpm,
```
yunfangjuan: IF_NULL

also you can do something like 
(total_reads - IFNULL((total_ercc_reads,0)) as real_reads, 
and use real_reads for the following calculation to make it more readable.

cdebourcy: ok. Will make adjusted_total_reads just like for report: need to also account for subsampling in a later diff. Not applicable to any of the backgrounds that currently exist

PR #1097, opened 2018-03-30
adjust total reads by subsampling fraction in background calculation

cdebourcy:


app/models/background.rb
```diff
@@ -22,8 +22,9 @@ def summarize
         tax_id,
         count_type,
         tax_level,
-        sum((1.0*1e6*count)/(total_reads-IFNULL(total_ercc_reads,0))) AS sum_rpm,
-        sum((1.0*1e6*count*1e6*count)/((total_reads-IFNULL(total_ercc_reads,0))*(total_reads-IFNULL(total_ercc_reads,0)))) AS sum_rpm2
+        @adjusted_total_reads := (total_reads - IFNULL(total_ercc_reads, 0)) * fraction_subsampled AS adjusted_total_reads,
```
boris-dimitrov: Probably don't need "AS adjusted_total_reads".   Make sure not to group by or order by something like this.   I hope it works fine under SUM, unclear from the docs.

PR #1101, opened 2018-03-30
Quick fix for checkbox issue

jshoe:
- Main issue was that htmlFor was generating ID's that were the same for multiple checkboxes across the page.
- Will do more to make the checkboxes consistent later?

### DIFF
--- a/app/assets/src/components/Samples.jsx
+++ b/app/assets/src/components/Samples.jsx
@@ -223,9 +223,11 @@ class Samples extends React.Component {
   selectHostFilter(e) {
     // current array of options
     const hostList = this.state.selectedHostIndices.slice(0);
+
+    let target_id = parseInt(e.target.getAttribute("data-host-id"));
+
     let index;
     // check if the check box is checked or unchecked
-    let target_id = e.target.id.slice(5); // Remove 'host-' from front.
     if (e.target.checked) {
       // add the numerical value of the checkbox to options array
       hostList.push(+target_id);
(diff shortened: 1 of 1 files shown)


app/assets/src/components/Samples.jsx
```diff
@@ -225,12 +225,13 @@ class Samples extends React.Component {
     const hostList = this.state.selectedHostIndices.slice(0);
     let index;
     // check if the check box is checked or unchecked
+    let target_id = e.target.id.slice(5); // Remove 'host-' from front.
```
cdebourcy: How about you get target_id from a new attribute, which you could name data-host-id, via 
`let target_id = e.target.getAttribute("data-host-id")`?
It would be cleaner

app/assets/src/components/Samples.jsx
```diff
@@ -1939,31 +1940,32 @@ function SampleDetailedColumns({
...
+            let label_i = `host-${host.id}`;
             let res = (
-              <div key={i} className="options-wrapper">
+              <div key={"host-"+i} className="options-wrapper">
                 <input
                   name="host"
                   type="checkbox"
-                  data-id={host.id}
+                  data-id={"host-"+host.id}
                   checked={indices.indexOf(host.id) < 0 ? "" : "checked"}
```
cdebourcy: can use `label_i` in lines 1961 and 1965 to avoid repeating the code?

PR #1105, opened 2018-03-31
Revert "Avoid failure in populate_fraction_subsampled.rake"

cdebourcy:
See if it fixes the build.
Boris, you'll have to reapply "redis locks are no longer necessary"


cdebourcy:
Making this change indeed fixed the build.

PR #1120, opened 2018-04-02
add kingdom

cdebourcy:
Goes with: https://github.com/chanzuckerberg/ncbitax2lin/pull/10


app/models/pipeline_run.rb
```diff
@@ -257,15 +257,6 @@ def update_names
   end
 
   def update_genera
-    # Make sure to run update_genera after generate_aggregate_counts
-    # HACK This should probably have been accomplished with schema DEFAULTs
-    TaxonCount.connection.execute("
```
jshoe: would it be possible to write a migration for these?

PR #1121, opened 2018-04-02
Closes #1122: hide concordance column

### LINKED ISSUE #1122: Hide % conc column on report page

cdebourcy:
until we fix its correctness issue


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1185,10 +1187,10 @@ function ThresholdMap() {
       name: "NT log(1/e)",
       value: "NT_neglogevalue"
     },
-    {
-      name: "NT %conc",
-      value: "NT_percentconcordant"
-    },
+    //{
```
yunfangjuan: Instead of comment this out, 
check  this.showConcordance to decide to hide it or not?

PR #1143, opened 2018-04-04
(Do not merge) new hit calling

cdebourcy:
Not intended to be merged. Keep as separate branch to test new hit calling on alpha when pipeline branch "charles/hit-calling" is used for processing.
(see idseq-pipeline: https://github.com/chanzuckerberg/idseq-pipeline/pull/120)

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -18,7 +18,7 @@ class PipelineRun < ApplicationRecord
   accepts_nested_attributes_for :ercc_counts
 
   DEFAULT_SUBSAMPLING = 1_000_000 # number of reads to subsample to, after host filtering
-  OUTPUT_JSON_NAME = 'idseq_web_sample.json'.freeze
+  OUTPUT_JSON_NAME = 'multihit_idseq_web_sample.json'.freeze
   STATS_JSON_NAME = 'stats.json'.freeze
   VERSION_JSON_NAME = 'versions.json'.freeze
   ERCC_OUTPUT_NAME = 'reads_per_gene.star.tab'.freeze
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -18,7 +18,11 @@ class PipelineRun < ApplicationRecord
   accepts_nested_attributes_for :ercc_counts
 
   DEFAULT_SUBSAMPLING = 1_000_000 # number of reads to subsample to, after host filtering
-  OUTPUT_JSON_NAME = 'multihit_idseq_web_sample.json'.freeze
+  OUTPUT_JSON_NAME = if pr.pipeline_branch == 'charles/hit-calling'
+                       'multihit_idseq_web_sample.json'
+                     else
+                       'idseq_web_sample.json'
+                     end
   STATS_JSON_NAME = 'stats.json'.freeze
   VERSION_JSON_NAME = 'versions.json'.freeze
   ERCC_OUTPUT_NAME = 'reads_per_gene.star.tab'.freeze
--- a/app/models/pipeline_run_stage.rb
+++ b/app/models/pipeline_run_stage.rb
@@ -281,13 +281,23 @@ class PipelineRunStage < ApplicationRecord
     version_s3_path = "#{pipeline_run.alignment_output_s3_path}/#{PipelineRun::VERSION_JSON_NAME}"
     pr.version = `aws s3 cp #{version_s3_path} -`
 
+    # only keep counts at certain taxonomic levels
+    taxon_counts_attributes_filtered = []
+    acceptable_tax_levels = [TaxonCount::TAX_LEVEL_SPECIES]
+    acceptable_tax_levels << TaxonCount::TAX_LEVEL_GENUS if pr.pipeline_branch == 'charles/hit-calling'
+    pipeline_output_dict['taxon_counts_attributes'].each do |tcnt|
+      if acceptable_tax_levels.include?(tcnt['tax_level'].to_i)
+        taxon_counts_attributes_filtered << tcnt
+      end
+    end
+
     pr.job_stats.delete_all
     pr.job_stats_attributes = stats_array
     pr.taxon_counts_attributes = taxon_counts_attributes_filtered
     pr.updated_at = Time.now.utc
     pr.save
     # aggregate the data at genus level
-    pr.generate_aggregate_counts('genus')
+    pr.generate_aggregate_counts('genus') unless pr.pipeline_branch == 'charles/hit-calling'
     # merge more accurate name information from lineages table
     pr.update_names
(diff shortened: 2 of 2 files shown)


app/models/taxon_lineage.rb
```diff
@@ -1,5 +1,6 @@
 # The TaxonLineage model gives the taxids forming the taxonomic lineage of any given species-level taxid.
 class TaxonLineage < ApplicationRecord
+  INVALID_CALL_BASE_ID = -1e8 # don't run into -2e9 limit
```
boris-dimitrov: -10_000_000 please

app/models/pipeline_run.rb
```diff
@@ -19,6 +19,7 @@ class PipelineRun < ApplicationRecord
 
   DEFAULT_SUBSAMPLING = 1_000_000 # number of reads to subsample to, after host filtering
   OUTPUT_JSON_NAME = 'idseq_web_sample.json'.freeze
+  MULTIHIT_OUTPUT_JSON_NAME = 'multihit_idseq_web_sample.json'.freeze
```
boris-dimitrov: Seems unwise to rename this file.

boris-dimitrov: But if it has to be renamed, the decision on which filename to use should be made based on pipeline version not pipeline branch.

app/models/sample.rb
```diff
@@ -356,7 +356,7 @@ def kickoff_pipeline
     # The subsample field of "sample" is currently used as a simple flag (UI checkbox),
     # but was made an integer type in case we want to allow users to enter the desired number
     # of reads to susbample to in the future
-    pr.pipeline_branch = pipeline_branch.blank? ? "master" : pipeline_branch
+    pr.pipeline_branch = pipeline_branch.blank? ? "charles/hit-calling" : pipeline_branch
```
boris-dimitrov: Let's never change the default branch like this.   When a branch is ready to become default for an environment, it can be called release-{environment} (or until then, just master).  An admin can always special-request a branch.   We should avoid making versions of idseq-web and idseq-pipeline that interlock.

app/models/pipeline_run.rb
boris-dimitrov: An alternative way of implementing such a rename without complicating the code with version checks is as follows

1)  Make pipeline emit both files, compatible with old webapp

2) In a week or two, make webapp only look for the new file, and make pipeline no longer generate the old one.

We should also get rid of that "wait for all files to be generated with timestamps after the job submission" approach, as it leads to cumbersome situations like this one.

app/models/pipeline_run_stage.rb
```diff
@@ -296,7 +308,7 @@ def db_load_alignment
     pr.updated_at = Time.now.utc
     pr.save
     # aggregate the data at genus level
-    pr.generate_aggregate_counts('genus')
+    pr.generate_aggregate_counts('genus') unless pr.pipeline_branch == 'charles/hit-calling'
```
boris-dimitrov: yeah this too needs to be based on something else, pipeline version perhaps

app/models/pipeline_run_stage.rb
```diff
@@ -282,10 +292,12 @@ def db_load_alignment
...
     pr.version = `aws s3 cp #{version_s3_path} -`
 
-    # only keep species level counts
+    # only keep counts at certain taxonomic levels
     taxon_counts_attributes_filtered = []
+    acceptable_tax_levels = [TaxonCount::TAX_LEVEL_SPECIES]
+    acceptable_tax_levels << TaxonCount::TAX_LEVEL_GENUS if pr.pipeline_branch == 'charles/hit-calling'
     pipeline_output_dict['taxon_counts_attributes'].each do |tcnt|
-      if tcnt['tax_level'].to_i == TaxonCount::TAX_LEVEL_SPECIES
+      if acceptable_tax_levels.include?(tcnt['tax_level'].to_i) && !invalid_genus_call?(tcnt)
```
boris-dimitrov: Why are we dropping hits that align poorly at the genus level but very well at a family level?

boris-dimitrov: The invalid_genus_call function also catches invalid species calls, doesn't it?

app/models/pipeline_run_stage.rb
```diff
@@ -256,10 +256,20 @@ def db_load_host_filtering
...
   end
 
+  def output_json_name
+    pipeline_run.pipeline_branch == 'charles/hit-calling' ? PipelineRun::MULTIHIT_OUTPUT_JSON_NAME : PipelineRun::OUTPUT_JSON_NAME
+  end
+
+  def invalid_genus_call?(tcnt)
+    tcnt['genus_taxid'].to_i < TaxonLineage::INVALID_CALL_BASE_ID
+  rescue
+    false
```
boris-dimitrov: Why is this `false` and not `true`?

boris-dimitrov: Doesn't this function return `true` for invalid calls at any level, whether genus, species, or family?

boris-dimitrov:
Made some of the necessary fixes in https://github.com/chanzuckerberg/idseq-web/pull/1182/files but an unresolved question remains regarding the filtering of taxons in the db loader.

app/models/sample.rb
cdebourcy: It was meant as a temporary hack for letting Josh play around on alpha without needing to give him a long list of instructions. Never meant to be merged to master

app/models/pipeline_run_stage.rb
cdebourcy: Yes, we could choose to display them. I omitted them for now because the current report page design does not include family-level information, only genus and species. So seeing "Non-genus-specific <family name> hits" looked out of place to me.

app/models/pipeline_run_stage.rb
cdebourcy: No, if the species call is invalid, the genus_taxid may still be valid (positive).

cdebourcy: The rescue case should never occur, but when in doubt, I'm returning false so that the entry doesn't get dropped. That way, we would at least see the offending entry on the report page and catch the bug. If we hide it, we may never know.

app/models/pipeline_run.rb
cdebourcy: I agree, now that we have versioned results folders, we could have kept the name the same. The name change was meant to avoid any confusion because the meaning of what's in the file changes drastically from the old pipeline version to the new one.

app/models/pipeline_run_stage.rb
cdebourcy: same comment as for the other occurrence

app/models/sample.rb
cdebourcy: Thanks for turning it in something production-ready

PR #1146, opened 2018-04-05
Project-specific background

cdebourcy:
When all samples in project have completed, create or update a background made of all the project's samples.
This background will be viewable by anyone who has access to the project.
Checkbox in "Add User" modal to enable or disable the background.


app/models/background.rb
```diff
@@ -67,4 +67,16 @@ def store_summary
   def compute_stdev(sum, sum2, n)
     Math.sqrt((sum2 - sum**2 / n.to_f) / (n - 1))
   end
+
+  def self.viewable(user)
+    if user.admin?
+      all
+    else
+      where("project_id in (select project_id from projects_users where user_id=?)
```
yunfangjuan: make it where(project_id: Project.viewable(user).pluck(:id)) in case project.viewable changed in the future?

app/controllers/samples_controller.rb
```diff
@@ -157,7 +157,7 @@ def show
     @project_info = @sample.project ? @sample.project : nil
     @project_sample_ids_names = @sample.project ? Hash[current_power.project_samples(@sample.project).map { |s| [s.id, s.name] }] : nil
     @host_genome = @sample.host_genome ? @sample.host_genome : nil
-    @background_models = Background.all
+    @background_models = current_power.backgrounds
```
yunfangjuan: since you are adding access control related code. unit test is required. Please add a few test cases to 
- make sure users who are allowed access have access
- make sure users who are not allowed access won't access it.

yunfangjuan: You will also need to make sure people don't access individual background through report_info, so make sure you use current_power.backgrounds.find in the code.

test/controllers/power_controller_test.rb
```diff
@@ -207,4 +207,22 @@ class PowerControllerTest < ActionDispatch::IntegrationTest
...
+  test 'joe can view joe_sample with background for public_project' do
+    access_sample_with_background(backgrounds(:background_for_public_project), samples(:joe_sample))
+  end
+
+  test 'joe can view joe_sample with background for joe_project' do
+    access_sample_with_background(backgrounds(:background_for_joe_project), samples(:joe_sample))
+  end
+
+  test 'joe cannot view joe_sample with background for project one' do
+    access_sample_with_background(backgrounds(:background_for_project_one), samples(:joe_sample))
```
yunfangjuan: why is this case assert success?

PR #1152, opened 2018-04-06
Versioned result folder

yunfangjuan:
# Description

Make the results folder versioned so the results will be under
samples/{project_id}/{sample_id}/results/{pipeline_version} 
samples/{project_id}/{sample_id}/postprocess/{pipeline_version} 
 
Make the rerun/archive better
## Type of change

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:


app/models/pipeline_run.rb
```diff
@@ -293,21 +293,49 @@ def subsample_fraction
   end
 
   def subsample_suffix
-    subsample? ? "/subsample_#{subsample}" : ""
+    subsample? ? "subsample_#{subsample}" : (pipeline_version ? "subsample_all" : "")
```
boris-dimitrov: good

app/models/pipeline_run_stage.rb
```diff
@@ -201,7 +201,7 @@ def host_filtering_command
     file_type = sample.input_files.first.file_type
     batch_command_env_variables = "INPUT_BUCKET=#{sample.sample_input_s3_path} OUTPUT_BUCKET=#{sample.sample_output_s3_path} " \
       "FILE_TYPE=#{file_type} DB_SAMPLE_ID=#{sample.id} " \
-      "COMMIT_SHA_FILE=#{COMMIT_SHA_FILE_ON_WORKER} "
+      "COMMIT_SHA_FILE=#{COMMIT_SHA_FILE_ON_WORKER} PIPELINE_VERSION=\\`idseq_pipeline --version\\` "
```
boris-dimitrov: This is a little bit too clever :)   What you actually need to capture is the git hash at the top of the pipeline branch being submitted, so that each pipeline stage will checkout that exact hash.   The pipeline version doesn't need to be passed, should be available as a variable to the actual pipeline code.

boris-dimitrov: So line 23 in this file, `git checkout #{pipeline_run.pipeline_branch};` needs to change so that it checks out a specific git hash.   This hash can be determined by rev-parse as on line 24 at the time of creating the batch command.

boris-dimitrov: The command you need to run in ruby for this is 
```
pipeline_branch ||= "master"
`git ls-remote https://github.com/chanzuckerberg/idseq-pipeline.git | grep refs/heads/#{pipeline_branch}`.split[0]
```

yunfangjuan: i need the pipeline_version to determine the output directory. it's also only give the major version.

boris-dimitrov: Right, I understand the pipeline code will emit a pipeline_version.txt file that will be picked up by rails so it knows where to look for output files. 

My additional points above are two:

1) Before submitting a job, the rails code needs to find and remember the commit SHA for the pipeline branch, in order to ensure that the exact same pipeline code runs in all stages.

2) The pipeline version is directly available to the pipeline code and need not be passed through the environment.   In fact, passing it is more error-prone.

PR #1161, opened 2018-04-10
freeze the pipeline commit for all stages in a pipeline run

boris-dimitrov:
freeze the pipeline commit for all stages in a pipeline run

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -101,6 +101,8 @@ class PipelineRun < ApplicationRecord
       output_func: 'postprocess_outputs'
     )
     self.pipeline_run_stages = run_stages
+    # we consider the job successful after stage 2 completes, even if subsequent stages fail
+    # this is the only meaning of "ready_step"
     self.ready_step = 2
   end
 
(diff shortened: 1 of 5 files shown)


app/models/pipeline_run.rb
```diff
@@ -34,8 +34,6 @@ class PipelineRun < ApplicationRecord
   STATUS_READY = 'READY'.freeze
   POSTPROCESS_STATUS_LOADED = 'LOADED'.freeze
 
-  before_create :create_run_stages
```
yunfangjuan: did you do a scan on PipelineRun to make sure there's only one place for create_run_stages?

boris-dimitrov: i did

but i now realize i don't need to change this

boris-dimitrov:
Yeah valid point about the fact that if there is a bug in the pipeline code a simple retry would not recover.   However, only admins can fix such a bug, and they have the power to retry with the fix (aka rerun vs resume).

boris-dimitrov:
I think we may be able to do better in choosing when to retry/resume.  I think resume is useful only when the failed stage is stage 2.  Basically resume=lazy.

PR #1170, opened 2018-04-12
Style tweaks

jshoe:
- Various style tweaks to match up with design references


jshoe:
@cdebourcy hmm interesting. Reverted that button for now / can be in a different PR.

cdebourcy:
Ok. Shouldn't be hard though -- just a matter of making sure clicking "New Project" doesn't trigger form onSubmit action.

PR #1172, opened 2018-04-13
Alignment viz refactor

yunfangjuan:
# Description

Converting the alignment visualization to React component.  This is ready to review~ 
screen shot 2018-04-13 at 5 44 21 pm

## Type of change

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:


yunfangjuan:
the url will be located at 
http://localhost:3000/samples/490/alignment_viz/nt_genus_1301 
basically change the existing url from alignment -> alignment_viz

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+          break;
+        default:
+          output += c
+      }
+    }
+    return output
+  }
+
+  render() {
+    const nopadding = {padding: '0px'}
```
jshoe: ideally would start another .scss file?

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -129,23 +129,40 @@ def generate_quality_string(ref_string, seq_string)
...
   end
 
-  def parse_tree(results, key, current_dict)
+  def parse_tree(results, key, current_dict, raw = false)
     if current_dict["reads"]
-      results[key] = parse_accession(current_dict)
+      if raw # no further parsing
+        # sort the coverage
+        if current_dict["coverage_summary"] && current_dict["coverage_summary"]["coverage"]
+          coverage = current_dict["coverage_summary"]["coverage"].sort_by { |k, _v| k.split("-")[0].to_i }
```
jshoe: did you mean |k, _| ?

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+    let white_space_right = this.repeatStr(' ', right_portion.length)
+    let ref_seq_display = readInfo.refInfo.join("|")
+    let read_seq_display = [left_portion, aligned_portion, right_portion].join("|")
+    let quality_string_display = [white_space_left, quality_string, white_space_right].join("|")
+
+    // generate quality string
+    return {
+      reversed: reversed,
+      read_part: readPart,
+      ref_display: ref_seq_display,
```
tihuan: snake_case to camelCase :D!

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+         </table>
+       </div>
+      </div>
+    );
+  }
+}
+
+class AccessionViz extends React.Component {
+  constructor(props) {
+    super(props);
```
tihuan: super()

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+  render_coverage_summary(cs) {
+    let read_length = cs.total_read_length / cs.num_reads;
+    let avg_aligned_length = cs.total_aligned_length/cs.num_reads
+    let avg_aligned_pct = avg_aligned_length/read_length * 100
+    let mismatched_pct = cs.total_mismatched_length / cs.total_aligned_length * 100
+    let coverage = cs.total_aligned_length / cs.ref_seq_len * 100
+    let distinct_coverage = cs.distinct_covered_length / cs.ref_seq_len * 100
+    return (
+      <div>
+        Read Length: <b> { read_length } </b>, Average Aligned Length: <b> {avg_aligned_length.toFixed(2)} ({avg_aligned_pct.toFixed(2)} %) </b>, Mismatched Percenage: <b> {mismatched_pct.toFixed(2)} % </b> <br/>
```
tihuan: typo: s/Percenage/Percentage

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+       </div>
+      </div>
+    );
+  }
+}
+
+class AccessionViz extends React.Component {
+  constructor(props) {
+    super(props);
+    this.allReads = props.reads
```
tihuan: is there a particular reason why we're assigning `this.allReads = props.reads` instead of just using `props.reads` directly in constructor() and in this class :D?

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+    this.readsPerPage = props.readsPerPage
```
tihuan: same for this. Seems like `this.readsPerPage` assignment is not necessary?

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+    this.coverageSummary = props.coverage_summary || {}
```
tihuan: to assign default value to optional props, React has a shortcut. Like so:

```tsx
AccessionViz.defaultProps = {
  coverageSummary: {},
}
```

So we can remove `this.coverageSummary = props.coverage_summary || {}` and use `this.props.coverage_summary` directly and it will default to `{}` for you!

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+    this.state = {
+      reads: this.allReads.slice(0, this.readsPerPage)
+    }
+  }
+
+  render_coverage_table(coverage_table) {
```
tihuan: nit: camelCase :D

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+    return  (
+      <div>
```
tihuan: also suggest following ReadViz `render()` example for `scss` and `html` here :D

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+import React from "react";
+
+class ReadViz extends React.Component {
+  constructor(props) {
+    super(props);
+    this.parsedAlignment = this.parseAlignment(props);
+  }
+
+  parseAlignment(readInfo) {
+    let readPart = parseInt(readInfo.name.split("/")[1])
```
tihuan: suggest destructuring `readInfo` like so:

```tsx
const { name, sequence, metrics, refInfo } = readInfo
```

and use the consts in the rest of the function.

(I checked that this syntax is available for `babel-preset-env, which IDseq has. So definitely use away!)

https://babeljs.io/docs/plugins/preset-env/

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
...
+      <div>
+        Read Length: <b> { read_length } </b>, Average Aligned Length: <b> {avg_aligned_length.toFixed(2)} ({avg_aligned_pct.toFixed(2)} %) </b>, Mismatched Percenage: <b> {mismatched_pct.toFixed(2)} % </b> <br/>
+       Total Aligned bps: <b> {cs.total_aligned_length } </b>, Coverage: <b> {coverage} % </b> <br/>
+       Distinct Aligned  bps: <b> {cs.distinct_covered_length} </b>, Distinct Coverage: <b>{distinct_coverage} %</b>
+      </div>
+    );
+  }
+
+  render() {
+    let coverage_table = this.coverageSummary.coverage ? this.render_coverage_table(this.coverageSummary.coverage) : null;
```
tihuan: s/let/const

jshoe:
ty @tihuan!

app/assets/src/components/AlignmentViz.jsx
yunfangjuan: great. I was doing inline until we have a design.

yunfangjuan:
Thanks for all the style suggestions. If I run prettier, would it make a lot of the changes for me?

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
+import axios from "axios";
+import React from "react";
+
+class ReadViz extends React.Component {
+  constructor(props) {
+    super(props);
+    this.parsedAlignment = this.parseAlignment(props);
+  }
+
+  parseAlignment(readInfo) {
```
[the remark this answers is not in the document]
tihuan: I'd recommend the following two steps:

1. If `parseAlignment` only depends on its input argument `readInfo` and not any other instance props (`this.props`), we can extract it to a separate file and import the function into this class to use. 

This way it's easier to write test. And also, since the function doesn't depend on class instances and their specific properties, keeping the function outside of the class will keep the class cleaner.

So we can create a file `parse-alignment.js` and move this function into it. Like so:

```js
function parseAlignment(readInfo) {
  ...
};

export default parseAlignment;
```

And then in the current class file, we can just import the function, like so:

```jsx
import parseAlignment from './parse-alignment'
```

2. Now that we have the function as a separate file, we can create a test file in the same directory like so: `parse-alignment.test.js`, and then code:

```js
import jest from 'jest'
import parseAlignment from './parse-alignment'

describe('parseAlignment', () => {
  it('renders the expected snapshot', () => {
    const readInfo = ... // Not sure what the input shape looks like
    expect(parseAlignment(readInfo)).toMatchSnapshot()
  })
})
```
3. Now you've created a test for the function, we need to run test to create a snapshot of the output object. So run `npm test` and Jest will generate a snapshot file for you!

4. Next time you need to modify the function, you can run test and see if the output is still expected :-)

Some syntax might be off, because I'm at a conference now and don't have access to my laptop XD Hope this helps!

app/assets/src/components/AlignmentViz.jsx
tihuan: Yes! Airbnb actually has a really decent JavaScript style guide that influences a lot of companies' own style guides, such as Google.

https://github.com/airbnb/javascript/blob/master/README.md#objects

The object section shows that the keys use lowerCamelCase in general. But then again, each project can have their own preference!

app/assets/src/components/AlignmentViz.jsx
tihuan: Ah that makes sense! Thanks for letting me know! 👍👍👍

tihuan:
@yunfangjuan np! Glad to help :D I believe prettier only deals with indentation, semicolon insertions, etc.. So for changes like lower_snake_case to camelCase would still need to be done manually. 

But please definitely run Prettier first and see what gets updated for you automatically!

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -0,0 +1,276 @@
+import axios from "axios";
+import React from "react";
+
+class ReadViz extends React.Component {
+  constructor(props) {
+    super(props);
```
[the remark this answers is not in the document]
yunfangjuan: what if i need to use this.props inside render() function?

app/assets/src/components/AlignmentViz.jsx
yunfangjuan: I am actually changing some content in the readInfo object so can't quite do const.

app/assets/src/components/AlignmentViz.jsx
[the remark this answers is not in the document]
tihuan: Yes! We just need to modify the code above slightly by instantiating an instance of the `ReadViz` class. Like so:

```tsx
import jest from 'jest'
import ReadViz from './ReadViz'

describe('ReadViz', () => {
  describe('parsedAlignment', () => {
    it('matches the expected snapshot', () => {
      const readInfo = ... // Not sure what the input shape looks like
      const instance = new ReadViz(readInfo)

      expect(instance.parsedAlignment).toMatchSnapshot()
    })
  })
})
```

This should work. If not, please let me know 👌

app/assets/src/components/AlignmentViz.jsx
tihuan: Ah, javascript `const` only signals that the value cannot be reassigned.

So,

```js
const timmy = { hairColor: 'blue', eyeColor: 'brown' }

timmy.hairColor = 'purple' // good. You can change the content within the object.

timmy = { hairColor: 'purple', eyeColor: 'brown' } // bad. Expected output: TypeError: invalid assignment to const `timmy'. Because we can't reassign timmy to a different object!
```

PR #1180, opened 2018-04-16
Allow sample name to be edited

cdebourcy:


app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -102,6 +102,27 @@ class PipelineSampleReads extends React.Component {
...
   }
 
+  render_metadata_textfield_wide(label, field) {
+    return (
+      <div className="col s12">
+        <div className="details-title note">{label}</div>
+        <div className="sample-notes note">
+          <pre
+            className="details-value"
+            suppressContentEditableWarning={true}
```
tihuan: Great PR 🎉 ! 

The only thing I can see is for prop `suppressContentEditableWarning={true}` we could just omit the `={true}` part and do `suppressContentEditableWarning`, per https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md.

E.g.,
```tsx
// No good
const Hello = <Hello personal={true} />;

// Good
const Hello = <Hello personal />;
```

app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -102,6 +102,27 @@ class PipelineSampleReads extends React.Component {
     );
   }
 
+  render_metadata_textfield_wide(label, field) {
```
tihuan: Nit: This is a great candidate for extracting as a standalone component `MetadataTextFieldWide`, like so:

```tsx
const MetadataTextFieldWide  = ({ label, field, contentEditable, sampleInfo }) => (
  ...
)
```

And then in this component, we can just use the component like so:
```tsx
<MetadataTextFieldWide 
    label="Unique ID"
    field="sample_host"
    contentEditable={this.can_edit}
    // perhaps create a helper method this.getSampleInfo(field) to return the string instead of inlining the expression for maintainability and readability.
    sampleInfo={this.sampleInfo['sample_host'] && this.sampleInfo['sample_host'].trim() !== "" ? this.sampleInfo['sample_host'] : this.TYPE_PROMPT}
/>

// and in other places 
```

And the reason we turn it into a component from a render function is testability and avoid unnecessary re-rendering in virtual DOM.

For re-rendering, when we use `MetadataTextFieldWide`, React will compare the props passed to the component, and when nothing changes, it will not re-render the component.

For testing, it's harder to test a component's methods than a standalone component.

For example, if we were to test this render method, we'd first need to instantiate an instance of this component class, and then call the render function with `label` and `field`.  And if we happen to update the render method name, we'd need to refactor the test file. So not impossible, but unnecessary extra steps we can eliminate!

Whereas as a standalone component `MetadataTextFieldWide`, we can just test it by importing the component and passing props to it directly. Like so:

```tsx
// test file
import { shallow } from 'enzyme'
import MetadataTextFieldWide from './components/MetadataTextFieldWide` // assuming this is the directory where we put the extracted component

// This is a jest test, which I'll go into tomorrow :D
describe('MetadataTextFieldWide', () => {
  const props = {
    label: 'test-label',
    field: 'test-field',
    ...
  }
  it('renders correctly', () => {
    const wrapper = shallow(<MetadataTextFieldWide {...props} />)

    expect(wrapper.getNode()).toMatchSnapshot()
  })
})
```

app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -322,7 +343,7 @@ class PipelineSampleReads extends React.Component {
                 .slideUp(200);
             } else {
               $(".note-save-failed")
-                .html(`<i class='fa fa-frown-o'></i> ${response.data.message}`)
+                .html(`<i class='fa fa-frown-o'></i> ${response.data.message} ${response.data.errors}`)
```
tihuan: This comment is outside the scope of this PR, but in general we should not use jQuery($) to update the DOM, because it defeats the purpose of using React in the first place (which is to enjoy the virtual DOM diff and only update the minimal amount of DOM).

And this also messes up with React's render cycle, since React is unaware of the DOM changes made by jQuery($), so next time React re-renders this component in virtual DOM, even if this part of the DOM node has not changed, jQuery will still touch the DOM unnecessarily)

Just headsup :)

app/assets/src/components/PipelineSampleReads.jsx
tihuan: Note: WHOOPS I forgot to update `MetadataTextFieldWide` to actually address the re-render part of the discussion lol

Stateless Functional Components aka SFC (React components that are pure functions) like so:

```tsx
const MetadataTextFieldWide  = ({ label, field, contentEditable, sampleInfo }) => (
  ...
)
```

Actually ALWAYS re-renders its virtual DOM representation and serves it to React diff algorithm. 

And unless your SFC tree is huge, it's actually okay to let it re-render its virtual DOM, since re-rendering virtual DOM is cheap. The expensive part is manipulating DOM (which jQuery does all the time).

For the extracting component recommendation above, mostly the benefits lie in testability and readability of the code.

However, if you notice a component's re-rendering is causing janks, then it is time to do performance refactoring by turning the SFC into a `class extends React.Component` or `class extends React.PureComponent`, depending on whether you have specific logic for method `shouldComponentUpdate()`. 

Since by default `PureComponent` does reference equality check on all prop values and state values. But if you only need to check equality on any specific prop values or state values, or you need to do deep equality test on a certain prop, to determine whether the component should re-render, then you can use `React.Component` directly :D 

And please feel free to throw more questions at me whenever!

PR #1182, opened 2018-04-17
fixes to hit-calling

boris-dimitrov:


app/models/pipeline_run_stage.rb
```diff
@@ -336,19 +333,14 @@ def db_load_postprocess
   def host_filtering_outputs
     pr = pipeline_run
     pr.pipeline_version = pr.fetch_pipeline_version
-
-    stats_json_s3_path = "#{pipeline_run.host_filter_output_s3_path}/#{PipelineRun::STATS_JSON_NAME}"
-    unmapped_fasta_s3_path = "#{pipeline_run.host_filter_output_s3_path}/unmapped.bowtie2.lzw.cdhitdup.priceseqfilter.unmapped.star.merged.fasta"
-    [stats_json_s3_path, unmapped_fasta_s3_path]
+    "#{pipeline_run.host_filter_output_s3_path}/#{PipelineRun::STATS_JSON_NAME}"
```
cdebourcy: Isn't changing the return value from array to string going to break https://github.com/chanzuckerberg/idseq-web/blob/0e242ec053424930986514ad5a9a1d25abfbbaf5/app/models/pipeline_run_stage.rb#L70 ?

boris-dimitrov: Good eye, Charles.   My change to output_ready? above addresses that.   The future direction here is that we will output one final file if and only if everything else has succeeded.   The only gotcha is if pipeline codebase changes so that file is updated before everything has succeeded, and currently looks like we are okay in that regard.

PR #1185, opened 2018-04-17
Main page tweaks

jshoe:
- More tweaks to the All Samples main page
- Biggest thing is just adding an SVG for the "Compare" button icon
- Checkboxes will be in a separate PR :p


app/assets/src/components/Samples.jsx
```diff
@@ -988,24 +989,36 @@ class Samples extends React.Component {
       </div>
     );
 
+    let compare_button_inner = (
```
tihuan: it looks like variable `compare_button_inner` is not being reassigned, so `const` would convey the meaning better :D? Unless I'm missing something, then nvm lol

app/assets/src/components/Samples.jsx
```diff
@@ -988,24 +989,36 @@ class Samples extends React.Component {
...
+      <span>
+        <span
+          className="img-container compare-container"
+          dangerouslySetInnerHTML={{
+            __html: IconComponent.compare("#c5c5c5")
```
tihuan: nit: For Meta, we have a color file that describes all colors being used in the app: https://github.com/chanzuckerberg/khepri/blob/master/src/styles/themes.js. 

For one, it eliminates the need to look up what color is for `"#c5c5c5"` (which is grey in this case), and another benefit is providing a central place for managing all colors being used through out the app. So it's easier to work with UX team.

Not sure if there's a need for that in this project. Just sharing thoughts :-)

app/assets/src/components/Samples.jsx
```diff
@@ -988,24 +989,36 @@ class Samples extends React.Component {
...
+          }}
+        />
+        <span>{"Compare"}</span>
```
tihuan: `<span>Compare</span>` should work. Unless `prettier` did that lol?

PR #1208, opened 2018-04-21
Assembly

cdebourcy:
This supposed to be fully backwards compatible. Samples run with the current master branch of idseq-pipeline will have a version below 1.11 and so none of the assembly operations in idseq-web will be visible to the user.


app/models/pipeline_run_stage.rb
```diff
@@ -142,6 +142,7 @@ def update_job_status
     end
     # The job appears to be in progress.  Check to make sure it hasn't been killed in AWS.   But not too frequently.
     return unless due_for_aegea_check?
+    db_load_assembly # update the list of taxids that have already been assembled by polling S3
```
boris-dimitrov: keep this quick

boris-dimitrov: pls ensure this doesn't pr.save

PR #1218, opened 2018-04-26
fix blank report with hit calling and support non-genus-specific hits

boris-dimitrov:
the blank reports were caused by lineage map version drift between
pipeline and webapp, e.g., species 1173061 got reclassified from
genus 43987 to genus 27316;  and also by a pipeline bug fixed in 1.6

non-genus-specific hits are important e.g. for some species of
escheria coli where the identical genome is uploaded to NCBI
under two distinct species under two distinct genera


boris-dimitrov:
Confirmed in testing that this at least changes the behavior from crashing and blank report to logs of singleton ad-hoc genera where the real one were not imported due to the pipeline bug and the lineage mismatch.   Now testing with the fixes for those bugs.

PR #1219, opened 2018-04-26
Closes #1210: fix "delete project" button

### LINKED ISSUE #1210: "Delete project" button should not show up when sample list is only empty due to filters
screen shot 2018-04-23 at 3 32 15 pm

cdebourcy:


app/assets/src/components/Samples.jsx
```diff
@@ -1891,7 +1893,10 @@ function ProjectInfoHeading({
         {state.selectedProjectId ? project_menu : null}
         {table_download_dropdown}
         {compare_button}
-        {state.selectedProjectId && state.allSamples.length == 0
+        {state.selectedProjectId &&
+        canEditProject(state.selectedProjectId) &&
```
jshoe: would this work even if you didn't pass it in? since canEditProject was already bound to 'this' at the top?

cdebourcy: It didn't...

PR #1229, opened 2018-04-27
Download host read counts for all samples in project

cdebourcy:


app/models/project.rb
```diff
@@ -39,6 +55,29 @@ def samples
     nil
   end
 
+  def cleaned_project_name
+    "project-#{name.downcase.split(' ').join('_')}"
+  end
+
+  def host_gene_counts_from_params(params)
```
jshoe: maybe more error handling in this function?

cdebourcy: ok. I'm making it admin only for now. We can add more error handling once we start showing it to users

PR #1243, opened 2018-04-30
Rename sample host to sample id to reflect usage

jshoe:
- Sample_host was repurposed to be like "unique ID"


app/models/sample.rb
```diff
@@ -107,7 +107,7 @@ def self.search(search)
         OR samples.sample_tissue LIKE :search
         OR samples.sample_location LIKE :search
         OR samples.sample_notes LIKE :search
-        OR samples.sample_host', search: "%#{search}%")
+        OR samples.sample_id', search: "%#{search}%")
```
yunfangjuan: what is sample_id? this can be confusing with sample.id

jshoe:
@yunfangjuan @cdebourcy  renamed to "sample_unique_id". E.g. patient ID for the human case

cdebourcy:
@jshoe Or update the CLI as well? :)

PR #1245, opened 2018-05-01
bump stage 2 resources to 60GB RAM and 8 vCPU

boris-dimitrov:
hit calling needs it


boris-dimitrov:
This passed rubocop and tests locally so merging to master to get a deployable build quicker.   Will revert if has problems.

boris-dimitrov:
Confirmed working.

PR #1249, opened 2018-05-01
(WIP) Taxon filter

jshoe:


cdebourcy:
It's conceptually the right start, but needs more work. Right now it breaks the species / genus level selector on the heatmap. You can do some debugging on the latest sql dump from alpha (samples that were run with the new hit calling)

app/lib/report_helper.rb
```diff
@@ -427,10 +427,26 @@ def top_taxons_details(samples, background_id, num_results, sort_by_key, only_sp
       taxon_counts = res["taxon_counts"]
       sample_id = pr.sample_id
       tax_2d = convert_2d(taxon_counts)
+      if only_species
+        puts "ONLY SPECIES"
+        puts "TAX LEVEL SPECIES: " + TaxonCount::TAX_LEVEL_SPECIES.to_s
+        tax_2d.keep_if { |_tax_id, tax_info| tax_info['tax_level'] = TaxonCount::TAX_LEVEL_SPECIES }
```
boris-dimitrov: @jshoe don't you mean '==' instead of '=' here?

boris-dimitrov: also this filter needs to be applied after cleaning and name validation, not before

boris-dimitrov: and you need to get rid of the filter in the sql query above

jshoe: holy the "==" was my main bug lol

jshoe:
@boris-dimitrov do we also want to remove_family_level_counts in top_taxons_details? / we can just use cleanup_all ?

boris-dimitrov:
For  remove_family_level_counts in top_taxons_details I think we want to have that separate as long as the two codes paths here are generally separate due to the separate capabilities of the two visualizations w.r.t. family level stuff.   In tomorrow's in-person review with Trevor we can decide how best to rearchitect and unify this stuff.

app/lib/report_helper.rb
boris-dimitrov: :)

app/lib/report_helper.rb
```diff
@@ -427,10 +427,26 @@ def top_taxons_details(samples, background_id, num_results, sort_by_key, only_sp
...
+      else
+        puts "NOT ONLY SPECIES"
+        puts "TAX LEVEL GENUS: " + TaxonCount::TAX_LEVEL_GENUS.to_s
+        puts "BEFORE"
+        puts tax_2d
+        tax_2d.keep_if { |_tax_id, tax_info| tax_info['tax_level'] = TaxonCount::TAX_LEVEL_GENUS }
+        puts "AFTER"
+        puts tax_2d
+      end
+      tax_2d = validate_names!(tax_2d)
```
boris-dimitrov: so this line needs to be above the above filter, not below

boris-dimitrov: and you are right why not use the same cleanup_all function for all the cleanup, instead of unfolding it --- may be there was a good reason for that, but probably the reason was not good :P)

app/lib/report_helper.rb
```diff
@@ -427,10 +421,21 @@ def top_taxons_details(samples, background_id, num_results, sort_by_key, only_sp
       taxon_counts = res["taxon_counts"]
       sample_id = pr.sample_id
       tax_2d = convert_2d(taxon_counts)
+      cleanup_genus_ids!(tax_2d)
+      validate_names!(tax_2d)
+      cleanup_missing_genus_counts!(tax_2d)
+
+      if only_species
+        tax_2d.keep_if { |_tax_id, tax_info| tax_info['tax_level'] == TaxonCount::TAX_LEVEL_SPECIES }
```
yunfangjuan: Not sure what the performance impact will be if we get both genus and species data.

app/lib/report_helper.rb
```diff
@@ -427,10 +421,21 @@ def top_taxons_details(samples, background_id, num_results, sort_by_key, only_sp
...
+      else
+        tax_2d.keep_if { |_tax_id, tax_info| tax_info['tax_level'] == TaxonCount::TAX_LEVEL_GENUS }
+      end
+
       rows = []
       tax_2d.each do |_tax_id, tax_info|
         rows << tax_info
       end
+
       compute_aggregate_scores_v2!(rows)
```
yunfangjuan: If you want to be completely be sync with the report page, you might want to go back to the original compute aggregate scores and move it above the genus/species selection.

yunfangjuan:
see comments.

boris-dimitrov:
I agree with Yun Fang, we should have unified data fetching for all visualizations.   Fetching higher taxon count ranks is probably at most 2x cost (obviously i am making a guess about the degree of branching but it seems justified looking at the reports).

app/lib/report_helper.rb
```diff
@@ -427,10 +421,21 @@ def top_taxons_details(samples, background_id, num_results, sort_by_key, only_sp
       taxon_counts = res["taxon_counts"]
       sample_id = pr.sample_id
       tax_2d = convert_2d(taxon_counts)
+      cleanup_genus_ids!(tax_2d)
+      validate_names!(tax_2d)
+      cleanup_missing_genus_counts!(tax_2d)
```
boris-dimitrov: Is this any different from cleanup_all?    This is good for now, but should be deduped further.

Issue #1236, opened 2018-04-30
expand/collapse all should expand all non-genus-specific buckets


boris-dimitrov:
currently it won't expand negative taxid buckets because it used to be the only such bucket "all taxa without genus classification" contained so much junk that expanding it automatically wasn't a good idea

the situation is different after hit calling

Issue #1234, opened 2018-04-30
"all taxa without genus classification" should now say "all taxa with neither family nor genus classification"


boris-dimitrov:

jshoe:
in progress

Issue #1224, opened 2018-04-27
non-genus-specific hits and non-species-specific hits need links to NCBI


boris-dimitrov:
those links are currently disabled for negative taxids such as these

in order to enable the links obtain the link taxid by taking the negative taxid modulo int(1e8)

cc @jshoe @cdebourcy

jshoe:
in progress

PR #1254, opened 2018-05-02
Hit calling UI tweaks

jshoe:
For:
#1224 
#1234 
#1236


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -847,7 +852,7 @@ class PipelineSampleReport extends React.Component {
     let ncbiDot, fastaDot, alignmentVizDot;
     if (taxInfo.tax_level == 1) tax_level_str = "species";
     else tax_level_str = "genus";
-    if (taxInfo.tax_id > 0)
+    if (taxInfo.tax_id != -200 && taxInfo.tax_id != -100)
```
boris-dimitrov: ```&& taxInfo.taxid != -300, -400, ...```

for now may be simpler to just say if `taxid > 0 || taxid < -1e8`

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -820,7 +820,12 @@ class PipelineSampleReport extends React.Component {
   // path to NCBI
   gotoNCBI(e) {
     const taxId = e.target.getAttribute("data-tax-id");
-    const ncbiLink = `https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=${taxId}`;
+    let num = parseInt(taxId);
+    if (num < 0) {
+      num = -num % 1e8;
```
boris-dimitrov: How well does JavaScript implement % 1e8?   I think with fp64 it's important that this be `(-num) % 1e8` rather than `-(num % 1e8)`.

boris-dimitrov: Also this should be `if (num < -1e8)` unless already implied.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1161,9 +1166,9 @@ class PipelineSampleReport extends React.Component {
     // expand all real genera
     this.expandAll = true;
     this.expandedGenera = [];
-    $(".report-row-species.real-genus").removeClass("hidden");
-    $(".report-arrow-down.real-genus").removeClass("hidden");
-    $(".report-arrow-right.real-genus").addClass("hidden");
+    $(".report-row-species").removeClass("hidden");
+    $(".report-arrow-down").removeClass("hidden");
+    $(".report-arrow-right").addClass("hidden");
```
boris-dimitrov: why this change?

jshoe: For #1236 since the ".real-genus" prevented it from opening the 'fake genus' lines

jshoe:
@boris-dimitrov fixed comments

PR #1261, opened 2018-05-07
Fix superkingdom assignment

cdebourcy:
Run this after deploy to update all the existing runs:
```
  pipeline_runs = PipelineRun.top_completed_runs
  pipeline_runs.each do |pr|
    pr.update_superkingdoms
  end
```


cdebourcy:
@jameshowardwang Test plan:
- load idseq-alpha database into local dev, go to http://localhost:3000/samples/530
- filter to show only Uncategorized. Aquabacterium is listed as uncategorized
- in the rails console, do: `Sample.find(530).pipeline_runs.first.update_superkingdoms`. Search for Aquabacterium. It is now listed as "bacterial"
- run the same on all completed pipeline_runs in a loop as described above and make sure nothing breaks

PR #1264, opened 2018-05-07
upload total_reads.json if present

boris-dimitrov:
needed for running post-filtered inputs

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -18,6 +18,7 @@ class Sample < ApplicationRecord
   SORTED_TAXID_ANNOTATED_FASTA_GENUS_NR = 'taxid_annot_sorted_genus_nr.fasta'.freeze
   SORTED_TAXID_ANNOTATED_FASTA_FAMILY_NT = 'taxid_annot_sorted_family_nt.fasta'.freeze
   SORTED_TAXID_ANNOTATED_FASTA_FAMILY_NR = 'taxid_annot_sorted_family_nr.fasta'.freeze
+  TOTAL_READS_JSON = "total_reads.json".freeze
 
   LOG_BASENAME = 'log.txt'.freeze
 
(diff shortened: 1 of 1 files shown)


app/models/sample.rb
```diff
@@ -179,11 +179,24 @@ def initiate_input_file_upload
...
     stderr_array = []
+    total_reads_json = "total_reads.json"
+    total_reads_json_path = nil
     input_files.each do |input_file|
       fastq = input_file.source
+      total_reads_json_path = File.join(File.dirname(fastq.to_s), total_reads_json)
       _stdout, stderr, status = Open3.capture3("aws", "s3", "cp", fastq.to_s, "#{sample_input_s3_path}/#{input_file.name}")
       stderr_array << stderr unless status.exitstatus.zero?
     end
+    if total_reads_json.present?
```
cdebourcy: did you mean `if total_reads_json_path.present?` ?

app/models/sample.rb
```diff
@@ -179,11 +179,24 @@ def initiate_input_file_upload
...
+      # For samples where we are only given fastas post host filtering, we need to input the total reads (before host filtering) from this file.
+      _stdout, _stderr, status = Open3.capture3("aws", "s3", "cp", total_reads_json_path, "#{sample_input_s3_path}/#{total_reads_json}")
+      # We don't have a good way to know if this file should be present or not, so we just try to upload it
+      # and if it fails, we try one more time;  if that fails too, it's okay, the file is optional.
```
cdebourcy: You can check whether file is present as follows?
```
_stdout, _stderr, status = Open3.capture3("aws", "s3", "ls", total_reads_json_path)
  ...
```

app/models/sample.rb
boris-dimitrov: Yes!

app/models/sample.rb
boris-dimitrov: Why would I want to do that?   If the file is present, I've just wasted an operation.  If the file is absent, I get an error the same way I would if I try to copy it.   Also, either approach can be a false negative for absence due to operational issues.   For this reason I am actually removing check_file_s3_presence from the python pipeline in https://github.com/chanzuckerberg/idseq-pipeline/pull/174

app/models/sample.rb
boris-dimitrov: Thanks for catching this :)

app/models/sample.rb
boris-dimitrov: Btw while working on the pipeline check_s3_file_presence this weekend I looked up the s3 api for that, something called head-object, with which you can distinguish more easily operational errors from file absence.  But it's too much work to be worth it at this time, I think.

app/models/sample.rb
```diff
@@ -179,11 +179,24 @@ def initiate_input_file_upload
   def initiate_s3_cp
     return unless status == STATUS_CREATED
     stderr_array = []
+    total_reads_json = "total_reads.json"
```
jameshowardwang: seems like convention is to put this into a frozen string

boris-dimitrov: thanks!  i often forget that

boris-dimitrov: so I moved it to the top as a frozen constant, and then I am not sure if I should do this
```
      total_reads_json_path = File.join(File.dirname(fastq.to_s), TOTAL_READS_JSON)
```
or add `.to_s` like so:
```
      total_reads_json_path = File.join(File.dirname(fastq.to_s), TOTAL_READS_JSON.to_s)
```
Definitely testing :)

boris-dimitrov: Btw another problem I had here with testing is that I can't figure out how to import File into the ruby shell `irb`.   I ended up just spinning up rails console as somehow it has that magically imported.

boris-dimitrov: The `.to_s` appears unnecessary.

boris-dimitrov:
Tested & implemented James' suggestion to freeze the filename constant.

PR #1266, opened 2018-05-08
add truncation stats

cdebourcy:
Goes with https://github.com/chanzuckerberg/idseq-pipeline/pull/176

### DIFF
--- a/app/models/pipeline_run_stage.rb
+++ b/app/models/pipeline_run_stage.rb
@@ -248,6 +248,11 @@ class PipelineRunStage < ApplicationRecord
     version_s3_path = "#{pr.host_filter_output_s3_path}/#{PipelineRun::VERSION_JSON_NAME}"
     pr.version = `aws s3 cp #{version_s3_path} -`
 
+    # Check if input was truncated
+    truncation_file = "#{pr.host_filter_output_s3_path}/#{PipelineRun::INPUT_TRUNCATED_FILE}"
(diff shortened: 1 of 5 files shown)


app/models/pipeline_run_stage.rb
```diff
@@ -248,6 +248,11 @@ def db_load_host_filtering
     version_s3_path = "#{pr.host_filter_output_s3_path}/#{PipelineRun::VERSION_JSON_NAME}"
     pr.version = `aws s3 cp #{version_s3_path} -`
 
+    # Check if input was truncated
+    truncation_file = "#{pr.host_filter_output_s3_path}/#{PipelineRun::INPUT_TRUNCATED_FILE}"
+    _stdout, _stderr, status = Open3.capture3("aws", "s3", "ls", truncation_file)
+    pr.truncated = `aws s3 cp #{truncation_file} -`.split("\n")[0].to_i if status.exitstatus.zero?
```
jameshowardwang: i'm not entirely sure about this operation - does this pull the entire file into memory to get the first line? is the file generally super small?

cdebourcy: Yeah, the file just contains a single number

jameshowardwang: ah ok all good then. otherwise worth reading https://felipeelias.github.io/ruby/2017/01/02/fast-file-processing-ruby.html

boris-dimitrov: cool reference, lazy grep take 10 :)

PR #1268, opened 2018-05-09
More taxon consolidation and fix lineages: null issue

jshoe:
- Basically same thing with negative taxons being converted so they can be found in the taxon counts.
- Some taxon_counts cleanup consolidation
- Fixes I believe all the visible issues with missing lineages and such

BEFORE:
screen shot 2018-05-08 at 5 00 51 pm

AFTER:
screen shot 2018-05-08 at 4 50 44 pm

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -232,23 +232,35 @@ class SamplesController < ApplicationController
     end
 
     @report_info = external_report_info(pipeline_run_id, background_id, params)
-    tax_details = @report_info[:taxonomy_details][2]
 
-    # Filtered rows
-    lineage_by_taxid = @report_info[:taxonomy_details][3]
+    tax_ids = @report_info[:taxonomy_details][2].map { |x| x['tax_id'] }
+    tax_ids = tax_ids | @report_info[:taxonomy_details][2].map { |x| x['family_taxid'] }
+    lineages = TaxonCount.connection.select_all(TaxonLineage.where(taxid: tax_ids)).to_hash
+    lineage_by_taxid = {}
+    lineages.each do |x|
+      lineage_by_taxid[x['taxid']] = x
+    end
 
-    # Fix visible genus/family 'Uncategorized' bug on tree view by checking
-    # species/genus/family taxids and using the first positive lineage info.
-    tax_details.each do |tax|
-      target_id = tax['tax_id']
-      if target_id < 0
-        if tax['genus_taxid'] > 0
-          target_id = tax['genus_taxid']
-        elsif tax['family_taxid'] > 0
-          target_id = tax['family_taxid']
-        end
+    @report_info[:taxonomy_details][2].each do |tax|
+      if tax['tax_id'] > 0
+        tax['lineage'] = lineage_by_taxid[tax['tax_id']]
+      elsif tax['tax_id'] < 0 && tax['tax_level'] == 1 && tax['genus_taxid'] > 0
+        tax['lineage'] = lineage_by_taxid[tax['genus_taxid']]
+        tax['lineage']['taxid'] = tax['tax_id']
+        tax['lineage']['species_taxid'] = tax['species_taxid']
+        tax['lineage']['species_name'] = tax['name']
+      elsif tax['tax_id'] < 0 && tax['tax_level'] == 1 && tax['genus_taxid'] < 0 && tax['family_taxid'] > 0
+        tax['lineage'] = lineage_by_taxid[tax['family_taxid']]
+        tax['lineage']['taxid'] = tax['tax_id']
+        tax['lineage']['species_taxid'] = tax['species_taxid']
+        tax['lineage']['species_name'] = tax['name']
+        tax['lineage']['genus_taxid'] = tax['genus_taxid']
+      elsif tax['tax_id'] < 0 && tax['tax_level'] == 2 && tax['family_taxid'] > 0
+        tax['lineage'] = lineage_by_taxid[tax['family_taxid']]
+        tax['lineage']['taxid'] = tax['tax_id']
+        tax['lineage']['genus_taxid'] = tax['genus_taxid']
+        tax['lineage']['genus_name'] = tax['name']
       end
-      tax['lineage'] = lineage_by_taxid[target_id]
     end
 
     render json: JSON.dump(@report_info)
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -233,18 +233,8 @@ class SamplesController < ApplicationController
 
     @report_info = external_report_info(pipeline_run_id, background_id, params)
 
-    # Get list of tax_ids to look up in TaxonLineage and TaxonCount rows. Include family_taxids.
-    tax_ids = @report_info[:taxonomy_details][2].map { |x| x['tax_id'] }
-    tax_ids |= @report_info[:taxonomy_details][2].map { |x| x['family_taxid'] }
-
-    # TODO: Should definitely be simplified with taxonomy/lineage refactoring.
-    lineage_by_taxid = {}
-    TaxonLineage.where(taxid: tax_ids).as_json.each { |x|
-      lineage_by_taxid[x['taxid']] = x
-    }
-
     # Fill lineage details into report info
-    fill_lineage_details(lineage_by_taxid)
+    fill_lineage_details
 
     render json: JSON.dump(@report_info)
   end
@@ -520,32 +510,47 @@ class SamplesController < ApplicationController
 
   private
 
-  def fill_lineage_details(lineage_by_taxid)
+  def fill_lineage_details
+    # Get list of tax_ids to look up in TaxonLineage and TaxonCount rows. Include family_taxids.
+    tax_ids = @report_info[:taxonomy_details][2].map { |x| x['tax_id'] }
+    tax_ids |= @report_info[:taxonomy_details][2].map { |x| x['family_taxid'] }
+
+    # TODO: Should definitely be simplified with taxonomy/lineage refactoring.
+    lineage_by_taxid = {}
+    TaxonLineage.where(taxid: tax_ids).each do |x|
+      lineage_by_taxid[x.taxid] = x.as_json
+    end
+
     # Set lineage info from the first positive tax_id of the species, genus, or family levels.
     # Preserve names of the negative 'non-specific' nodes.
     # Used for the tree view and sets appropriate lineage info at each node.
+
     @report_info[:taxonomy_details][2].each do |tax|
-      if tax['tax_id'] > 0
-        tax['lineage'] = lineage_by_taxid[tax['tax_id']]
-      elsif tax['tax_id'] < 0 && tax['tax_level'] == 1 && tax['genus_taxid'] > 0
-        # NOTE: start_tax_lineage fills missing info from TaxonLineage with an empty hash.
-        tax['lineage'] = lineage_by_taxid[tax['genus_taxid']] || {}
-        tax['lineage']['taxid'] = tax['tax_id']
-        tax['lineage']['species_taxid'] = tax['species_taxid']
-        tax['lineage']['species_name'] = tax['name']
-      elsif tax['tax_id'] < 0 && tax['tax_level'] == 1 && tax['genus_taxid'] < 0 && tax['family_taxid'] > 0
-        tax['lineage'] = lineage_by_taxid[tax['family_taxid']] || {}
-        tax['lineage']['taxid'] = tax['tax_id']
-        tax['lineage']['species_taxid'] = tax['species_taxid']
-        tax['lineage']['species_name'] = tax['name']
-        tax['lineage']['genus_taxid'] = tax['genus_taxid']
-      elsif tax['tax_id'] < 0 && tax['tax_level'] == 2 && tax['family_taxid'] > 0
-        tax['lineage'] = lineage_by_taxid[tax['family_taxid']] || {}
-        tax['lineage']['taxid'] = tax['tax_id']
-        tax['lineage']['genus_taxid'] = tax['genus_taxid']
-        tax['lineage']['genus_name'] = tax['name']
+      missing_vals = { species_taxid: TaxonLineage::MISSING_SPECIES_ID, genus_taxid: TaxonLineage::MISSING_GENUS_ID, family_taxid: TaxonLineage::MISSING_FAMILY_ID }
+      lineage_id = most_specific_positive_id(tax)
+
+      if lineage_id
+        tax['lineage'] = lineage_by_taxid[lineage_id] || {}
+      else
+        tax['lineage'] = missing_vals
+        next
       end
+
+      tax['lineage']['taxid'] = tax['tax_id']
+      tax['lineage']['species_taxid'] = tax['species_taxid']
+      tax['lineage']['genus_taxid'] = tax['genus_taxid']
+
+      name = level_name(tax['tax_level']) + "_name"
+      tax['lineage'][name] = tax['name']
     end
+  end
+
+  def most_specific_positive_id(tax)
+    targets = [tax['tax_id'], tax['species_taxid'], tax['genus_taxid'], tax['family_taxid']]
+    targets.each do |tentative_id|
+      return tentative_id if tentative_id && tentative_id > 0
+    end
+    nil
   end
 
   def clean_taxid_name(pipeline_run, taxid)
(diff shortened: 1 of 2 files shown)


jshoe:
Actually maybe will try to use validate_names ?

app/controllers/samples_controller.rb
```diff
@@ -232,15 +232,25 @@ def report_info
...
-    tax_ids = @report_info[:taxonomy_details][2].map { |x| x['tax_id'] }
+    thres = TaxonLineage::INVALID_CALL_BASE_ID
+    tax_ids = @report_info[:taxonomy_details][2].map do |x|
+      t = x['tax_id']
+      if t > thres
+        t
+      else
+        -(t % thres).to_i
+      end
+    end
```
cdebourcy: Since you reuse this logic below, please put it into a function

app/models/pipeline_run.rb
```diff
@@ -298,7 +298,7 @@ def update_superkingdoms
       SET taxon_counts.superkingdom_taxid = taxon_lineages.superkingdom_taxid
       WHERE taxon_counts.pipeline_run_id=#{id} AND
             (taxon_counts.created_at BETWEEN taxon_lineages.started_at AND taxon_lineages.ended_at) AND
-            taxon_lineages.taxid = taxon_counts.tax_id
+            taxon_lineages.taxid = MOD(ABS(taxon_counts.tax_id), #{TaxonLineage::INVALID_CALL_BASE_ID})
```
cdebourcy: @boris-dimitrov wrote an elaborate fix here that should go in first

app/controllers/samples_controller.rb
```diff
@@ -234,14 +234,34 @@ def report_info
...
+    tax_ids = tax_ids | @report_info[:taxonomy_details][2].map { |x| x['family_taxid'] }
     lineages = TaxonCount.connection.select_all(TaxonLineage.where(taxid: tax_ids)).to_hash
     lineage_by_taxid = {}
     lineages.each do |x|
       lineage_by_taxid[x['taxid']] = x
     end
 
     @report_info[:taxonomy_details][2].each do |tax|
-      tax['lineage'] = lineage_by_taxid[tax['tax_id']]
+      if tax['tax_id'] > 0
```
yunfangjuan: can you make 245-264 a function that can be changed around later? 
also can extract another function to extract the level of the match. But maybe there's an easier way for representation in TaxonCount. Also more comments for this function would be appreciated.

app/lib/report_helper.rb
```diff
@@ -865,20 +857,28 @@ def apply_filters!(rows, tax_2d, all_genera, params)
...
-    tax_2d.each do |_tax_id, tax_info|
-      tax_name = tax_info['name']
-      tax_level = tax_info['tax_level']
-      all_genera.add(tax_name) if tax_level == TaxonCount::TAX_LEVEL_GENUS
+    # Generate lineage info.
+    unfiltered_ids = []
+    tax_2d.each do |tid, _|
+      unfiltered_ids << tid
+    end
+    lineages = TaxonCount.connection.select_all(TaxonLineage.where(taxid: unfiltered_ids)).to_hash
```
yunfangjuan: similar to my commebt above. This Taxon.connection.selecg_all is not necessary. to_hash is also not.

app/lib/report_helper.rb
```diff
@@ -913,7 +912,17 @@ def taxonomy_details(pipeline_run_id, background_id, params)
     t5 = wall_clock_ms
     logger.info "Data processing took #{t5 - t1} seconds (#{t5 - t0} with I/O)."
 
-    [rows_passing_filters, rows_total, rows]
+    [rows_passing_filters, rows_total, rows, lineage_by_taxid]
+  end
+
+  def lineage_details(pipeline_run_id, background_id)
+    taxon_counts = fetch_taxon_counts(pipeline_run_id, background_id)
```
yunfangjuan: umm running another big query here? is it really necessary?

app/lib/report_helper.rb
```diff
@@ -927,6 +936,26 @@ def flat_hash(h, f = [], g = {})
...
+    res = []
+    input.each do |k, _|
+      k.each do |_, id|
+        res << id
+      end
+    end
+    res
+  end
+
+  def only_species_or_genus_counts!(tax_2d, species_selected)
```
yunfangjuan: It seems a bit excessive here. Isn't the difference between specieis_level_counts and genus_level_acounts just checking tax_level. 
why not have a function do
select_proper_level_counts(tax_2d, TaxonCounts::LEVEL_GENUS|SPECIES) etc

app/lib/report_helper.rb
```diff
@@ -308,13 +307,7 @@ def fetch_parent_ids(taxon_ids, samples)
         pipeline_run_id in (#{pipeline_run_ids.join(',')}) AND
         taxon_counts.tax_id in (#{taxon_ids.join(',')})
        ").to_hash
-
```
yunfangjuan: I don't think you even need line 310. You can just to to_a

app/controllers/samples_controller.rb
```diff
@@ -233,15 +233,38 @@ def report_info
...
+        tax['lineage']['taxid'] = tax['tax_id']
+        tax['lineage']['species_taxid'] = tax['species_taxid']
+        tax['lineage']['species_name'] = tax['name']
+        tax['lineage']['genus_taxid'] = tax['genus_taxid']
+      elsif tax['tax_id'] < 0 && tax['tax_level'] == 2 && tax['family_taxid'] > 0
+        tax['lineage'] = lineage_by_taxid[tax['family_taxid']] || {}
+        tax['lineage']['taxid'] = tax['tax_id']
+        tax['lineage']['genus_taxid'] = tax['genus_taxid']
+        tax['lineage']['genus_name'] = tax['name']
+      end
```
boris-dimitrov: This looks too complicated.   Shouldn't it just be as follows:
```
    def most_specific_positive_id(tax):
            [tax['species_taxid'], tax['genus_taxid'], tax['family_taxid']].each |tentative_id| do
            if tentative_id > 0:
                return tentative_id
            return 0

lineage_id = most_specific_positive_id(tax)
if lineage_id
    lineage = lineage_by_taxid[lineage_id]
else
    lineage = [-100, -200, ..., -700]
lineage['species_taxid'] = tax['species_taxid']
lineage['genus_taxid'] = tax['genus_taxid']
lineage['family_taxid'] = tax['family_taxid']
```

cdebourcy: Yes, it would be great to simplify in a similar way. However, it is crucial that we also set `tax['lineage']['#{level_name (tax['tax_level'])}_name'] = tax['name']`

boris-dimitrov: Also note that in the very near future, after migration, this should change to simply copy the lineage from the taxon.   Or better yet change JS to use the lineage from the taxon.   It's only a temporary situation that the taxon here doesn't have the complete lineage.
```
lineage['species_taxid'] = tax['species_taxid']
lineage['genus_taxid'] = tax['genus_taxid']
lineage['family_taxid'] = tax['family_taxid']
...
```

app/lib/report_helper.rb
```diff
@@ -613,20 +595,23 @@ def validate_names!(taxon_counts_2d)
...
-        if tax_id < TaxonLineage::INVALID_CALL_BASE_ID && (tax_info['tax_level'] == TaxonCount::TAX_LEVEL_GENUS || tax_info['tax_level'] == TaxonCount::TAX_LEVEL_SPECIES)
-          parent_taxid = -(tax_id % TaxonLineage::INVALID_CALL_BASE_ID)
-          if taxon_counts_2d[parent_taxid]
-            parent_name = taxon_counts_2d[parent_taxid]['name']
-            parent_level = level_name(taxon_counts_2d[parent_taxid]['tax_level'])
+
+        if tax_id < TaxonLineage::INVALID_CALL_BASE_ID && species_or_genus(tax_info['tax_level'])
+          parent_id = convert_neg_taxid(tax_id)
+          if tax_2d[parent_id]
+            tax_info["superkingdom_taxid"] = tax_2d[parent_id]["superkingdom_taxid"]
```
boris-dimitrov: This line should no longer be necessary, now that pr.update_superkingdom was fixed.

app/controllers/samples_controller.rb
cdebourcy: What do you mean by "doesn't seem to fill out all the lineages"? We nee lineage_by_taxid to fill out the part of the lineage that is above family level, and make sure to keep the info from the taxon_counts to fill the lineages below or at family level

app/lib/report_helper.rb
```diff
@@ -865,20 +857,28 @@ def apply_filters!(rows, tax_2d, all_genera, params)
...
-    # Pull out all genera names in sample (before filters are applied).
-    all_genera = Set.new
-    tax_2d.each do |_tax_id, tax_info|
-      tax_name = tax_info['name']
-      tax_level = tax_info['tax_level']
-      all_genera.add(tax_name) if tax_level == TaxonCount::TAX_LEVEL_GENUS
+    # Generate lineage info.
+    unfiltered_ids = []
+    tax_2d.each do |tid, _|
+      unfiltered_ids << tid
```
boris-dimitrov: Here you may restrict to just the non-negative ones, but that's a very minor optimization.

app/lib/report_helper.rb
jshoe: The lineage info needs to come from TaxonCounts because "-200 and -100 are usually from the lineage table not from taxon counts. In taxon counts they will be like -200000543 and -10000543."  cc: Boris

app/lib/report_helper.rb
```diff
@@ -865,20 +857,28 @@ def apply_filters!(rows, tax_2d, all_genera, params)
...
+    end
+    lineages = TaxonCount.connection.select_all(TaxonLineage.where(taxid: unfiltered_ids)).to_hash
+    lineage_by_taxid = {}
+    lineages.each do |x|
+      lineage_by_taxid[x['taxid']] = x
```
boris-dimitrov: I must be blind but I don't see where this `lineage_by_taxid` is used.

boris-dimitrov:
This is relatively clean and well contained as far as these types of changes go.   I just have some minor comments/questions.

app/controllers/samples_controller.rb
boris-dimitrov: Oh my bad I missed completely that we were also setting the names not just the ids here.

boris-dimitrov: Yeah I like the line 
```
tax['lineage']["#{level_name(tax['tax_level'])}_name"] = tax['name']
```
Is that the only name that needs to be passed up?   I am actually curious how the tree viz makes up the higher category names --- does it get them from here, or does it get them from the lineage map?  If the latter, it will need the modulus arithmetic there.   What might be most efficient, of course, is to return to it the subset of the lineage tree that it needs, but that may be in the scope of another pull request.

boris-dimitrov: Right, so my recommendation is to get rid of those if statements and simplify as suggested, including Charles' way to copy the name at the tax level, since that's easier to maintain (in future we just delete the lineage lookup and all comes from that tax info).

cdebourcy: The higher rank names come from lineage_by_taxid, which is directly derived from TaxonLineage. 
`tax` here is derived from the same hash that corresponds to a row in the report: it contains only the name of its own level.

app/controllers/samples_controller.rb
```diff
@@ -233,16 +233,23 @@ def report_info
...
     tax_ids = @report_info[:taxonomy_details][2].map { |x| x['tax_id'] }
-    lineages = TaxonCount.connection.select_all(TaxonLineage.where(taxid: tax_ids)).to_hash
+    tax_ids |= @report_info[:taxonomy_details][2].map { |x| x['family_taxid'] }
+
+    # Get lineage records from the TaxonLineage table and get more info from TaxonCount rows.
+    # This is because taxon counts contains transformed results such as -10000543 to indicate
+    # non-specific hits.
+    # TODO: Should definitely be simplified with taxonomy/lineage refactoring.
+    lineage_records = TaxonLineage.where(taxid: tax_ids)
+    lineages = TaxonCount.connection.select_all(lineage_records).to_hash
```
yunfangjuan: I don't think you need this line. you can just do lineage_records.each do |x|

app/controllers/samples_controller.rb
boris-dimitrov: Thanks.  I also realized we may need to handle the case then the lineage info doesn't exist for the most specific taxid.   In those cases we still have the specices, genus, family from the tax hash (ultimately from pipeline) but nothing higher up.  Also the superkingdom won't be available.  Probably shouldn't worry too much about such legitimately uncategorized things though;  as they are also supposed to disappear if lineage is updated in sql (because should only happen if pipeline lineage is ahead of webapp lineage).

app/controllers/samples_controller.rb
```diff
@@ -237,16 +237,11 @@ def report_info
...
-    # non-specific hits.
     # TODO: Should definitely be simplified with taxonomy/lineage refactoring.
-    lineage_records = TaxonLineage.where(taxid: tax_ids)
-    lineages = TaxonCount.connection.select_all(lineage_records).to_hash
     lineage_by_taxid = {}
-    lineages.each do |x|
+    TaxonLineage.where(taxid: tax_ids).as_json.each { |x|
       lineage_by_taxid[x['taxid']] = x
-    end
+    }
```
boris-dimitrov: beautiful change

yunfangjuan: why do you need as_json, you really dont need that. 
just do 

lineage_by_taxid[x.taxid]  = x.to_hash

yunfangjuan: why not move 236 to 244 to fill_lineage_details?

jshoe: The TaxonLineage ActiveRecord result actually doesn't have .to_hash so as_json was basically the same thing? https://stackoverflow.com/questions/15427936/how-to-convert-activerecord-results-into-an-array-of-hashes

app/controllers/samples_controller.rb
```diff
@@ -518,6 +510,49 @@ def remove_taxon_confirmation
...
+    # Set lineage info from the first positive tax_id of the species, genus, or family levels.
+    # Preserve names of the negative 'non-specific' nodes.
+    # Used for the tree view and sets appropriate lineage info at each node.
+
+    @report_info[:taxonomy_details][2].each do |tax|
+      missing_vals = { species_taxid: TaxonLineage::MISSING_SPECIES_ID, genus_taxid: TaxonLineage::MISSING_GENUS_ID, family_taxid: TaxonLineage::MISSING_FAMILY_ID }
+      lineage_id = most_specific_positive_id(tax)
+
+      if lineage_id
+        tax['lineage'] = lineage_by_taxid[lineage_id] || {}
```
boris-dimitrov: shouldn't this be `|| missing_vals` instead of `|| {}`?

app/controllers/samples_controller.rb
```diff
@@ -518,6 +510,49 @@ def remove_taxon_confirmation
...
+      else
+        tax['lineage'] = missing_vals
+        next
```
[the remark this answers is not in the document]
boris-dimitrov: And the name.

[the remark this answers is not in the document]
boris-dimitrov: And the name.

app/controllers/samples_controller.rb
```diff
@@ -518,6 +510,49 @@ def remove_taxon_confirmation
...
+    TaxonLineage.where(taxid: tax_ids).each do |x|
+      lineage_by_taxid[x.taxid] = x.as_json
+    end
+
+    # Set lineage info from the first positive tax_id of the species, genus, or family levels.
+    # Preserve names of the negative 'non-specific' nodes.
+    # Used for the tree view and sets appropriate lineage info at each node.
+
+    @report_info[:taxonomy_details][2].each do |tax|
+      missing_vals = { species_taxid: TaxonLineage::MISSING_SPECIES_ID, genus_taxid: TaxonLineage::MISSING_GENUS_ID, family_taxid: TaxonLineage::MISSING_FAMILY_ID }
```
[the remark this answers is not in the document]
jshoe: I don't think it actually needs them for some reason..

app/controllers/samples_controller.rb
```diff
@@ -518,6 +510,49 @@ def remove_taxon_confirmation
...
+
+      name = level_name(tax['tax_level']) + "_name"
+      tax['lineage'][name] = tax['name']
+    end
+  end
+
+  def most_specific_positive_id(tax)
+    targets = [tax['tax_id'], tax['species_taxid'], tax['genus_taxid'], tax['family_taxid']]
+    targets.each do |tentative_id|
+      return tentative_id if tentative_id && tentative_id > 0
```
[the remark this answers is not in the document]
jshoe: yes I get some Nil vals for some reason

PR #1273, opened 2018-05-10
 Version background

yunfangjuan:
# Description

- create a new background model when the list of pipeline_run ids change. 
## Type of change

- [ X] Breaking change (fix a feature that knowingly causes existing functionality to not work as expected)

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:


app/models/background.rb
```diff
@@ -44,11 +44,23 @@ def summarize
     results
   end
 
+  def pipeline_run_ids
+    updated_pipeline_run_ids = []
+    pipeline_runs.each do |pr|
+      pr_new = pr.sample.pipeline_runs.find_by(job_status: PipelineRun::STATUS_CHECKED)
```
jameshowardwang: can you add a comment here about what's going on? i think it's important to explain:
- find_by is returning the first result
- the first result is always the most recent *completed* run because of the way the table is indexed, which we're assuming to be with the latest pipeline ID (though this may not be the case)
- we might change this function down the road if we allow backgrounds to be assembled using older pipeline runs of a sample

app/views/backgrounds/index.html.erb
```diff
@@ -18,6 +20,8 @@
         <td><%= background.name %></td>
         <td><%= background.pipeline_runs.count %></td>
         <td><%= background.created_at %></td>
+        <td><%= background.pipeline_version %></td>
+        <td><%= background.time_version %></td>
```
jameshowardwang: isn't this column duplicating "created_at"

app/controllers/backgrounds_controller.rb
```diff
@@ -42,20 +42,15 @@ def create
...
   def update
-    current_data = @background.to_json(include: [{ pipeline_runs: { only: [:id, :sample_id] } }])
-    current_data_full_string = @background.to_json(include: [{ pipeline_runs: { only: [:id, :sample_id] } },
-                                                             :taxon_summaries])
-    background_changed = assign_attributes_and_has_changed?(background_params)
-    if background_changed
-      s3_archive_successful, s3_backup_path = archive_background_to_s3(current_data_full_string)
-      db_archive_successful = archive_background_to_db(current_data, s3_backup_path)
-      update_successful = db_archive_successful && s3_archive_successful ? @background.save : false # this triggers recomputation of @background's taxon_summaries if archiving was successful
+    if background_params[:pipeline_run_ids].map(&:to_i).select { |p| p > 0 } .sort == @background.pipeline_runs.pluck(:id).sort
```
jameshowardwang: i'm confused about the equality here -- is this to allow a background to be updated only if the pipeline_run ids exactly match? and then the only thing that is updated is the name? so this allows rename functionality, is that right?

jameshowardwang: if that's the case, a comment would really help

yunfangjuan: yes. but actual logic is more complicated. I am still working on it. will add comments.

app/views/backgrounds/index.html.erb
```diff
@@ -18,6 +20,8 @@
         <td><%= background.name %></td>
         <td><%= background.pipeline_runs.count %></td>
         <td><%= background.created_at %></td>
+        <td><%= background.pipeline_version %></td>
+        <td><%= background.created_at %></td>
```
jshoe: `<td><%= background.created_at %></td>` is repeated

db/schema.rb
```diff
@@ -35,7 +37,7 @@
     t.index ["pipeline_output_id"], name: "index_backgrounds_pipeline_outputs_on_pipeline_output_id"
   end
 
-  create_table "backgrounds_pipeline_runs", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
+  create_table "backgrounds_pipeline_runs", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=latin1" do |t|
```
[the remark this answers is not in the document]
jshoe: I can't find the commits but I remember seeing it change back and forth to `DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci` :3

app/models/background.rb
```diff
@@ -44,11 +77,22 @@ def summarize
     results
   end
 
+  def pipeline_run_ids
+    # get the list of successful and most recent pipeline_runs that belong to the same samples
+    updated_pipeline_run_ids = []
+    pipeline_runs.each do |pr|
+      pr_new = pr.sample.pipeline_runs.find_by(job_status: PipelineRun::STATUS_CHECKED)
```
cdebourcy: Is there a difference between this and just selecting from PipelineRun.top_completed_runs based on sample_id?

yunfangjuan:
table for now.

PR #1279, opened 2018-05-10
Developer utility for making backgrounds given a list of sample names

cdebourcy:


lib/tasks/make_background.rake
```diff
@@ -0,0 +1,40 @@
+task make_background: :environment do
```
jameshowardwang: what does  ":environment" do here?

cdebourcy: This gives the task access to the models, like `Background` and `Sample`

PR #1283, opened 2018-05-12
Add back sql results unfolding

jshoe:
- Adds back hash unfolding code to parent_ids, which should fix some SQL errors


cdebourcy:
https://github.com/chanzuckerberg/idseq-web/pull/1284

PR #1285, opened 2018-05-12
Closes #1282: disallow uploads

### LINKED ISSUE #1282: [security/privacy] any user can gain access to another user input fastqs by guessing the path
Virtually every input is s3://idseq-samples-production/samples/<project_id>/<sample_id>fastqs/<sample_name>.fastq.gz where the project id and sample id and sample name are obvious from urls when browsing the tool.   So I can submit a sample with inputs that are the same as some other user's inputs, and specify no host filtering, then recover all the dna.

cc @eggerdesigns @jameshowardwang

cdebourcy:
Prevents non-admins from choosing files in the private idseq S3 bucket as inputs for their own samples

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -401,6 +401,7 @@ class SamplesController < ApplicationController
     end
 
     params[:input_files_attributes] = params[:input_files_attributes].reject { |f| f["source"] == '' }
+    params[:input_files_attributes] = params[:input_files_attributes].reject { |f| f["source"].include?(SAMPLES_BUCKET_NAME) } unless current_user && current_user.role == 1
     @sample = Sample.new(params)
     @sample.project = project if project
     @sample.input_files.each { |f| f.name ||= File.basename(f.source) }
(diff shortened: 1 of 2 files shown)


app/controllers/samples_controller.rb
```diff
@@ -401,6 +401,7 @@ def create
     end
 
     params[:input_files_attributes] = params[:input_files_attributes].reject { |f| f["source"] == '' }
+    params[:input_files_attributes] = params[:input_files_attributes].reject { |f| f["source"].include?(SAMPLES_BUCKET_NAME) } unless current_user && current_user.role == 1
```
jshoe: what does `current_user.role == 1` mean?

jshoe: oh admin?

cdebourcy: Correct

yunfangjuan: can we just do current_user.admin?  
this makes only admin can upload?

yunfangjuan: Actually, I think we should do 

 { |f| f["source"] == '' && current_user.can_upload( f["source"]) } 

and implement the can_upload in the user model to make it more modular and generic. 
Also, current_user should always be set in this case.

cdebourcy: ok!

PR #1287, opened 2018-05-14
Tweak if logic for filling lineage for Tree View

jshoe:
- The simplified if logic was a little more incorrect than the more verbose first cut and was giving incorrect results in the tree view on some edge samples. Can test on CHRF_RNA_0094_S94. Highlighted the wrong node / hid species that were actually supposed to show.
- Added back `tax['tax_id']` in most_specific_positive_id targets because it should use that if it's positive? / was changing the displayed results.


app/controllers/samples_controller.rb
```diff
@@ -525,37 +525,40 @@ def fill_lineage_details
...
       tax['lineage'][name] = tax['name']
+
+      if tax['tax_id'] < 0
+        if tax['tax_level'] == 1
+          tax['lineage']['species_taxid'] = tax['species_taxid']
+        end
+        if tax['family_taxid'] > 0
+          tax['lineage']['genus_taxid'] = tax['genus_taxid']
+        end
+      end
```
boris-dimitrov: Let's discuss this afternoon.   I'm afraid this may be a symptom of another bug somewhere.   If so, we should fix that bug instead of working around it here.

app/controllers/samples_controller.rb
```diff
@@ -525,37 +525,40 @@ def fill_lineage_details
     # Preserve names of the negative 'non-specific' nodes.
     # Used for the tree view and sets appropriate lineage info at each node.
 
+    missing_vals = {
```
yunfangjuan: I think fill_lineage_details function should be in the TaxonLineage model. we should make samples_controller as simple as possible.

boris-dimitrov: good idea

app/lib/report_helper.rb
```diff
@@ -198,6 +199,7 @@ def fetch_taxon_counts(pipeline_run_id, background_id)
         taxon_counts.tax_id              AS  tax_id,
         taxon_counts.count_type          AS  count_type,
         taxon_counts.tax_level           AS  tax_level,
+
```
boris-dimitrov: this one doesn't need species_taxid?

boris-dimitrov: oh yes it's not yet in the table, right

app/models/taxon_lineage.rb
```diff
@@ -37,4 +38,57 @@ def self.get_genus_info(genus_tax_id)
...
+    TaxonLineage.where(taxid: tax_ids).find_each do |x|
+      lineage_by_taxid[x.taxid] = x.as_json
+    end
+
+    # Set lineage info from the first positive tax_id of the species, genus, or family levels.
+    # Preserve names of the negative 'non-specific' nodes.
+    # Used for the tree view and sets appropriate lineage info at each node.
+
+    # Make a new hash with 'species_taxid', 'genus_taxid', etc.
+    missing_vals = Hash[MISSING_LINEAGE_ID.map { |k, v| [k.to_s + "_taxid", v] }]
```
boris-dimitrov: cool

app/models/taxon_lineage.rb
```diff
@@ -37,4 +38,57 @@ def self.get_genus_info(genus_tax_id)
...
+
+    tax_map.each do |tax|
+      # Grab the appropriate lineage info by the first positive tax level
+      tax['lineage'] = missing_vals
```
boris-dimitrov: this needs to create a copy/clone of missing_vals, otherwise below when you copy species_taxid etc into it, you will be overwriting elements of the missing_vals hash

app/models/taxon_lineage.rb
```diff
@@ -37,4 +38,57 @@ def self.get_genus_info(genus_tax_id)
...
+      lineage_id = most_specific_positive_id(tax)
+      lin = lineage_by_taxid[lineage_id] if lineage_id
+      tax['lineage'] = lin if lin
```
boris-dimitrov: i am unclear on the scope of `lin`, is it a new variable for every iteration of the loop, or is the same variable and the value from one iteration can be seen by the next?

here lin woud be uninitialized if the very first iteration o the loop did not have a positive lineage_id

boris-dimitrov: May I recommend
```
tax['lineage'] = lineage_by_taxid[lineage_id] if lineage_id else missing_values.dup
```

PR #1295, opened 2018-05-15
(Small) Make report download button download with the right pipeline version

jshoe:
- Small change to download the right "pipeline version report" with the report download button.
- Doesn't include the cookie changes.

### DIFF
--- a/app/assets/src/components/PipelineSampleReads.jsx
+++ b/app/assets/src/components/PipelineSampleReads.jsx
@@ -318,7 +318,12 @@ class PipelineSampleReads extends React.Component {
     const csv_background_id_param = param_background_id
       ? `?background_id=${param_background_id}`
       : defaultBackground;
-    return `/samples/${this.sampleId}/report_csv${csv_background_id_param}`;
+    const pipeline_version_param = this.pipelineRun.pipeline_version
+      ? `&pipeline_version=${this.pipelineRun.pipeline_version}`
+      : "";
+    return `/samples/${
+      this.sampleId
+    }/report_csv${csv_background_id_param}${pipeline_version_param}`;
   }
 
   componentDidMount() {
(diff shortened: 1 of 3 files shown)


app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -318,7 +318,12 @@ class PipelineSampleReads extends React.Component {
     const csv_background_id_param = param_background_id
       ? `?background_id=${param_background_id}`
       : defaultBackground;
```
jameshowardwang: this code seems kind of dangerous in its use of '?' and '&' -- is there a way to write this that clearly puts only one ? and the rest of the parameters joined by &? i'm imagining the use of an array and join.('&')

cdebourcy:
Where does it consume the `pipeline_version` param? It seems to me like that part is missing

jshoe:
it's in

```
  def select_pipeline_run(sample)
    if params[:pipeline_version].blank?
      sample.pipeline_runs.first
    else
      sample.pipeline_run_by_version(params[:pipeline_version])
    end
  end
```

app/controllers/samples_controller.rb
```diff
@@ -592,7 +592,7 @@ def sample_params
                                    :sample_memory, :sample_location, :sample_date, :sample_tissue,
                                    :sample_template, :sample_library, :sample_sequencer,
                                    :sample_notes, :job_queue, :search, :subsample, :pipeline_branch,
-                                   :sample_input_pg, :sample_batch, :sample_diagnosis, :sample_organism, :sample_detection,
+                                   :sample_input_pg, :sample_batch, :sample_diagnosis, :sample_organism, :sample_detection, :pipeline_version,
```
cdebourcy: Adding pipeline_version here is not necessary

cdebourcy:
I see. I got thrown off by the that `select_pipeline_run` in the helper doesn't take `params` as an argument -- didn't realize it's a controller function

PR #1309, opened 2018-05-21
Add RuboCop to precommit hook

jshoe:
- Adds 'RuboCop autocorrect' to our 'husky' pre-commit hook for staged files since I think we've all wasted a few minutes by forgetting to run it and then failing Travis for a trivial reason.
- Can still be skipped with "git commit --no-verify", which is also suggested on a fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,11 @@
     "setupTestFrameworkScriptFile": "./enzyme_config.js"
   },
   "lint-staged": {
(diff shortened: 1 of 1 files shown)


yunfangjuan:
nice

cdebourcy:
Worth remembering that the changes made by "rubocop auto_correct" are not always trivial and have introduced a bug in my code on one occason (due to it removing parentheses and getting precedence wrong)

PR #1311, opened 2018-05-21
Initial pass: remove the most redundant queries

cdebourcy:
Initial pass at refactoring project page.

(a) If a project is specified, retrieve only that project's samples.
(b) Remove queries that are repeated multiple times just to give the exact same information.

Next step will be to do only one SQL query for each type of information across all samples

### DIFF
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -94,20 +94,19 @@ module SamplesHelper
     HostGenome.all.map { |h| h.slice('name', 'id') }
   end
 
-  def get_summary_stats(jobstats)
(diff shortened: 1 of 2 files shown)


app/helpers/samples_helper.rb
```diff
@@ -262,13 +260,12 @@ def sample_uploader(sample)
     user
   end
 
-  def sample_derived_data(sample)
+  def sample_derived_data(sample, pipeline_run)
```
yunfangjuan: actually this won't make a difference. 
if you call sample.pipeline_runs.first on the sample sample multiple times, the second time and on it the results will be cached.

PR #1313, opened 2018-05-21
Add pushstate functionality and also set the cookies properly

jshoe:
- Redone to use the axios async stuff so it's cleaner.
- Went with adding the background info to the "report_info" call from samples_controller because it seemed cleaner rather than worry about background name and background id not matching in the JS.

### DIFF
--- a/app/assets/src/components/PipelineSampleReads.jsx
+++ b/app/assets/src/components/PipelineSampleReads.jsx
@@ -361,61 +361,6 @@ class PipelineSampleReads extends React.Component {
         this.handleDropdownChange
       );
     }
-
-    this.fillUrlParams();
-  }
-
-  // Fill in desired URL parameters so user's can copy and paste URLs.
-  // Ex: Add ?pipeline_version=1.7&background_id=4 to /samples/545
-  // This way links can still be to '/samples/545' in the rest of the app
-  // but the URL will be filled in without triggering another page reload
-  // usually.
-  fillUrlParams() {
-    // Skip if report is not present or a background ID and pipeline version
-    // are explicitly specified in the URL.
-    let reportPageParams = this.props.reportPageParams;
-    if (
-      !this.reportPresent ||
-      !reportPageParams ||
-      (this.fetchParams("pipeline_version") &&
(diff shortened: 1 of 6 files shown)


app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -349,6 +347,61 @@ class PipelineSampleReads extends React.Component {
...
+      // Select the background with the matching name.
+      let match = this.allBackgrounds.filter(
+        b => b["name"] === Cookies.get("background_name")
+      );
+      if (match && match[0] && match[0]["id"]) {
+        let cookie_id = match[0]["id"];
+        let loaded_id = reportPageParams.background_id;
+        if (loaded_id !== cookie_id) {
+          // Refresh the page using this background id.
+          this.refreshPage({ background_id: cookie_id });
```
yunfangjuan: this shouldn't happen here actually. this function should be called in componentWillMount before fetchreport data and fetchbackground list

jshoe:
Going to rewrite some parts of this since YF explained how to avoid the page refresh to me..

jshoe:
Updated this one if you want to take a look! Cleaner now

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1205,6 +1211,57 @@ class PipelineSampleReport extends React.Component {
...
+  fetchParams(param) {
+    let url = new URL(window.location);
+    return url.searchParams.get(param);
+  }
+
+  // Select the background with the matching name.
+  getBackgroundIdByName(name) {
+    let match = this.all_backgrounds.filter(b => b["name"] === name);
+    if (match && match[0] && match[0]["id"]) {
+      this.props.reportPageParams.background_id = match[0]["id"];
```
yunfangjuan: you can probably just return match[0]["id"] 
and do 
const background_id = getBackgroundIdByName(name) ; 
if (background_id) this.props.reportPageParams.background_id = background_id 

so getBackgroundIdByName(name) might be reusable in the future.

PR #1315, opened 2018-05-22
Charles/project page

cdebourcy:
Speed up the project page:
- only 1 SQL query for job_stats
- only 1 SQL query for pipeline_run_stages
- do not poll S3 to retrieve pipeline_version for each sample...

Tested locally, looking at query traces with `tail -f log/development.log`. Speed-up very noticeable.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -155,7 +155,8 @@ class SamplesController < ApplicationController
 
     @pipeline_run_display = curate_pipeline_run_display(@pipeline_run)
     @sample_status = @pipeline_run ? @pipeline_run.job_status : nil
-    job_stats_hash = make_job_stats_hash([@pipeline_run.id])[@pipeline_run.id]
+    pipeline_run_id = @pipeline_run ? @pipeline_run.id : nil
+    job_stats_hash = make_job_stats_hash([pipeline_run_id])[pipeline_run_id]
     @summary_stats = job_stats_hash.present? ? get_summary_stats(job_stats_hash, @pipeline_run) : nil
     @project_info = @sample.project ? @sample.project : nil
     @project_sample_ids_names = @sample.project ? Hash[current_power.project_samples(@sample.project).map { |s| [s.id, s.name] }] : nil
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -68,7 +68,7 @@ class SamplesController < ApplicationController
     results = filter_by_tissue_type(results, tissue_type_query) if tissue_type_query.present?
     results = filter_by_host(results, host_query) if host_query.present?
 
-    @samples = sort_by(results, sort).paginate(page: params[:page], per_page: params[:per_page] || PAGE_SIZE).includes([:user, :host_genome, :input_files, :pipeline_runs])
+    @samples = sort_by(results, sort).paginate(page: params[:page], per_page: params[:per_page] || PAGE_SIZE).includes([:user, :host_genome, :pipeline_runs])
     @samples_count = results.size
     @all_samples = format_samples(@samples)
 
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -68,7 +68,7 @@ class SamplesController < ApplicationController
     results = filter_by_tissue_type(results, tissue_type_query) if tissue_type_query.present?
     results = filter_by_host(results, host_query) if host_query.present?
 
-    @samples = sort_by(results, sort).paginate(page: params[:page], per_page: params[:per_page] || PAGE_SIZE).includes([:user, :host_genome, :pipeline_runs])
+    @samples = sort_by(results, sort).paginate(page: params[:page], per_page: params[:per_page] || PAGE_SIZE).includes([:user, :host_genome, :pipeline_runs, :input_files])
     @samples_count = results.size
     @all_samples = format_samples(@samples)
 
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -156,7 +156,7 @@ class SamplesController < ApplicationController
     @pipeline_run_display = curate_pipeline_run_display(@pipeline_run)
     @sample_status = @pipeline_run ? @pipeline_run.job_status : nil
     pipeline_run_id = @pipeline_run ? @pipeline_run.id : nil
-    job_stats_hash = make_job_stats_hash([pipeline_run_id])[pipeline_run_id]
+    job_stats_hash = job_stats_db_get(pipeline_run_id)
     @summary_stats = job_stats_hash.present? ? get_summary_stats(job_stats_hash, @pipeline_run) : nil
     @project_info = @sample.project ? @sample.project : nil
     @project_sample_ids_names = @sample.project ? Hash[current_power.project_samples(@sample.project).map { |s| [s.id, s.name] }] : nil
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -68,7 +68,7 @@ class SamplesController < ApplicationController
     results = filter_by_tissue_type(results, tissue_type_query) if tissue_type_query.present?
     results = filter_by_host(results, host_query) if host_query.present?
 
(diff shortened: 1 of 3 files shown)


app/helpers/samples_helper.rb
```diff
@@ -262,26 +260,59 @@ def sample_uploader(sample)
...
     output_data[:pipeline_run] = pipeline_run
     output_data[:host_genome_name] = sample.host_genome ? sample.host_genome.name : nil
-    output_data[:job_stats] = job_stats
     output_data[:summary_stats] = summary_stats
 
     output_data
   end
 
+  def make_job_stats_hash(pipeline_run_ids)
+    all_job_stats = JobStat.where(pipeline_run_id: pipeline_run_ids).as_json
```
yunfangjuan: don't need as_json. it will create a big blob instead of just having an iterator.

app/helpers/samples_helper.rb
```diff
@@ -262,26 +260,59 @@ def sample_uploader(sample)
...
+    job_stats_by_pipeline_run_id = {}
+    all_job_stats.each do |entry|
+      job_stats_by_pipeline_run_id[entry['pipeline_run_id']] ||= {}
+      job_stats_by_pipeline_run_id[entry['pipeline_run_id']][entry['task']] = entry
```
yunfangjuan: entr.pipeline_run_id entry.task

app/helpers/samples_helper.rb
```diff
@@ -262,26 +260,59 @@ def sample_uploader(sample)
...
+        pipeline_runs.id
+      FROM pipeline_runs LEFT OUTER JOIN pipeline_run_stages
+      ON pipeline_runs.id = pipeline_run_stages.pipeline_run_id
+      WHERE
+        pipeline_runs.id IN (#{pipeline_run_ids.join(',')}) AND
+        (
+          pipeline_runs.job_status = '#{PipelineRun::STATUS_CHECKED}' OR
+          (pipeline_run_stages.job_status = '#{PipelineRun::STATUS_LOADED}' AND pipeline_run_stages.step_number = pipeline_runs.ready_step)
+        )
+    ").to_hash.map(&:values).flatten
```
yunfangjuan: try not to use to_hash, it should be to_a (the result is an array). 

Also, I will redo the query as follows: 
PipelineRun.where(id: pipeline_run_ids).where(" status = CHECKED" or id in (select pipeline_run_id from pipeline_run_stages where  pipeline_run_stages.step_number = pipeline_runs.ready_step and pipeline_run_stages.job_status = 'loaded'").pluck(:id)

app/helpers/samples_helper.rb

app/helpers/samples_helper.rb

app/helpers/samples_helper.rb
```diff
@@ -262,26 +260,50 @@ def sample_uploader(sample)
...
+    summary_stats = job_stats_hash.present? ? get_summary_stats(job_stats_hash, pipeline_run) : nil
     output_data[:pipeline_run] = pipeline_run
     output_data[:host_genome_name] = sample.host_genome ? sample.host_genome.name : nil
-    output_data[:job_stats] = job_stats
     output_data[:summary_stats] = summary_stats
 
     output_data
   end
 
+  def make_job_stats_hash(pipeline_run_ids)
```
jameshowardwang: i think this pattern is going to become more common as we run into other perf bottlenecks. at FB we used to call these functions "..._db_get" and "..._db_multiget" to make it explicit that these would trigger a db hit, and you could use the multiget version to fetch a lot at once. we might want to consider something similar here (this function would then be called job_stats_db_multiget, and there could be a job_stats_db_get function that just calls job_stats_db_multiget and returns the contained, unhashed result

app/helpers/samples_helper.rb
```diff
@@ -262,26 +260,50 @@ def sample_uploader(sample)
...
+
   def format_samples(samples)
     formatted_samples = []
+    return formatted_samples if samples.empty?
+
+    # Do major SQL queries
+    sample_ids = samples.map(&:id)
+    pipeline_run_ids = PipelineRun.where("id in (select max(id) from pipeline_runs where
+                  sample_id in (#{sample_ids.join(',')}) group by sample_id)").pluck(:id)
+    job_stats_by_pipeline_run_id = make_job_stats_hash(pipeline_run_ids)
```
jameshowardwang: see other comment, this would be job_stats_db_multiget

app/controllers/samples_controller.rb
```diff
@@ -155,8 +155,9 @@ def show
 
     @pipeline_run_display = curate_pipeline_run_display(@pipeline_run)
     @sample_status = @pipeline_run ? @pipeline_run.job_status : nil
-    @job_stats = @pipeline_run ? @pipeline_run.job_stats : nil
-    @summary_stats = @job_stats ? get_summary_stats(@job_stats) : nil
+    pipeline_run_id = @pipeline_run ? @pipeline_run.id : nil
+    job_stats_hash = make_job_stats_hash([pipeline_run_id])[pipeline_run_id]
```
jameshowardwang: see other comment, this would be a call to job_stats_db_get

app/helpers/samples_helper.rb
```diff
@@ -262,26 +260,50 @@ def sample_uploader(sample)
...
+    all_job_stats = JobStat.where(pipeline_run_id: pipeline_run_ids)
+    job_stats_by_pipeline_run_id = {}
+    all_job_stats.each do |entry|
+      job_stats_by_pipeline_run_id[entry.pipeline_run_id] ||= {}
+      job_stats_by_pipeline_run_id[entry.pipeline_run_id][entry.task] = entry
+    end
+    job_stats_by_pipeline_run_id
+  end
+
+  def check_report_ready(pipeline_run_ids)
```
jameshowardwang: along those lines, should this be report_ready_db_multiget ?

app/controllers/samples_controller.rb
cdebourcy: ok

app/helpers/samples_helper.rb
cdebourcy: ok!

app/helpers/samples_helper.rb
cdebourcy: ok!

app/helpers/samples_helper.rb

app/controllers/samples_controller.rb
```diff
@@ -68,7 +68,7 @@ def index
     results = filter_by_tissue_type(results, tissue_type_query) if tissue_type_query.present?
     results = filter_by_host(results, host_query) if host_query.present?
 
-    @samples = sort_by(results, sort).paginate(page: params[:page], per_page: params[:per_page] || PAGE_SIZE).includes([:user, :input_files], pipeline_runs: [:job_stats, :pipeline_run_stages])
+    @samples = sort_by(results, sort).paginate(page: params[:page], per_page: params[:per_page] || PAGE_SIZE).includes([:user, :host_genome, :input_files, :pipeline_runs])
```
yunfangjuan: do we need input_files for index page?

app/helpers/samples_helper.rb
```diff
@@ -262,26 +260,54 @@ def sample_uploader(sample)
...
+    PipelineRun.where(id: pipeline_run_ids).where("job_status = '#{PipelineRun::STATUS_CHECKED}' or id in (select pipeline_run_id from pipeline_run_stages where pipeline_run_stages.step_number = pipeline_runs.ready_step and pipeline_run_stages.job_status = '#{PipelineRun::STATUS_LOADED}')").pluck(:id)
+  end
+
   def format_samples(samples)
     formatted_samples = []
+    return formatted_samples if samples.empty?
+
+    # Do major SQL queries
+    sample_ids = samples.map(&:id)
+    pipeline_run_ids = PipelineRun.where("id in (select max(id) from pipeline_runs where
```
yunfangjuan: I think you can get full pipeline_runs here and you can contruct a sample_id to pipeline_run hash say pipeline_run_by_sample_id 
and then you can go pipeline_run_by_sample_id.values.map(&:id) to get list of pipeline_run_ids and reuse the pipeline_runs object below.

app/controllers/samples_controller.rb
cdebourcy: turns out we do

PR #1316, opened 2018-05-22
remove assembly stage

cdebourcy:
Although I did my best to cover all cases at the time this was pushed, maintaining backwards compatibility is more trouble than it's worth!

Latest issue -- Sample Details page:
screen shot 2018-05-21 at 7 14 17 pm


yunfangjuan:
looks fine to me but the CI breaks.

PR #1318, opened 2018-05-23
Separate result monitor from job monitor

cdebourcy:
Separates result monitoring from the current pipeline_monitor / update_job_status functionality.
Goes with
- https://github.com/chanzuckerberg/idseq-pipeline/pull/216 which makes sure the status files get uploaded
-  https://github.com/chanzuckerberg/idseq-infra/pull/99 which makes sure both result_monitor and pipeline_monitor get deployed

Tested on small water sample which got successfully uploaded and processed from scratch without need for manual intervention. Report, fasta download, alignment viz all available. Status displays correct.

To do: test on an entire project of large samples on staging.


app/models/pipeline_run_stage.rb
```diff
@@ -122,27 +120,36 @@ def due_for_aegea_check?
     rand < 0.1
   end
 
-  def update_job_status
-    if !id || !started? || succeeded? || failed?
-      # This is called only from PipelineRun.update_status and only when the guard above is false.
-      Airbrake.notify("Invalid precondition for PipelineRunStage.update_job_status #{id} #{job_id} #{job_status}.")
-      return
-    end
+  def monitor_results
```
yunfangjuan: I think we want to move monitor_results out of the pipeline_run_stages. I actually think you can just implement it in pipeline_runs with hardcoded expected results for now.

yunfangjuan:
Looks good so far. see my comment.

app/models/pipeline_run_stage.rb
cdebourcy: ok!

app/models/pipeline_run.rb
```diff
@@ -160,31 +166,168 @@ def succeeded?
...
+    end
+    if output_ready?("db_load_alignment") && ![LOADING_QUEUED, ALIGNMENT_LOADED].include?(result_status)
+      update(result_status: LOADING_QUEUED)
+      Resque.enqueue(ResultMonitorLoad, id, "db_load_alignment", ALIGNMENT_LOADED)
+    end
+    if output_ready?("db_load_postprocess") && ![LOADING_QUEUED, POSTPROCESS_LOADED].include?(result_status)
+      update(result_status: LOADING_QUEUED)
+      Resque.enqueue(ResultMonitorLoad, id, "db_load_postprocess", POSTPROCESS_LOADED)
+    end
+    if job_status == POSTPROCESS_LOADED # last output has been loaded
```
yunfangjuan: I think 305-315 shouldn't be here actually. job_status and result_status should be separate. 
the finalize flag should only be set by job_monitor. 

I think you can do 
if result_status == POSTPROCESS_LOADED 
self.update(result_status: STATUS_CHECKED) 
end

app/models/pipeline_run.rb
```diff
@@ -160,31 +166,168 @@ def succeeded?
...
+      "#{postprocess_output_s3_path}/#{TAXID_BYTERANGE_JSON_NAME}"
+    end
+  end
+
+  def output_ready?(db_load_command)
+    file_generated_since_run(s3_output_for(db_load_command))
+  end
+
+  def monitor_results
+    return if completed?
```
yunfangjuan: change to depend on results only 

return if result_status == STATUS_CHECKED

app/models/pipeline_run.rb
```diff
@@ -160,31 +166,168 @@ def succeeded?
...
+    update(pipeline_version: fetch_pipeline_version) if pipeline_version.blank?
+    if output_ready?("db_load_host_filtering") && ![LOADING_QUEUED, HOST_FILTER_LOADED].include?(result_status)
+      update(result_status: LOADING_QUEUED)
+      Resque.enqueue(ResultMonitorLoad, id, "db_load_host_filtering", HOST_FILTER_LOADED)
+    end
+    if output_ready?("db_load_alignment") && ![LOADING_QUEUED, ALIGNMENT_LOADED].include?(result_status)
```
yunfangjuan: I think this might be a bit buggy. 

If result_status == ALIGNMENT_LOADED, wouldn't db_load_host_filtering being run again?

app/models/pipeline_run.rb
```diff
@@ -160,31 +166,168 @@ def succeeded?
...
+      sed -e 's/$/,#{id}/' -i taxon_byteranges;
+      mysqlimport --replace --local --user=$DB_USERNAME --host=#{rds_host} --password=$DB_PASSWORD --columns=taxid,hit_type,first_byte,last_byte,pipeline_run_id --fields-terminated-by=',' idseq_#{Rails.env} taxon_byteranges;
+    `
+    _stdout, _stderr, _status = Open3.capture3("rm -f #{downloaded_byteranges_path}")
+  end
+
+  def s3_output_for(db_load_command)
+    case db_load_command
+    when "db_load_host_filtering"
+      "#{host_filter_output_s3_path}/#{STATS_JSON_NAME}"
```
yunfangjuan: I think we should take opportunity to not make the result loading based on stats.json. 
Since we don't care about the stages any more.  I think we should reload stats.json constantly.  

For other files, we just need to load it once and we can make the result_monitor  
1. load stats.json constantly 
2. load the files we want whenever it's available. When all the files are loaded set the result_status to 'CHECKED'

app/models/pipeline_run.rb
```diff
@@ -160,31 +166,164 @@ def succeeded?
...
+    `
+    _stdout, _stderr, _status = Open3.capture3("rm -f #{downloaded_byteranges_path}")
+  end
+
+  def output_ready?(db_load_command)
+    case db_load_command
+    when "db_load_host_filtering"
+      "#{host_filter_output_s3_path}/#{STATS_JSON_NAME}"
+    when "db_load_alignment"
+      "#{alignment_output_s3_path}/#{STATS_JSON_NAME}"
```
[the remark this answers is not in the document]
cdebourcy: yes

app/models/pipeline_run_stage.rb
```diff
@@ -53,8 +53,8 @@ def failed?
     job_status == STATUS_FAILED
   end
 
-  def succeeded? # The whole thing completed successfully
-    db_load_status == 1
+  def succeeded?
+    pipeline_run.output_ready?(load_db_command_func)
```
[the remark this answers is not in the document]
cdebourcy: Because eventually we will have more load functions (on the result monitor side) than run stages (job monitor side). So I wanted to write `output_ready?` in such a way that it relates directly to a `load` command.

Main use of `output_ready?` is actually in pipeline_run / result_monitor. The function is only secondarily used here in pipeline_run_stages (job monitor) because we know we can kill the batch job if the last output of the stage is present

app/models/pipeline_run.rb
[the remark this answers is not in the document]
cdebourcy: Although... I just looked into it more and it is not that easy. Right now the batch job gets killed once the output is present. So if I made the change you suggested, the alignment stage would get killed after taxon_counts json is produced, even though there are some files that still need to be generated after that.

In order not to make this refactor too messy, I suggest sticking with `"#{alignment_output_s3_path}/#{STATS_JSON_NAME}"` (copied from current master) and keeping the target file change for a later diff.

app/models/pipeline_run.rb
```diff
@@ -160,31 +166,168 @@ def succeeded?
...
+  end
+
+  def invalid_family_call?(tcnt)
+    # TODO:  Better family support.
+    tcnt['family_taxid'].to_i < TaxonLineage::INVALID_CALL_BASE_ID
+  rescue
+    false
+  end
+
+  def db_load_alignment
```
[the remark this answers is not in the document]

app/models/pipeline_run.rb
cdebourcy: ok, I changed it so job_status is not here and only belongs to pipeline monitor. 

I've also added a `results_complete?` function and I think this should be used for anything to do with the web app. Web cares only about whether results are present, not about what happened with the job execution. For this reason, I left `notify_users` and `create_or_update_project_background` here in `monitor_results`, now conditional on `results_complete?` rather than `complete?` (=`job_complete?`).

app/models/pipeline_run.rb

app/models/pipeline_run.rb
cdebourcy: ah yes, thanks! Solved it for now by going through the stages in reverse order.

app/models/pipeline_run.rb
cdebourcy: Ok. I will remove the logic that kills the job when `output_ready?` so we can set the target output in the web to anything we want without having to make sure it is the last file generated in idseq-pipeline. Instead, I will kill any zombie jobs when `results_completed?`, which means that ALL results are there.

app/models/pipeline_run.rb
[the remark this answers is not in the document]
cdebourcy: Ok, even this won't work with the run_stages. Because when a batch container gets stuck after it has already produced its last output, we need to know that it is time to submit the next batch job (pipeline_run_stage). I will denormalize so that `pipeline_run_stage.succeeded?` has its own list of files, `last_output_in_stage`, distinct from the outputs we use as arguments to the load functions.

yunfangjuan:
Let's discuss on Tuesday more on the design side of stuff before you make more sweeping changes ;-)

app/assets/src/components/Samples.jsx
```diff
@@ -566,15 +569,11 @@ class Samples extends React.Component {
...
-      let statusClass = !descrip
-        ? this.applyChunkStatusClass(runInfo)
-        : this.applyClass(descrip);
-      let status = !descrip ? this.getChunkedStage(runInfo) : descrip;
+      let status = runInfo.result_status_description;
 
       const stageStatus = (
-        <div className={`${statusClass} status`}>
-          {this.appendStatusIcon(status)}
+        <div className={`${this.statusClass(status)} status`}>
```
jshoe: maybe statusClass() and statusIcon() could also be combined if this is the only placed they're used

app/helpers/samples_helper.rb
```diff
@@ -28,7 +28,7 @@ def generate_sample_list_csv(formatted_samples)
                         location: db_sample ? db_sample[:sample_location] : '',
                         host_genome: derived_output ? derived_output[:host_genome_name] : '',
                         notes: db_sample ? db_sample[:sample_notes] : '',
-                        overall_job_status: run_info ? run_info[:job_status_description] : '',
+                        overall_job_status: run_info ? run_info[:result_status_description] : '',
                         host_filtering_status: run_info ? run_info['Host Filtering'] : '',
```
jshoe: is it planned that these (host_filtering_status, nonhost_alignment_status, postprocessing_status) will also go away from here eventually?

app/jobs/result_monitor_load.rb
```diff
@@ -0,0 +1,19 @@
+# Load a result from S3 into the db
+require 'logger'
+class ResultMonitorLoad
```
jshoe: nit: maybe ResultMonitorLoader to be the actual object name

app/jobs/result_monitor_load.rb
```diff
@@ -0,0 +1,19 @@
...
+  @queue = :q03_pipeline_run
+  @logger = Logger.new(STDOUT)
+
+  def self.perform(pipeline_run_id, load_db_command_func)
+    @logger.info("#{load_db_command_func} for pipeline run #{pipeline_run_id}")
+    pr = PipelineRun.find(pipeline_run_id)
```
jshoe: this will raise RecordNotFound if it can't find the id, so maybe this can be in the 'begin' block for better handling or you can use find_by id and check for nil

app/models/pipeline_run.rb
```diff
@@ -127,6 +130,10 @@ def completed?
     return true if pipeline_run_stages.blank? && (job_status == STATUS_FAILED || job_status == STATUS_CHECKED)
   end
 
+  def results_completed?
```
jshoe: is #completed? going to go away later?

app/models/pipeline_run.rb
```diff
@@ -153,38 +160,249 @@ def retry
...
-  def update_job_status
-    prs = active_stage
-    all_stages_succeeded = prs.nil?
-    if all_stages_succeeded
-      self.finalized = 1
-      self.job_status = STATUS_CHECKED
+  def db_load_host_filtering
+    # Load version info applicable to host filtering
+    version_s3_path = "#{host_filter_output_s3_path}/#{VERSION_JSON_NAME}"
+    self.version = `aws s3 cp #{version_s3_path} -`
```
jshoe: I know this was there before, but this is a little weird isn't it? is it supposed to be self.version == 0 meaning the version file was copied successfully?

app/helpers/samples_helper.rb
cdebourcy: sure

app/jobs/result_monitor_load.rb
cdebourcy: ok

app/jobs/result_monitor_load.rb
cdebourcy: Hmm I don't think this case can ever occur. Unless someone deleted the pipeline_run while the job was waiting in the queue. In which case I think we SHOULD see the error directly in the resque log rather than going through the rescue clause

app/models/pipeline_run.rb
cdebourcy: sure

app/models/pipeline_run.rb
cdebourcy: Hm? It's actually a JSON that contains info about the versions of the host genomes used during filtering etc

cdebourcy: What this command does is copy the file from S3 and insert the contents directly into self.version without the need for a temp file :)

app/helpers/samples_helper.rb
```diff
@@ -222,32 +207,24 @@ def filter_by_host(samples, query)
...
-        run_stages.each do |rs|
-          pipeline_run_entry[rs.name] = rs.job_status
-        end
         pipeline_run_entry[:total_runtime] = get_total_runtime(pipeline_run, run_stages)
         pipeline_run_entry[:with_assembly] = pipeline_run.assembly? ? 1 : 0
+        pipeline_run_entry[:result_status_description] = if pipeline_run.result_status
+                                                           pipeline_run.status_display
+                                                         else
+                                                           # data processed before result monitor was instated
+                                                           pipeline_run.status_display_pre_result_monitor(run_stages)
```
boris-dimitrov: Anything we can do to avoid maintaining the older way?   Migrating existing data perhaps?   That can happen in a separate change if needed.

app/helpers/samples_helper.rb
```diff
@@ -289,7 +266,8 @@ def job_stats_get(pipeline_run_id)
 
   def report_ready_multiget(pipeline_run_ids)
     # query db to get ids of pipeline_runs for which "report_ready?" is true
-    PipelineRun.where(id: pipeline_run_ids).where("job_status = '#{PipelineRun::STATUS_CHECKED}' or id in (select pipeline_run_id from pipeline_run_stages where pipeline_run_stages.step_number = pipeline_runs.ready_step and pipeline_run_stages.job_status = '#{PipelineRun::STATUS_LOADED}')").pluck(:id)
+    # TODO: remove pipeline_run_stages from the equation?
```
boris-dimitrov: I would like that.

app/helpers/samples_helper.rb
```diff
@@ -289,7 +266,8 @@ def job_stats_get(pipeline_run_id)
...
+    PipelineRun.where(id: pipeline_run_ids).where("results_finalized = 1 or id in (select pipeline_run_id from pipeline_run_stages where pipeline_run_stages.step_number = pipeline_runs.ready_step and pipeline_run_stages.job_status = '#{PipelineRunStage::STATUS_SUCCEEDED}')").pluck(:id)
```
boris-dimitrov: Now that you are tracking results separately from stages, it seems the `report ready` property belongs in the pipeline run.   The goal of this change is to allow us to decouple results from stages, and make stages a concept that is private to the pipeline, and one the webapp shouldn't care about.

app/models/pipeline_run_stage.rb
```diff
@@ -53,8 +59,26 @@ def failed?
...
+  def last_output_in_stage
+    basename = "#{job_id}.done"
+    s3_folder = case name
+                when HOST_FILTERING_STAGE_NAME
+                  pipeline_run.host_filter_output_s3_path
+                when ALIGNMENT_STAGE_NAME
+                  pipeline_run.alignment_output_s3_path
+                when POSTPROCESS_STAGE_NAME
+                  pipeline_run.postprocess_output_s3_path
+                end
```
boris-dimitrov: This is nice and readable, but I am wondering is it more or less conventional than defining a hash?

app/helpers/samples_helper.rb
cdebourcy: Yes. I propose after the current PR, we make an additional change where we remove the old code and use information from pipeline_run_stages to populate result_status for all the old samples.

app/models/pipeline_run_stage.rb
```diff
@@ -53,8 +59,26 @@ def failed?
...
+    "#{s3_folder}/#{basename}"
+  end
+
+  def succeeded?
+    # Return whether stage succeeded AND record the result in the database if it is not there already
+    answer = pipeline_run.file_generated_since_run(last_output_in_stage)
```
boris-dimitrov: Why does it have to be "generated since last run"?   We have versioning now, if it was generated in a previous run, that's fine.  In fact what would happen with your change if you drop this is that rerunning from top when lazy results exist for host filtering will cause the host filtering batch command to be killed immediately which is fantastic as then the alignment would start.

boris-dimitrov: The only thing is that in your pipeline status reporting that might show up as a "failed" status for the rerun, but that's no problem, since there would have been a "succeeded" status for the previous run (the one that produced the results).

app/models/pipeline_run_stage.rb
```diff
@@ -123,26 +127,21 @@ def due_for_aegea_check?
...
-      Rails.logger.info "Job #{id} #{job_id} checked and waiting to load results."
+    if failed?
+      terminate_job
       return
     end
-    if output_ready?
-      self.job_status = STATUS_CHECKED
-      # save before enqueue, to prevent minor race
-      save
-      Resque.enqueue(LoadResultForRunStage, id)
```
boris-dimitrov: Where did this move to?

app/models/pipeline_run_stage.rb
```diff
@@ -123,26 +127,21 @@ def due_for_aegea_check?
...
 
   def update_job_status
-    if !id || !started? || succeeded? || failed?
-      # This is called only from PipelineRun.update_status and only when the guard above is false.
+    if !id || !started?
       Airbrake.notify("Invalid precondition for PipelineRunStage.update_job_status #{id} #{job_id} #{job_status}.")
       return
     end
-    if checked?
-      # This can happen due to overly-eager frequent calls from pipeline_monitor through PipelineRun.update_job_status.
```
boris-dimitrov: How are you dealing with this race after your change?

app/models/pipeline_run_stage.rb
```diff
@@ -123,26 +127,21 @@ def due_for_aegea_check?
...
+      terminate_job
       return
     end
-    if output_ready?
-      self.job_status = STATUS_CHECKED
-      # save before enqueue, to prevent minor race
-      save
-      Resque.enqueue(LoadResultForRunStage, id)
+    if succeeded?
+      update(job_status: STATUS_SUCCEEDED)
```
boris-dimitrov: It would be best to have a single place that sets status to SUCCEEDED, just like there should be a single place that sets  it to CHECKED, or whatever value.   Makes it easier to prove correctness of a concurrent program.

lib/tasks/result_monitor.rake
```diff
@@ -0,0 +1,93 @@
...
+    sleep 30
+    until MonitorPipelineResults.shutdown_requested
+      system("rake result_monitor[finite_duration]")
+      sleep wait_before_respawn
+      unless $CHILD_STATUS.exitstatus.zero?
+        sleep additional_wait_after_failure
+      end
+    end
+  end
+end
```
boris-dimitrov: All this stuff looks fine.

app/jobs/result_monitor_loader.rb
```diff
@@ -0,0 +1,19 @@
...
+require 'logger'
+class ResultMonitorLoader
+  @queue = :q03_pipeline_run
+  @logger = Logger.new(STDOUT)
+
+  def self.perform(pipeline_run_id, load_db_command_func)
+    @logger.info("#{load_db_command_func} for pipeline run #{pipeline_run_id}")
+    pr = PipelineRun.find(pipeline_run_id)
+    begin
+      pr.send(load_db_command_func) unless pr.completed?
```
boris-dimitrov: This reads like a bug to me.    Seems you want to skip doing this work if it's been done before.   If `pr.completed?` is the signal for that, then you should skip everything conditional on that signal, not just the `pr.send(load_db_command_func)` but also the `pr.update_result_status`?

boris-dimitrov: Also I can't seem to find the enque for this anywhere.

app/models/pipeline_run_stage.rb
```diff
@@ -123,26 +127,21 @@ def due_for_aegea_check?
   end
 
   def update_job_status
-    if !id || !started? || succeeded? || failed?
-      # This is called only from PipelineRun.update_status and only when the guard above is false.
+    if !id || !started?
```
boris-dimitrov: It appears the changes to this function are not all here yet.   I look forward to reviewing the more complete version.

app/models/pipeline_run_stage.rb
cdebourcy: I see your point and it seems like a good change to make. What if lazy_run is set to False in the pipeline though? As is the case for the entire postprocessing stage? The web app change would supersede that. Maybe that's fine

app/jobs/result_monitor_loader.rb
cdebourcy: It's in the file whose diff was hidden because it is too large ;)

cdebourcy: agree on the conditional skipping

app/models/pipeline_run_stage.rb
cdebourcy: This check originally existed to prevent a loader job from being enqueued twice. Enqueuing of loader jobs has now been moved to monitor_results. In the new version, result_status indicates whether a loader job is in the queue, and we refrain from enqueuing new jobs until that status has been changed.

app/models/pipeline_run_stage.rb
cdebourcy: to monitor_results in the pipeline_run model

app/models/pipeline_run_stage.rb
cdebourcy: agreed - now done as part of `succeeded?`

app/models/pipeline_run_stage.rb
cdebourcy: Actually they're all there :) Just, github was probably hiding part of the diff from you because it was to large (pipeline_run.rb)

lib/tasks/result_monitor.rake
cdebourcy: thx

app/models/pipeline_run_stage.rb
cdebourcy: Not sure. Hash could be nice. I think I'm going to leave it like this for now, for the simple reason that it allows us to have everything in one place

app/models/pipeline_run.rb
```diff
@@ -153,38 +157,255 @@ def retry
   end
 
   def report_ready?
-    job_status == STATUS_CHECKED || (ready_step && active_stage && active_stage.step_number > ready_step)
+    clause_for_old_runs = (job_status == STATUS_CHECKED || (ready_step && pipeline_run_stages.find_by(step_number: ready_step) && pipeline_run_stages.find_by(step_number: ready_step).job_status == STATUS_LOADED))
+    # TODO: migrate old runs so we don't need to deal with them separately in the code
+    results_finalized == 1 || result_status_for(REPORT_READY_LOADER) || clause_for_old_runs
```
boris-dimitrov: why check results_finalized == 1 here?

cdebourcy: If EVERYTHING has finished, we don't need to bother with the other checks

app/models/pipeline_run.rb
```diff
@@ -153,38 +157,255 @@ def retry
...
+      'FAILED'
+    elsif %w[RUNNING LOADED].include?(job_status)
+      'IN PROGRESS'
+    else
+      'WAITING'
+    end
+  end
+
+  def check_and_enqueue(db_load_command_name)
+    # TODO: handle case where resque crashes and needs to be restarted. What happens to runs that were queued to load results?
```
boris-dimitrov: is it true that this can run only from one process/thread?

app/helpers/samples_helper.rb
yunfangjuan: add a TODO so we will actually do it ;-)

app/models/pipeline_run_stage.rb
```diff
@@ -49,12 +59,35 @@ def started?
...
+                  pipeline_run.host_filter_output_s3_path
+                when ALIGNMENT_STAGE_NAME
+                  pipeline_run.alignment_output_s3_path
+                when POSTPROCESS_STAGE_NAME
+                  pipeline_run.postprocess_output_s3_path
+                end
+    "#{s3_folder}/#{basename}"
+  end
+
+  def assess_and_update(status_file_suffix, job_status_value)
```
boris-dimitrov: In other words, you check for file existence, and cache that in the SQL DB?

app/models/pipeline_run_stage.rb
```diff
@@ -49,12 +59,35 @@ def started?
...
+    if answer && job_status != job_status_value
+      update(job_status: job_status_value)
+    end
+    answer
   end
 
-  def succeeded? # The whole thing completed successfully
-    db_load_status == 1
+  def succeeded?
+    assess_and_update(JOB_SUCCEEDED, STATUS_SUCCEEDED)
```
boris-dimitrov: it's not clear here that JOB_SUCCEEDED is a file suffix

app/models/pipeline_run.rb
yunfangjuan: i thought we are getting rid of this?

app/models/pipeline_run_stage.rb
cdebourcy: Yes. It assesses `succeeded?` or `failed?` by looking at a file in S3. No need to query S3 if we have already done it before. That was one of the problems that caused the project page to slow down -- `fetch_pipeline_version` always had to make a request to S3 from scratch because it never stored the result it had just found.

app/models/pipeline_run_stage.rb
cdebourcy: ok, renaming to JOB_SUCCEEDED_FILE_SUFFIX

app/models/pipeline_run.rb
```diff
@@ -153,38 +157,255 @@ def retry
...
+    if output_ready?(db_load_command_name) && ![STATUS_LOADED, STATUS_LOADING].include?(result_status_for(db_load_command_name))
```
yunfangjuan: I think this is still problematic but we can deal with it later (namely, i think we should come up with the list of files that we want to load rather than still using the stages as the key to monitor.)   The most important thing though is to get rid of the dependency of the stats.json file as the time to "load" the results".

app/models/pipeline_run.rb
yunfangjuan: i think it's kinda problematic here to use stats.json for any stage. we should use a different file.

boris-dimitrov:
I am having difficulty loading this in my browser, but the recent "assess_and_update" stuff is concerning.   We should try to make it simpler.

cdebourcy:
Ok let's discuss

app/models/pipeline_run.rb
```diff
@@ -153,38 +157,255 @@ def retry
...
+    host_filtering_job_stats_s3 = "#{host_filter_output_s3_path}/#{STATS_JSON_NAME}"
+    alignment_job_stats_s3 = "#{alignment_output_s3_path}/#{STATS_JSON_NAME}"
+    if file_generated_since_jobstats?(alignment_job_stats_s3)
+      load_job_stats(alignment_job_stats_s3)
+    elsif file_generated_since_jobstats?(host_filtering_job_stats_s3)
+      load_job_stats(host_filtering_job_stats_s3)
+    end
+
+    # Finally, load any new outputs that have become available:
+    check_and_enqueue("db_load_host_filtering")
```
yunfangjuan: instead of doing this, I think you can make it true result_monitor by 
load_version_file() unless pipeline_version # blocking call  
check_and_enqueue("ercc") 
check_and_enqueue("taxon_counts") 
check_and_enqueue("taxon_byte_ranges") .... 
load_stats_file
and set finalized=1 when all the required files are generated

app/models/pipeline_run.rb
cdebourcy: Ok! I will use a reads_per_gene.tab (ercc counts)

app/models/pipeline_run.rb
cdebourcy: Ok! I will make this change today

app/helpers/samples_helper.rb

app/models/pipeline_run.rb
cdebourcy: OK. Although the result monitor here still uses the load commands we originally wrote for the pipeline_run_stages, it no longer makes any reference to the run stages directly. I can change it so it reads `output_ready?("taxon_counts")` instead of `output_ready?("db_load_alignment")`

cdebourcy:
Thanks all for your reviews -- I'm moving this to https://github.com/chanzuckerberg/idseq-web/pull/1331 to have fewer codeclimate comments and load faster

Issue #1324, opened 2018-05-27
Report download button broke again


cdebourcy:
Error:
```
undefined method `select_pipeline_run' for #<Project:0x000000000117e2a0>
/usr/local/bundle/gems/activemodel-5.1.6/lib/active_model/attribute_methods.rb:432:in `method_missing'
/app/app/lib/report_helper.rb:962:in `report_csv_from_params'
/app/app/models/project.rb:88:in `block in bulk_report_csvs_from_params'
/usr/local/bundle/gems/activerecord-5.1.6/lib/active_record/relation/delegation.rb:39:in `each'
/usr/local/bundle/gems/activerecord-5.1.6/lib/active_record/relation/delegation.rb:39:in `each'
/app/app/models/project.rb:87:in `bulk_report_csvs_from_params'
/app/app/jobs/generate_project_reports_csv.rb:5:in `perform'
```

cdebourcy:
This is because select_pipeline_run is defined only in the samples controller, but here the call originates from the project controller. I think select_pipeline_run should be moved to report_helper and params should be passed as an additional argument.

PR #1326, opened 2018-05-29
Fix project report downloads

jshoe:
- As reported here: https://github.com/chanzuckerberg/idseq-web/issues/1324

- Fixes the Project-specific report CSV download
- Tested by clicking all the download buttons (table, report, gene counts) on the Project page, the Report page, and the All Samples page.


yunfangjuan:
I wonder if moving the select_pipeline_run into models/sample.rb will be better.

yunfangjuan:
otherwise, looks reasonable.

PR #1331, opened 2018-05-31
Separate result monitor from job monitor

cdebourcy:
Separates result monitoring from the current pipeline_monitor / update_job_status functionality.
Goes with
-  https://github.com/chanzuckerberg/idseq-infra/pull/99 which makes sure both result_monitor and pipeline_monitor get deployed (adapt this for prod)

Test run:
https://staging.idseq.net/?project_id=82

AFTER DEPLOY:
- do `PipelineRun.where(job_status: "4.De-Novo Assembly-FAILED|READY").update_all(job_status: "CHECKED")`


app/helpers/samples_helper.rb
```diff
@@ -222,32 +210,26 @@ def filter_by_host(samples, query)
...
-        run_stages.each do |rs|
-          pipeline_run_entry[rs.name] = rs.job_status
-        end
         pipeline_run_entry[:total_runtime] = get_total_runtime(pipeline_run, run_stages)
         pipeline_run_entry[:with_assembly] = pipeline_run.assembly? ? 1 : 0
+        pipeline_run_entry[:result_status_description] = if pipeline_run.result_status
+                                                           pipeline_run.status_display
+                                                         else
+                                                           # data processed before result monitor was instated
+                                                           # TODO: migrate old runs and delete this code
```
boris-dimitrov: Great.   A nice convention in these cases is to file an issue and include the issue number in the comment.

app/helpers/samples_helper.rb
```diff
@@ -288,8 +270,14 @@ def job_stats_get(pipeline_run_id)
   end
 
   def report_ready_multiget(pipeline_run_ids)
-    # query db to get ids of pipeline_runs for which "report_ready?" is true
-    PipelineRun.where(id: pipeline_run_ids).where("job_status = '#{PipelineRun::STATUS_CHECKED}' or id in (select pipeline_run_id from pipeline_run_stages where pipeline_run_stages.step_number = pipeline_runs.ready_step and pipeline_run_stages.job_status = '#{PipelineRun::STATUS_LOADED}')").pluck(:id)
+    # query db to get ids of pipeline_runs for which the report is ready
+    report_ready_result_status = "\"#{PipelineRun::REPORT_READY_OUTPUT}\":\"#{PipelineRun::STATUS_LOADED}\""
```
boris-dimitrov: Is there a comment documenting the states and state transitions of a pipeline run?    I am finding these strings difficult to read.   Not obvious why we have REPORT_READY_OUTPUT as well as STATUS_LOADED here.   Having one place that lists the states and state transitions would greatly aid in understanding.

app/jobs/result_monitor_loader.rb
```diff
@@ -0,0 +1,22 @@
+# Load a result from S3 into the db
+require 'logger'
+class ResultMonitorLoader
+  @queue = :q03_pipeline_run
+  @logger = Logger.new(STDOUT)
+
+  def self.perform(pipeline_run_id, output)
+    @logger.info("Loading #{output} for pipeline run #{pipeline_run_id}")
+    pr = PipelineRun.find(pipeline_run_id)
+    return unless pr.result_status_for(output) == PipelineRun::STATUS_LOADING_QUEUED
```
boris-dimitrov: Given that this shouldn't happen, I'd probably print a warning when it does.   Not worth making a change if this is the only change, just a thought for future.

app/jobs/result_monitor_loader.rb
```diff
@@ -0,0 +1,22 @@
...
+    begin
+      pr.update_result_status(output, PipelineRun::STATUS_LOADING)
+      pr.send(PipelineRun::LOADERS_BY_OUTPUT[output])
```
boris-dimitrov: This is fine by me but you should be aware that a technique often used for dynamic dispatch like this is virtual methods.

app/jobs/result_monitor_loader.rb
```diff
@@ -0,0 +1,22 @@
...
+      pr.update_result_status(output, PipelineRun::STATUS_LOADED)
```
boris-dimitrov: I don't quite understand how the pipeline result status is used to track the status of different loadable entities.    This looks like you are updating a status on the entire pipeline run, but all you really have information about is a specific loadable entity.

app/models/pipeline_run.rb
```diff
@@ -153,38 +162,317 @@ def retry
...
+
+  def output_ready?(output)
+    file_generated_since_run(s3_file_for(output))
+  end
+
+  def update_result_status(field, value)
+    status_hash = result_status_hash
+    status_hash[field] = value
+    update(result_status: status_hash.to_json)
+  end
```
boris-dimitrov: How do you ensure atomicity for these updates?

app/jobs/result_monitor_loader.rb
boris-dimitrov: Oh I see, `output` is a tag that describes the loadable entity.   In that case the only problem here is ensuring atomicity for these updates.

app/helpers/samples_helper.rb

app/helpers/samples_helper.rb
cdebourcy: ok, added description in pipeline_run.rb

app/jobs/result_monitor_loader.rb
boris-dimitrov: Do you also want to exit early if the pipeline run has failed?   Or perhaps not, because you want a failure in postprocessing not to prevent alignment results from getting loaded?

app/jobs/result_monitor_loader.rb
```diff
@@ -0,0 +1,22 @@
...
+    rescue
+      pr.update_result_status(output, PipelineRun::STATUS_FAILED)
+      pr.update(results_finalized: 1)
```
boris-dimitrov: What happens if postprocessing result loading fails before alignmnet result loading has started running on a resque worker?

app/models/pipeline_run.rb
```diff
@@ -33,12 +33,20 @@ class PipelineRun < ApplicationRecord
   STATUS_RUNNABLE = 'RUNNABLE'.freeze
   STATUS_ERROR = 'ERROR'.freeze # when aegea batch describe failed
   STATUS_LOADED = 'LOADED'.freeze
+  STATUS_EMPTY = 'EMPTY'.freeze
+  STATUS_LOADING = 'LOADING'.freeze
```
boris-dimitrov: Whose status?   Can we separate the status token namespaces for pipeline monitor vs result monitor.

app/models/pipeline_run.rb
```diff
@@ -58,6 +66,10 @@ def self.in_progress
       .where(finalized: 0)
   end
 
+  def self.results_in_progress
+    where(results_finalized: 0)
```
boris-dimitrov: Yeah this seems not to be the case if alignment loader and postprocessing loader are executed out of order by resque, and the postprocessing loader fails.

app/jobs/result_monitor_loader.rb
cdebourcy: ok, changed

app/models/pipeline_run.rb
```diff
@@ -71,54 +83,51 @@ def finalized?
...
+    target_outputs.each do |o|
+      h[o] = STATUS_EMPTY
+    end
+    self.result_status = h.to_json
+  end
+
   def create_run_stages
     # Host Filtering
     run_stages = []
     unless sample.host_genome && sample.host_genome.name == HostGenome::NO_HOST_NAME
```
boris-dimitrov: NO_HOST_NAME is no longer an option

app/jobs/result_monitor_loader.rb
cdebourcy: thanks!

app/models/pipeline_run.rb
cdebourcy: ok, I separated them in the code when I added a description. Some, such as STATUS_FAILED, are still shared between various entities. That can be the subject of another diff.

app/models/pipeline_run.rb
```diff
@@ -153,38 +162,317 @@ def retry
...
+
+    # Load version info applicable to host filtering
+    version_s3_path = "#{host_filter_output_s3_path}/#{VERSION_JSON_NAME}"
+    self.version = `aws s3 cp #{version_s3_path} -`
+
+    # Check if input was truncated
+    truncation_file = "#{host_filter_output_s3_path}/#{INPUT_TRUNCATED_FILE}"
+    _stdout, _stderr, status = Open3.capture3("aws", "s3", "ls", truncation_file)
+    self.truncated = `aws s3 cp #{truncation_file} -`.split("\n")[0].to_i if status.exitstatus.zero?
+    save
```
boris-dimitrov: How does this save not clobber updates from other resque jobs that are trying to modify the same pipeline run?

app/models/pipeline_run.rb
cdebourcy: Actually, I can still see it in idseq.net/host_genomes. However, it was renamed to "No host subtraction __test__" a while ago. Changing that in the code

app/models/pipeline_run.rb
cdebourcy: However, in that case, there is still no need to run the result monitor again, which is the point of the results_in_progress method. Alignment loader is already enqueued.

app/models/pipeline_run.rb
boris-dimitrov: The pipeline no longer supports it.   We always do at least ERCC.

app/jobs/result_monitor_loader.rb
cdebourcy: The result monitor stops polling for updates and the alignment result loading will complete eventually

app/models/pipeline_run.rb
cdebourcy: Ok. Will add an issue to make this change in a separate diff

app/models/pipeline_run.rb
boris-dimitrov: Got it.   Just slightly confused by the name `results_finalized`.

app/models/pipeline_run.rb
```diff
@@ -26,19 +26,57 @@ class PipelineRun < ApplicationRecord
...
-  before_create :create_run_stages
+  # The RESULT MONITOR is responsible for keeping status of available outputs
+  # and for loading those outputs in from S3.
+  # It accomplishes this using the following:
+  #    function "monitor_results"
+  #    columns "result_status", "results_finalized".
+  # The result status is a JSON string where keys are outputs and values are status,
+  # e.g.  "{\"ercc_counts\":\"LOADED\",\"taxon_counts\":\"LOADING\",\"taxon_byteranges\":\"EMPTY\"}".
+  # The progression for an output's status is as follows:
+  # EMPTY -> LOADING_QUEUED -> LOADING -> LOADED / FAILED.
```
boris-dimitrov: Could we change EMPTY to UNKNOWN?

cdebourcy: sure

app/models/pipeline_run.rb
cdebourcy: Fixed clobbering of result_status according to our discussion. The columns modified in db_load_ercc_counts should not get clobbered, because there is only one job executing this function for each pipeline_run.

app/models/pipeline_run.rb
cdebourcy: fixed according to our discussion

cdebourcy:
Added new table `output_states` to handle result status updates in an atomic way.

app/controllers/samples_controller.rb
```diff
@@ -172,7 +172,7 @@ def show
 
     background_id = check_background_id(@sample)
     @report_page_params = { pipeline_version: @pipeline_version, background_id: background_id } if background_id
-    if @pipeline_run && (((@pipeline_run.remaining_reads.to_i > 0 || @pipeline_run.finalized?) && !@pipeline_run.failed?) || @pipeline_run.report_ready?)
+    if @pipeline_run && (((@pipeline_run.remaining_reads.to_i > 0 || @pipeline_run.results_finalized?) && !@pipeline_run.failed?) || @pipeline_run.report_ready?)
```
boris-dimitrov: Will pipeline_run.failed? be true when all batch jobs completed successfully but some results failed to load?

app/lib/report_helper.rb
```diff
@@ -182,13 +182,12 @@ def report_details(pipeline_run, user_id)
...
     sample = pipeline_run.sample
     taxon_confirmation_hash = taxon_confirmation_map(sample.id, user_id)
-    postprocess_stage = pipeline_run.pipeline_run_stages.find_by(name: "Post Processing")
     {
       pipeline_info: pipeline_run,
       subsampled_reads: pipeline_run.subsampled_reads,
       sample_info: sample,
       default_background: Background.find(pipeline_run.sample.default_background_id),
-      taxon_fasta_flag: postprocess_stage && postprocess_stage.db_load_status == 1,
+      taxon_fasta_flag: !pipeline_run.taxon_byteranges.empty?,
```
boris-dimitrov: Are these loaded atomically?

app/models/host_genome.rb
```diff
@@ -1,7 +1,7 @@
 class HostGenome < ApplicationRecord
   has_many :samples
 
-  NO_HOST_NAME = 'No host subtraction'.freeze
+  NO_HOST_NAME = 'No host subtraction __test__'.freeze
```
boris-dimitrov: Delete this test.

app/models/pipeline_run_stage.rb
```diff
@@ -123,26 +132,18 @@ def due_for_aegea_check?
...
+    if failed? || succeeded?
+      terminate_job
       return
     end
     # The job appears to be in progress.  Check to make sure it hasn't been killed in AWS.   But not too frequently.
     return unless due_for_aegea_check?
-    # db_load_assembly(false) # update the list of taxids that have already been assembled by polling S3
     stdout, stderr, status = Open3.capture3("aegea", "batch", "describe", job_id.to_s)
     unless status.exitstatus.zero?
       Airbrake.notify("Error for update job status for pipeline run #{id} with error #{stderr}")
```
boris-dimitrov: In this case you need to update the status as though the jobid.failed file had been created.

app/models/project.rb
```diff
@@ -7,8 +7,8 @@ class Project < ApplicationRecord
   validates :name, presence: true, uniqueness: true
   include ReportHelper
 
-  def complete?
-    incomplete_runs = PipelineRun.where("id in (select max(id) from pipeline_runs group by sample_id) and sample_id in (select id from samples where project_id = #{id.to_i})").where("job_status != ?", PipelineRun::STATUS_CHECKED)
+  def results_complete?
+    incomplete_runs = PipelineRun.where("id in (select max(id) from pipeline_runs group by sample_id) and sample_id in (select id from samples where project_id = #{id.to_i})").where("results_finalized != ? and job_status != ?", PipelineRun::FINALIZED_SUCCESS, PipelineRun::STATUS_CHECKED)
```
boris-dimitrov: Why the check for job_status?

app/models/pipeline_run.rb
```diff
@@ -153,38 +243,311 @@ def retry
...
+    # ]
+    output = output_state.output
+    state = output_state.state
+    if finalized? && !output_ready?(output)
+      output_state.update(state: STATUS_FAILED)
+      return
+    end
+    # Otherwise, if the output is ready and has not yet been loaded or enqueued
+    # for loading, enqueue a loader job.
+    if output_ready?(output) && state == STATUS_UNKNOWN
```
boris-dimitrov: This is doing a lot of unnecessary aws S3 ops to check if output is ready.    You only need to check if STATUS is UNKNOWN.   I recommend rewriting as follows.

```
output = output_state.output
state = output_state.state
return unless state == STATUS_UNKNOWN
if output_ready?(output)
    output_state.update(state: STATUS_LOADING_QUEUED)
    Resque.enqueue(ResultMonitorLoader, id, output)
else if finalized?
    output_state.update(state: STATUS_FAILED)
end
```

boris-dimitrov: The explicit `return unless state == STATUS_UNKNOWN` is also consistent with the state transition diagram.   You might mention this function implements the Resource Monitor's state transitions from that diagram.  The only state the Resource Monitor is permitted to transition out of is the UNKNOWN state.   Which that diagram calls POLLING.   How about we just call it POLLING instead of UNKNOWN.

app/models/pipeline_run.rb
```diff
@@ -153,38 +243,311 @@ def retry
...
+    # Update job stats:
+    load_stats_file
+
+    # Check if run is complete:
+    if output_states.pluck(:state).all? { |s| s == STATUS_LOADED }
+      update(results_finalized: FINALIZED_SUCCESS)
+      handle_success
+    elsif all_output_states_terminal?
+      update(results_finalized: FINALIZED_FAIL)
+    end
```
boris-dimitrov: This is fine, but would probably be more efficient to not repeat the query in the common case (polling).
```
if all_output_states_terminal?
    if all_output_states_loaded?
        update(results_finalized: FINALIZED_SUCCESS)
        handle_success
    else
        update(results_finalized: FINALIZED_FAIL)
    end
end
```

app/models/pipeline_run.rb
```diff
@@ -153,38 +243,311 @@ def retry
...
+    if prs.nil?
+      # all stages succeeded
       self.finalized = 1
       self.job_status = STATUS_CHECKED
       save
-      if sample.project.complete?
-        notify_users
-        sample.project.create_or_update_project_background if sample.project.background_flag == 1
-      end
+      return
```
boris-dimitrov: I do not like early returns or scattered saves.    Those can be error prone.

app/models/pipeline_run.rb
```diff
@@ -210,6 +573,22 @@ def file_generated_since_run(s3_path)
     end
   end
 
+  def file_generated_since_jobstats?(s3_path)
+    # If there is no file, return false
+    stdout, _stderr, status = Open3.capture3("aws", "s3", "ls", s3_path.to_s)
+    return false unless status.exitstatus.zero?
+    # If there is a file and there are no existing job_stats yet, return true
+    existing_jobstats = job_stats
+    return true if existing_jobstats.empty?
```
boris-dimitrov: Why not crash?

boris-dimitrov: job_stats isn't an explicit argument to this function, probably should be

app/models/pipeline_run.rb
```diff
@@ -210,6 +573,22 @@ def file_generated_since_run(s3_path)
...
+    # If there is a file and there are job_stats, check if the file supersedes the job_stats:
+    begin
+      s3_file_time = DateTime.strptime(stdout[0..18], "%Y-%m-%d %H:%M:%S")
+      return (s3_file_time && existing_jobstats.created_at && s3_file_time > existing_jobstats.created_at)
+    rescue
```
boris-dimitrov: Isn't this a programmer error?   If so, shouldn't we see something in the log?

boris-dimitrov:
@cdebourcy don't forget to update status to failed when a job crashes in aegea

app/models/pipeline_run.rb
cdebourcy: ok changed

app/models/pipeline_run.rb
cdebourcy: ok changed

app/models/pipeline_run_stage.rb
```diff
@@ -123,26 +132,18 @@ def due_for_aegea_check?
...
-      Rails.logger.info "Job #{id} #{job_id} checked and waiting to load results."
-      return
-    end
-    if output_ready?
-      self.job_status = STATUS_CHECKED
-      # save before enqueue, to prevent minor race
-      save
-      Resque.enqueue(LoadResultForRunStage, id)
+    check_status_file_and_update(JOB_SUCCEEDED_FILE_SUFFIX, STATUS_SUCCEEDED)
+    check_status_file_and_update(JOB_FAILED_FILE_SUFFIX, STATUS_FAILED)
```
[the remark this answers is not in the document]
cdebourcy: ok

app/models/pipeline_run_stage.rb
cdebourcy: that's what `self.job_status = STATUS_FAILED if stderr =~ /IndexError/ # job no longer exists` does?

app/models/project.rb
cdebourcy: for old runs. added comment & todo

app/models/pipeline_run.rb
cdebourcy: I don't see the point in giving it job_stats as an argument. They are being treated as properties of the pipeline_run.

app/controllers/samples_controller.rb
cdebourcy: ok I changed it so that the answer is yes

app/lib/report_helper.rb
cdebourcy: yes

app/models/pipeline_run.rb
cdebourcy: ok changed

PR #1341, opened 2018-06-07
Add version check for CLI

jshoe:
- Add a version check for samples_controller#create
- Tested locally, will test more once it makes it to staging
- The CLI for single and bulk uploads both hit samples_controller#create so it's the only thing that really needs a version flag.

- Accompanying CLI PR: https://github.com/chanzuckerberg/idseq-cli/pull/28

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -382,27 +382,21 @@ class SamplesController < ApplicationController
   # POST /samples.json
   def create
     # Single sample upload path
+    params = sample_params
+
     # Check if the client is up-to-date. "web" is always valid whereas the
     # CLI client should provide a version string to-be-checked against the
     # minimum version here. Bulk upload from CLI goes to this method.
-    valid_client = false
-    if client_params && client_params[:client]
-      client = client_params[:client]
-      min_version = '0.3.2'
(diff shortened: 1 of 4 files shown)


app/controllers/samples_controller.rb
```diff
@@ -381,6 +381,27 @@ def edit
...
+      end
+    end
+    unless valid_client
+      respond_to do |format|
+        msg = "Outdated command line client. Please run `pip install --upgrade git+https://github.com/chanzuckerberg/idseq-cli.git `"
+        format.json { render json: { status: msg }, status: :upgrade_required }
+        format.html { render json: { status: msg }, status: :upgrade_required }
+      end
+      return
+    end
```
cdebourcy: I think you can accomplish this with less code.
1. add :client in the permit list of `sample_params`
2. move `params = sample_params` to the top of `create`
3. change your block to:

```
    client = params[:client]
    unless client == "web" || Gem::Version.new(client) >= Gem::Version.new('0.3.2')
        render json: {
          message: "Outdated command line client. Please run `pip install --upgrade git+https://github.com/chanzuckerberg/idseq-cli.git `",
          status: :upgrade_required
        }
      return
    end
```

PR #1348, opened 2018-06-19
Pass AWS authentication environment variables to web node. 

tfrcarvalho:
# Description

Pass AWS authentication environment variables to web node. Allows authentication thr…ough env variables (vs. disk default profile).
Update lint-staged (bug due to a combination of node and package versions).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -71,6 +71,8 @@ services:
       - db
       - redis
     environment:
+      - AWS_ACCESS_KEY_ID
+      - AWS_SECRET_ACCESS_KEY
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -71,6 +71,8 @@ services:
       - db
       - redis
     environment:
+      - AWS_ACCESS_KEY_ID
+      - AWS_SECRET_ACCESS_KEY
```
yunfangjuan: you might want to do this for resque, job_monitor and result_monitor too.

PR #1351, opened 2018-06-19
Add AWS credentials environment variable for all nodes that need it.

tfrcarvalho:


jshoe:
not sure how these are currently getting set actually then.. through VPC settings?

tfrcarvalho:
No, docker actually maps a local volume to the containers. 
```
    volumes:
      - .:/app
      - ~/.aws:/root/.aws:ro
```
But using environment variables allows me to use `aws-vault` and not have the credentials written to disk locally.

PR #1352, opened 2018-06-20
WIP - use DAG to execute host filtering

cdebourcy:


app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -632,12 +632,16 @@ class PipelineSampleReads extends React.Component {
...
                 <div
                   className={`details-value col s6 plain ${
-                    !this.summary_stats.remaining_reads ? BLANK_TEXT : ""
+                    !this.summary_stats.adjusted_remaining_reads
+                      ? BLANK_TEXT
+                      : ""
                   }`}
                 >
-                  {!this.summary_stats.remaining_reads
+                  {!this.summary_stats.adjusted_remaining_reads
```
cdebourcy: Why are we displaying adjusted remaining reads here and in other places? Isn't that going to be confusing when the number of reads in the non-host fasta is different from the displayed value? I think adjusted values should only be used for rpm, where they are actually important for normalization

jshoe: Oh the idea was that they'd get too confused by seeing a tiny remaining reads now that subsampling is moved up? But should we change some back?

cdebourcy: Well, if the number of reads fed into the alignment stage was tiny, I think they should be aware of that, no? No point in artificially inflating it... @yunfangjuan ?

PR #1359, opened 2018-07-05
Fix background selection for bulk report downloads

jshoe:
- More of a feature than a bug fix but lets the users bulk download with the background model stored in Cookies.
- Fetching background ID by name since we store the name in the Cookies
- Asked for by Akshaya
- Tested locally a couple of ways

### DIFF
--- a/app/lib/report_helper.rb
+++ b/app/lib/report_helper.rb
@@ -957,7 +957,8 @@ module ReportHelper
   def report_csv_from_params(sample, params)
     params[:is_csv] = 1
     params[:sort_by] = "highest_nt_aggregatescore"
-    background_id = params[:background_id] || sample.default_background_id
+    background_id = params[:background_id] || params["background_id"] || sample.default_background_id
+    background_id = background_id.to_i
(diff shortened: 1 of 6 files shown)


app/lib/report_helper.rb
```diff
@@ -957,7 +957,8 @@ def generate_report_csv(tax_details)
   def report_csv_from_params(sample, params)
     params[:is_csv] = 1
     params[:sort_by] = "highest_nt_aggregatescore"
-    background_id = params[:background_id] || sample.default_background_id
+    background_id = params[:background_id] || params["background_id"] || sample.default_background_id
```
yunfangjuan: Why both symbol and string?

jshoe: Tweaked. Apparently when it goes through like `Resque.enqueue(GenerateProjectReportsCsv, params)` the 'params' changes from ActionController::Parameters to 'Hash' without indifferent access for symbol vs. string. So added `hash =  HashWithIndifferentAccess.new(hash)`

PR #1363, opened 2018-07-10
Closes #1272: restrict access to source files

### LINKED ISSUE #1272: Exclude source fastqs from 1-year privacy expiration

cdebourcy:


app/models/power.rb
```diff
@@ -33,6 +33,11 @@ def initialize(user)
     Sample.editable(@user)
   end
 
+  power :host_viewable_samples do
```
yunfangjuan: I don't think you really need this. see comments above.

app/controllers/samples_controller.rb
```diff
@@ -358,7 +358,7 @@ def results_folder
   end
 
   def fastqs_folder
-    @file_list = @sample.fastqs_folder_files
+    @file_list = current_power.host_viewable_samples.include?(@sample) ? @sample.fastqs_folder_files : []
```
yunfangjuan: I think you can just do current_power.updatable_samples.find_by(id: @sample.id)  or just use current_power.updatable_samples

PR #1382, opened 2018-07-13
UI for creating backgrounds

cdebourcy:


app/assets/src/components/Samples.jsx
```diff
@@ -858,6 +860,27 @@ class Samples extends React.Component {
...
+        description: description,
+        sample_ids: sample_ids,
+        authenticity_token: this.csrf
+      })
+      .then(response => {
+        that.setState({
+          background_creation_response: response.data
+        });
+      })
+      .catch(error => {
```
tfrcarvalho: Shouldnt we give some feedback to the user?

app/assets/src/components/Samples.jsx
```diff
@@ -987,32 +1010,12 @@ class Samples extends React.Component {
...
-              {compare_button_inner}
-            </a>
-          ) : (
-            <a className="compare center btn-disabled">
-              {compare_button_inner}
-            </a>
-          )}
-        </div>
-      </div>
+      <CohortButton
```
tfrcarvalho: Why Cohort? 
We should give more descriptive names.

app/assets/src/components/Samples.jsx
```diff
@@ -987,32 +1010,12 @@ class Samples extends React.Component {
...
+        label="Compare"
+        action={this.compareSamples}
```
tfrcarvalho: action -> onClick
We should keep descriptive names and in this case, jut inherit the name of the underlying element.

app/assets/src/components/Samples.jsx
```diff
@@ -1708,6 +1712,125 @@ function TableDownloadDropdown({ project_id, parent }) {
...
+    );
+  }
+
+  render() {
+    let background_creation_response = this.props.parent.state
+      .background_creation_response;
+    return (
+      <Modal
+        trigger={
+          <CohortButton
```
tfrcarvalho: Do we use the same button as compare? Or is this temporary until you get feedback from Jenn?

app/assets/src/components/Samples.jsx
```diff
@@ -2180,4 +2305,31 @@ function AddUserModalMemberArea({ state, parent }) {
   );
 }
 
+function CohortButton({ label, action, enabled }) {
```
tfrcarvalho: Can you move this to a 'buttons' subfolder? We should start organizing any components that might be shared.

app/assets/src/styles/samples.scss
```diff
@@ -1107,7 +1107,7 @@ input[type="button"] {
           padding-top: 20px;
         }
         .compare {
-          padding: 5px 30px;
+          padding: 5px;
```
tfrcarvalho: Why is the padding change required?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
```
tfrcarvalho: Can you manually edit this so that encoding is not changed, but only the fields you wanted to change?

app/assets/src/components/Samples.jsx
```diff
@@ -1708,6 +1712,125 @@ function TableDownloadDropdown({ project_id, parent }) {
...
+      sample_names: []
+    };
+    this.handleOpen = this.handleOpen.bind(this);
+    this.handleClose = this.handleClose.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.renderTextField = this.renderTextField.bind(this);
+    this.sample_ids = props.parent.state.selectedSampleIds;
+  }
+  handleOpen() {
```
tfrcarvalho: Why do you need a new request to get sample names? 
You can just pass them as attributes to the component as you loaded them to the table, right?

db/schema.rb
tfrcarvalho: I am still investigating why rails tries to change the encoding.

app/models/background.rb
```diff
@@ -74,10 +74,15 @@ def compute_stdev(sum, sum2, n)
   end
 
   def self.viewable(user)
+    viewable_pipeline_run_ids = PipelineRun.where(sample_id: Sample.viewable(user).pluck(:id)).pluck(:id)
     if user.admin?
       all
     else
-      where(project_id: Project.viewable(user).pluck(:id) + [nil])
+      # backgrounds which are marked as public or for which none of the pipeline_run_ids are ABSENT from the viewable_pipeline_run_ids
```
tfrcarvalho: very confusing... 
Does it meant that: "public backgrounds, for which all pipeline_run_ids are viewable"?

app/views/backgrounds/_form.html.erb
```diff
@@ -1,31 +0,0 @@
-<%= form_with(model: background, local: true) do |form| %>
```
tfrcarvalho: Do you need to delete the previous background create actions? 
As we discuss, we should keep this and make a revamped dedicated page for backgrounds later.

app/assets/src/components/Samples.jsx
cdebourcy: yes, will add

app/assets/src/components/Samples.jsx
cdebourcy: The idea was that this is a button representing an action on a group of samples, i.e. a "cohort" -- such as comparing them in a heatmap or aggregating them in a background. Better name suggestions welcome -- what did you have in mind?

app/assets/src/components/Samples.jsx
cdebourcy: Hmm, I can't just pass the names from the sample table to the component for lookup, because whenever the user clicks on a project, the sample table from the previous project is lost -- but some samples from that project may still be selected and need to be added to the new background too. 

What I could do is add the sample name into the state whenever a sample is selected, just like we currently add the sample ID to keep track of it... but I feel this makes the organization of the code less clear. 

The sample selection checkbox feature is shared between multiple buttons ("Compare" / "Make background" / potentially others in the future), so it is currently based on sample_id only (rather than adding all the attributes needed by all of the CohortButtons into the state). The "Compare" (heatmap) action then retrieves all the information it needs (names, top taxa, scores, etc) once you click -- I think it is most consistent if the "Background" button functions similarly.

app/assets/src/components/Samples.jsx
cdebourcy: Yes to both :D Conceptually it should be a similar button because it represents an action on a group of samples. But it is also temporary until I get a design from Jenn

app/assets/src/components/Samples.jsx
cdebourcy: sure

app/assets/src/styles/samples.scss
cdebourcy: It is temporary. Padding caused the text in the button to spread over 2 lines.

app/assets/src/components/Samples.jsx
tfrcarvalho: I do not think the idea is to use the same exact button for "Comparing" and "Create Background". The icon was design for comparing only. 
I am working on organizing all our UI assets. I believe we should have generic buttons: "PrimaryButton", "SecondaryButton" that have specific designs. And for button with specific icons, we will give the name of the action "CompareButton"... Currently, I think you should a generic primary button instead of the compare.

app/models/background.rb
cdebourcy: No. It means:
"Background is viewable by user because
either
(a) user can view all pipeline_runs that went into the background (because he/she is part of the projects that contain them)
or
(b) background is marked as public, so any user can see it (even when they don't have access to the individual pipeline_runs that went into it).

app/views/backgrounds/_form.html.erb
cdebourcy: This page has become unusable because it loads all samples that are in the database onto the page. Therefore the page is so slow and unwieldy that we've been unable to use it for background creation.
The new feature in this PR replicates the same functionality (select samples, enter name, create background) but in a way that is actually usable

app/assets/src/components/Samples.jsx

app/assets/src/components/Samples.jsx
cdebourcy: sounds good

app/assets/src/components/Samples.jsx
tfrcarvalho: I believe the icon makes not sense for this particular action.

app/assets/src/components/Samples.jsx
tfrcarvalho: "the sample table from the previous project is lost -- but some samples from that project may still be selected" - we need to discuss this, makes no sense from a UX perspective.
"compare" leads to a different page, "create background" creates a modal. It is a very different behavior.

app/models/background.rb
tfrcarvalho: Ok, you should add this new explanation to the comments instead.

app/assets/src/components/Samples.jsx
cdebourcy: Sure, let's discuss. The reason is, people need to be able to select samples from multiple projects. So selection needs to stick even if user switches projects.

app/views/backgrounds/_form.html.erb
tfrcarvalho: Yeah, in any case this does not use react as all other components.

app/assets/src/styles/samples.scss
cdebourcy: removed padding change as part of switching to generic button

app/controllers/backgrounds_controller.rb
```diff
@@ -26,16 +30,32 @@ def edit
...
 
-    respond_to do |format|
-      if @background.save
-        format.html { redirect_to @background, notice: 'Background was successfully created.' }
-        format.json { render :show, status: :created, location: @background }
-      else
-        format.html { render :new }
-        format.json { render json: @background.errors, status: :unprocessable_entity }
-      end
+    if sample_ids.any? { |sid| !current_power.samples.pluck(:id).include?(sid) }
```
tfrcarvalho: Shouldn't these checks be done in the model ?

cdebourcy: Yes, but the background creation is asynchronous. We need to let the user know what's wrong as soon as they submit their request

app/assets/src/components/Samples.jsx
cdebourcy: ok removed icon

app/assets/src/components/Samples.jsx
```diff
@@ -2180,4 +2328,34 @@ function AddUserModalMemberArea({ state, parent }) {
...
+function ActiveInactiveButton({
+  label,
+  onClick,
+  enabled,
+  icon,
+  outerClass,
+  enabledClass,
+  disabledClass
+}) {
+  let cohort_button_inner = (
```
tfrcarvalho: Can you just remove the `cohort_` prefix?

cdebourcy: ah yes, thanks

test/controllers/backgrounds_controller_test.rb
```diff
@@ -12,29 +12,16 @@ class BackgroundsControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
-  test "should get new" do
-    get new_background_url
-    assert_response :success
-  end
-
   test "should create background" do
```
tfrcarvalho: Could you create a test "should not create background if it cannot view all samples" ?

app/models/background.rb
```diff
@@ -77,7 +82,16 @@ def self.viewable(user)
...
     else
-      where(project_id: Project.viewable(user).pluck(:id) + [nil])
+      # Background is viewable by user if either
+      # (a) user is allowed to view all pipeline_runs that went into the background, or
+      # (b) background is marked as public (regardless of whether user is allowed to view individual pipeline_runs).
+      condition = "public_access = 1"
+      viewable_pipeline_run_ids = PipelineRun.where(sample_id: Sample.viewable(user).pluck(:id)).pluck(:id)
+      unless viewable_pipeline_run_ids.empty?
+        condition += " or id not in (select background_id from backgrounds_pipeline_runs
+                                    where pipeline_run_id not in (#{viewable_pipeline_run_ids.join(',')}))"
```
jshoe: optional but you may be able to keep this in ActiveRecord?

config/routes.rb
```diff
@@ -34,6 +34,7 @@
   get 'samples/:id/assembly/:taxid', to: 'samples#assembly'
   get 'samples/:id/alignment/:taxon_info', to: 'samples#show_taxid_alignment'
   get 'samples/:id/alignment_viz/:taxon_info', to: 'samples#show_taxid_alignment_viz'
+  get 'show_sample_names', to: 'samples#show_sample_names'
```
jshoe: maybe for hierarchy 'samples/show_names'

app/assets/src/components/Samples.jsx
yunfangjuan: agreed that a sample_id => name mapping will be ideal. It can just be stored in a class variable. it doesn't have to be in the state.

app/controllers/backgrounds_controller.rb
```diff
@@ -1,7 +1,11 @@
 class BackgroundsController < ApplicationController
   include BackgroundsHelper
+  before_action :login_required
   before_action :set_background, only: [:show, :edit, :update, :destroy]
-  before_action :admin_required
+
+  before_action :admin_required, except: :create
+  before_action :authenticate_user!, only: :create
```
yunfangjuan: i think line 7 and 8, the only: should be removed. DANGER!

app/controllers/backgrounds_controller.rb
```diff
@@ -26,15 +30,27 @@ def edit
   # POST /backgrounds
   # POST /backgrounds.json
   def create
-    @background = Background.new(background_params)
+    name = params[:name]
+    description = params[:description]
+    sample_ids = params[:sample_ids]
 
-    respond_to do |format|
+    if sample_ids.any? { |sid| !current_power.samples.pluck(:id).include?(sid) }
```
yunfangjuan: I think you can simplify this by doing sample_ids.to_set - current_power.samples.pluck(:id).to_set and make sure it's empty.

app/controllers/samples_controller.rb
```diff
@@ -377,6 +377,12 @@ def show_taxid_alignment_viz
     end
   end
 
+  def show_sample_names
```
yunfangjuan: per discussion, perhaps we don't need this.

app/assets/src/components/Samples.jsx
```diff
@@ -840,6 +849,12 @@ class Samples extends React.Component {
       if (checked) {
         if (sampleList.indexOf(sample_id) === -1) {
           sampleList.push(sample_id);
+          // Also keep track of certain data for the sample that was just added
+          let attributeHash = {};
+          for (let key of this.sampleAttributeKeys) {
+            attributeHash[key] = this.getSampleAttribute(sample.db_sample, key);
```
tfrcarvalho: getSampleAttibute is not neccessary.

What do you want to store here specially if the key does not exist?

app/assets/src/components/Samples.jsx
```diff
@@ -245,6 +248,12 @@ class Samples extends React.Component {
     });
   }
 
+  getSampleAttribute(dbSample, key) {
```
tfrcarvalho: As per comment above, this is not necessary.

app/assets/src/components/Samples.jsx
```diff
@@ -35,7 +35,10 @@ class Samples extends React.Component {
     this.favoriteProjects = props.favorites || [];
     this.allProjects = props.projects || [];
     this.pageSize = props.pageSize || 30;
+    this.sampleAttributeHelper = {};
```
tfrcarvalho: Use more descriptive names, e.g. 
sampleAttributeHelper -> selectedSamples
sampleAttributeKeys -> sampleAttributesToStore

helps understand what these variables are for

app/assets/src/components/Samples.jsx
cdebourcy: It is work in progress. I am planning to add pipeline_run_id to the dictionary, which will need to be retrieved in a different way. For that, I'll need the function

app/assets/src/components/Samples.jsx
cdebourcy: ok for sampleAttributeToStore. I feel selectedSamples has too much potential for confusion with selectedSampleIds.

cdebourcy: Plus, it isn't actually just selected samples. It could contain more than the selected samples

tfrcarvalho: You should merge both.
"It could contain more than the selected samples" How so?

app/assets/src/components/Samples.jsx
tfrcarvalho: Ok, let me know when you are done...

test/controllers/backgrounds_controller_test.rb
cdebourcy: done

app/assets/src/components/Samples.jsx
cdebourcy: Ok now that's no longer the case. Still, I prefer to keep selectedSampleIds and sampleAttributeHelper separate. I need selectedSampleIds in the state to control visuals. sampleAttributeHelper is used for looking up certain additional information for the modal

PR #1389, opened 2018-07-18
integrating amr step

Meera-Srinivasan:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change. Explain both the expected functionality and the edge cases. 

## Type of change

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -1,5 +1,6 @@
 require 'open3'
 require 'json'
+
 class PipelineRun < ApplicationRecord
   include ApplicationHelper
   include PipelineOutputsHelper
@@ -26,9 +27,11 @@ class PipelineRun < ApplicationRecord
   PIPELINE_VERSION_FILE = "pipeline_version.txt".freeze
   STATS_JSON_NAME = "stats.json".freeze
   ERCC_OUTPUT_NAME = 'reads_per_gene.star.tab'.freeze
+  AMR_OUTPUT_NAME = 'output__fullgenes__ARGannot_r2__results.txt'.freeze
   TAXID_BYTERANGE_JSON_NAME = 'taxid_locations_combined.json'.freeze
   ASSEMBLY_STATUSFILE = 'job-complete'.freeze
   LOCAL_JSON_PATH = '/app/tmp/results_json'.freeze
+  LOCAL_TXT_PATH = '/app/tmp/results_txt'.freeze
   PIPELINE_VERSION_WHEN_NULL = '1.0'.freeze
 
   # The PIPELINE MONITOR is responsible for keeping status of AWS Batch jobs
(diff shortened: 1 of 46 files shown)


app/models/pipeline_run.rb
```diff
@@ -291,6 +297,29 @@ def invalid_family_call?(tcnt)
     false
   end
 
+  def db_load_amr_counts
+    # amr_s3_path = "#{postprocess_output_s3_path}/#{AMR_OUTPUT_NAME}"
+    amr_test_s3_path = "s3://idseq-database/test/AMR/__fullgenes__ARGannot_r2__results.txt"
+    amr_downloaded_path = PipelineRun.download_file(amr_test_s3_path, local_txt_path)
```
cdebourcy: I think you'll want to change local_txt_path to something pipeline_run-specific so that there are no clashes between runs. E.g. "local_txt_path = app/temp/amr_results/#{id}"

app/models/pipeline_run.rb
```diff
@@ -386,11 +417,19 @@ def pre_result_monitor?
   def status_display_pre_result_monitor(run_stages)
     # TODO: remove the need for this function by migrating old runs
     state_by_output = {}
-    old_loaders_by_output = { "db_load_host_filtering" => "ercc_counts",
-                              "db_load_alignment" => "taxon_counts",
-                              "db_load_postprocess" => "taxon_byteranges" }
+    # old_loaders_by_output = { "db_load_host_filtering" => "ercc_counts",
```
cdebourcy: I think you should rebase on current master. Legacy function `status_display_pre_result_monitor` no longer exists ;) so no need to bother with it

app/models/pipeline_run.rb
cdebourcy: Oh, just saw you did this.

db/migrate/20180719162241_create_amr_counts.rb
```diff
@@ -0,0 +1,13 @@
+class CreateAmrCounts < ActiveRecord::Migration[5.1]
+  def change
+    create_table :amr_counts do |t|
+      t.string :sample_id
```
cdebourcy: Since you say amr_count belongs to pipeline_run, you need to have a column pipeline_run_id here (bigint). Also, remove sample_id. If I rerun the pipeline for a sample (new pipeline_run), the resulting amr_counts should be associated with that new pipeline_run, and the old pipeline_run would still have its own amr_counts. So the relationship is on the pipeline_run level, not the sample level.

cdebourcy: Ok, saw you have pipeline_run_id. But sample_id still needs to be removed.

Meera-Srinivasan:
@cdebourcy thanks for the comments! Will fix and for sure, am currently trying to rebase from master :)

PR #1390, opened 2018-07-19
Rename fq to fastq

jshoe:
- Name sub of .fq to .fastq in case people upload with .fq. Our pipeline expects .fastq in a few locations apparently.


jshoe:
Tweaked but yeah for some reason the dag template was specifying "AC1_S1_R1_val_1.fastq"

PR #1394, opened 2018-07-20
Fix heatmap bugs and change categories filter behavior

tfrcarvalho:
Few fixes and UI changes:
* Categories acts like a filter: select options are filtered out from the original results (no options selected means all results are shown)
* Move updating heatmap to a callback of setState
* LabeledSlider manages its own state now.

### DIFF
--- a/app/assets/src/components/modules/LabeledSlider.jsx
+++ b/app/assets/src/components/modules/LabeledSlider.jsx
@@ -1,17 +1,46 @@
+import PropTypes from "prop-types";
 import React from "react";
 import Slider from "rc-slider";
 
 class LabeledSlider extends React.Component {
+  constructor(props) {
+    super(props);
+    this.onChangeParent = this.props.onChange;
+
+    this.state = {
+      value: this.props.value
+    };
+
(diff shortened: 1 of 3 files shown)


app/assets/src/components/modules/LabeledSlider.jsx
```diff
@@ -1,17 +1,46 @@
...
+      value: this.props.value
+    };
+
+    this.onChange = this.onChange.bind(this);
+  }
+
+  onChange(newValue) {
+    this.setState({ value: newValue });
+    if (typeof this.onChangeParent == "function") {
+      this.onChangeParent();
```
jshoe: what is the onChangeParent needed for actually?

tfrcarvalho: If the parent (in this case the heatmap page) wants to do any else when the LabeledSlider value is changed.

Labeled Slider hooks to the on change of Labeled Slider to update its inner fields.
The SamplesHeatmap hooks to the on change of Labeled Slider to update the heatmap when value changes.

PR #1400, opened 2018-07-24
Redo CLI instructions modal

jshoe:
- Mostly a static change
- Based on direction from Rebecca/Jenn
- Here's what it looks like now:
screen shot 2018-07-24 at 2 28 52 pm

### DIFF
--- a/app/assets/src/components/CliUserInstructionsModal.jsx
+++ b/app/assets/src/components/CliUserInstructionsModal.jsx
@@ -1,0 +1,27 @@
+import React from "react";
+import { Modal } from "semantic-ui-react";
+import PropTypes from "prop-types";
+
+class CliUserInstructionsModal extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    return (
+      <Modal
+        trigger={this.props.trigger}
+        className="project-popup generic-centered-modal"
+      >
+        <Modal.Header>Command Line Upload Instructions</Modal.Header>
+        <Modal.Content />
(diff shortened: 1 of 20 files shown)


app/assets/src/components/CliUserInstructionsModal.jsx
```diff
@@ -0,0 +1,191 @@
...
+          <Button primary onClick={this.close}>
+            Close
+          </Button>
+        </Modal.Actions>
+      </Modal>
+    );
+  }
+}
+
+CliUserInstructionsModal.propTypes = {
```
tfrcarvalho: +1

app/assets/src/styles/_header.scss
```diff
@@ -268,22 +268,27 @@ a {
   }
 }
 
+.instruction-heading {
```
tfrcarvalho: Please move the css for the new component to its own file, i.e. `cli_user_instructions_modal.scss`

app/assets/src/components/CliUserInstructionsModal.jsx
```diff
@@ -0,0 +1,191 @@
...
+  open() {
+    this.setState({ open: true });
+  }
+
+  close() {
+    this.setState({ open: false });
+  }
+
+  render() {
+    const singleUploadCmd = `idseq -e ${this.props.email} -t ${
```
tfrcarvalho: be consistent with strings definition: `` ` `` -> `"` unless you need to avoid escaping quotes.

PR #1403, opened 2018-07-26
Add rudimentary error log on failed sample report page

jshoe:
- Add a rudimentary error log to the failed sample report page
- Plan is to add lines to the pipeline code to add a cleaner log summary at the bottom that is more human interpretable, and then those will also show up here. Even as simple as "failed at the STAR_OUT step"
- This more raw information will be present in the hopefully rare failed samples so that users can get some feedback when we don't have a cleaner error message

### DIFF
--- a/app/assets/src/components/PipelineSampleReads.jsx
+++ b/app/assets/src/components/PipelineSampleReads.jsx
@@ -592,6 +592,12 @@ class PipelineSampleReads extends React.Component {
               ? waitingSpinner
               : this.state.rerunStatusMessage}
           </h6>
+          <div className="log-block">
+            {
+              "Rudimentary debugging information (from the end of the pipeline run logs):\n"
+            }
+            {this.props.logSummary}
+          </div>
         </div>
       );
     }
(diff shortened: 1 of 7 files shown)


cdebourcy:
Not sure we want to show this level of raw data to users? I don't think anyone will be able to interpret it (they don't even know what the steps of the pipeline are). I think it might make more sense to handle frequent error cases (e.g. no reads left after filtering), record the name of the stage that failed, and implement a framework for showing this info in an organized manner.

Also, I think it would be better to populate `log_summary` as part of pipeline_monitor rather than in the samples_controller.

jshoe:
Hah ok I'm going to rework this since people disagree on how to do this.. YF and I were kind of OK with tailing the logs + adding nicer error messages since there are a lot of unexpected errors that could come up. Nobody seems to like the idea of log parsing so maybe the better thing is -> pipeline generates nice error messages when encountered -> spit it out at the bottom of the log -> display only the pretty error messages in the report page without any raw logs

tfrcarvalho:
Actually, what it spits out could depend on the time of user. No matter how we output we can generate: (1) user friendly message (2) admin friendly message (3) traceback information if required.

jshoe:
hm yeah that could be a good idea

PR #1404, opened 2018-07-27
Meera/add amr

Meera-Srinivasan:
# Description

 Added code to run AMR detection as part of the host filtering step, dump results in a database, and visualize on a separate page in table form. 
## Type of change

- [ x] New feature (non-breaking change which adds functionality)

# Test and Reproduce

1.  

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

- [x ] I have run through the testing script to make sure current functionality is unchanged
- [in progress ] I have done relevant tests that prove my fix is effective or that my feature works
- [in progress ] I have spent time testing out edge cases for my feature

### DIFF
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -7,18 +7,27 @@ const columns = [
   {
     Header: "Antibiotic Class",
     accessor: "drug_family",
+    style: {
+      textAlign: "center"
+    },
     filterable: false
   },
   {
     Header: "Gene",
     accessor: "gene",
     aggregate: vals => _.size(vals),
+    style: {
+      textAlign: "center"
+    },
     filterable: false
   },
   {
     Header: "Gene Family",
     accessor: "gene_family",
     aggregate: vals => _.size(vals),
+    style: {
+      textAlign: "center"
+    },
     filterable: false
   },
   {
@@ -46,20 +55,16 @@ const columns = [
 class AMRView extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { data: props.amr ? props.amr : [] };
-    if (this.state.data) {
+    this.data = props.amr ? props.amr : (this.data = []);
+    if (this.data) {
       // TODO: Talk with Chaz to determine how to improve gene family identification further.
-      for (var i = 0; i < Object.keys(this.state.data).length; i++) {
-        var key = Object.keys(this.state.data)[i];
-        this.state.data[key].gene_family = this.state.data[key].gene.slice(
-          0,
-          5
-        ); // first four characters of gene family
+      for (var i = 0; i < Object.keys(this.data).length; i++) {
+        var key = Object.keys(this.data)[i];
+        this.data[key].gene_family = this.data[key].gene.slice(0, 5); // first four characters of gene family
       }
     }
   }
   render() {
-    const { data } = this.state;
     return (
       <div>
         <ReactTable
@@ -72,7 +77,7 @@ class AMRView extends React.Component {
                   .includes(filter.value.toLowerCase())
               : true;
           }}
-          data={data}
+          data={this.data}
           columns={columns}
           defaultPageSize={5}
           pivotBy={["drug_family", "gene_family"]}
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -19,6 +19,7 @@ const columns = [
     aggregate: vals => _.size(vals)
   },
   {
+    // TODO: Make average for gene family level
     Header: "Coverage",
     accessor: "coverage",
     aggregate: vals => _.sum(vals),
@@ -36,9 +37,15 @@ class AMRView extends React.Component {
   constructor(props) {
     super(props);
     this.state = { data: props.amr };
-    for (var i = 0; i < Object.keys(this.state.data).length; i++) {
-      var key = Object.keys(this.state.data)[i];
-      this.state.data[key].gene_family = this.state.data[key].gene.slice(0, 5); // first four characters of gene family
+    if (this.state.data) {
+      // TODO: Use regex to parse out gene family, i.e. is [.]*_ -- everything until the underscore
+      for (var i = 0; i < Object.keys(this.state.data).length; i++) {
+        var key = Object.keys(this.state.data)[i];
+        this.state.data[key].gene_family = this.state.data[key].gene.slice(
+          0,
+          5
+        ); // first four characters of gene family
+      }
     }
   }
   render() {
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -40,8 +40,8 @@ class AMRView extends React.Component {
   constructor(props) {
     super(props);
     this.state = { data: props.amr };
-    for (var z = 0; z < Object.keys(this.state.data).length; z++) {
-      var key = Object.keys(this.state.data)[z];
+    for (var i = 0; i < Object.keys(this.state.data).length; i++) {
+      var key = Object.keys(this.state.data)[i];
       this.state.data[key].gene_family = this.state.data[key].gene.slice(0, 5); // first four characters of gene family
     }
     console.log(this.state.data);
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -1,5 +1,4 @@
 import React from "react";
-import ReactDOM from "react-dom";
 import ReactTable from "react-table";
 import "react-table/react-table.css";
 import _ from "lodash";
@@ -7,10 +6,7 @@ import _ from "lodash";
 const columns = [
   {
     Header: "Antibiotic Class",
-    accessor: "drug_family",
-    PivotValue: ({ value }) => (
-      <span style={{ color: "darkblue" }}>{value}</span>
-    )
+    accessor: "drug_family"
   },
   {
     Header: "Gene",
@@ -45,7 +41,6 @@ class AMRView extends React.Component {
       this.state.data[key].gene_family = this.state.data[key].gene.slice(0, 5); // first four characters of gene family
     }
   }
-
   render() {
     const { data } = this.state;
     return (
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -6,29 +6,40 @@ import _ from "lodash";
 const columns = [
   {
     Header: "Antibiotic Class",
-    accessor: "drug_family"
+    accessor: "drug_family",
+    filterable: false
   },
   {
     Header: "Gene",
     accessor: "gene",
-    aggregate: vals => _.size(vals)
+    aggregate: vals => _.size(vals),
+    filterable: false
   },
   {
     Header: "Gene Family",
     accessor: "gene_family",
-    aggregate: vals => _.size(vals)
+    aggregate: vals => _.size(vals),
+    filterable: false
   },
   {
     Header: "Coverage",
     accessor: "coverage",
+    style: {
+      textAlign: "center"
+    },
     aggregate: vals => _.round(_.mean(vals)),
-    filterMethod: (filter, row) => row[filter.id] >= filter.value
+    filterMethod: (filter, row) => row[filter.id] >= filter.value,
+    Cell: props => parseFloat(props.value).toFixed(1)
   },
   {
     Header: "Depth",
     accessor: "depth",
     aggregate: vals => _.round(_.sum(vals)),
-    filterMethod: (filter, row) => row[filter.id] >= filter.value
+    style: {
+      textAlign: "center"
+    },
+    filterMethod: (filter, row) => row[filter.id] >= filter.value,
+    Cell: props => parseFloat(props.value).toFixed(1)
   }
 ];
 
@@ -53,6 +64,14 @@ class AMRView extends React.Component {
       <div>
         <ReactTable
           filterable
+          defaultFilterMethod={(filter, row, column) => {
+            const id = filter.pivotId || filter.id;
+            return row[id] !== undefined
+              ? String(row[id])
+                  .toLowerCase()
+                  .includes(filter.value.toLowerCase())
+              : true;
+          }}
           data={data}
           columns={columns}
           defaultPageSize={5}
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -19,7 +19,6 @@ const columns = [
     aggregate: vals => _.size(vals)
   },
   {
-    // TODO: Make average for gene family level
     Header: "Coverage",
     accessor: "coverage",
     aggregate: vals => _.round(_.mean(vals)),
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -44,7 +44,6 @@ class AMRView extends React.Component {
       var key = Object.keys(this.state.data)[i];
       this.state.data[key].gene_family = this.state.data[key].gene.slice(0, 5); // first four characters of gene family
     }
-    console.log(this.state.data);
   }
 
   render() {
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -38,7 +38,7 @@ class AMRView extends React.Component {
     super(props);
     this.state = { data: props.amr ? props.amr : [] };
     if (this.state.data) {
-      // TODO: Use regex to parse out gene family, i.e. is [.]*_ -- everything until the underscore
+      // TODO: Talk with Chaz to determine how to improve gene family identification further.
       for (var i = 0; i < Object.keys(this.state.data).length; i++) {
         var key = Object.keys(this.state.data)[i];
         this.state.data[key].gene_family = this.state.data[key].gene.slice(
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -22,13 +22,13 @@ const columns = [
     // TODO: Make average for gene family level
     Header: "Coverage",
     accessor: "coverage",
-    aggregate: vals => _.sum(vals),
+    aggregate: vals => _.round(_.mean(vals)),
     filterMethod: (filter, row) => row[filter.id] >= filter.value
   },
   {
     Header: "Depth",
     accessor: "depth",
-    aggregate: vals => _.sum(vals),
+    aggregate: vals => _.round(_.sum(vals)),
     filterMethod: (filter, row) => row[filter.id] >= filter.value
   }
 ];
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -1,0 +1,73 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import ReactTable from "react-table";
+import "react-table/react-table.css";
+import _ from "lodash";
+
+const columns = [
+  {
+    Header: "Antibiotic Class",
+    accessor: "drug_family",
+    PivotValue: ({ value }) => (
+      <span style={{ color: "darkblue" }}>{value}</span>
+    )
+  },
+  {
+    Header: "Gene",
+    accessor: "gene",
+    aggregate: vals => _.size(vals)
+  },
+  {
+    Header: "Gene Family",
+    accessor: "gene_family",
+    aggregate: vals => _.size(vals)
+  },
+  {
+    Header: "Coverage",
+    accessor: "coverage",
+    aggregate: vals => _.sum(vals),
+    filterMethod: (filter, row) => row[filter.id] >= filter.value
+  },
+  {
+    Header: "Depth",
+    accessor: "depth",
+    aggregate: vals => _.sum(vals),
+    filterMethod: (filter, row) => row[filter.id] >= filter.value
+  }
+];
+
+class AMRView extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { data: props.amr };
+    for (var z = 0; z < Object.keys(this.state.data).length; z++) {
+      var key = Object.keys(this.state.data)[z];
+      this.state.data[key].gene_family = this.state.data[key].gene.slice(0, 5); // first four characters of gene family
+    }
+    console.log(this.state.data);
+  }
+
+  render() {
+    const { data } = this.state;
+    return (
+      <div>
+        <ReactTable
+          filterable
+          data={data}
+          columns={columns}
+          defaultPageSize={5}
+          pivotBy={["drug_family", "gene_family"]}
+          defaultSorted={[
+            {
+              id: "coverage",
+              desc: true
+            }
+          ]}
+          className="-striped -highlight"
+        />
+        <br />
+      </div>
+    );
+  }
+}
+export default AMRView;
--- a/app/assets/src/components/AMRView.jsx
+++ b/app/assets/src/components/AMRView.jsx
@@ -36,7 +36,7 @@ const columns = [
 class AMRView extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { data: props.amr };
+    this.state = { data: props.amr ? props.amr : [] };
     if (this.state.data) {
       // TODO: Use regex to parse out gene family, i.e. is [.]*_ -- everything until the underscore
       for (var i = 0; i < Object.keys(this.state.data).length; i++) {
--- a/app/assets/src/components/PipelineSampleReads.jsx
+++ b/app/assets/src/components/PipelineSampleReads.jsx
@@ -32,7 +32,6 @@ class PipelineSampleReads extends React.Component {
     this.reportDetails = props.reportDetails;
     this.pipelineRunRetriable = props.pipelineRunRetriable;
     this.pipelineVersions = props.pipeline_versions;
-    this.pipelineVersion = props.pipeline_version;
 
     this.jobStatistics = props.jobStatistics;
     this.summary_stats = props.summary_stats;
@@ -841,7 +840,7 @@ class PipelineSampleReads extends React.Component {
     let amr_tab = show_amr ? (
       <li className="tab">
         <a href="#amr" className="">
-          Antimicrobial Resistance Report
+          Antimicrobial Resistance
         </a>
       </li>
     ) : null;
--- a/app/assets/src/components/PipelineSampleReads.jsx
+++ b/app/assets/src/components/PipelineSampleReads.jsx
@@ -32,6 +32,7 @@ class PipelineSampleReads extends React.Component {
     this.reportDetails = props.reportDetails;
     this.pipelineRunRetriable = props.pipelineRunRetriable;
     this.pipelineVersions = props.pipeline_versions;
+    this.pipelineVersion = props.pipeline_version;
 
     this.jobStatistics = props.jobStatistics;
     this.summary_stats = props.summary_stats;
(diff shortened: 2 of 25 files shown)


app/controllers/samples_controller.rb
```diff
@@ -188,6 +188,11 @@ def show
     end
   end
 
+  def show_amr_counts
+    @sample = samples_scope.find(params[:id])
```
cdebourcy: Since you listed show_amr_counts as a read_action (which makes sense), I think you should replace `samples_scope` by `current_power.samples`. As you can see in the definition of samples_scope, it selects just the "updatable" samples (tighter access control appropriate for edit actions), but what we want is instead all the "viewable" samples, which are accessible as `current_power.samples`. If you inspect `app/models/power.rb`, you'll see that the latter expression through the `viewable` method implemented in the Sample model.

app/lib/dags/host_filter.json.erb
```diff
@@ -5,6 +5,7 @@
     <% if @attribute_dict[:fastq2] %>
       "fastqs": ["<%= @attribute_dict[:fastq1] %>", "<%= @attribute_dict[:fastq2] %>"],
       "star_out": ["unmapped1.<%= @attribute_dict[:file_type] %>", "unmapped2.<%= @attribute_dict[:file_type] %>"],
+      "srst2_out": ["output.log", "output__genes__ARGannot_r2__results.txt", "output__fullgenes__ARGannot_r2__results.txt"],
```
cdebourcy: Since this line is the same in both the `if` and the `else` clause, you should take it out of the conditional to make the code more maintainable. Otherwise, any change to srst2 output files will require 2 lines of code to be modified instead of 1.

app/models/pipeline_run.rb
```diff
@@ -26,9 +28,11 @@ class PipelineRun < ApplicationRecord
   PIPELINE_VERSION_FILE = "pipeline_version.txt".freeze
   STATS_JSON_NAME = "stats.json".freeze
   ERCC_OUTPUT_NAME = 'reads_per_gene.star.tab'.freeze
+  AMR_OUTPUT_NAME = 'output__fullgenes__ARGannot_r2__results.txt'.freeze
```
cdebourcy: Since you define this constant here, you should reuse it in host_filter.json.erb, rather than typing it out again. That way, the consistency of the filename you need in pipeline_run.rb and the filename that gets generated in idseq-dag is always guaranteed. It's more maintainable if you only need to modify the filename in 1 constant, rather than 2 places of the codebase.

app/models/pipeline_run.rb
```diff
@@ -26,9 +28,11 @@ class PipelineRun < ApplicationRecord
...
   TAXID_BYTERANGE_JSON_NAME = 'taxid_locations_combined.json'.freeze
   ASSEMBLY_STATUSFILE = 'job-complete'.freeze
   LOCAL_JSON_PATH = '/app/tmp/results_json'.freeze
+  LOCAL_TXT_PATH = '/app/tmp/results_txt'.freeze
```
cdebourcy: I know you just wrote something analogous to what already existed (LOCAL_JSON_PATH), but I actually think it would be better to name it something more descriptive, such as `LOCAL_AMR_PATH = '/app/tmp/amr_results'.freeze`.

app/models/pipeline_run.rb
```diff
@@ -276,6 +281,30 @@ def db_load_ercc_counts
...
   end
 
+  def download_amr_results
+    amr_s3_path = "#{host_filter_output_s3_path}/#{AMR_OUTPUT_NAME}"
+    amr_results = PipelineRun.download_file(amr_s3_path, local_txt_path)
+    amr_results
+  end
+
+  def db_load_amr_counts
+    amr_results = download_amr_results
```
cdebourcy: I'd rename `amr_results` to indicate that it is a file path.

Also, the fact that the code for downloading this file is spread over 2 different functions makes it a bit confusing to see what's going on. I'd almost just write 
```
amr_results_file = PipelineRun.download_file("#{host_filter_output_s3_path}/#{AMR_OUTPUT_NAME}",
  "#{LOCAL_AMR_PATH}/#{id}")
```
and remove the functions `local_txt_path` and `download_amr_results`. That way it is obvious what we're doing: download the AMR output from S3 to a local folder unique to the pipeline_run id. [If `local_txt_path` were to stay, it should definitely be renamed to something that indicates its purpose, which is to make sure we have a unique folder for each pipeline_run id.]

Even better, write:
```
amr_results_file = PipelineRun.download_file(s3_file_for("amr_counts"), "#{LOCAL_AMR_PATH}/#{id}")
```
so that you don't have to write `"#{host_filter_output_s3_path}/#{AMR_OUTPUT_NAME}"` out twice (again, this makes it more maintainable).

app/views/samples/show_amr_counts.html.erb
```diff
@@ -0,0 +1,45 @@
+<p>
```
cdebourcy: We'll work more on this display later -- good enough for now

app/views/samples/show_amr_counts.html.erb
```diff
@@ -0,0 +1,45 @@
...
+            <td> <%= amr_count.allele  %></td>
+            <td><%= amr_count.coverage %> &nbsp;</td>
+            <td><%= amr_count.depth %> &nbsp;</td>
+          </tr>
+        <% end %>
+      </tbody>
+</table>
+
+<p>
+<%= link_to 'See all samples', samples_path %> |
```
cdebourcy: What happens when you click on this? Not sure it is a page we'd want the user to see. Could change so that it takes you back to this specific sample's report page?

Meera-Srinivasan:
Thanks for the comments @cdebourcy ! Will change + add.

app/models/pipeline_run.rb
Meera-Srinivasan: sounds good

db/schema.rb
```diff
@@ -25,6 +25,16 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "amr_counts", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
```
cdebourcy: make sure you add a unique index on [pipeline_run_id, allele] so as to avoid your earlier issue of having duplicate entries. (See https://github.com/chanzuckerberg/idseq-web/blob/b8e5edde16c9d4d9bfca828870699ea30cafcb8e/db/schema.rb#L331 -- do it using an add_index migration)

app/models/pipeline_run.rb
Meera-Srinivasan: Resolved -- from a testing perspective, though, I thought perhaps two functions would be easier? i.e. can test downloading separately that way. But this definitely makes sense from a readability perspective and have made this change!

app/controllers/samples_controller.rb
Meera-Srinivasan: Sounds good. I'm going to work on rewriting the display end for amr so hopefully I can just get rid of this soon -- will keep it in mind as I'm making my next set of display related changes.

app/models/pipeline_run.rb
Meera-Srinivasan: for sure, I don't think the constant was recognized when I moved it to the json.erb? Did <%= @AMR_OUTPUT_NAME%> (now this constant is called some thing else, but still)

cdebourcy: You can access it from outside the class using `PipelineRun:: AMR_OUTPUT_NAME`

app/lib/dags/host_filter.json.erb
Meera-Srinivasan: for sure, TODO: remove from here

app/controllers/samples_controller.rb
cdebourcy: Actually never mind. You should remove that line entirely, because `set_sample` is listed as a `before_action` for all `READ_ACTIONS`. So `@sample = samples_scope.find(params[:id])` happens automatically when show_amr_counts is called (just before)

app/lib/dags/experimental.json.erb
```diff
@@ -0,0 +1,32 @@
+{
+  "output_dir_s3":
+    "s3://<%= @attribute_dict[:bucket] %>/samples/<%= @project_id %>/<%= @sample_id %>/expt",
+  "targets": {
+    <% if @attribute_dict[:fastq2] %>
+      "fastqs": ["<%= @attribute_dict[:fastq1] %>", "<%= @attribute_dict[:fastq2] %>"],
+      "srst2_out": ["output.log", "output__genes__ARGannot_r2__results.txt", "output__fullgenes__ARGannot_r2__results.txt", "amr_processed_results.csv", "amr_summary_results.csv"]
```
cdebourcy: same comment about deduplicating by moving the `srst2_out` line after the if-else block

app/models/pipeline_run.rb
```diff
@@ -775,6 +844,13 @@ def subsample_suffix
 
   delegate :sample_output_s3_path, to: :sample
 
+  def expt_output_s3_path
```
cdebourcy: If it works and you have time, this would be a good opportunity to dedup and refactor some code.  `alignment_output_s3_path`, `postprocess_output_s3_path` and now `expt_output_s3_path` all contain essentially the same code. So you could make a helper function to which you would pass `sample.sample_expt_s3_path` as an argument

app/models/pipeline_run_stage.rb
```diff
@@ -269,8 +270,28 @@ def postprocess_command
...
+    batch_command = [install_pipeline, upload_version, dag_commands].join("; ")
+
+    # Dispatch job
+    memory = sample.sample_memory.present? ? sample.sample_memory : Sample::DEFAULT_MEMORY_IN_MB
+    aegea_batch_submit_command(batch_command, memory)
+  end
+
   def assembly_command
     # TODO: Change the following to DAG
+    # TODO: DOES this have anything to do with experimental stage
```
cdebourcy: no, it's deprecated and being removed

db/migrate/20180808030435_add_drug_gene_hits_to_amr_counts.rb
```diff
@@ -0,0 +1,5 @@
+class AddDrugGeneHitsToAmrCounts < ActiveRecord::Migration[5.1]
+  def change
+    add_column :amr_counts, :drug_gene_hits, :integer
```
cdebourcy: this is good for now until you get everything working. But before merging to master, we should consolidate all of these migrations into a single one. Probably inside create_amr_counts. You can rollback your local migrations that have already been applied using `rails db:rollback` multiple times. Then edit the file `db/migrate/20180726220853_create_amr_counts.rb` to add these additional columns. Then `rails db:migrate` again.
You should know that there was no need to do `rails generate migration` multiple times -- you could have done it ones and then edited the migration file to add more `add_column` statements

app/models/pipeline_run_stage.rb
Meera-Srinivasan: sounds good

db/migrate/20180808030435_add_drug_gene_hits_to_amr_counts.rb
Meera-Srinivasan: Great, thanks for that suggestion !

app/models/pipeline_run.rb
Meera-Srinivasan: Hmm I got a JSON format error when doing it that way -- will mark this as a TODO and figure out how to get it in there.

app/models/pipeline_run.rb
```diff
@@ -276,6 +291,48 @@ def db_load_ercc_counts
     update(total_ercc_reads: ercc_counts_array.map { |entry| entry[:count] }.sum)
   end
 
+  def db_load_amr_counts
+    # test files: "s3://idseq-database/test/AMR/amr_processed_results.csv", "s3://idseq-database/test/AMR/amr_summary_results.csv"
+    amr_results = PipelineRun.download_file(s3_file_for("amr_full_results"), local_amr_full_results_path)
```
cdebourcy: We noticed that loading failed because amr_results was nil. Given that this function only gets called once `check_and_enqueue` detects that the S3 file is present, my only explanation is that `PipelineRun.download_file` randomly failed. My recommendation is: change it to use `download_file_with_retries`, and return early from `db_load_amr_counts` if `amr_results` is still nil. It should then keep doing `db_load_amr_counts` every couple minutes until it succeeds in downloading the file and `amr_results` is not nil. Only then should it move on to the next block.

cdebourcy: Try that and see how it goes.

cdebourcy: You'll notice that `db_load_taxon_counts` functions in the way I just described

db/migrate/20180726220853_create_amr_counts.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.float :coverage
+      t.float :depth
+      t.bigint :pipeline_run_id
+      t.string :drug_family
+      t.integer :level
+      t.float :drug_gene_coverage
+      t.float :drug_gene_depth
+      t.timestamps
+    end
+    add_index :amr_counts, :allele, unique: true
```
cdebourcy: This isn't the right unique index to add. This would mean there could only be 1 entry for each allele in the entire table. Instead, it should be restricted to one entry per allele PER PIPELINE RUN. So: `add_index :amr_counts, [:pipeline_run_id, :allele], unique: true`

Meera-Srinivasan: thanks!

app/assets/src/components/AMRView.jsx
```diff
@@ -0,0 +1,74 @@
...
+    accessor: "depth",
+    aggregate: vals => _.sum(vals),
+    filterMethod: (filter, row) => row[filter.id] >= filter.value
+  }
+];
+
+class AMRView extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { data: props.amr ? props.amr : [] };
```
cdebourcy: Why is the data put in `state`? Is it a requirement of ReactTable? Normally `state` is used for variables that will change during the session. The AMR data on the other hand seems static.

app/models/pipeline_run.rb
```diff
@@ -278,6 +293,52 @@ def db_load_ercc_counts
     update(total_ercc_reads: ercc_counts_array.map { |entry| entry[:count] }.sum)
   end
 
+  def db_load_amr_counts
+    # test files: "s3://idseq-database/test/AMR/amr_processed_results.csv", "s3://idseq-database/test/AMR/amr_summary_results.csv"
+    # amr_results = PipelineRun.download_file_with_retries("s3://idseq-database/test/AMR/amr_processed_results.csv", local_amr_full_results_path, 3)
+    amr_results = PipelineRun.download_file_with_retries(s3_file_for("amr_full_results"), local_amr_full_results_path, 3)
```
cdebourcy: Problem is that you have an inconsistency in the name of the output. In `s3_file_for`, you define the path for `amr_full_results` (https://github.com/chanzuckerberg/idseq-web/blob/157a91386f645c5139b65ba655b5d60b59ea8866/app/models/pipeline_run.rb#L420), but in output_states it is `amr_counts` (https://github.com/chanzuckerberg/idseq-web/blob/157a91386f645c5139b65ba655b5d60b59ea8866/app/models/pipeline_run.rb#L183). They need to be the same. Your loading errors stem from the fact that the check for the presence of `s3_file_for("amr_counts")` succeeded but downloading `s3_file_for("amr_full_results")` failed. Once you make the names consistent, it won't try to load until the file is ACTUALLY there

app/assets/src/components/AMRView.jsx
Meera-Srinivasan: That's true, I can look into it more and get back to you.

app/models/pipeline_run.rb
Meera-Srinivasan: I see, thanks for pointing this out.

app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -868,12 +869,22 @@ class PipelineSampleReads extends React.Component {
                       Details
                     </a>
                   </li>
+                  <li className="tab">
```
cdebourcy: check here whether user is admin. If yes, show your <li>. If no, show null.

app/helpers/pipeline_runs_helper.rb
```diff
@@ -4,7 +4,7 @@ module PipelineRunsHelper
   def aegea_batch_submit_command(base_command,
                                  memory = Sample::DEFAULT_MEMORY_IN_MB,
                                  job_queue: nil,
-                                 docker_image: "idseq_dag")
+                                 docker_image: "idseq_dag_meera1")
```
cdebourcy: Don't change the default here -- call the function with argument "idseq_dag_meera1" in your pipeline_run_stage command

Meera-Srinivasan: resolved

app/models/pipeline_run_stage.rb
```diff
@@ -212,4 +213,33 @@ def postprocess_command
...
+    file_type = sample.fasta_input? ? 'fasta' : 'fastq'
+    attribute_dict = {
+      fastq1: sample.input_files[0].name,
+      file_type: file_type
+    }
+    attribute_dict[:fastq2] = sample.input_files[1].name if sample.input_files[1]
+    dag_commands = prepare_dag("experimental", attribute_dict)
+
+    # upload_version = "idseq_dag --version | cut -f2 -d ' ' | aws s3 cp  - #{pipeline_run.pipeline_version_file} "
+    batch_command = [install_pipeline(pipeline_run.pipeline_commit), dag_commands].join("; ")
```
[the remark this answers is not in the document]
Meera-Srinivasan: resolved

app/models/pipeline_run_stage.rb
```diff
@@ -212,4 +213,33 @@ def postprocess_command
...
+
+    # Dispatch job
+    # memory = sample.sample_memory.present? ? sample.sample_memory : Sample::DEFAULT_MEMORY_IN_MB
+    aegea_batch_submit_command(batch_command, job_queue: pipeline_run.sample.job_queue)
```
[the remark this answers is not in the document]
Meera-Srinivasan: resolved

app/controllers/samples_controller.rb
```diff
@@ -150,7 +150,9 @@ def report_csv
   # GET /samples/1.json
 
   def show
+    @sample = samples_scope.find(params[:id])
```
yunfangjuan: I don't think you need this. set_sample should be run by default.

app/models/amr_count.rb
```diff
@@ -0,0 +1,3 @@
+class AmrCount < ApplicationRecord
```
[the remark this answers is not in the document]
yunfangjuan: fix this?

app/models/amr_count.rb
```diff
@@ -0,0 +1,14 @@
...
+  belongs_to :pipeline_run
+  def self.viewable(user)
+    if user.admin?
+      all
+    end
```
yunfangjuan: so the functions here are not implemented?

yunfangjuan: If you are not calling these functions, I don't think you need to implement them here.

app/models/sample.rb
```diff
@@ -292,6 +292,10 @@ def sample_postprocess_s3_path
     "s3://#{SAMPLES_BUCKET_NAME}/#{sample_path}/postprocess"
   end
 
+  def sample_expt_s3_path
```
yunfangjuan: add some comments so people understand expt means experimental.

app/models/sample.rb
```diff
@@ -466,7 +470,7 @@ def kickoff_pipeline
     # The subsample field of "sample" is currently used as a simple flag (UI checkbox),
     # but was made an integer type in case we want to allow users to enter the desired number
     # of reads to susbample to in the future
-    pr.pipeline_branch = pipeline_branch.blank? ? "master" : pipeline_branch
+    pr.pipeline_branch = pipeline_branch.blank? ? "meera-add-amr" : pipeline_branch
```
yunfangjuan: this is dangerous. please change back to master.

app/models/pipeline_run_stage.rb
```diff
@@ -212,4 +213,33 @@ def postprocess_command
...
+  end
+
+  def assembly_command
```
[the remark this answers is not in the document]
Meera-Srinivasan: resolved

app/controllers/samples_controller.rb
Meera-Srinivasan: Resolved

app/models/sample.rb
Meera-Srinivasan: resolved

app/models/sample.rb
Meera-Srinivasan: resolved

app/models/amr_count.rb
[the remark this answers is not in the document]
Meera-Srinivasan: resolved

app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -30,6 +32,7 @@ class PipelineSampleReads extends React.Component {
     this.reportDetails = props.reportDetails;
     this.pipelineRunRetriable = props.pipelineRunRetriable;
     this.pipelineVersions = props.pipeline_versions;
+    this.pipelineVersion = props.pipeline_version;
```
cdebourcy: remove this?

app/controllers/samples_controller.rb
```diff
@@ -189,6 +196,10 @@ def show
     end
   end
 
+  def show_amr_counts
```
cdebourcy: Remove this?

app/models/pipeline_run.rb
```diff
@@ -28,9 +30,14 @@ class PipelineRun < ApplicationRecord
   PIPELINE_VERSION_FILE = "pipeline_version.txt".freeze
   STATS_JSON_NAME = "stats.json".freeze
   ERCC_OUTPUT_NAME = 'reads_per_gene.star.tab'.freeze
+  AMR_DRUG_SUMMARY_RESULTS = 'amr_summary_results.csv'.freeze
+  AMR_FULL_RESULTS_NAME = 'amr_processed_results.csv'.freeze
   TAXID_BYTERANGE_JSON_NAME = 'taxid_locations_combined.json'.freeze
   ASSEMBLY_STATUSFILE = 'job-complete'.freeze
   LOCAL_JSON_PATH = '/app/tmp/results_json'.freeze
+  LOCAL_AMR_FULL_RESULTS_PATH = '/app/tmp/amr_full_results'.freeze
+  # Note -- not currently being used, but leaving here for reference
```
cdebourcy: If not used, please remove it

app/models/pipeline_run.rb
```diff
@@ -285,6 +300,41 @@ def db_load_ercc_counts
...
+                              depth:  amr_result_fields[3],
+                              drug_family: amr_result_fields[12] }
+      end
+      update(amr_counts_attributes: amr_counts_array)
+    end
+  end
+
+  # Can remove as aggregation is being done on the UI side,
+  # but leaving as useful function for possible future changes (e.g. if UI table)
+  def summarize_amr_drug_family_counts
```
cdebourcy: Since it is not being used, please remove it

app/assets/src/components/AMRView.jsx
tfrcarvalho: As recommended by @cdebourcy, you should move the data out of state as you are not using it to influence any local elements.
This could even be a functional component.

app/views/samples/show_amr_counts.html.erb
```diff
@@ -0,0 +1,7 @@
+<div id="AMRView_page">
```
cdebourcy: Since we have the tab now, we don't need this page anymore. Please remove it

config/routes.rb
```diff
@@ -12,6 +12,7 @@
     put :retry_pipeline, on: :member
     get :all, on: :collection
     get :pipeline_runs, on: :member
+    get :show_amr_counts, on: :member
```
cdebourcy: remove this

app/controllers/samples_controller.rb
```diff
@@ -14,7 +14,7 @@ class SamplesController < ApplicationController
   ##########################################
   skip_before_action :verify_authenticity_token, only: [:create, :update]
 
-  READ_ACTIONS = [:show, :report_info, :search_list, :report_csv, :assembly, :show_taxid_fasta, :nonhost_fasta, :unidentified_fasta, :results_folder, :fastqs_folder, :show_taxid_alignment, :show_taxid_alignment_viz].freeze
+  READ_ACTIONS = [:show, :report_info, :search_list, :report_csv, :assembly, :show_taxid_fasta, :nonhost_fasta, :unidentified_fasta, :results_folder, :fastqs_folder, :show_taxid_alignment, :show_taxid_alignment_viz, :show_amr_counts].freeze
```
cdebourcy: remove this change

app/models/pipeline_run_stage.rb
```diff
@@ -212,4 +213,23 @@ def postprocess_command
...
+    sample = pipeline_run.sample
+    file_type = sample.fasta_input? ? 'fasta' : 'fastq'
+    attribute_dict = {
+      fastq1: sample.input_files[0].name,
+      file_type: file_type
+    }
+    attribute_dict[:fastq2] = sample.input_files[1].name if sample.input_files[1]
+    dag_commands = prepare_dag("experimental", attribute_dict)
+
+    # upload_version = "idseq_dag --version | cut -f2 -d ' ' | aws s3 cp  - #{pipeline_run.pipeline_version_file} "
```
cdebourcy: remove this commented line

app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -834,6 +836,21 @@ class PipelineSampleReads extends React.Component {
           </div>
         </div>
       );
+
+    let show_amr = this.amr != null;
+    let amr_tab = show_amr ? (
+      <li className="tab">
+        <a href="#amr" className="">
+          Antimicrobial Resistance Report
```
cdebourcy: This is a bit wordy. Change it to just "Antimicrobial Resistance"

app/assets/src/components/AMRView.jsx
cdebourcy: So for now, just literally change `this.state` to `this.data` or whatever

Meera-Srinivasan: Sounds good, resolved

app/assets/src/components/PipelineSampleReads.jsx
Meera-Srinivasan: resolved

app/controllers/samples_controller.rb
Meera-Srinivasan: done

config/routes.rb
Meera-Srinivasan: done

app/controllers/samples_controller.rb
Meera-Srinivasan: done

app/models/pipeline_run.rb
Meera-Srinivasan: done

app/models/pipeline_run.rb
Meera-Srinivasan: done

app/views/samples/show_amr_counts.html.erb
Meera-Srinivasan: done

app/models/pipeline_run_stage.rb
Meera-Srinivasan: done

app/assets/src/components/PipelineSampleReads.jsx
Meera-Srinivasan: done

PR #1407, opened 2018-08-02
Model scoring functions

yunfangjuan:
# Description

Implementing a representation of scoring models to easily integrate katrina's scoring model. 

## Type of change

- [ X] New feature (non-breaking change which adds functionality)

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [X ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

### DIFF
--- a/app/lib/report_helper.rb
+++ b/app/lib/report_helper.rb
@@ -22,7 +22,7 @@ module ReportHelper
   # the fake genus IDs start here:
   FAKE_GENUS_BASE = -1_900_000_000
 
-  DECIMALS = 1
+  DECIMALS = 7
 
   DEFAULT_SORT_PARAM = 'highest_nt_aggregatescore'.freeze
   DEFAULT_TAXON_SORT_PARAM = 'highest_nt_aggregatescore'.freeze
@@ -34,7 +34,7 @@ module ReportHelper
 
   SORT_DIRECTIONS = %w[highest lowest].freeze
   # We do not allow underscores in metric names, sorry!
-  METRICS = %w[r rpm zscore percentidentity alignmentlength neglogevalue percentconcordant aggregatescore maxzscore].freeze
+  METRICS = %w[r rpm zscore percentidentity alignmentlength neglogevalue percentconcordant aggregatescore maxzscore r_pct].freeze
   COUNT_TYPES = %w[NT NR].freeze
   # Note: no underscore in sortable column names. Add to here to protect from data cleaning.
   PROPERTIES_OF_TAXID = %w[tax_id name common_name tax_level species_taxid genus_taxid family_taxid superkingdom_taxid category_name is_phage].freeze
@@ -203,7 +203,7 @@ module ReportHelper
   def fetch_taxon_counts(pipeline_run_id, background_id)
     pipeline_run = PipelineRun.find(pipeline_run_id)
     adjusted_total_reads = (pipeline_run.total_reads - pipeline_run.total_ercc_reads.to_i) * pipeline_run.subsample_fraction
-    raw_non_host_reads = pipeline_run.adjusted_remaining_reads * pipeline_run.subsample_fraction
+    raw_non_host_reads = pipeline_run.adjusted_remaining_reads.to_f * pipeline_run.subsample_fraction
 
     # NOTE:  If you add more columns to be fetched here, you really should add them to PROPERTIES_OF_TAXID above
     # otherwise they will not survive cleaning.
--- a/app/lib/report_helper.rb
+++ b/app/lib/report_helper.rb
@@ -852,7 +852,7 @@ module ReportHelper
       next unless species_info['tax_level'] == TaxonCount::TAX_LEVEL_SPECIES
       genus_id = species_info['genus_taxid']
       genus_info = tax_2d[genus_id]
-      taxon_info = TaxonScoringModel.flatten_taxon("genus" => genus_info, "species" => species_info)
+      taxon_info = { "genus" => genus_info, "species" => species_info }
       species_score = tsm.score(taxon_info)
       species_info['NT']['aggregatescore'] = species_score.to_f
       species_info['NR']['aggregatescore'] = species_score.to_f
--- a/app/lib/report_helper.rb
+++ b/app/lib/report_helper.rb
@@ -834,6 +834,7 @@ module ReportHelper
       species_score = species_info['NT']['aggregatescore']
       genus_score = genus_info['NT']['aggregatescore']
       unless genus_score && genus_score > species_score
+        # genus aggregate score is the max of the specifies scores
         genus_info['NT']['aggregatescore'] = species_score.to_f
         genus_info['NR']['aggregatescore'] = species_score.to_f
       end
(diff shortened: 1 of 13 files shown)


app/models/taxon_scoring_model.rb
```diff
@@ -0,0 +1,96 @@
...
+#          "op": "abs",
+#          "on": { "op": "attr", "on": "genus.NR.zscore" }
+#        },
+#        { "op": "attr", "on": "species.NR.zscore" },
+#        { "op": "attr", "on": "species.NR.rpm" }
+#      ]
+#    }
+#  ]
+# }
+class TaxonScoringModel < ApplicationRecord
```
tfrcarvalho: Is this tree model generic enough? I see this as an unnecessary complexity. 

Why not just have a ruby library of scoring functions and store the name of the scoring function?
Seems simpler to me and provide way more flexibility.

yunfangjuan: yes. it's very generic. it's a general representation of any arithmatic formula

yunfangjuan: i would imagine people can plug in their function down the road and we can just store their formula in the database. also, katrina's model probably will change all the time, i don't want to change the code every time she (or anyone) updates the model.

boris-dimitrov: Some random thoughts or suggestions, mostly just to see what you think about those, not really pushing to get you to change this.

1) Can we store a lisp function in the database instead of a json abstract syntax tree?   Then we can use https://github.com/maryrosecook/littlelisp to run the stored lisp function in the browser or https://github.com/mooreryan/lisby to run the same lisp function in Ruby.

2) I am afraid to ask about performance here, but I see several approaches that could work to improve performance:
     a) The ruby backend could generate javascript that computes the formula, then ship that javascript to the browser, where it would get compiled and executed fast.
     b) Experimenting with novel formulas would be expected to run slow, but once the formulas are vetted and blessed, we can add ruby/js code for them, so they run natively and bypass this AST interpreter.

Overall I am slightly concerned that if we reinvent a lot of things then we might incur some maintenance cost -- but on the other hand if this lets you move faster and experiment faster in the product, we can have it this way in the short term and eventually replace with a more standard solution.

boris-dimitrov: Oh btw if we go the way of custom formulas, we should also do custom filters, all the way :)

app/models/taxon_scoring_model.rb
```diff
@@ -0,0 +1,96 @@
...
+    when "const"
+      return score_func["on"].to_f
+    when "attr"
+      return taxon[score_func["on"]].to_f
+    else
+      # atomic transformation function like abs, sqrt, log10
+      result = score(taxon, score_func["on"])
+      begin
+        # eg. result.abs
+        return result.send(op)
```
boris-dimitrov: Could you validate that the ops are safe before you do send?   I actually thought this ability to validate, for security purposes, was the chief advantage of interpreting the json AST formula yourself, rather than just getting a big math formula and sending it all to ruby.  If this is addressed, I think I'd be happy to approve the pull request.

app/models/taxon_scoring_model.rb
boris-dimitrov: FWIW, interana had its own similar interpreter for query filters represented in a custom format (i think that was reverse polish and not json ast, but same difference);  That interpreter was written in C++, but eventually had to be replaced with on-demand LLVM compiler so the filters could run natively (much faster).  Here we have a lot less data to run this on, up to a few million rows perhaps?   So if this AST evaluator was implemented in C++, I wouldn't raise a  perf concern.   Only because ruby is already slow to begin with, before we ask it to eval a tree, am I slightly concerned --- but not overly concerned since this may still not be the slowest bit.  Just worth noting/measuring perhaps.

tfrcarvalho: Up to you...
But I would not call it generic, if it constrains you to arithmetic (eventually logic) functions. And in this case introduces unnecessary complexity to constrain what you can use. 
I guess it is for making customization of scoring more accessible to users...

If you want to follow this path anyway, I also agree with Boris that we should not reinvent the wheel: there are solutions out there with multi operations and multi language support, e.g. http://jsonlogic.com/

Also, could this be ran as part of the pipeline, as a post processing step? If it is run on demand as part of the web, every bit counts.

yunfangjuan: This thing has to be done on the server side. (There might be some privacy/security issue of exposing it on the client side.) And we can't do it on the pipeline side, because it changes dynamically. 

We have to do it on the web part. I did a similar implementation in C++ before for delopying (generalized) linear models so I think this will be more than sufficient after adding a bit more features ( '/', identity functions etc). It's generic in a way that we dont have to change code every single time people update the model.  Sure if we were to do tree ensembles or deep learning models, we will have to do something else but for now and at our stage, i believe this is more than enough.

yunfangjuan: exposing it : it means model coefficients.

app/models/taxon_scoring_model.rb
yunfangjuan: OK. I have issues with this too. I will try coming up a solution.

db/migrate/20180801183223_create_taxon_scoring_models.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTaxonScoringModels < ActiveRecord::Migration[5.1]
+  def change
+    create_table :taxon_scoring_models do |t|
```
tfrcarvalho: Would it make sense now to associate this with a user/owner? Would we ever expose this?
I kind of want to add that ability to all our content for accountability for security purposes..

app/lib/report_helper.rb
```diff
@@ -834,20 +834,23 @@ def compute_genera_aggregate_scores!(rows, tax_2d)
...
       end
     end
   end
 
   def compute_species_aggregate_scores!(rows, tax_2d)
+    tsm = TaxonScoringModel.find_by(name: TaxonScoringModel::DEFAULT_MODEL_NAME)
     rows.each do |species_info|
       species_info['NT']['maxzscore'] = [species_info['NT']['zscore'], species_info['NR']['zscore']].max
       species_info['NR']['maxzscore'] = species_info['NT']['maxzscore']
       next unless species_info['tax_level'] == TaxonCount::TAX_LEVEL_SPECIES
```
tfrcarvalho: Can you invert this to an if that encompasses the code below?
ifs always preferred to unless and continues.

app/models/taxon_scoring_model.rb
```diff
@@ -0,0 +1,96 @@
...
+    when "*"
+      result = 1.0
+      score_func["on"].each { |sf| result *= score(taxon, sf) }
+      return result
+    when "const"
+      return score_func["on"].to_f
+    when "attr"
+      return taxon[score_func["on"]].to_f
+    else
+      # atomic transformation function like abs, sqrt, log10
```
tfrcarvalho: Please check all supported functions explicitly.
The else should be a failure case (hopefully rarely or never triggered). All the others should be validated.

tfrcarvalho: This will solve your issue below...

app/models/taxon_scoring_model.rb
tfrcarvalho: Changes dynamically because of the background model?
We can compute that too...

db/migrate/20180801183223_create_taxon_scoring_models.rb
yunfangjuan: I can add that but probably won't fill it for a while.

PR #1408, opened 2018-08-02
Prototype for outbreak phylogenetic trees

cdebourcy:


app/assets/src/components/PhyloTree.jsx
```diff
@@ -0,0 +1,76 @@
+import axios from "axios";
+import React from "react";
+import { Button } from "semantic-ui-react";
+import PhyloTreeViz from "./PhyloTreeViz";
+
+class PhyloTree extends React.Component {
```
tfrcarvalho: Can you do a little refactoring for future compliance:
1) Create a folder `views/phylo_tree`
2) Move all phylo tree js files in there
3) Renave `PhyloTree` -> `PhyloTreeView`

app/assets/src/components/PhyloTreeList.jsx
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+class PhyloTreeList extends React.Component {
```
tfrcarvalho: We should probably have a component for "list views". Is this type of views used anywhere else already?

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -896,12 +903,14 @@ class PipelineSampleReport extends React.Component {
...
-      assemblyDot = (
+    if (taxInfo.tax_id > 0 && taxInfo.tax_level == 1 && taxInfo.NT.r > 0)
+      // right now, tree pipeline is only implemented for tax_level 1, but may change in future
+      phyloTreeDot = (
         <i
           data-tax-id={taxInfo.tax_id}
-          onClick={this.downloadAssemblyLink}
-          className="fa fa-gg"
+          data-tax-level={taxInfo.tax_level}
+          onClick={this.gotoTreeLink}
```
tfrcarvalho: It is better to have explicit calls: `onClick={()=> this.goToTreeLink(taxInfo)}`

app/assets/src/styles/phylo_tree_viz.scss
```diff
@@ -0,0 +1,105 @@
+.dropdown-menu {
```
tfrcarvalho: Also create same structure as for the js components: `styles/views/phylo_tree/`

app/views/phylo_trees/show.html.erb
```diff
@@ -0,0 +1,12 @@
+<div id="PhyloTree_page">
+  <%= javascript_tag do %>
+    react_component('PhyloTree', {
+    csrf: '<%= form_authenticity_token %>',
+    taxon: JSON.parse('<%= raw escape_json(@taxon) %>'),
+    project: JSON.parse('<%= raw escape_json(@project) %>'),
+    samples: JSON.parse('<%= raw escape_json(@samples) %>'),
+    pipeline_runs: JSON.parse('<%= raw escape_json(@pipeline_runs) %>'),
+    phylo_tree: JSON.parse('<%= raw escape_json(@phylo_tree) %>')
```
tfrcarvalho: Use camelCase on JSON too.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -896,12 +903,14 @@ class PipelineSampleReport extends React.Component {
           aria-hidden="true"
         />
       );
-    if (reportDetails.assembled_taxids.indexOf(taxInfo.tax_id.toString()) >= 0)
-      assemblyDot = (
+    if (taxInfo.tax_id > 0 && taxInfo.tax_level == 1 && taxInfo.NT.r > 0)
```
tfrcarvalho: As discussed, constrain to admin for now.

app/models/phylo_tree.rb
```diff
@@ -0,0 +1,154 @@
...
+  end
+
+  def fetch_dag_version
+    # Needed to build full idseq-dag output path.
+    return dag_version if dag_version.present?
+    # If not present yet, need to fetch it.
+    # TEMP HACK:
+    # Assume version folder is the only folder in phylo_trees folder.
+    # Right now, assume each tree can only run once, so only 1 version ever.
+    stdout, _stderr, status = Open3.capture3("aws s3 ls #{phylo_tree_output_s3_path}/ | grep PRE | awk '{print $2}' | head -n1")
```
yunfangjuan: this is a bit cryptic. example input and output?

app/models/phylo_tree.rb
```diff
@@ -0,0 +1,154 @@
...
+    job_status, self.job_log_id, _job_hash, self.job_description = PipelineRunStage.job_info(job_id, id)
+    if job_status == PipelineRunStage::STATUS_FAILED ||
+       (job_status == "SUCCEEDED" && !Open3.capture3("aws", "s3", "ls", newick_s3_path)[2].exitstatus.zero?)
+      self.status = STATUS_FAILED
+      save
+    end
+  end
+
+  def aegea_command(base_command)
+    "aegea batch submit --command=\"#{base_command}\" " \
```
yunfangjuan: this seems to be copied from somewhere?

cdebourcy: refactor - make common helper with pipeline_run

app/lib/dags/phylo_tree.json.erb
```diff
@@ -0,0 +1,37 @@
+{
+  "output_dir_s3": "<%= @attribute_dict[:phylo_tree_output_s3_path] %>",
+  "targets": {
+    "taxon_fasta_in": [
+      <% @attribute_dict[:taxon_fasta_files].each do |tff| %>
```
cdebourcy: move fasta generation inside dag step

app/models/phylo_tree.rb
```diff
@@ -0,0 +1,154 @@
...
+    " --storage /mnt=#{Sample::DEFAULT_STORAGE_IN_GB} --volume-type gp2 --ecr-image idseq_phylo --memory #{Sample::DEFAULT_MEMORY_IN_MB}" \
```
cdebourcy: put into same docker image

app/controllers/phylo_trees_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+    if project_id
+      @project = current_power.projects.find(project_id)
+      @phylo_trees = PhyloTree.where(project_id: project_id.to_i)
+    else
+      @project = []
+      @phylo_trees = PhyloTree.where(project_id: current_power.projects.pluck(:id))
+    end
+  end
+
+  def show
```
cdebourcy: separate into "new" and "show"

app/controllers/phylo_trees_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      pipeline_run_ids = params[:pipeline_run_ids].split(",").map(&:to_i)
+      pt = PhyloTree.create(taxid: taxid, tax_level: tax_level, tax_name: tax_name, user_id: current_user.id, project_id: @project.id, pipeline_run_ids: pipeline_run_ids)
+      Resque.enqueue(KickoffPhyloTree, pt.id)
+      render json: { status: :ok, message: "tree creation job submitted" }
+    end
+  end
+
+  private
+
+  def set_project
```
cdebourcy: can keep tree permissions at project level. E.g. I create a tree in my project, which is to include some public samples from another project

cdebourcy: have to have edit access to project to create tree in that project

app/assets/src/components/PipelineSampleReport.jsx
cdebourcy: ok

app/assets/src/components/PhyloTreeViz.jsx
```diff
@@ -0,0 +1,63 @@
+import axios from "axios";
+import d3 from "d3";
+import phylotree from "phylotree";
+import React from "react";
+
+class PhyloTreeViz extends React.Component {
```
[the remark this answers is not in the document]
cdebourcy: yes, have state now

app/assets/src/components/PhyloTreeViz.jsx
```diff
@@ -0,0 +1,63 @@
...
+  constructor(props) {
+    super();
+    this.phylo_tree = props.phylo_tree;
+    this.createTreeViz = this.createTreeViz.bind(this);
+  }
+
+  componentDidMount() {
+    this.createTreeViz();
+  }
+  componentDidUpdate() {
```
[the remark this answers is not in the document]
cdebourcy: no

app/assets/src/components/PhyloTreeList.jsx
cdebourcy: not really. Well, Samples.jsx is one, but it's not very reusable in this case

app/assets/src/components/PipelineSampleReport.jsx
cdebourcy: done

app/assets/src/components/PhyloTree.jsx
```diff
@@ -0,0 +1,76 @@
...
+  constructor(props) {
+    super();
+    this.csrf = props.csrf;
+    this.taxon = props.taxon;
+    this.project = props.project;
+    this.samples = props.samples;
+    this.pipeline_runs = props.pipeline_runs;
+    this.phylo_tree = props.phylo_tree;
```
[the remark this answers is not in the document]
cdebourcy: ok

app/views/phylo_trees/show.html.erb
cdebourcy: ok

app/assets/src/styles/views/phylo_tree/phylo_tree_viz.scss
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.tree-scale-bar line,
+.tree-scale-bar path {
+  fill: none;
+  stroke: #000;
+  shape-rendering: crispEdges;
+}
+
+.node circle,
```
tfrcarvalho: Can you add these bellow a tree's specific class, i.e. add a phylo-tree-view class to the div containing the tree? 
(add all the styling on this file as children of the class you use for the container div for you visualization).

For instace `.node circle` would be applied to all our d3 visualizations that have a node.

app/assets/src/styles/views/phylo_tree_viz.scss
```diff
@@ -0,0 +1,105 @@
+.dropdown-menu {
```
tfrcarvalho: This is the same file as above, no?

app/assets/src/components/views/phylo_tree/PhyloTreeViz.jsx
```diff
@@ -0,0 +1,96 @@
...
+        );
+      });
+  }
+
+  render() {
+    let status_display, newick, tree;
+    if (this.phyloTree) {
+      switch (this.phyloTree.status) {
+        case 1:
+          status_display = "TREE IS READY";
```
tfrcarvalho: Why the ALL caps?

test/controllers/power_controller_test.rb
```diff
@@ -248,4 +248,59 @@ class PowerControllerTest < ActionDispatch::IntegrationTest
       access_sample_with_background(backgrounds(:background_for_project_one), samples(:joe_sample))
     end
   end
+
+  # phylo_trees
+  test 'joe can see joe_phylo_tree' do
```
tfrcarvalho: Is using `joe` as test subject ID Seq a standard? :)

tfrcarvalho:
Please, also wait for @yunfangjuan approvals on the backend parts of the code.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -854,14 +856,16 @@ class PipelineSampleReport extends React.Component {
...
   }
 
-  downloadAssemblyLink(e) {
-    const taxId = e.target.getAttribute("data-tax-id");
-    location.href = `/samples/${this.sample_id}/assembly/${taxId}`;
+  gotoTreeLink(taxid) {
+    window.open(
+      `/phylo_trees/index?taxid=${taxid}&project_id=${this.projectId}`,
+      "_blank"
+    );
```
jshoe: add "noopener", "hide_referrer" after "_blank"

https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

app/assets/src/components/views/phylo_tree/PhyloTreeInputs.jsx
```diff
@@ -0,0 +1,237 @@
...
+    this.taxon = this.phyloTree
+      ? {
+          name: this.phyloTree.tax_name,
+          tax_level: this.phyloTree.tax_level,
+          taxid: this.phyloTree.taxid
+        }
+      : props.taxon;
+    this.project = props.project;
+    this.samples = props.samples;
+    this.MinReads = 5;
```
jshoe: nit: minReads ?

app/assets/src/components/views/phylo_tree/PhyloTreeInputs.jsx
```diff
@@ -0,0 +1,237 @@
+import axios from "axios";
+import React from "react";
+import { Input, Checkbox, Button, Accordion, Menu } from "semantic-ui-react";
+
+class PhyloTreeInputs extends React.Component {
```
jshoe: could you add a comment summarizing what a 'PhyloTreeInputs' represents

app/assets/src/components/views/phylo_tree/PhyloTreeInputs.jsx
```diff
@@ -0,0 +1,237 @@
...
+    let project_names = Object.keys(samples_by_project_name);
+
+    let sample_display_outside_project =
+      project_names.length === 0 ? null : (
+        <div>
+          <br />
+          <b>
+            {this.disabled
+              ? "Samples from other projects:"
+              : `You may wish samples from other projects that contain ${
```
jshoe: is this a typo? "You may wish samples"?

test/controllers/phylo_trees_controller_test.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  test 'should get index' do
+    get "/phylo_trees/index"
+    assert_response :success
+  end
+
+  test 'should get new' do
+    taxid = 573
+    get "/phylo_trees/new?taxid=#{taxid}&project_id=#{@project.id}"
```
jshoe: wait will this fail if 573 already exists?

jshoe:
- Looks pretty good to me although would've been easier to review in smaller chunks? (e.g. with dag templating re-arrangement broken off)
- nit on keeping JS naming camelCase w.r.t. https://czi.quip.com/c7sGAU2bOXVX/IDSeq-Frontend-Development-Guidelines

app/assets/src/components/Samples.jsx
```diff
@@ -2092,6 +2102,16 @@ function ProjectInfoHeading({
         {table_download_dropdown}
         {compare_button}
         <BackgroundModal parent={parent} />
+        {parent.admin ? (
+          <ActiveInactiveButton
```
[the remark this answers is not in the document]

app/assets/src/components/views/phylo_tree/PhyloTreeInputs.jsx
cdebourcy: sure

app/assets/src/components/views/phylo_tree/PhyloTreeInputs.jsx
cdebourcy: ok

app/assets/src/components/views/phylo_tree/PhyloTreeInputs.jsx
cdebourcy: forgot "to include"

app/assets/src/components/views/phylo_tree/PhyloTreeViz.jsx
cdebourcy: Because the status display on the front page is also all caps

app/assets/src/components/views/phylo_tree/PhyloTreeViz.jsx
```diff
@@ -0,0 +1,96 @@
...
+          break;
+        case 2:
+          status_display = "TREE GENERATION FAILED";
+          break;
+        default:
+          status_display = "TREE GENERATION IN PROGRESS";
+      }
+    }
+    let retry_button = (
+      <Button primary onClick={this.retryTree}>
```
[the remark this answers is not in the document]
cdebourcy: ok

app/assets/src/styles/views/phylo_tree/phylo_tree_viz.scss
cdebourcy: I'll leave that for you in your UI work. Since I assume you'll replace the tree viz anyway

app/assets/src/styles/views/phylo_tree_viz.scss
cdebourcy: thanks, removing

app/assets/src/components/views/phylo_tree/PhyloTreeViz.jsx
tfrcarvalho: You can leave it as is for now but, for future reference, this should be a styling decision, i.e. you write normally, and then on styling you set all caps.

app/models/phylo_tree.rb
```diff
@@ -0,0 +1,147 @@
...
+    return if throttle && rand >= 0.1 # if throttling, do time-consuming aegea checks only 10% of the time
+    job_status, self.job_log_id, _job_hash, self.job_description = job_info(job_id, id)
+    if job_status == PipelineRunStage::STATUS_FAILED ||
+       (job_status == "SUCCEEDED" && !Open3.capture3("aws", "s3", "ls", newick_s3_path)[2].exitstatus.zero?)
+      self.status = STATUS_FAILED
+      save
+    end
+  end
+
+  def job_command
```
boris-dimitrov: On which thread is this executed?   On a webapp request response thread, or on the monitor main thread?

boris-dimitrov: If monitor, then we should make sure it doesn't take too long.    If webapp, I don't care.

app/models/pipeline_run.rb
```diff
@@ -432,13 +434,7 @@ def monitor_results
...
     # Get pipeline_version, which determines S3 locations of output files.
     # If pipeline version is not present, we cannot load results yet.
-    # [ We use "file_generated_since_run(pipeline_version_file)" because "pipeline_version_file"
-    #   is not in the versioned result folder, so it gets overwritten with each new run.
-    #   TODO: change that, so that we can get rid of "file_generated_since_run".
-    # ]
-    if pipeline_version.blank? && file_generated_since_run(pipeline_version_file)
-      update(pipeline_version: fetch_pipeline_version)
-    end
+    update_pipeline_version(self, :pipeline_version, pipeline_version_file)
```
boris-dimitrov: Can you make sure this does "aws s3 cp" only if it has to?   That is, if the version has already been retrieved earlier, it doesn't have to go to s3 for it again.

boris-dimitrov: Oh I see that's how it works already.

app/helpers/pipeline_runs_helper.rb
```diff
@@ -0,0 +1,96 @@
...
+    return false unless status.exitstatus.zero?
+    begin
+      s3_file_time = DateTime.strptime(stdout[0..18], "%Y-%m-%d %H:%M:%S")
+      return (s3_file_time && record.created_at && s3_file_time > record.created_at)
+    rescue
+      return nil
+    end
+  end
+
+  def update_pipeline_version(record, version_column, s3_version_file)
```
boris-dimitrov: Orthogonal to this pull request, but this is an awkward way to get the version.   When we dispatch a job, we know what version we are dispatching.  Extract the version there and store it;  then we don't have to poll and look if the timestamp is current.

test/controllers/phylo_trees_controller_test.rb
cdebourcy: no

app/models/phylo_tree.rb
cdebourcy: Neither -- it's executed as an asynchronous resque job enqueued upon the user's request. But the action responds immediately

app/models/pipeline_run.rb
cdebourcy: Yes

app/helpers/pipeline_runs_helper.rb
cdebourcy: Actually when we dispatch the job we don't know the version yet. As far as I know, we need to clone idseq-dag before we can check for the version. But yes, this has always felt awkward. If there were a way to just read a single file from github, that would allow us to easily get the version immediately upon dispatch rather than cloning the whole repo in the worker.

app/models/pipeline_run_stage.rb
```diff
@@ -226,12 +171,11 @@ def host_filtering_command
...
     dag_commands = prepare_dag("host_filter", attribute_dict)
 
-    upload_version = "idseq_dag --version | cut -f2 -d ' ' | aws s3 cp  - #{pipeline_run.pipeline_version_file} "
-    batch_command = [install_pipeline, upload_version, dag_commands].join("; ")
+    batch_command = [install_pipeline(pipeline_run.pipeline_commit), upload_version(pipeline_run.pipeline_version_file), dag_commands].join("; ")
 
     # Dispatch job
     memory = sample.sample_memory.present? ? sample.sample_memory : Sample::DEFAULT_MEMORY_IN_MB
-    aegea_batch_submit_command(batch_command, memory)
+    aegea_batch_submit_command(batch_command, memory, job_queue: pipeline_run.sample.job_queue)
```
boris-dimitrov: This change in particular looks like it might affect the batch queue to which host filtering jobs are sent.   Can you confirm those jobs would still go to the idseq_himem queue?   What need is, quite simply, host filtering and alignment viz on idseq_himem, and non-host-alignment on idseq.   Non-host-alignment should share queue with anything else.   Ideally host filtering would also not share queues with anything else, so some day we should create additional queues to support that.   Also add autoscaling for those.

cdebourcy: resolved

app/helpers/pipeline_runs_helper.rb
boris-dimitrov: Worthy of an issue.   I think there is a way to read a single file.  It's also possible to run the idseq-dag command with argument that is its githash, and then it would dump the version into a file named with the githash, so we don't have to check its timestamp.  But that's also cumbersome.

PR #1416, opened 2018-08-13
Fix bug with subcategory in sample view.

tfrcarvalho:
Fix bug with subcategory in sample view.


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -583,14 +586,21 @@ class PipelineSampleReport extends React.Component {
     let includedSubcategories = newIncludedCategories.filter(category => {
       return category in this.category_child_parent;
     });
+
+    const allCategorySet = new Set(
```
cdebourcy: it seems this block appears twice

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -54,6 +60,8 @@ class PipelineSampleReport extends React.Component {
       { text: "NR %id", value: "NR_percentidentity" },
       { text: "R log(1/e)", value: "NR_neglogevalue" }
     ];
+    this.category_child_parent = { Phage: "Viruses" };
+    this.category_parent_child = { Viruses: ["Phage"] };
```
cdebourcy: Can this be expressed so that the relationship only needs to be specified once?

tfrcarvalho: This a temp hack mostly for performance. I will refactor this once I refactor the rest of the page.

app/assets/src/components/PipelineSampleReport.jsx
tfrcarvalho: Indeed, they are local blocks. Again, a little hackish to make it work with old code. Will refactor too with the whole page.

PR #1417, opened 2018-08-13
Tweak to allow users to delete failed samples

jshoe:
- This is a tweak that Product's been asking for for awhile. Allowing users to delete failed samples (same as when they can delete right after uploading for a short bit).
- A lot of users ask for this when they upload like an incorrect file and then try to upload again but run into the name conflict. Probably cleaner for their project view too to not have the failed samples.
- The trade off is that the failed run info will also go away. I'm thinking this is OK b/c usually users will delete for minor things like uploading with the wrong name or uploading an index file. We can discourage people from using it if they end up using it too much.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -497,9 +497,10 @@ class SamplesController < ApplicationController
   def destroy
     # Will also delete from job_stats, ercc_counts, backgrounds_pipeline_runs, pipeline_runs, input_files, and backgrounds_samples
     deletable = @sample.deletable?(current_user)
-    @sample.destroy if deletable
+    success = false
+    success = @sample.destroy if deletable
     respond_to do |format|
-      if deletable
+      if deletable && success
         format.html { redirect_to samples_url, notice: 'Sample was successfully destroyed.' }
         format.json { head :no_content }
       else
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -496,8 +496,12 @@ class SamplesController < ApplicationController
   # DELETE /samples/1.json
   def destroy
     # Will also delete from job_stats, ercc_counts, backgrounds_pipeline_runs, pipeline_runs, input_files, and backgrounds_samples
+    # Current job_status values look like "1.Host Filtering-FAILED" or "2.GSNAPL/RAPSEARCH alignment-FAILED"
+    # Allow deletion if pipeline runs are empty or the latest run failed, and the sample belongs to you, or you're an admin.
+    deletable = ((@sample.pipeline_runs.empty? || @sample.pipeline_runs[0].job_status.include?("FAILED")) && @sample.user_id == current_user.id) || current_user.admin?
+    @sample.destroy if deletable
     respond_to do |format|
-      if @sample.destroy
+      if deletable
         format.html { redirect_to samples_url, notice: 'Sample was successfully destroyed.' }
         format.json { head :no_content }
       else
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -495,15 +495,14 @@ class SamplesController < ApplicationController
   # DELETE /samples/1
   # DELETE /samples/1.json
   def destroy
-    deletable = @sample.pipeline_runs.empty?
-    @sample.destroy if deletable
+    # Will also delete from job_stats, ercc_counts, backgrounds_pipeline_runs, pipeline_runs, input_files, and backgrounds_samples
     respond_to do |format|
-      if deletable
+      if @sample.destroy
         format.html { redirect_to samples_url, notice: 'Sample was successfully destroyed.' }
         format.json { head :no_content }
       else
         format.html { render :edit }
-        format.json { render json: { message: 'Cannot delete this sample' }, status: :unprocessable_entity }
+        format.json { render json: { message: 'Cannot delete this sample. Something went wrong.' }, status: :unprocessable_entity }
       end
     end
   end
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -500,7 +500,7 @@ class SamplesController < ApplicationController
     success = false
     success = @sample.destroy if deletable
     respond_to do |format|
-      if deletable && success
+      if success
         format.html { redirect_to samples_url, notice: 'Sample was successfully destroyed.' }
         format.json { head :no_content }
       else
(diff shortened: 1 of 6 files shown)


jshoe:
👍

cdebourcy:
@eggerdesigns thoughts?

tfrcarvalho:
In general, with respect to data privacy, users should be able to delete data if they change their minds about sharing. In this case, I do not know if we have a clear message that they will not be able to do this.

app/controllers/samples_controller.rb
```diff
@@ -495,15 +495,14 @@ def update
   # DELETE /samples/1
   # DELETE /samples/1.json
   def destroy
-    deletable = @sample.pipeline_runs.empty?
```
cdebourcy: I think there should still be a guard here, only a different one -- i.e. check if the pipeline_run is indeed failed. Probably not enough to have the condition only shallowly in the frontend

jshoe: good idea

jshoe:
Ready for another pass

app/controllers/samples_controller.rb
```diff
@@ -495,15 +495,18 @@ def update
...
+    # Will also delete from job_stats, ercc_counts, backgrounds_pipeline_runs, pipeline_runs, input_files, and backgrounds_samples
+    # Current job_status values look like "1.Host Filtering-FAILED" or "2.GSNAPL/RAPSEARCH alignment-FAILED"
+    # Allow deletion if pipeline runs are empty or the latest run failed, and the sample belongs to you, or you're an admin.
+    deletable = ((@sample.pipeline_runs.empty? || @sample.pipeline_runs[0].job_status.include?("FAILED")) && @sample.user_id == current_user.id) || current_user.admin?
```
cdebourcy: What if postprocessing failed but the report is ready?
This is probably complex enough to warrant its own function like https://github.com/chanzuckerberg/idseq-web/blob/master/app/models/sample.rb#L381

jshoe: Hm, ok

app/models/sample.rb
```diff
@@ -387,6 +387,21 @@ def self.editable(user)
     end
   end
 
+  def deletable?(user)
+    if user.admin?
+      return true
+    elsif user_id == user.id
+      # Sample belongs to the user
+      if pipeline_runs.empty? ||
+         (pipeline_runs[0].job_status.include?("Filtering-FAILED") || pipeline_runs[0].job_status.include?("alignment-FAILED"))
```
cdebourcy: is there a way to avoid hard-coding this? How about checking the status of the output_states

cdebourcy: You can use REPORT_READY_OUTPUT to check whether the failure should allow the sample to be deletable or not

app/controllers/samples_controller.rb
```diff
@@ -495,15 +495,16 @@ def update
   # DELETE /samples/1
   # DELETE /samples/1.json
   def destroy
-    deletable = @sample.pipeline_runs.empty?
+    # Will also delete from job_stats, ercc_counts, backgrounds_pipeline_runs, pipeline_runs, input_files, and backgrounds_samples
+    deletable = @sample.deletable?(current_user)
     @sample.destroy if deletable
     respond_to do |format|
       if deletable
```
cdebourcy: the `if` also needs to include whether the destroy actually succeeded, no?

app/models/pipeline_run.rb
```diff
@@ -252,6 +252,11 @@ def report_ready?
     output_states.find_by(output: REPORT_READY_OUTPUT).state == STATUS_LOADED
   end
 
+  def report_failed?
+    # If the report ready output was marked as STATUS_FAILED
+    output_states.find_by(output: REPORT_READY_OUTPUT).state == STATUS_FAILED
```
cdebourcy: Thanks for adding the check :) FYI I don’t think the way it is written right now would give the correct behavior when host filtering fails. The host-filtering output_state would be failed, but the report-ready output_state would still be unknown

jshoe: Thanks, defined some constants for clarity

app/models/pipeline_run.rb
```diff
@@ -82,7 +82,10 @@ class PipelineRun < ApplicationRecord
                         "taxon_byteranges" => "db_load_byteranges" }.freeze
   # Note: reads_before_priceseqfilter, reads_after_priceseqfilter, reads_after_cdhitdup
   #       are the only "job_stats" we actually need for web display.
-  REPORT_READY_OUTPUT = "taxon_counts".freeze
+  HOST_FILTERING_OUTPUT = "ercc_counts".freeze
+  ALIGNMENT_OUTPUT = "taxon_counts".freeze
```
cdebourcy: nit: the philosophy of result_monitor / output_states is that they are in principle decoupled from the run_stages / pipeline_monitor. So you could have multiple outputs per stage in the future. So more appropriate names for the constants would be: ERCC_OUTPUT, TAXON_COUNT_OUTPUT, TAXON_BYTERANGE_OUTPUT. But otherwise looks more or less good to go

app/models/pipeline_run.rb
```diff
@@ -252,6 +255,15 @@ def report_ready?
     output_states.find_by(output: REPORT_READY_OUTPUT).state == STATUS_LOADED
   end
 
+  def report_failed?
+    # The report failed if host filtering or alignment failed.
+    host_filtering_status = output_states.find_by(output: HOST_FILTERING_OUTPUT).state
+    alignment_status = output_states.find_by(output: ALIGNMENT_OUTPUT).state
+    if host_filtering_status == STATUS_FAILED || alignment_status == STATUS_FAILED
```
cdebourcy: nit: just return `host_filtering_status == STATUS_FAILED || alignment_status == STATUS_FAILED`?

jshoe: changed, thnx

app/controllers/samples_controller.rb
```diff
@@ -495,15 +495,17 @@ def update
...
   def destroy
-    deletable = @sample.pipeline_runs.empty?
-    @sample.destroy if deletable
+    # Will also delete from job_stats, ercc_counts, backgrounds_pipeline_runs, pipeline_runs, input_files, and backgrounds_samples
+    deletable = @sample.deletable?(current_user)
+    success = false
+    success = @sample.destroy if deletable
     respond_to do |format|
-      if deletable
+      if deletable && success
```
yunfangjuan: just if success should suffice?

app/models/sample.rb
```diff
@@ -387,6 +387,16 @@ def self.editable(user)
     end
   end
 
+  def deletable?(user)
+    if user.admin?
+      true
+    elsif user_id == user.id
+      # Sample belongs to the user
+      # Allow deletion if no pipeline runs, or report failed.
+      pipeline_runs.empty? || pipeline_runs[0].report_failed?
```
yunfangjuan: loop through the pipeline_runs and get the pipeline_runs that didn't fail and make sure it's not empty?

PR #1420, opened 2018-08-14
Impose lineage date check

jshoe:
- Tested by clicking on a couple of samples and checking the results. Same as current report pages.
- Tested locally showing the right dates filled in to the ActiveRecord queries. Ex: ` 568386, 474942, 5148, 1890422, 563835, 1506295, 4637, 31984, 203492, 603431, 332247, 119093, 1639119, 403, 1962910, 1892254, 186827, 41873, 84107, 361606, 1799696, 1804623, 34450) AND (started_at < '2018-07-20 21:02:14' AND ended_at > '2018-07-20 21:02:14')`

### DIFF
--- a/app/models/taxon_lineage.rb
+++ b/app/models/taxon_lineage.rb
@@ -39,14 +39,23 @@ class TaxonLineage < ApplicationRecord
     return { query: r.genus_name, tax_id: genus_tax_id } if r
   end
 
-  def self.fill_lineage_details(tax_map)
-    # Get list of tax_ids to look up in TaxonLineage and TaxonCount rows. Include family_taxids.
+  def self.fill_lineage_details(tax_map, pipeline_run_id)
+    # Get list of tax_ids to look up in TaxonLineage rows. Include family_taxids.
     tax_ids = tax_map.map { |x| x['tax_id'] }
     tax_ids |= tax_map.map { |x| x['family_taxid'] }
 
+    # Get created_at date for our TaxonCount entries. Same for all TaxonCounts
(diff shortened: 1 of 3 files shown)


app/models/taxon_lineage.rb
```diff
@@ -39,14 +39,23 @@ def self.get_genus_info(genus_tax_id)
...
+
     # TODO: Should definitely be simplified with taxonomy/lineage refactoring.
     lineage_by_taxid = {}
-    TaxonLineage.where(taxid: tax_ids).find_each do |x|
+
+    # Since there may be multiple TaxonLineage entries with the same taxid
+    # now, we only select the valid entry based on started_at and ended_at.
+    # The valid lineage entry has start and end dates that include the valid
+    # taxon count entry date.
+    TaxonLineage.where(taxid: tax_ids).where("started_at < ? AND ended_at > ?", valid_date, valid_date).find_each do |x|
```
cdebourcy: why use find_each instead of each? find_each is less efficient

jshoe: Oh weird RuboCop is actually changing the each to find_each. https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/FindEach

Maybe I'll change the RuboCop rule? I'd prefer to keep it in one query as well

cdebourcy: Interesting. Yes, let's change that rule

jshoe: changed now

PR #1426, opened 2018-08-15
Improve phylo controls and add beta label

cdebourcy:
image
image


app/assets/src/components/ui/controls/tooltips/BetaLabel.jsx
```diff
@@ -0,0 +1,12 @@
+import { Label as BaseLabel } from "semantic-ui-react";
```
tfrcarvalho: Can you move this one level down, i.e. `ui/tooltips`? 
Maybe just call it labels, since this is not necessarily something that appears on interaction/hover.

app/assets/src/styles/ui/controls/buttons/primary_button.scss
```diff
@@ -7,6 +7,7 @@
   color: $white;
 
   .icon-label {
+    position: relative;
```
tfrcarvalho: Was this a requirement for the label to position correctly?

cdebourcy: yes

app/assets/src/components/ui/controls/tooltips/BetaLabel.jsx
cdebourcy: ok

app/assets/src/components/ui/controls/buttons/Button.jsx
```diff
@@ -7,17 +7,19 @@ const Button = ({
   disabled,
   onClick,
   icon,
+  label,
   primary,
   secondary,
   text,
   type
```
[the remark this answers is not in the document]
cdebourcy: ok

PR #1428, opened 2018-08-16
Move zscore parameters into the TaxonScoringModel

jshoe:
- A little messy but more flexible for further attributes in the future.
- So now a TaxonScoringModel.attributes_json is like `{ "zscore_min": -99, "zscore_max": 99, "zscore_when_absent_from_sample": -100, "zscore_when_absent_from_bg": 100 }` (same values as default before)

- Tested by loading a couple reports and heatmaps. Confirmed that the Rails logs are using the right values (with different TaxonScoringModels locally).
- Performance on report loads is about the same.
- So this PR should not have any user visible changes.


app/lib/report_helper.rb
```diff
@@ -200,7 +197,20 @@ def all_categories
...
+    # Load scoring model attributes
+    scoring_model ||= TaxonScoringModel::DEFAULT_MODEL_NAME
+    tsm = TaxonScoringModel.find_by(name: scoring_model)
+    JSON.parse(tsm.attributes_json)
+  end
+
+  def fetch_taxon_counts(pipeline_run_id, background_id, scoring_model)
+    # Use scoring model attributes
+    model_attributes = fetch_scoring_model_attributes(scoring_model)
+    zscore_min = model_attributes["zscore_min"]
```
cdebourcy: What if a new scoring model doesn't specify these limits? Should it break?

app/lib/report_helper.rb
```diff
@@ -311,9 +327,9 @@ def fetch_top_taxons(samples, background_id, categories)
         result_hash[pipeline_run_id] = { "pr" => pr, "taxon_counts" => [] }
       end
       row["rpm"] = row["r"] / ((pr.total_reads - pr.total_ercc_reads.to_i) * pr.subsample_fraction) * 1_000_000.0
-      row["zscore"] = row["stdev"].nil? ? ZSCORE_WHEN_ABSENT_FROM_BACKGROUND : ((row["rpm"] - row["mean"]) / row["stdev"])
-      row["zscore"] = ZSCORE_MAX if row["zscore"] > ZSCORE_MAX && row["zscore"] != ZSCORE_WHEN_ABSENT_FROM_BACKGROUND
-      row["zcore"] = ZSCORE_MIN if row["zscore"] < ZSCORE_MIN
+      row["zscore"] = row["stdev"].nil? ? zscore_when_absent_from_bg : ((row["rpm"] - row["mean"]) / row["stdev"])
```
cdebourcy: Can this be refactored so that the code for computing zscore from the scoring model exists only once? Right now it is duplicated in `fetch_top_taxons` / `fetch_taxon_counts`. Maybe you can refactor `fetch_top_taxons` so that it is a single SQL query just like in `fetch_taxon_counts`, and then deduplicate.

Since you're now establishing a relationship between scoring model and z-score (when before it was just about the aggregate score), I think you should encapsulate the z-score calculation in a well-documented function that explains how it links back to the scoring model.

Also, since zscore_min / zscore_max / zscore_when_absent_from_bg are basically mandatory now with your change, I think perhaps you should make them explicit columns in the schema (maybe with a default).

boris-dimitrov: Good suggestions about cleaning up the code, except I was hoping that adding parameters can be achieved without adding columns to the DBs schema, just adding keys and values to a json dict.

As it is, the lets you define a function.  What are doing here, is letting you define a class.   The class has some fields that represent parameters, and some fields that represent computations, like the aggregate score.

This simply takes the existing work and extends it.   Fundamentally, this direction is for adding easier programmability.   Its essentially a general purpose program module or class represented as a json data structure.

boris-dimitrov: The super specific need is that this allows users to experiment easily with different thresholds for improving the signal to noise for the highest scores, which are the ones shown at the top of the report page.   We don't want to force another choice on users without letting them evaluate different params vs each other, and since Yunfang already laid the foundation, this looked like the straightforward step to build on that.

PR #1435, opened 2018-08-17
Bug fix for report trying to load when sample is still pending

jshoe:
Basically remove `@pipeline_run.adjusted_remaining_reads.to_i > 0` b/c that's generated on the fly and no longer indicates the report is ready.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -175,20 +175,20 @@ class SamplesController < ApplicationController
     @report_page_params = { pipeline_version: @pipeline_version, background_id: background_id } if background_id
     @report_page_params[:scoring_model] = params[:scoring_model] if params[:scoring_model]
 
-    # Report is ready if the pipeline run has results finalized and did not
-    # fail, or if all its output states are loaded.
-    if @pipeline_run && ((@pipeline_run.results_finalized? && !@pipeline_run.failed?) || @pipeline_run.report_ready?)
-      if background_id
-        @report_present = 1
-        @report_ts = @pipeline_run.updated_at.to_i
-        @all_categories = all_categories
-        @report_details = report_details(@pipeline_run, current_user.id)
-        @ercc_comparison = @pipeline_run.compare_ercc_counts
-      end
+    # Report is ready if pipeline run has results finalized and did not fail
+    # (success), or if all its output states are loaded successfully.
+    if @pipeline_run &&
+       ((@pipeline_run.results_finalized? && !@pipeline_run.failed?) || @pipeline_run.report_ready?) &&
+       background_id
+      @report_present = 1
+      @report_ts = @pipeline_run.updated_at.to_i
+      @all_categories = all_categories
+      @report_details = report_details(@pipeline_run, current_user.id)
+      @ercc_comparison = @pipeline_run.compare_ercc_counts
+    end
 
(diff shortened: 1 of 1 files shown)


app/controllers/samples_controller.rb
```diff
@@ -174,18 +174,21 @@ def show
...
-      end
 
-      if @pipeline_run.failed?
-        @pipeline_run_retriable = true
-      end
+    # Report is ready if pipeline run has results finalized and did not fail
+    # (success), or if all its output states are loaded successfully.
+    if @pipeline_run &&
+       ((@pipeline_run.results_finalized? && !@pipeline_run.failed?) || @pipeline_run.report_ready?) &&
+       background_id
```
tfrcarvalho: Can you try to reformat the code here like
```
if @pipeline_run 
   && ((@pipeline_run.results_finalized? && !@pipeline_run.failed?) || @pipeline_run.report_ready?) 
   && background_id
```

tfrcarvalho: Or does it get formatted automatically?

app/controllers/samples_controller.rb
```diff
@@ -174,18 +174,21 @@ def show
...
-        @report_details = report_details(@pipeline_run, current_user.id)
-        @ercc_comparison = @pipeline_run.compare_ercc_counts
-      end
 
-      if @pipeline_run.failed?
-        @pipeline_run_retriable = true
-      end
+    # Report is ready if pipeline run has results finalized and did not fail
+    # (success), or if all its output states are loaded successfully.
+    if @pipeline_run &&
```
tfrcarvalho: So, report can be ready event if the `report_ready` flag is not set?

app/controllers/samples_controller.rb
```diff
@@ -174,18 +174,21 @@ def show
...
+       ((@pipeline_run.results_finalized? && !@pipeline_run.failed?) || @pipeline_run.report_ready?) &&
+       background_id
+      @report_present = 1
+      @report_ts = @pipeline_run.updated_at.to_i
+      @all_categories = all_categories
+      @report_details = report_details(@pipeline_run, current_user.id)
+      @ercc_comparison = @pipeline_run.compare_ercc_counts
+    end
+
+    if @pipeline_run && @pipeline_run.results_finalized? && @pipeline_run.failed?
```
tfrcarvalho: You do not need to check `pipeline_run` here.

app/controllers/samples_controller.rb
jshoe: Oh you can't do that in Ruby actually apparently. Says unexpected &&

app/controllers/samples_controller.rb
jshoe: Actually it doesn't look like it anymore

jshoe: Oh wait actually this is so that some "COMPLETE*" reports that failed post processing can still be loaded.. report_ready? is only true if all output states succeed so it's hacky but the COMPLETE* ones will still load with results_finalized and !failed..

app/controllers/samples_controller.rb
tfrcarvalho: ok, never mind then

jshoe:
- Going to address Travis

PR #1436, opened 2018-08-18
WIP: View for list of phylogenetic trees.

tfrcarvalho:
Creates a view to visualizate pre-created phylo trees.


tfrcarvalho:
@cdebourcy 
* **reroot**: It has the reroot feature already on double click the node;
* **scale span**: yeah, I will work on the scale in a follow up PR. Do not want to extend this PR right now.
* **null tree**: last commit should avoid the crash. In any case, maybe the backend should not return unavailable trees for this page?

cdebourcy:
Thanks! Awesome about the rerooting, sorry I missed it! Agreed, the scale change can be for later. I do think we should surface trees that are not yet available -- with an appropriate status description (just like samples that haven't finished processing)

tfrcarvalho:
Ok, I will consider the workflow for not completed trees in the next iteration.
Can you approve this as is for now?

tfrcarvalho:
Shows "tree not available" for when newick string is empty or null.

PR #1441, opened 2018-08-20
disable rubocop "each -> find_each" rule

cdebourcy:
Rubocop's auto-conversion of `each` to `find_each` in https://github.com/chanzuckerberg/idseq-web/blob/4489573c2a973a8af90df09450f67a2a369679bd/app/models/taxon_lineage.rb#L58 more than doubles the load time of the report page.
Example: 
idseq-staging sample 905, on my local machine, takes 13 seconds to load with "each",
30 seconds with "find_each".

Also: each one of the TaxonLineage queries takes of order ~1 second. Next, will work on refactoring to change that.

### DIFF
--- a/app/models/taxon_lineage.rb
+++ b/app/models/taxon_lineage.rb
@@ -46,9 +46,8 @@ class TaxonLineage < ApplicationRecord
 
     # Get created_at date for our TaxonCount entries. Same for all TaxonCounts
     # in a PipelineRun.
-    # TODO: change this to get the created_at date from the smaller pipeline_run table.
-    #       And/or move the lineage selection mechanism into alignment_config.
-    valid_date = TaxonCount.where(pipeline_run_id: pipeline_run_id).select(:created_at).last.created_at
+    # TODO: Move the lineage selection mechanism into alignment_config.
+    valid_date = PipelineRun.find(pipeline_run_id).created_at
 
     # TODO: Should definitely be simplified with taxonomy/lineage refactoring.
     lineage_by_taxid = {}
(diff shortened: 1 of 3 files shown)


app/models/taxon_lineage.rb
```diff
@@ -46,7 +46,9 @@ def self.fill_lineage_details(tax_map, pipeline_run_id)
 
     # Get created_at date for our TaxonCount entries. Same for all TaxonCounts
     # in a PipelineRun.
-    valid_date = PipelineRun.find(pipeline_run_id).taxon_counts.last.created_at
+    # TODO: change this to get the created_at date from the smaller pipeline_run table.
+    #       And/or move the lineage selection mechanism into alignment_config.
+    valid_date = TaxonCount.where(pipeline_run_id: pipeline_run_id).select(:created_at).last.created_at
```
yunfangjuan: I thought you are going to change
PipelineRun.find(pipeline_run_id).created_at?

db/migrate/20180820164834_add_index_to_taxon_counts.rb
```diff
@@ -0,0 +1,4 @@
+class AddIndexToTaxonCounts < ActiveRecord::Migration[5.1]
```
yunfangjuan: don't need this?

app/models/taxon_lineage.rb
cdebourcy: Yes, I thought so too. But there are other places in the code where it uses the taxon_count date, so I left it in the interest of consistency. It will be a bigger undertaking to refactor all the places to use pipeline_run date -- want to do it properly (probably using alignment_config). That's why I added it as a TODO. Please let me know what you think.

db/migrate/20180820164834_add_index_to_taxon_counts.rb
cdebourcy: yep sorry checked in by mistake

PR #1444, opened 2018-08-23
Charles/taxon lookup perf

cdebourcy:
Speed up phylo_tree related pages.
Also, add an index to taxon_lineages to speed up the report page.

### DIFF
--- a/app/controllers/phylo_trees_controller.rb
+++ b/app/controllers/phylo_trees_controller.rb
@@ -58,16 +58,17 @@ class PhyloTreesController < ApplicationController
     # Retrieve pipeline runs that contain the specified taxid.
     eligible_pipeline_runs = current_power.pipeline_runs.top_completed_runs
     all_pipeline_run_ids_with_taxid = TaxonByterange.where(taxid: taxid).where(hit_type: 'NT').pluck(:pipeline_run_id)
-    eligible_pipeline_runs_with_taxid = eligible_pipeline_runs.where(id: all_pipeline_run_ids_with_taxid)
+    eligible_pipeline_run_ids_with_taxid = eligible_pipeline_runs.where(id: all_pipeline_run_ids_with_taxid).pluck(:id)
 
     # Retrieve information for displaying the tree's sample list.
-    @samples = sample_details_json(eligible_pipeline_runs_with_taxid, taxid)
+    @samples = sample_details_json(eligible_pipeline_run_ids_with_taxid, taxid)
 
     # Retrieve information about the taxon
-    example_taxon_count = eligible_pipeline_runs_with_taxid.first.taxon_counts.find_by(tax_id: taxid)
-    taxon_name = example_taxon_count ? example_taxon_count.name : nil
-    tax_level = example_taxon_count ? example_taxon_count.tax_level : nil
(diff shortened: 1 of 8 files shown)


app/controllers/phylo_trees_controller.rb
```diff
@@ -35,18 +35,34 @@ def index
...
+      taxon_lineage = TaxonLineage.where(taxid: taxid).last
+      @taxon = { taxid: taxid,
+                 # Hardcode the species-level for now.
+                 # TODO(charles): clean up when we actually implement genus-level support.
+                 name: taxon_lineage.species_name }
     end
 
-    @phylo_trees = @phylo_trees.as_json(include: :pipeline_runs)
+    # Augment tree data with number of pipeline_runs
+    run_counts_by_tree_id = ActiveRecord::Base.connection.select_all("
```
yunfangjuan: move this to pylo_tree.rb?

app/controllers/phylo_trees_controller.rb
```diff
@@ -120,27 +137,41 @@ def set_phylo_tree
...
+        (select
+           pipeline_runs.id as pipeline_run_id,
+           samples.name,
+           samples.project_id,
+           projects.name as project_name
+         from
+           ((pipeline_runs inner join samples on pipeline_runs.sample_id = samples.id)
+            inner join projects on samples.project_id = projects.id)
+         where pipeline_runs.id in (#{pipeline_run_ids.join(',')})) pipeline_runs_samples_projects
     ").to_hash
```
yunfangjuan: why to_hash not to_a, 
Also, instead of doing the multiple inner joins, you can really simplify the queries as the following. It's more readable and more optimizable by mysql. 

select xxxx from pipeline_runs, projects, samples where pipeline_runs.id in (the list) and pipeline_runs.sample_id = samples.id and samples.project_id = projects.id.

app/controllers/phylo_trees_controller.rb
```diff
@@ -120,27 +137,41 @@ def set_phylo_tree
...
+
+    # Also add:
+    # - number of reads matching the specified taxid in NT
+    # Do not include the query on taxon_counts in the previous query above using a join,
+    # because the taxon_counts table is large.
+    taxon_counts = TaxonCount.where(pipeline_run_id: pipeline_run_ids).where(tax_id: taxid).where(count_type: 'NT').index_by(&:pipeline_run_id)
```
yunfangjuan: make sure there's index for querying tax_id in taxon_counts.

cdebourcy: yes, there is 
```
t.index %w[pipeline_run_id tax_id count_type], name: "taxon_counts_pr_index", unique: true
```

cdebourcy: https://github.com/chanzuckerberg/idseq-web/blob/master/db/schema.rb#L349

app/controllers/phylo_trees_controller.rb
cdebourcy: Thanks for the simplification. The output of to_a and to_hash is the same in this case -- I'v changed it to to_a.

app/controllers/phylo_trees_controller.rb
cdebourcy: ok

PR #1445, opened 2018-08-24
Put postprocessing back in himem queue

jshoe:
- Boris's suggestion to put postprocessing back in the himem queue: https://czi.quip.com/64xeAX9iyMeZ/Learnings-from-cMAL-2-rerun-on-idseq-dag-and-new-result-monitor

### DIFF
--- a/app/models/pipeline_run_stage.rb
+++ b/app/models/pipeline_run_stage.rb
@@ -211,7 +211,6 @@ class PipelineRunStage < ApplicationRecord
     batch_command = [install_pipeline(pipeline_run.pipeline_commit), dag_commands].join("; ")
     # Run it
     memory = sample.sample_memory.present? ? sample.sample_memory : Sample::DEFAULT_MEMORY_IN_MB
-    job_queue = pipeline_run.sample.job_queue || Sample::DEFAULT_QUEUE_HIMEM
-    aegea_batch_submit_command(batch_command, memory, job_queue: job_queue)
+    aegea_batch_submit_command(batch_command, memory, job_queue: pipeline_run.sample.job_queue)
   end
 end
--- a/app/models/pipeline_run_stage.rb
+++ b/app/models/pipeline_run_stage.rb
@@ -175,7 +175,7 @@ class PipelineRunStage < ApplicationRecord
 
     # Dispatch job
     memory = sample.sample_memory.present? ? sample.sample_memory : Sample::DEFAULT_MEMORY_IN_MB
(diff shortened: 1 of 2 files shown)


app/models/pipeline_run_stage.rb
```diff
@@ -210,6 +210,8 @@ def postprocess_command
     dag_commands = prepare_dag("postprocess", attribute_dict)
     batch_command = [install_pipeline(pipeline_run.pipeline_commit), dag_commands].join("; ")
     # Run it
-    aegea_batch_submit_command(batch_command, job_queue: pipeline_run.sample.job_queue)
+    memory = sample.sample_memory.present? ? sample.sample_memory : Sample::DEFAULT_MEMORY_IN_MB
+    job_queue = pipeline_run.sample.job_queue || Sample::DEFAULT_QUEUE_HIMEM
+    aegea_batch_submit_command(batch_command, memory, job_queue: job_queue)
```
cdebourcy: Given the PR description, I'm curious -- what are the reasons for:
(a) adding the memory argument
(b) writing `pipeline_run.sample.job_queue || Sample::DEFAULT_QUEUE_HIMEM` instead of just `job_queue: Sample::DEFAULT_QUEUE_HIMEM`?

jshoe: (a) Ah great question... this might sound weird but the frontend currently sets sample.sample_memory to 240000, so by passing in the memory argument aegea_batch_submit_command sets `vcpus = Sample::DEFAULT_VCPUS_HIMEM`

(b) I did `pipeline_run.sample.job_queue || Sample::DEFAULT_QUEUE_HIMEM` to be consistent with `sample.sample_memory.present? ? sample.sample_memory : Sample::DEFAULT_MEMORY_IN_MB` / because pipeline_run.sample.job_queue is usually nil, so it'd respect the override if someone wanted to set it manually (e.g. for a test run)?

BUT on second thought maybe I'll just remove the job_queue line so it's the same as the host_filtering line?

cdebourcy: Interesting. The field `sample.sample_memory` is set based on the `host_genome`, the idea being that larger host genomes will require more memory to load the STAR index. So it only really applies to host filtering. The resource usage of postprocessing is entirely different and I believe Boris's suggestion was to use the HIMEM queue always.

jshoe: Ah ok so not always 240000.. lemme rework this

jshoe:
Now it'll just explicitly set `vcpus: Sample::DEFAULT_VCPUS_HIMEM` and `job_queue: Sample::DEFAULT_QUEUE_HIMEM`

PR #1446, opened 2018-08-24
Genus-level trees

cdebourcy:
Extend phylo_tree creation to taxids of any level. In practice right now, this means both species and genus levels can be used

### DIFF
--- a/app/models/taxon_lineage.rb
+++ b/app/models/taxon_lineage.rb
@@ -23,6 +23,8 @@ class TaxonLineage < ApplicationRecord
     superkingdom: MISSING_SUPERKINGDOM_ID
   }.freeze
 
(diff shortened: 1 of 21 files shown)


app/models/taxon_lineage.rb
```diff
@@ -23,6 +23,8 @@ class TaxonLineage < ApplicationRecord
     superkingdom: MISSING_SUPERKINGDOM_ID
   }.freeze
 
+  LEVEL_2_NAME = TaxonCount::NAME_2_LEVEL.invert
```
jshoe: maybe this should be next to NAME_2_LEVEL?

yunfangjuan: agreed~

cdebourcy: ok

PR #1449, opened 2018-08-30
Some Rails logger changes

jshoe:
- Replace Airbrake.notify with LogUtil.log_err_and_notify just to get things in the Rails log as well.
- LongRunningSampleEvent block will just write that into the log when a sample has been running for over the threshold.

- So the plan is to set up a simple alert in Datadog to email or Slack the on call person when it sees SampleFailedEvent or LongRunningSampleEvent in the log, for example.


jshoe:
need to fix the schema

PR #1454, opened 2018-09-05
use both NT and NR hits for phylo trees

cdebourcy:


app/assets/src/components/views/phylo_tree/PhyloTreeInputs.jsx
```diff
@@ -29,12 +30,13 @@ class PhyloTreeInputs extends React.Component {
             this.samples
               .filter(
                 s =>
-                  s.taxid_nt_reads >= this.minReads &&
+                  this.passes_create_condition(s.taxid_reads) &&
```
jshoe: nit for keeping in camelCase

app/assets/src/components/views/phylo_tree/PhyloTreeInputs.jsx
```diff
@@ -3,6 +3,8 @@ import React from "react";
 import { Input, Checkbox, Accordion, Menu } from "semantic-ui-react";
 import PrimaryButton from "../../ui/controls/buttons/PrimaryButton";
 
+const MIN_READS = 100;
```
tfrcarvalho: Humm, not a fan of this constraint here: for one, we need the same filter on the report page, to decide if we show the link to create the tree (and we would need to keep them in sync); also, this can be changed/bypassed by an user.

- Are we enforcing this limit on the server side too?
- Can we get this value from the server?

cdebourcy: For now it's purely cosmetic and not enforced on the server side. If it is not satisfied it won't break the pipeline in any way. I think it will make sense to move this constant to server side once we decide on it dynamically -- for now it's just an arbitrary threshold

cdebourcy: It's just a filter that helps you select relevant samples more quickly

cdebourcy: I'll consider defining the condition on the server side in the next iteration

tfrcarvalho: ok, but make sure you enforce it in the PipelineSampleReport too.

PR #1457, opened 2018-09-06
Switch AMR and/or phylo_trees on for select users

cdebourcy:


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -24,6 +24,7 @@ class PipelineSampleReport extends React.Component {
       class: "prog-bar"
     });
     this.admin = props.admin;
+    this.allowedFeatures = props.allowedFeatures;
```
yunfangjuan: maybe you can also set another object variable 
this.allowPhyloTree = this.allowedFeatures.indexOf("phylo_trees") >= 0  
and use it as follows. This will make it slightly easier if you have other way to check it in the future. 

Also, do you have to check if this project is readonly or not for the user?

app/controllers/samples_controller.rb
```diff
@@ -152,7 +152,7 @@ def report_csv
   def show
     @pipeline_run = select_pipeline_run(@sample, params)
     @amr_counts = nil
-    if current_user.admin? && @pipeline_run
+    if current_user.allowed_feature_list.include?("AMR") && @pipeline_run
```
yunfangjuan: admin might not see this feature?

yunfangjuan:
looks good. See comments and make sure you have the project level read/write permissions right for the allowed users.

app/controllers/samples_controller.rb
cdebourcy: thx

app/assets/src/components/PipelineSampleReport.jsx
cdebourcy: Thanks, changed both

PR #1459, opened 2018-09-06
Add Lograge gem for more concise logging in deployed envs

jshoe:
Explanation:
- Lograge is a gem for cleaner logs. Example of new log requests in PROD: `{"method":"GET","path":"/samples/888","format":"html","controller":"SamplesController","action":"show","status":200,"duration":995.79,"view":103.32,"db":95.34,"time":"2018-09-06 22:30:19 +0000","ddsource":["ruby"],"params":{"id":"888"}}`
- https://github.com/roidrage/lograge
- Comes from following the Datadog recommended steps here: https://docs.datadoghq.com/logs/log_collection/ruby/#adding-the-gems
- In development mode your logs will basically look the same so you can still see the full stacktraces of things.
- Most of the boilerplate comes from running `rails generate logging:install` but I already culled it down to the lines I think we need. The files already had similar boilerplate actual from the initial setup of idseq-web.

- Lots of repeat config lines in staging/prod but it seems to be the Rails way and it didn't seem to work to have a base logging.rb initializer with shared lines :3. Like the logging.rb configuration would override the other settings..
- Most of the time in both dev and staging/prod we're just going to be logging to stdout.
- Will add to these things over time to get all the fields in there like current_user.id and stuff...


config/environments/development.rb
```diff
@@ -68,4 +68,45 @@
     domain: 'mg.idseq.net'
   }
   ActiveRecordQueryTrace.enabled = true
+
+  # Development logging configuration
+  logger           = ActiveSupport::Logger.new(STDOUT)
+  logger.formatter = config.log_formatter
+  config.logger    = ActiveSupport::TaggedLogging.new(logger)
+  config.log_to = %w[stdout file]
```
tfrcarvalho: Why do we need both ?

jshoe: Ah you mean stdout and file? or you mean `logger` and `config.logger`?

tfrcarvalho: stdout and file

jshoe: Ah it was in the sample code but I figured this way people can keep their development.log if they like?

jshoe: but open to removing it

tfrcarvalho: Not very important. I approved as is... was just curious if there is any particular reason as one can always tail the file. 
But I guess it can make it easy if you run the server in the foreground.

PR #1463, opened 2018-09-07
Quiet stack trace for invalid routes

jshoe:
- Getting random requests for /ws or robots.txt or favicons that are spamming our logs..


tfrcarvalho:
@jsheu I was thinking, maybe we can silence the ones we can explain (robots.txt or favicon).

PR #1464, opened 2018-09-10
Add colormap util to read arrays of colors and create dynamically-sized scales.

tfrcarvalho:
Add viridis color map.


app/assets/src/components/utils/colormaps/Colormap.js
```diff
@@ -0,0 +1,18 @@
+export class Colormap {
```
jshoe: could you add comments to these

tfrcarvalho: Done... what is the main confusion?
is the comment sufficiently clear?

PR #1465, opened 2018-09-10
Change to CLI instructions to add Mouse

jshoe:
- Gets from HostGenomes instead of hardcoding


app/models/host_genome.rb
```diff
@@ -4,4 +4,8 @@ class HostGenome < ApplicationRecord
   def default_background
     Background.find(default_background_id) if default_background_id
   end
+
+  def self.exposed_genomes
+    all.map(&:name).reject { |n| n.downcase.include?("__test__") || n.downcase.include?("no host subtraction") }
```
tfrcarvalho: Not sure I like this black listing...
Did you consider any alternatives to classify the genome as exposable?

jshoe: Well, we have Human, Mosquito, Tick, ERCC only, and Mouse, and there's nothing really special about them from the app's perspective. So the other alternative is whitelisting them?

cdebourcy: Yeah this is a relic. We have host genomes like "Human __test__" which is the same as "Human", except it specified that the job should go to a queue that doesn't exist. Then the job will fail and we don't incur compute costs, but we can still check if the upload worked. I'm not even sure if the queues even work like that anymore. So I think the solution might be to remove the code and also remove those "__test__" genomes. We will lose the feature of "admin-only" host genomes, but it doesn't seem like we've needed that for a while

cdebourcy: see https://idseq.net/host_genomes

tfrcarvalho: I would either do what @cdebourcy says, if you we are comfortable with deleting those, or add an explicit boolean field of `expose` it to the model. 
I prefer Charles' suggestion if possible...

jshoe: Can we also remove "No host subtraction" Charles? Is that replaced by "ERCC only"?

cdebourcy: I think so. @boris-dimitrov ?

jshoe:
@tfrcarvalho cool I made the db change

PR #1468, opened 2018-09-11
Modify "Read Specificity" filter behavior on Report Page

jshoe:
- Changes the behavior to remove the species-level lines like "Non-species-specific reads in genus X".
- Also removes empty Genus rows when enabled like "Plantibacter (0 bacterial species)" would go away.

- See: https://jira.czi.team/browse/IDSEQ-313

sep-11-2018 10-19-07

### DIFF
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -128,7 +128,9 @@ class PipelineSampleReport extends React.Component {
       loading: true,
       activeThresholds: this.defaultThresholdValues,
       countType: "NT",
-      readSpecificity: cachedReadSpecificity ? cachedReadSpecificity : 0
+      readSpecificity: cachedReadSpecificity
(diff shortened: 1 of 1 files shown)


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -295,6 +295,9 @@ class PipelineSampleReport extends React.Component {
     let selected_taxons = [];
     const thresholded_taxons = input_taxons || this.state.thresholded_taxons;
     const active_thresholds = this.state.activeThresholds;
+    const specificOnly =
+      this.state.readSpecificity.toLowerCase() === "specific only";
```
tfrcarvalho: I know it is not new, but how hard would it be to change it to 0/1?

jshoe: Seems reasonable. I will change it

jshoe:
@tfrcarvalho should be better now O:

PR #1475, opened 2018-09-11
choose idseq-dag branch

cdebourcy:
Need this so we can experiment with different parameters for the tree pipeline


app/assets/src/components/views/phylo_tree/PhyloTreeInputs.jsx
```diff
@@ -214,15 +233,11 @@ class PhyloTreeInputs extends React.Component {
...
-        id="treeName"
-        placeholder="Name"
-        disabled={this.disabled}
-        value={this.disabled ? this.phyloTree.name : undefined}
-        onChange={this.handleInputChange}
-      />
-    );
+    let tree_name = this.renderTextInput("treeName", "Name", "name");
+    let dagBranchInput =
+      this.admin != 1
```
jshoe: can just do `this.admin`?

PR #1482, opened 2018-09-12
Add `kingdom_taxid` column to `taxon_counts`

tfrcarvalho:
* Add `kingdom_taxid` column to `taxon_counts`
* Will update database manually:
`UPDATE taxon_counts tc
	LEFT JOIN taxon_lineages AS tl ON (
		MOD(ABS(tc.tax_id), 100000000) = tl.taxid
	)
SET tc.kingdom_taxid = tl.kingdom_taxid
WHERE 
	tc.created_at BETWEEN tl.started_at AND tl.ended_at;`

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -368,6 +368,7 @@ class PipelineRun < ApplicationRecord
     update_names
     # denormalize superkingdom_taxid into taxon_counts
     if multihit?
+      update_kingdoms
       update_superkingdoms
     else
       update_genera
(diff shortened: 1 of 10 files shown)


app/models/pipeline_run.rb
```diff
@@ -368,6 +368,7 @@ def db_load_taxon_counts
     update_names
     # denormalize superkingdom_taxid into taxon_counts
     if multihit?
+      update_kingdoms
```
cdebourcy: Can we squash update_kingdoms & update_superkingdoms into one?

tfrcarvalho: done

cdebourcy: Thanks. I meant put it into a single SQL query though. Can't you just add another column/value pair to the SET clause? Am I missing something?

cdebourcy:
Can this be closed?

PR #1483, opened 2018-09-12
Create new Metadata table + add a new create method

jshoe:
- Create a new Metadata table and a new add_to_sample method for Metadata.
- We will start using the new Metadata table and eventually migrate / stop using the old "metadata columns" on Sample.
- For now we basically define Metadata as being either a string or a number. The strings can have limited string options that will be checked if present. It seems like the vast majority of metadata can just be treated as strings.
- raw_value is for the original inputted value. validated_value is for the value that passed validation.
- Tested pretty well locally.

- See this doc for reference: https://czi.quip.com/fZtIAkROUlqB/Metadata-in-IDseq

### DIFF
--- a/app/models/metadatum.rb
+++ b/app/models/metadatum.rb
@@ -1,14 +1,14 @@
 class Metadatum < ApplicationRecord
   # ActiveRecord related
   belongs_to :sample
+  STRING_TYPE = 0
+  NUMBER_TYPE = 1
   # When using an ActiveRecord enum, the type returned from reading records is String.
-  enum data_type: [:string, :number]
+  enum data_type: { string: STRING_TYPE, number: NUMBER_TYPE }
   validates :text_validated_value, length: { maximum: 250 }
   validates :number_raw_value, :number_validated_value, numericality: true, allow_nil: true
   validate :set_validated_values
 
-  STRING_TYPE = 0
-  NUMBER_TYPE = 1
   # Key to the metadatum type. Supporting strings and numbers currently.
   KEY_TO_TYPE = {
     unique_id: STRING_TYPE,
--- a/app/models/metadatum.rb
+++ b/app/models/metadatum.rb
@@ -19,8 +19,8 @@ class Metadatum < ApplicationRecord
   def self.add_to_sample(sample, key, raw_value)
     key = key.to_sym
 
-    # Raise RuntimeError if the raw_value is invalid. Caller can handle and
-    # provide feedback.
+    # validate raises RuntimeError if invalid. Caller can handle and provide
+    # feedback.
     validated_value = validate(key, raw_value)
 
     # Create the entry
--- a/app/models/metadatum.rb
+++ b/app/models/metadatum.rb
@@ -1,0 +1,3 @@
+class Metadatum < ApplicationRecord
+  belongs_to :sample
+end
--- a/app/models/metadatum.rb
+++ b/app/models/metadatum.rb
@@ -1,6 +1,7 @@
 class Metadatum < ApplicationRecord
   # ActiveRecord related
   belongs_to :sample
+  # When using an ActiveRecord enum, the type returned from reading records is String.
   enum data_type: [:string, :number]
   validates :text_validated_value, length: { maximum: 250 }
   validates :number_raw_value, :number_validated_value, numericality: true, allow_nil: true
@@ -53,19 +54,21 @@ class Metadatum < ApplicationRecord
   # Custom validator called on save or update. Writes to the *_validated_value column.
   def set_validated_values
     # Check if the key is valid
-    errors.add(:key, "not a supported metadatum") unless KEY_TO_TYPE.key?(key)
+    unless key && KEY_TO_TYPE.key?(key.to_sym)
+      errors.add(:key, "#{key} is not a supported metadatum")
+    end
 
-    if data_type == NUMBER_TYPE
+    if data_type == "number"
       # Set number types. ActiveRecord validated.
       self.number_validated_value = number_raw_value
-    elsif data_type == STRING_TYPE
+    elsif data_type == "string"
       check_and_set_string_type
     end
-    save
   end
 
   # Called by set_validated_values custom validator
   def check_and_set_string_type
+    key = self.key.to_sym
     if KEY_TO_STRING_OPTIONS.key?(key)
       # If there are explicit string options, match the value to one of them.
       matched = false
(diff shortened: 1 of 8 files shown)


app/models/metadatum.rb
```diff
@@ -0,0 +1,101 @@
...
+    sequencer: %w[MiSeq NextSeq HiSeq NovaSeq Other]
+  }.freeze
+
+  # Add a new Metadatum entry to the sample
+  def self.add_to_sample(sample, key, raw_value)
+    key = key.to_sym
+
+    # validate raises RuntimeError if invalid. Caller can handle and provide
+    # feedback.
+    validated_value = validate(key, raw_value)
```
cdebourcy: I think you should look into rail-style validations e.g. the validates_with hook (https://guides.rubyonrails.org/active_record_validations.html#validates-with). That way you could call a standard create method and the validations would be carried out automatically

app/models/metadatum.rb
```diff
@@ -0,0 +1,101 @@
+class Metadatum < ApplicationRecord
+  belongs_to :sample
+
+  # Types: 0 for string, 1 for number, 2 for lists of values (as string)
+  KEY_TO_TYPE = { unique_id: 0, sample_type: 0, nucleotide_type: 0, collection_date: 0, collection_location: 0, collected_by: 0, age: 1, gender: 0, race: 0, primary_diagnosis: 0, antibiotic_administered: 0, admission_date: 0, admission_type: 0, discharge_date: 0, discharge_type: 0, immunocomp: 0, other_infections: 0, comorbitity: 0, known_organism: 0, infection_class: 0, detection_method: 0, library_prep: 0, sequencer: 0, rna_dna_input: 1, library_prep_batch: 0, extraction_batch: 0 }.freeze
```
tfrcarvalho: I would suggestion formatting this one per line. Makes it easier to read.

tfrcarvalho: Would also suggest using enums for readability: https://api.rubyonrails.org/v5.2.1/classes/ActiveRecord/Enum.html

app/models/metadatum.rb
```diff
@@ -0,0 +1,101 @@
...
+          validated_value = opt
+          break
+        end
+      end
+      unless validated_value
+        msg = "#{raw_value} for #{key} did not match options #{options.join(', ')}."
+        raise msg
+      end
+    else
+      # Otherwise any string without restricted options is OK.
```
tfrcarvalho: Length constraints?

db/schema.rb
```diff
@@ -138,7 +151,7 @@
     t.index %w[pipeline_run_id output], name: "index_output_states_on_pipeline_run_id_and_output", unique: true
   end
 
-  create_table "phylo_trees", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=latin1" do |t|
+  create_table "phylo_trees", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
```
tfrcarvalho: You are all well "collated" :)

app/models/metadatum.rb
cdebourcy: There’s no conflict. When inserting a single metadatum, you would call a standard create method with a validates_with referring to your custom validation function. When inserting multiple metadata, you would run your

cdebourcy: There’s no conflict. When inserting a single metadatum, you would call a standard create method with a validates_with referring to your custom validation function. When inserting multiple metadata, you would run your

tfrcarvalho: Coming from rails ignorance here: would you have to disabled validation to do the bulk insert? Or were you saying to bypass the model?

cdebourcy: I meant bypass the model -- I don't think there is a standard rails way to create objects in bulk.
However:
(a) There is a gem that lets you create in bulk through ActiveRecord, although I don't know if we would want to use it. See Joao's answer here: https://stackoverflow.com/questions/39064067/activerecord-skip-validation-when-saving-multiple-objects). To be tested.
(b) Another option I had never heard of would be to wrap all the saves in a transaction to make it faster -- see Harish's answer here: https://stackoverflow.com/questions/2509320/saving-multiple-objects-in-a-single-call-in-rails. Sounds like it is not as good as a single query, but much better than without the transaction. To be tested.

app/models/metadatum.rb
yunfangjuan: i think you should also define constant for type like METADATA_TYPE_INTEGER , etc. agreed with Tiago on one per line

app/models/metadatum.rb
yunfangjuan: i am not sure about the rails validate. my understanding is that if validates fail, the active record will fail to save. Since we are doing a softcheck here, I am not sure if it;s necessary. The validation context here is quite different from the rails validation context i think

yunfangjuan: also instead of calling validate directly,  add a before_save hook to calculate validated value.

jshoe:
- Thanks all for the comments. I reworked it to run as a Rails validation function to fit with that model better. The only trade-off is to be aware that the validation function now has a side effect in setting the *_validated_value column but that's not too bad.
- Also changed things to use constants and ActiveRecord "enum".
- Re bulk SQL inserts I've made a note of it for later but in my experience the metadata inserts have been pretty fast / have a limited scope to insert into the db.

app/models/metadatum.rb
```diff
@@ -0,0 +1,114 @@
...
+  # ActiveRecord related
+  belongs_to :sample
+  # When using an ActiveRecord enum, the type returned from reading records is String.
+  enum data_type: [:string, :number]
+  validates :text_validated_value, length: { maximum: 250 }
+  validates :number_raw_value, :number_validated_value, numericality: true, allow_nil: true
+  validate :set_validated_values
+
+  STRING_TYPE = 0
+  NUMBER_TYPE = 1
```
tfrcarvalho: you did not find a way to reference it for other class variables?
if there is no alternative do at least: `
 STRING_TYPE = 0
 NUMBER_TYPE = 1
enum data_type: {:string => STRING_TYPE, :number => NUMBER_TYPE}
`
So that you do not have to keep the two lists in sync.

jshoe: yeah I couldn't find a way to use `enum` directly apparently. Changed to above

PR #1484, opened 2018-09-13
add SNP annotations

cdebourcy:
- Modify DAG to generate a SNP annotation file (using the SNPs on which the tree was based + an annotated reference genome to compare against)
- Make the file available for download

### DIFF
--- a/app/assets/src/components/ui/controls/buttons/DownloadButton.jsx
+++ b/app/assets/src/components/ui/controls/buttons/DownloadButton.jsx
@@ -3,10 +3,14 @@ import PropTypes from "prop-types";
 import SecondaryButton from "./SecondaryButton";
 import React from "react";
 
-const DownloadButton = ({ disabled, onClick }) => {
+const DownloadButton = ({ what, disabled, onClick }) => {
+  let text = "Download";
+  if (what) {
+    text += ` ${what}`;
+  }
   return (
     <SecondaryButton
-      text="Download"
+      text={text}
       disabled={disabled}
       onClick={onClick}
       icon={<Icon size="large" className={"cloud download alternate"} />}
@@ -15,6 +19,7 @@ const DownloadButton = ({ disabled, onClick }) => {
 };
 
 DownloadButton.propTypes = {
+  what: PropTypes.string,
   disabled: PropTypes.bool,
   onClick: PropTypes.func
 };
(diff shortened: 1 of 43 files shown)


app/assets/src/components/ui/controls/buttons/DownloadButton.jsx
```diff
@@ -15,6 +19,7 @@ const DownloadButton = ({ disabled, onClick }) => {
 };
 
 DownloadButton.propTypes = {
+  what: PropTypes.string,
```
tfrcarvalho: Can you check with Jenn on this ? 
If we need to set the label, I would probably would go with the DropdownDownloadButton to keep consistency of the UI.

app/models/phylo_tree.rb
```diff
@@ -61,12 +64,29 @@ def job_command
...
+      top_taxid_by_run_id = {}
+      species_counts.each do |sc|
+        top_taxid_by_run_id[sc.pipeline_run_id] ||= sc.tax_id
+      end
+      reference_taxids = top_taxid_by_run_id.values
+    else
+      reference_taxids = [taxid]
+    end
+    # Retrieve superkigdom name for idseq-dag
+    superkingdom_name = TaxonLineage.where(taxid: 573).last.superkingdom_name
```
tfrcarvalho: What is 573 taxid? Can we add it to a constant for readability?

app/assets/src/components/ui/controls/buttons/DownloadButton.jsx
cdebourcy: ok will go with DropdownDownloadButton

app/models/phylo_tree.rb
cdebourcy: oops that's a glitch, it should be `where(taxid: taxid)`, good catch

app/controllers/phylo_trees_controller.rb
```diff
@@ -97,6 +97,12 @@ def retry
     end
   end
 
+  def download_snps
+    snp_file = Tempfile.new
+    Open3.capture3("aws", "s3", "cp", @phylo_tree.s3_outputs["SNP_annotations"], snp_file.path)
+    send_file snp_file.path, filename: "#{@phylo_tree.name.downcase.gsub(/\W/, '-')}__SNP-annotations.txt"
```
jshoe: for Open3.capture3 could you do something `unless status.success? LogUtil.log_err_and_airbrake("downloading #{@phylo_tree.s3_outputs["SNP_annotations"]} failed")`

cdebourcy: sure

PR #1485, opened 2018-09-13
JIRA-324 and JIRA-12

jameshowardwang:
Fixes Threshold filter dropdown to preserve state and enables 'X' on the threshold tags to work

JIRA-12 Also fixes Search filter to preserve Threshold and Category filters and vice versa

Note: new behavior of Search is that it behaves like the other filters: it is an additional filter on the results currently visible. So if the user searches for something that is in the sample but is hidden due to other filters, it will not show up until the user removes the other filters

## Type of change

# Test and Reproduce

1. Lots of testing on http://localhost:3000/samples/1019?pipeline_version=2.7&background_id=4 and http://localhost:3000/samples/27?pipeline_version=1.0&background_id=4 -- enabled/disabled all types of filters, refreshed page, did console logs, react inspections, etc.

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [x] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

### DIFF
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -76,15 +76,15 @@ class PipelineSampleReport extends React.Component {
     this.INVALID_CALL_BASE_TAXID = -1e8;
 
     this.thresholdTextByValue = this.allThresholds.reduce(
-      (labels, threshold) => {
-        labels[threshold.value] = threshold.text;
-        return labels;
+      (metrics, threshold) => {
+        metrics[threshold.value] = threshold.text;
+        return metrics;
       },
       {}
     );
 
     this.defaultThreshold = {
-      label: this.allThresholds[0]["value"],
+      metric: this.allThresholds[0]["value"],
       operator: ">=",
       value: ""
     };
@@ -156,8 +156,13 @@ class PipelineSampleReport extends React.Component {
     this.handleBackgroundModelChange = this.handleBackgroundModelChange.bind(
       this
     );
-    this.resetAllFilters = this.resetAllFilters.bind(this);
+    this.removeCategory = this.removeCategory.bind(this);
+    this.removeThresholdFilter = this.removeThresholdFilter.bind(this);
     this.renderMore = this.renderMore.bind(this);
+    this.resetAllFilters = this.resetAllFilters.bind(this);
+    this.saveAndApplyThresholdFilters = this.saveAndApplyThresholdFilters.bind(
+      this
+    );
     this.setSortParams = this.setSortParams.bind(this);
     this.sortCompareFunction = this.sortCompareFunction.bind(this);
     this.sortResults = this.sortResults.bind(this);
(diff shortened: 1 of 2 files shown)


app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
```diff
@@ -70,7 +64,9 @@ class ThresholdFilterDropdown extends React.Component {
 
   handleClose() {
     this.setState({ popupIsOpen: false });
```
tfrcarvalho: Should merge the setState calls.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -288,26 +293,24 @@ class PipelineSampleReport extends React.Component {
     );
   }
 
-  applySearchFilter(
-    searchTaxonId,
-    includedCategories,
-    input_taxons,
-    includedSubcategories = []
-  ) {
+  applySearchFilter(searchTaxonId = this.state.search_taxon_id) {
```
tfrcarvalho: Why cant we move searchTaxonId below like the remaining arguments?

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -288,26 +293,24 @@ class PipelineSampleReport extends React.Component {
...
-    const thresholded_taxons = input_taxons || this.state.thresholded_taxons;
-    const active_thresholds = this.state.activeThresholds;
     const specificOnly = this.state.readSpecificity === 1;
 
     if (searchTaxonId > 0) {
       let genus_taxon = {};
       let matched_taxons = [];
-      for (let i = 0; i < this.state.taxonomy_details.length; i++) {
-        const taxon = this.state.taxonomy_details[i];
+      let new_input_taxons = [];
```
tfrcarvalho: Just trying to understand the changes: what did taxonomy_details contained (vs thresholded_taxons) before?

app/assets/src/components/PipelineSampleReport.jsx
jameshowardwang: you're right, we totally can. in a subsequent diff i think i will rename some of these functions to be more accurately descriptive of what they're doing (since applySearchFilter also applies category filters)

app/assets/src/components/PipelineSampleReport.jsx
jameshowardwang: taxonomy_details contains ALL the taxons, while thresholded taxons only contains ones passing the threshold filters. so this is part of the change that text search won't return anything not already visible in the table based on the other filters

PR #1489, opened 2018-09-14
move compile stats to result monitor

boris-dimitrov:


boris-dimitrov:
wow rubucop didn't pull me over for this

app/models/pipeline_run.rb
```diff
@@ -641,6 +650,8 @@ def compile_stats_file
     unless status.exitstatus.zero?
       Rails.logger.warn("Failed to write compiled stats file: #{stderr}")
     end
+
+    save
```
jshoe: we don't need this save here, right?

PR #1492, opened 2018-09-17
Add bulk loader for Metadata

jshoe:
- Add functions for Metadatum#bulk_load_from_s3_csv, basically replacing SamplesHelper#populate_metadata_bulk
- Works pretty well on the templates given so far although will need to be fine-tuned depending on what people actually put in their CSVs...

### DIFF
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -49,6 +49,7 @@ module SamplesHelper
   end
 
   # Load bulk metadata from a CSV file
+  # TODO: Succeeded by Metadatum#load_csv_from_s3 for new Metadatum objects.
   def populate_metadata_bulk(csv_s3_path)
     # Load the CSV data. CSV should have columns "sample_name", "project_name", and any desired columns from Sample::METADATA_FIELDS.
     csv_data = get_s3_file(csv_s3_path)
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -49,7 +49,7 @@ module SamplesHelper
   end
 
   # Load bulk metadata from a CSV file
-  # TODO: Succeeded by Metadatum#load_csv_from_s3 for new Metadatum objects.
+  # NOTE: Succeeded by Metadatum#load_csv_from_s3 for new Metadatum objects.
   def populate_metadata_bulk(csv_s3_path)
     # Load the CSV data. CSV should have columns "sample_name", "project_name", and any desired columns from Sample::METADATA_FIELDS.
     csv_data = get_s3_file(csv_s3_path)
--- a/app/models/metadatum.rb
+++ b/app/models/metadatum.rb
@@ -34,7 +34,7 @@ class Metadatum < ApplicationRecord
     discharge_type: STRING_TYPE,
     immunocomp: STRING_TYPE,
     other_infections: STRING_TYPE,
-    comorbitity: STRING_TYPE,
+    comorbidity: STRING_TYPE,
     known_organism: STRING_TYPE,
     infection_class: STRING_TYPE,
     detection_method: STRING_TYPE,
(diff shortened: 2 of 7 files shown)


app/models/metadatum.rb
```diff
@@ -90,25 +111,113 @@ def check_and_set_string_type
...
+    # validator.
   end
 
-  def str_to_basic_chars(res)
+  def self.str_to_basic_chars(res)
     res.downcase.gsub(/[^0-9A-Za-z]/, '')
   end
+
+  # Load bulk metadata from a CSV file from S3
+  def self.bulk_load_from_s3_csv(path)
```
cdebourcy: So if I try to upload a 20-column metadata file for project "Uganda - all - 2" (214 samples), this will make 4280 database queries?

cdebourcy: I think this would be a good time to improve on that

cdebourcy: the previous implementation at least updated all the columns in one go for each row ;). But better would be: do 1 query to retrieve the data needed to make the validations. Iterate through the data and enforce the validations. Then do 1 query to write all the data. It will require some reorganization of the code, but if we ever seriously want to use metadata adding/editing functionality, it will be indispensable.

app/helpers/samples_helper.rb
```diff
@@ -49,6 +49,7 @@ def generate_sample_list_csv(formatted_samples)
   end
 
   # Load bulk metadata from a CSV file
+  # TODO: Succeeded by Metadatum#load_csv_from_s3 for new Metadatum objects.
```
tfrcarvalho: nit: What is the TODO? convert and merge for old data?

app/models/metadatum.rb
```diff
@@ -51,6 +57,21 @@ class Metadatum < ApplicationRecord
...
+    :"rna/dna_input(ng)" => "rna_dna_input",
+    :"rna/dna_input (ng)" => "rna_dna_input",
+    :"rna/dna_input" => "rna_dna_input",
+    :"rna/dna input" => "rna_dna_input",
+    :"rna/dna input (ng)" => "rna_dna_input",
+    antibiotics_administered: "antibiotic_administered",
+    :"known_organism(s)" => "known_organism",
+    :known_organisms => "known_organism"
+  }.freeze
+
```
tfrcarvalho: Are we planning to (eventually) move all this configuration into db ?

app/models/metadatum.rb
tfrcarvalho: There are more queries in the process, like loading the project per row and and the sample per row that can be optimized by grouping everything.

app/helpers/samples_helper.rb
jshoe: more like a NOTE than a TODO

app/models/metadatum.rb
jshoe: I think eventually yes but I kind of like it in code like this..

app/models/metadatum.rb
jshoe: - Rough benchmark on staging, inserting 5000 Metadata will take about 15-30 seconds. So to enable a better user experience for the sake of uploading in the frontend I'll try to get that down with bulk inserts.
- I'd like to keep things in ActiveRecord though for keeping things the Rails way as much as possible.. the performance gain will be from not having to do the roundtrip transaction/commit/etc for each Metadata.

app/helpers/samples_helper.rb
tfrcarvalho: Can you move it to Note:  then ? 
TODOs are typically used for tracking.

jshoe:
- So now it'll basically insert something like `INSERT INTO `metadata` (`id`,`key`,`data_type`,`text_raw_value`,`text_validated_value`,`number_raw_value`,`number_validated_value`,`sample_id`,`created_at`,`updated_at`) VALUES (NULL,'unique_id',0,'a','a',NULL,NULL,897,'2018-09-18 18:59:37','2018-09-18 18:59:37'),(NULL,'sample_type',0,'b','b',NULL,NULL,897,'2018-09-18 18:59:37','2018-09-18 18:59:37'),(NULL,'nucleotide_type',0,'DNA','DNA',NULL,NULL,897,'2018-09-18 18:59:37','2018-09-18 18:59:37'),(NULL,'collection_date',0,'c','c',NULL,NULL,897,'2018-09-18 18:59:37','2018-09-18 18:59:37'),(NULL,'collection_location',0,'d','d',NULL,NULL,897,'2018-09-18 18:59:37','2018-09-18 18:59:37'),(NULL,'collected_by',0,'e','e',NULL,NULL,897,'2018-09-18 18:59:37','2018-09-18 18:59:37'),(NULL,'age',1,NULL,NULL,20.0,20.0,897,'2018-09-18 18:59:37','2018-09-18 18:59:37'),(NULL,'gender',0,'Male','Male',NULL,NULL,897,'2018-09-18 18:59:37','2018-09-18 18:59:37'),(NULL,'race',0,'Asian','Asian',NULL,NULL,897,'2018-09-18 18:59:37','2018-09-18 18:59:37'),(NULL,'primary_diagnosis',0,'i','i',NULL,NULL,897,'2018-09-18 18:59:37','2018-09-18 18:59:37'),(NULL,'admission_date',0...`

cdebourcy:
Awesome! How does that affect the timing of your 5000-line benchmark?

tfrcarvalho:
It seems like activerecord-import still adds correct items even if some errors occur. How does it handle duplicate items on retry?

jshoe:
- @cdebourcy yeah it takes about 5 seconds locally so probably even faster deployed..
- @tfrcarvalho I configured the `on_duplicate_key_update: update_keys` to update the value entries on duplicates, so I think it works as intended? People could upload the same CSV / fix some values / upload again until things are right.

tfrcarvalho:
Sounds good! LGTM

PR #1500, opened 2018-09-19
pathogen tags and executive summary

cdebourcy:


app/assets/src/components/ui/labels/PathogenLabel.jsx
```diff
@@ -0,0 +1,24 @@
+import Label from "./Label";
+import React from "react";
+
+const PathogenLabel = ({ type, number }) => {
+  let type2text = {
+    category_A: "priority A pathogen",
```
jshoe: is the "priority A pathogen" labeling more of a prototype thing?

cdebourcy: yeah. We haven't gotten definitive text yet. We felt the term "pathogenic | A" in the mocks was too vague. I need to restyle the colors too

tfrcarvalho: category_A -> categoryA

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -100,6 +100,7 @@ class PipelineSampleReport extends React.Component {
     // Starting state is default values which are to be set later.
     this.state = {
       taxonomy_details: [],
+      topScoringTaxa: [],
```
tfrcarvalho: Should the top scoring come from server? Could you have do this on client side?

app/assets/src/components/ui/labels/PathogenLabel.jsx
```diff
@@ -0,0 +1,24 @@
+import Label from "./Label";
+import React from "react";
+
+const PathogenLabel = ({ type, number }) => {
+  let type2text = {
```
tfrcarvalho: nit: categoryLabels?

app/assets/src/components/ui/labels/PathogenLabel.jsx
tfrcarvalho: We still set on "Pathogenic | A" right?

app/assets/src/components/ui/labels/PathogenLabel.jsx
```diff
@@ -0,0 +1,24 @@
...
+    category_A: "priority A pathogen",
+    category_B: "priority B pathogen",
+    category_C: "priority C pathogen"
+  };
+  let text = type2text[type];
+  if (number) {
```
tfrcarvalho: Just compare number >1 here.
the extra check in the line below is not necessary.

app/assets/src/components/ui/labels/PathogenSummary.jsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
```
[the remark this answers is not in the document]
tfrcarvalho: Maybe we can move PipelineSampleReport there too... later we can start breaking it apart there.

app/lib/report_helper.rb
```diff
@@ -169,6 +169,31 @@ def external_report_info(pipeline_run_id, background_id, params)
...
+    #   top 3 hits by NT.aggregatescore that satisfy:
+    #   NT.zscore > 1 AND
+    #   NR.zscore > 1 AND
+    #   NT.rpm > 1 AND
+    #   NR.rpm > 1 AND
+    #   tax_id > 0
+    # Current implementation assumes tax_map is already sorted by aggregate score.
+    # This assumption is valid because of the current value of DEFAULT_SORT_PARAM.
+    # All user-triggered sorting is implemented in the JSX frontend, so won't affect sorting of tax_map
+    # (despite some unused relics of code that accepted sort_by as a param in samples_controller).
```
tfrcarvalho: We might have to recover `sort_by` on server... in case we ever want load the page in chunks, no? 
Also, why not do this on client side?

app/lib/report_helper.rb
```diff
@@ -169,6 +169,31 @@ def external_report_info(pipeline_run_id, background_id, params)
...
+    top_taxa = []
+    i = 0
+    while top_taxa.length < 3
```
tfrcarvalho: Always define numbers as constants instead of using magic numbers.

app/models/taxon_lineage.rb
```diff
@@ -34,6 +34,49 @@ class TaxonLineage < ApplicationRecord
   PHAGE_FAMILIES_TAXIDS = [10_472, 10_474, 10_477, 10_656, 10_659, 10_662, 10_699, 10_744, 10_841, 10_860,
                            10_877, 11_989, 157_897, 292_638, 324_686, 423_358, 573_053, 1_232_737].freeze
 
+  # From https://www.niaid.nih.gov/research/emerging-infectious-diseases-pathogens
+  # Accessed 9/18/2018.
+  # Categories are listed in order of priority (A, B, C). So if a taxon is included by the rules of both
```
tfrcarvalho: We should eventually put this in the database, probably as metadata.

app/assets/src/components/ui/labels/PathogenLabel.jsx
cdebourcy: sure, although it doesn't make any sense

tfrcarvalho: It is just a representation we set on. 
I would agree if you say we should explained it somewhere... maybe in a tooltip in the label or in the header of the column.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1354,53 +1366,26 @@ class PipelineSampleReport extends React.Component {
 
     const advanced_filter_tag_list = this.state.activeThresholds.map(
       (threshold, i) => (
-        <AdvancedFilterTagList
-          threshold={threshold}
-          key={i}
-          i={i}
-          parent={this}
-        />
+        <AdvancedFilterTag threshold={threshold} key={i} i={i} parent={this} />
```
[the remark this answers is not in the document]
cdebourcy: It was initially going to be related be cause I was implementing a filter for priority pathogens, which I subsequently decided to postpone. Ok, I have reverted the deduplication.

app/assets/src/components/PipelineSampleReport.jsx
cdebourcy: I think it should. We're trying to serve resource-limited areas so shouldn't do too much data processing in the browser. Semantically the logic also fits well in that place of the code.

app/assets/src/components/ui/labels/PathogenLabel.jsx
cdebourcy: ok

app/models/taxon_lineage.rb
cdebourcy: sure

app/assets/src/components/ui/labels/PathogenLabel.jsx
cdebourcy: ok changed

cdebourcy: ok

app/lib/report_helper.rb
cdebourcy: ok i made it an argument to the function instead since we might want to play with it in the future

app/assets/src/components/ui/labels/PathogenLabel.jsx
cdebourcy: resolved because I removed the `number` prop

app/assets/src/loader.scss
```diff
@@ -33,6 +33,10 @@
 @import "styles/ui/controls/buttons/secondary_button";
 @import "styles/ui/controls/slider";
 
+// labels
+@import "styles/ui/labels/pathogen_label";
+@import "styles/ui/labels/pathogen_summary";
```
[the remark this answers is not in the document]
cdebourcy: ok moved

app/assets/src/components/ui/labels/PathogenSummary.jsx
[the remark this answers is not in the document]
cdebourcy: ok moved PathogenSummary to views/report

app/assets/src/components/PipelineSampleReport.jsx
tfrcarvalho: The most scarce resource is actually network :)
But it is ok to do it server side.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -15,7 +15,10 @@ import OurDropdown from "./ui/controls/dropdowns/Dropdown";
 import MultipleDropdown from "./ui/controls/dropdowns/MultipleDropdown";
 import ThresholdFilterDropdown from "./ui/controls/dropdowns/ThresholdFilterDropdown";
 import BetaLabel from "./ui/labels/BetaLabel";
+import PathogenLabel from "./ui/labels/PathogenLabel";
+import PathogenSummary from "./views/report/PathogenSummary";
 import PhyloTreeInputs from "./views/phylo_tree/PhyloTreeInputs.jsx";
+import ReportInsightIcon from "./views/report/ReportInsightIcon";
```
tfrcarvalho: Can you add this to `ui/icons` and call it BulbIcon or something more descriptive of the image?

PR #1512, opened 2018-09-21
JIRA-369 - Fix threshold filter not adding thresholds

tfrcarvalho:
Fix threshold filter dropdown.
* It does a deep comparison of all threshold filters.
* This getDerivedStateFromProps is called everytime that react determines that props might have change or when state changes. This could lead to some ineficiency if we have many active thresholds but should make no difference on our current use cases.

Jira: https://jira.czi.team/browse/IDSEQ-369

Actions checked:
- [x] Report page: Added multiple filters and verified filters had effect on table
- [x] Report page: Removed some filters and verified the new set of filters had effect on table
- [x] Report page: Removed the target tags and verified that removed the threshold filters on the dropdown
- [x] Report page: Reloaded report page and verified that threshold filter were preloaded from local storage
- [x] Heatmap page: Added multiple filters and verified filters had effect on table
- [x] Heatmap page: Removed some filters and verified the new set of filters had effect on table


app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
```diff
@@ -23,12 +23,42 @@ class ThresholdFilterDropdown extends React.Component {
...
-      return { thresholds: newThresholds };
+    if (
+      !ThresholdFilterDropdown.areThresholdsFiltersEqual(
+        newThresholds,
+        state.oldThresholds
+      )
+    ) {
+      return {
+        thresholds: newThresholds,
+        oldThresholds: newThresholds
```
jameshowardwang: why do we need to return oldThresholds?

package.json
```diff
@@ -10,10 +10,7 @@
     "setupTestFrameworkScriptFile": "./enzyme_config.js"
   },
   "lint-staged": {
-    "*.{js,jsx,json,css,scss}": [
-      "prettier --write",
-      "git add"
-    ],
+    "*.{js,jsx,json,css,scss}": ["prettier --write", "git add"],
```
jameshowardwang: you can probably revert this right?

app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
```diff
@@ -23,12 +23,42 @@ class ThresholdFilterDropdown extends React.Component {
...
+      };
     }
     return null;
   }
 
+  static areThresholdsFiltersEqual(tfs1, tfs2) {
```
jameshowardwang: should this go into some helper library for deep comparison of objects?

app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
tfrcarvalho: Its a consequence of getDerivedPropsFromState being static. Does not have access to the old props. So the advised way is to keep track of the old props.
You should only check if the old props changed as state might change within the component.
- React's advice: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#alternative-1-reset-uncontrolled-component-with-an-id-prop
- Why old props are not included: https://github.com/facebook/react/issues/12188

package.json
tfrcarvalho: This formatting is done somewhere on a post commit... i did not include this file, so i have no way to undo it.
Keeps going back and forth... need to look into it.

app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
tfrcarvalho: This one is very specific of this object... so I would say no. A full object comparison is more complicated.

PR #1521, opened 2018-09-25
Detailed rpm list

yunfangjuan:
# Description

Add the detailed rpms (including 0's into taxon_summaries) 

## Type of change

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

### DIFF
--- a/app/models/background.rb
+++ b/app/models/background.rb
@@ -6,6 +6,7 @@ class Background < ApplicationRecord
   validate :validate_size
   validates :name, presence: true, uniqueness: true
   after_save :submit_store_summary_job
+  attr_accessor :just_updated
 
   DEFAULT_BACKGROUND_MODEL_NAME = "default".freeze
   TAXON_SUMMARY_CHUNK_SIZE = 100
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -466,30 +466,30 @@
     t.float "stdev", limit: 24
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["background_id", "count_type", "tax_level", "tax_id"], name: "index_taxon_summaries_detailed", unique: true
+    t.text "rpm_list"
+    t.index %w[background_id count_type tax_level tax_id], name: "index_taxon_summaries_detailed", unique: true
     t.index ["background_id"], name: "index_taxon_summaries_on_background_id"
```
cdebourcy: I think with the index you added, we can and should remove the one on just background_id

app/models/background.rb
```diff
@@ -19,42 +20,72 @@ def validate_size
...
-      h[:updated_at] = date
-      h[:mean] = (h["sum_rpm"] || 0.0) / n.to_f
-      h[:stdev] = compute_stdev(h["sum_rpm"] || 0.0, h["sum_rpm2"] || 0.0, n)
+    key = ""
+    res = {}
+    results = []
+    rows.each do |row|
+      current_key = [row["tax_id"], row["count_type"], row["tax_level"]].join('-')
+      if current_key != key
+        if res[:tax_id]
```
cdebourcy: what does this check do?

cdebourcy: nvm

cdebourcy: Is this just to prevent it from doing summarize_taxon in the first iteration? If so, wouldn't it be more readable to say so explicitly?

app/models/background.rb
```diff
@@ -19,42 +20,72 @@ def validate_size
...
     n = pipeline_runs.count
     date = DateTime.now.in_time_zone
-    results.each do |h|
-      h[:background_id] = id
-      h[:created_at] = date
-      h[:updated_at] = date
-      h[:mean] = (h["sum_rpm"] || 0.0) / n.to_f
-      h[:stdev] = compute_stdev(h["sum_rpm"] || 0.0, h["sum_rpm2"] || 0.0, n)
+    key = ""
+    res = {}
```
cdebourcy: would prefer a name like `single_taxon_results` (and then `all_taxa_results` instead of `results`) -- it would do a lot for readability -- took me sometime to figure out what the code does

app/models/background.rb
yunfangjuan: i can add a comment.

app/models/background.rb
yunfangjuan: i think taxon_results and result_list is good enough. I don't like long names.

db/schema.rb
yunfangjuan: I didn't ass any index. 
the ["background_id"]  index is completely redundant. MySQL has the prefix index that having both [background_id count_type tax_level tax_id] and [background_id]   is completely unnecessary.

cdebourcy: agreed. Oh yeah I just saw that line 470 is in fact not new.

PR #1523, opened 2018-09-26
Tweak to phylo tree coloring

jshoe:
Improvements are:
- Just showing everything as grey and no legend if there's no valid info
- Not showing the absent value if the absent value is not actually used

### DIFF
--- a/app/assets/src/components/visualizations/dendrogram/Dendogram.js
+++ b/app/assets/src/components/visualizations/dendrogram/Dendogram.js
@@ -160,23 +160,33 @@ export default class Dendogram {
     let attrName = this.options.colorGroupAttribute;
 
     // Set up all attribute values. Colors end up looking like:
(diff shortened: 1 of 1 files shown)


app/assets/src/components/visualizations/dendrogram/Dendogram.js
```diff
@@ -158,25 +158,34 @@ export default class Dendogram {
     }
 
     let attrName = this.options.colorGroupAttribute;
+    let absentName = this.options.colorGroupAbsentName;
 
     // Set up all attribute values. Colors end up looking like:
-    // Uncolored (grey) | Real values.. | Absent attribute color (e.g. for External Sources)
+    // Uncolored (grey) | Absent attribute color (e.g. for NCBI References) + Real values..
```
tfrcarvalho: what do you mean by real values?

jshoe: values that were actually seen? I could change to "Values found"

PR #1524, opened 2018-09-26
Deploy using czecs

mbarrien:
This PR converts idseq-web and resque to using https://github.com/chanzuckerberg/cellxgene-rest-api/pull/czecs for deployment, and AWS Parameter Store via Chamber for both secrets and for communicating resource IDs allocated by Terraform. It is paired with <https://github.com/chanzuckerberg/idseq-infra/pull/136>. Both must be merged at the same time.
—
czecs (Chan Zuckerberg Elastic Container Service, pronounced like “checks”) puts the definition of the task definition into the code repos, no longer requiring Terraform to pre-define the task. (See other PR for more details.) This PR includes the task definition, which is a template where some of the values are defined via the included balances.json file, or via command line overrides as seen in .travis.yml.

This PR adds the czecs.json and czecs-resque.json service templates, along with balances.json for staging use. For prod, appropriate command line overrides are provided in the new deploy.sh; an alternate option would be to create a balances.prod.sh. We also use czecs’s new task command to be able to run the migrations as before.

deploy has been changed to mostly delegate functionality to czecs through system calls.
—
We also introduce chamber for secrets and Terraform resources. Secrets are expected to be stored with the prefix `idseq-${ENV}-web/`. Secrets should be passed via chamber, but in local dev, the same pre-existing variable names can still be passed at the command line/docker-compose without needing to invoke chamber.

We modify the Dockerfile both to download Chamber and to add an entrypoint that will read from AWS Parameter Store using chamber (using the ENVIRONMENT environment variable to determine where to read from)  and stick values into the environment automatically. (The entrypoint will need to be manually run if running docker exec to enter an existing container.)
—
In miscellaneous changes:
* docker-compose.yml is modified to use the correct Airbrake environment variables.
* The pip installs are coalesced into a single call to reduce the number of layers in the resulting docker image


balances.json
```diff
@@ -0,0 +1,8 @@
+{
+    "db_username": "idseqmaster",
```
jameshowardwang: i don't see where this is referenced, also is it okay for this to be exposed publicly?

docker-compose.yml
```diff
@@ -74,8 +74,8 @@ services:
       - AWS_ACCESS_KEY_ID
       - AWS_SECRET_ACCESS_KEY
       - SAMPLES_BUCKET_NAME=idseq-samples-development
-      - IDSEQ_AIRBRAKE_PROJECT_ID
-      - IDSEQ_AIRBRAKE_PROJECT_KEY
+      - AIRBRAKE_PROJECT_ID
+      - AIRBRAKE_PROJECT_KEY
```
jameshowardwang: weird, how has this been working before?

mbarrien: I'd guess that in local dev, Airbrake doesn't get used.

balances.json
mbarrien: Removed; snuck in from earlier drafts. People would still need a password anyways for this to be useful, and that didn't make it in here.

mbarrien:
Testing for similar deploys in the past involved having some time to go through a try-it-then-debug loop across both infra and application repos, but in those cases, staging wasn't as in demand at the time. I'd probably need a few hours (possibly overnight) to do that, and would have the ability to rollback if things didn't get to a working state quickly.

mbarrien:
Just reconverted the deploy.sh script back to Ruby for consistency with the rest of the existing scripts, but still delegates main work to czecs.

mbarrien:
I've successfully tested and debugged the deploy script, Terraform state, and czecs tool. This is now ready for merge, but I need an authorized user to do the merge. If someone authorized does get around to merging this, please also merge chanzuckerberg/idseq-infra#136 (already approved) at the same time.

Once this is merged, chanzuckerberg/idseq-infra#137 should be migrated and run before the next deploy to prod.

PR #1528, opened 2018-09-27
Get rid of honeycomb errors in rails logs

jameshowardwang:
Tired of seeing these in our error logs:
[2018-09-27T23:31:38] WARN  Rails : No write key defined! (Check your config's `writekey` value in config/initializers/honeycomb.rb) No events will be sent to Honeycomb.


jameshowardwang:
abandoning this for now

PR #1532, opened 2018-09-28
Add tooltips to the phylo tree for both NCBI and Sample entries

tfrcarvalho:
# Description

* Add tooltips to the phylo tree for both NCBI and Sample entries
* New DataTooltip container component for generic data tooltips
* Links to NCBI and samples page
* Save session state to allow refreshing same page

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change. Explain both the expected functionality and the edge cases. 

## Type of change


app/assets/src/components/views/phylo_tree/PhyloTreeListView.jsx
```diff
@@ -28,7 +28,11 @@ class PhyloTreeListView extends React.Component {
   }
 
   getDefaultSelectedTreeId(urlParams, phyloTrees = []) {
-    return urlParams.treeId || (phyloTrees[0] || {}).id;
+    return (
+      urlParams.treeId ||
+      parseInt(window.sessionStorage.getItem("treeId")) ||
```
jshoe: did we all agree that window.sessionStorage is OK? is it supported on all the modern browsers

app/assets/src/components/views/phylo_tree/PhyloTreeVis.jsx
```diff
@@ -33,26 +46,91 @@ class PhyloTreeVis extends React.Component {
...
+  }
+
+  handleNodeHover(node) {
+    this.setState({ hoveredNode: node });
+  }
+
+  handleNodeClick(node) {
+    if (node.data.accession) {
+      let url = `https://www.ncbi.nlm.nih.gov/nuccore/${node.data.accession}`;
+      window.open(url, "_blank");
```
jshoe: Add rel="noopener noreferrer" to all external links https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

app/assets/src/components/views/phylo_tree/PhyloTreeListView.jsx
tfrcarvalho: Should be fine: https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage. We should do a more through css/js audit for issues in older browsers.

app/assets/src/components/views/phylo_tree/PhyloTreeVis.jsx
jshoe: or https://mathiasbynens.github.io/rel-noopener/

app/assets/src/components/views/phylo_tree/PhyloTreeVis.jsx
```diff
@@ -2,25 +2,38 @@ import React from "react";
...
-        newick: props.newick,
-        props: props.nodeData
-      };
-    }
+    this.handleNodeHover = this.handleNodeHover.bind(this);
+
+    this.sampleFields = [
+      { name: "project_name", label: "Project" },
+      { name: "sample_location", label: "Location" },
+      { name: "created_at", label: "Upload Date" },
```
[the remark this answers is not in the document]
tfrcarvalho: Good catch. I thought of that when I was adding the field, but forgot to add. Done.

app/assets/src/components/views/phylo_tree/PhyloTreeVis.jsx
```diff
@@ -2,25 +2,38 @@ import React from "react";
...
+      { name: "sample_date", label: "Collection Date" },
+      { name: "sample_tissue", label: "Tissue" },
+      { name: "sample_template", label: "Template" },
+      { name: "sample_library", label: "Library" },
+      { name: "sample_sequencer", label: "Sequencer" },
+      { name: "sample_unique_id", label: "Unique ID" },
+      { name: "sample_notes", label: "Notes" }
```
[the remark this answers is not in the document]
tfrcarvalho: Added.

tfrcarvalho:
@cdebourcy I think that can be addressed during stabilization. It might take me a while to figure out a good solution for that. Scrolling would not work because as you scroll, you probably lose the tooltip.

PR #1535, opened 2018-10-01
Remodeled Taxon Tree

tfrcarvalho:
# Description

**Type of Change**: New/Remodeled/Refactored Feature

* Overall code refactoring into JS-D3/React elements
* D3 code refactoring
* Restyled Tree
* Added Pathogen labels as React overlay on JS (keep same code for the labels)
* Highlighting top taxa branches
* Ability to collapse only less relevant children.

## Testing

* Did some preliminary testing, but **more testing is needed** before mergind.


app/assets/src/components/views/TaxonTreeVis.jsx
```diff
@@ -0,0 +1,296 @@
...
+
+    this.tree = null;
+    this.treeVis = null;
+
+    this.metrics = {
+      aggregatescore: {
+        label: "Aggregate Score",
+        agg: arr => Math.max(...arr)
+      },
+      nt_r: { label: "NT - r", agg: arr => arr.reduce((a, b) => a + b, 0) },
```
cdebourcy: can we use something other than "-"? It looks like "minus"

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -0,0 +1,547 @@
...
+        addOverlays: true,
+        attribute: "aggregatescore",
+        leafNodeHeight: 35,
+        minWidth: 1000,
+        minHeight: 300,
+        transitionDuration: 500,
+        onNodeHover: null,
+        onCreatedTree: null,
+        tooltipContainer: null,
+        collapseThreshold: 0.4,
```
cdebourcy: It would be amazing if you could take this as a prop coming straight from the database, because we might want to tweak it for demos on short notice without deploying code changes. There's a table called `ui_configs`

app/assets/src/components/views/TaxonTreeVis.jsx
tfrcarvalho: Sure. What do you suggest?

cdebourcy: I'd have suggested no delimiter at all, since that's how it is for the threshold filters in the report

app/assets/src/components/visualizations/TidyTree.js
tfrcarvalho: Maybe, I would prefer to make a UI choice with a slider or something.

cdebourcy: that would also be cool

db/schema.rb
```diff
@@ -25,13 +25,16 @@
...
+  create_table "amr_counts", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=latin1" do |t|
     t.string "gene"
     t.string "allele"
     t.float "coverage", limit: 24
     t.float "depth", limit: 24
     t.bigint "pipeline_run_id"
     t.string "drug_family"
+    t.integer "level"
+    t.float "drug_gene_coverage", limit: 24
+    t.float "drug_gene_depth", limit: 24
```
jshoe: what are these additions for again? intentional?

app/assets/src/components/views/TaxonTreeVis.jsx
tfrcarvalho: Sure, will remove.

db/schema.rb
tfrcarvalho: This comes from some merge ... Did not add this.

PR #1543, opened 2018-10-01
minor fixes mostly to development env

boris-dimitrov:
wrong way to fix it (should probably use host_genomes.rake?) but at least now

  * clicking on a seed sample no longer crashes the web page

  * the pipeline monitor is no longer prevented from running its state machine for each seed sample by the null host genome id

 more details on the entire constellation of broken seed db https://jira.czi.team/browse/IDSEQ-391

in addition, we fix a minor bug in the result monitor, and adapt the docker build script to a recent update in aegea (without which it fails to build...)

### DIFF
--- a/lib/tasks/pipeline_monitor.rake
+++ b/lib/tasks/pipeline_monitor.rake
@@ -38,7 +38,8 @@ class CheckPipelineRuns
     # Force refresh well before autoscaling.EXPIRATION_PERIOD_MINUTES.
     # prod does it more often because it needs to pick up updates from
     # all other environments and adjust the autoscaling groups.
-    Rails.env == "prod" ? 60 : 600
+    cloud_env = ["prod", "staging"].include?(Rails.env)
+    Rails.env == cloud_env ? 60 : 600
   end
 
   def self.autoscaling_update(autoscaling_state, t_now)
@@ -109,7 +110,8 @@ task "pipeline_monitor", [:duration] => :environment do |_t, args|
   # spawn a new finite duration process every 60 minutes
   respawn_interval = 60 * 60
   # rate-limit status updates
-  checks_per_minute = 4.0
+  cloud_env = ["prod", "staging"].include?(Rails.env)
+  checks_per_minute = cloud_env ? 4.0 : 1.0
   # make sure the system is not overwhelmed under any cirmustances
   wait_before_respawn = 5
   additional_wait_after_failure = 25
--- a/lib/tasks/result_monitor.rake
+++ b/lib/tasks/result_monitor.rake
@@ -34,10 +34,6 @@ class MonitorPipelineResults
     end
   end
 
-  def self.forced_update_interval
-    600
-  end
-
   def self.run(duration, min_refresh_interval)
     Rails.logger.info("Monitoring results for the active pipeline runs every #{min_refresh_interval} seconds over the next #{duration / 60} minutes.")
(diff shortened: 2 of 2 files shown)


app/models/pipeline_run.rb
```diff
@@ -493,8 +493,9 @@ def monitor_results
 
     # Get pipeline_version, which determines S3 locations of output files.
     # If pipeline version is not present, we cannot load results yet.
+    # Except, if the pipeline run is finalized, we have to (this is a failure case).
     update_pipeline_version(self, :pipeline_version, pipeline_version_file)
-    return if pipeline_version.blank?
+    return if pipeline_version.blank? && !finalized
```
boris-dimitrov: @jameshowardwang  I am sneaking a tricker one past you.   This was actually responsible for some of the problems in the seed env, and can also happen in production if the batch job gets killed before it starts, or otherwise fails to produce a version file.  The result monitor would keep iterating over that job, without ever marking its status as "failed".   With this fix, it will check the (lack of) results, and mark it as failed.

jameshowardwang: 😮

jameshowardwang: very nice catch, thanks boris!

boris-dimitrov:
This does appear to fix the pipeline and RM issues in my development environment, yay.

lib/tasks/pipeline_monitor.rake
```diff
@@ -38,7 +38,8 @@ class CheckPipelineRuns
     # Force refresh well before autoscaling.EXPIRATION_PERIOD_MINUTES.
     # prod does it more often because it needs to pick up updates from
     # all other environments and adjust the autoscaling groups.
-    Rails.env == "prod" ? 60 : 600
+    cloud_env = ["prod", "staging"].include?(Rails.env)
+    Rails.env == cloud_env ? 60 : 600
```
boris-dimitrov: staging is now an independent clound env and we should make a scaling decision for it every 60 seconds

lib/tasks/pipeline_monitor.rake
```diff
@@ -109,7 +110,8 @@ task "pipeline_monitor", [:duration] => :environment do |_t, args|
   # spawn a new finite duration process every 60 minutes
   respawn_interval = 60 * 60
   # rate-limit status updates
-  checks_per_minute = 4.0
+  cloud_env = ["prod", "staging"].include?(Rails.env)
+  checks_per_minute = cloud_env ? 4.0 : 1.0
```
boris-dimitrov: 4 checks per minute makes laptop hot

lib/tasks/result_monitor.rake
```diff
@@ -76,7 +72,8 @@ task "result_monitor", [:duration] => :environment do |_t, args|
   # spawn a new finite duration process every 60 minutes
   respawn_interval = 60 * 60
   # rate-limit status updates
-  checks_per_minute = 4.0
+  cloud_env = ["prod", "staging"].include?(Rails.env)
+  checks_per_minute = cloud_env ? 4.0 : 0.2
```
boris-dimitrov: 4 checks per minute makes laptop very hot!

1 check every 5 minutes is barely acceptable... due to the result monitor inefficiency

this wasn't noticeable when the environment was totally broken, but since the rest of this pull request fixes the environment, heat becomes a concern

PR #1544, opened 2018-10-01
speed up taxon search

cdebourcy:


tfrcarvalho:
Just by curiosity: how much data is this?

cdebourcy:
> Just by curiosity: how much data is this?

It's 85,000 rows

PR #1552, opened 2018-10-02
Add placeholder announcements bar for later + UserMailer fixes

jshoe:
- Add placeholder "bulletin" for the announcements bar (see pic). Commented out for now
- Use React SUI Container for smaller left/right margins in mobile view
- Fix bug in UserMailer form
- Standardize app emails to accounts@idseq.net

screen shot 2018-10-02 at 2 55 43 pm

### DIFF
--- a/app/assets/src/components/views/Landing.jsx
+++ b/app/assets/src/components/views/Landing.jsx
@@ -8,6 +8,7 @@ import {
   Grid,
   Image,
   Message,
+  Container,
   Divider
 } from "semantic-ui-react";
 import TransparentButton from "../ui/controls/buttons/TransparentButton";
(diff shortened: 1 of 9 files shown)


app/assets/src/components/views/Landing.jsx
```diff
@@ -8,6 +8,7 @@ import {
   Grid,
   Image,
   Message,
+  Container,
```
tfrcarvalho: I actually missed this. This goes agains our policy of wrapping semantic ui element in our elements.

app/assets/src/components/views/Landing.jsx
```diff
@@ -203,6 +206,22 @@ class Landing extends React.Component {
       </div>
     );
 
+    // const bulletinLink = () => (location.href = "/");
```
tfrcarvalho: Why not add this for admins only instead of commenting?
At least we can also see how it is going to look.

What is the purpose of merging this as a comment?

tfrcarvalho: But you can put it as a javascript flag, right? Maybe a URL parameter. Not necessarily admin.

PR #1555, opened 2018-10-03
only one aggregatescore on heatmap metric selector

cdebourcy:
Show only one aggregatescore, not NT.aggregatescore and NR.aggregatescore (they are the same) -- JIRA-337. 
Tiago's & Jonathan's comment that I should define a display name led to the realization that such names are defined 5 times over in different parts of the app. This is a first attempt at standardization so we only need to change code in 1 place and all metric names are consistent.

### DIFF
--- a/app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
+++ b/app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
@@ -3,13 +3,17 @@ import { forbidExtraProps } from "airbnb-prop-types";
 import PrimaryButton from "../buttons/PrimaryButton";
 import PropTypes from "prop-types";
 import RemoveIcon from "../../icons/RemoveIcon";
+import { METRIC_NAMES } from "../../../utils/Metrics";
 import React from "react";
 
 class ThresholdFilterDropdown extends React.Component {
   constructor(props) {
     super(props);
 
-    this.metrics = (this.props.options || {}).targets || [];
+    let targets = (this.props.options || {}).targets || [];
+    this.metrics = targets.map(metric => {
+      return { text: METRIC_NAMES[metric], value: metric };
+    });
     this.operators = (this.props.options || {}).operators || [];
     this.label = this.props.label || "Threshold Filters:";
 
--- a/app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
+++ b/app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
@@ -4,7 +4,6 @@ import PrimaryButton from "../buttons/PrimaryButton";
 import PropTypes from "prop-types";
 import RemoveIcon from "../../icons/RemoveIcon";
 import {
-  METRIC_NAMES,
   THRESHOLD_METRICS,
   MetricTextAndValue
 } from "../../../utils/Metrics";
--- a/app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
+++ b/app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
@@ -3,16 +3,18 @@ import { forbidExtraProps } from "airbnb-prop-types";
 import PrimaryButton from "../buttons/PrimaryButton";
 import PropTypes from "prop-types";
 import RemoveIcon from "../../icons/RemoveIcon";
-import { METRIC_NAMES, THRESHOLD_METRICS } from "../../../utils/Metrics";
+import {
+  METRIC_NAMES,
+  THRESHOLD_METRICS,
+  MetricTextAndValue
+} from "../../../utils/Metrics";
 import React from "react";
 
 class ThresholdFilterDropdown extends React.Component {
   constructor(props) {
     super(props);
 
-    this.metrics = THRESHOLD_METRICS.map(metric => {
-      return { text: METRIC_NAMES[metric], value: metric };
-    });
+    this.metrics = MetricTextAndValue(THRESHOLD_METRICS);
     this.operators = [">=", "<="];
     this.label = this.props.label || "Threshold Filters:";
 
--- a/app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
+++ b/app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
@@ -3,18 +3,17 @@ import { forbidExtraProps } from "airbnb-prop-types";
 import PrimaryButton from "../buttons/PrimaryButton";
 import PropTypes from "prop-types";
 import RemoveIcon from "../../icons/RemoveIcon";
-import { METRIC_NAMES } from "../../../utils/Metrics";
+import { METRIC_NAMES, THRESHOLD_METRICS } from "../../../utils/Metrics";
 import React from "react";
 
 class ThresholdFilterDropdown extends React.Component {
   constructor(props) {
     super(props);
 
-    let targets = (this.props.options || {}).targets || [];
-    this.metrics = targets.map(metric => {
+    this.metrics = THRESHOLD_METRICS.map(metric => {
       return { text: METRIC_NAMES[metric], value: metric };
     });
-    this.operators = (this.props.options || {}).operators || [];
+    this.operators = [">=", "<="];
     this.label = this.props.label || "Threshold Filters:";
 
     this.state = {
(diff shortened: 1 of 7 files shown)


jshoe:
should NT.aggregatescore be renamed?

app/controllers/samples_controller.rb
```diff
@@ -228,7 +228,6 @@ def heatmap
         "NT.r",
```
tfrcarvalho: We should remove the NT too for the aggregate score.
The plan here is to send a full object: `[{name: ..., value: ...}]` so that metrics show up properly and values are still understanded by the server.

app/assets/src/components/utils/Metrics.js
```diff
@@ -0,0 +1,20 @@
+let metric_names = {
```
tfrcarvalho: replace by `const MetricName`;

app/controllers/samples_controller.rb
```diff
@@ -223,36 +223,18 @@ def heatmap
         Viruses: ["Phage"]
       },
       metrics: [
+        "NT.r", # first entry will be default selection
```
tfrcarvalho: Better do default selection in the client. Just select the value instead of the first one, so that the metrics are ordered.

cdebourcy: ok

app/assets/src/components/utils/Metrics.js
```diff
@@ -0,0 +1,54 @@
...
+for (var m in metric_names) {
+  metric_names_augmented[m] = metric_names[m];
+  metric_names_augmented[m.replace(".", "_")] = metric_names[m];
+  metric_names_augmented[m.replace(".", "_").toLowerCase()] = metric_names[m];
+}
+metric_names_augmented["aggregatescore"] =
+  metric_names_augmented["NT.aggregatescore"];
+export const METRIC_NAMES = metric_names_augmented;
+
+export const THRESHOLD_METRICS = [
```
tfrcarvalho: These are the name that will show up? 
Should we use proper spacing and words, i.e. for instance `NT Percent Identity`

app/assets/src/components/views/TaxonTreeVis.jsx
```diff
@@ -30,15 +31,33 @@ class TaxonTreeVis extends React.Component {
 
     this.metrics = {
       aggregatescore: {
-        label: "Aggregate Score",
+        label: METRIC_NAMES["aggregatescore"],
```
tfrcarvalho: No need to add label here. Remove the labels here and call `METRIC_NAMES[metric]` below instead.

app/controllers/samples_controller.rb
```diff
@@ -223,36 +223,18 @@ def heatmap
...
         "NT.rpm",
-        "NT.r",
         "NT.zscore",
         "NT.maxzscore",
-        "NR.aggregatescore",
         "NR.rpm",
         "NR.r",
-        "NR.zscore",
-        "NR.maxzscore"
+        "NR.zscore"
```
tfrcarvalho: Why did you remove NR.maxzscore?

app/controllers/samples_controller.rb
```diff
@@ -223,36 +223,18 @@ def heatmap
         Viruses: ["Phage"]
       },
       metrics: [
+        "NT.r",
```
tfrcarvalho: You already have default value above. You can reorder these.

app/assets/src/components/utils/Metrics.js
```diff
@@ -0,0 +1,54 @@
+const metric_names = {
```
tfrcarvalho: Naming to camel case.

app/assets/src/components/utils/Metrics.js
```diff
@@ -0,0 +1,54 @@
...
+  "NT.maxzscore": "Max z-score",
+  "NT.percentidentity": "NT %id",
+  "NT.neglogevalue": "NT log(1/e)",
+  "NR.rpm": "NR rPM",
+  "NR.r": "NR r (reads)",
+  "NR.zscore": "NR z-score",
+  "NR.percentidentity": "NR %id",
+  "NR.neglogevalue": "NR log(1/e)"
+};
+const metric_names_augmented = {};
```
tfrcarvalho: Naming to camelCase.

app/controllers/samples_controller.rb
cdebourcy: because it is the same as NT.maxzscore. `maxzscore == max(NT.zscore, NR.zscore)`

app/assets/src/components/views/TaxonTreeVis.jsx
cdebourcy: thanks

app/controllers/samples_controller.rb
cdebourcy: ok

app/assets/src/components/utils/Metrics.js
cdebourcy: ok

app/assets/src/components/utils/Metrics.js
cdebourcy: ok

app/controllers/samples_controller.rb
cdebourcy: fixed

app/controllers/samples_controller.rb
tfrcarvalho: I see. It confused me because here still has the NT prefix. Thanks

app/assets/src/components/views/TaxonTreeVis.jsx
```diff
@@ -30,15 +31,26 @@ class TaxonTreeVis extends React.Component {
...
-        label: "Aggregate Score",
         agg: arr => Math.max(...arr)
       },
-      nt_r: { label: "NT r", agg: arr => arr.reduce((a, b) => a + b, 0) },
-      nt_rpm: { label: "NT rpm", agg: arr => arr.reduce((a, b) => a + b, 0) },
-      nt_zscore: { label: "NT Z-Score", agg: arr => Math.max(...arr) },
-      nr_r: { label: "NR r", agg: arr => arr.reduce((a, b) => a + b, 0) },
-      nr_rpm: { label: "NR rpm", agg: arr => arr.reduce((a, b) => a + b, 0) },
-      nr_zscore: { label: "NR Z-Score", agg: arr => Math.max(...arr) }
+      nt_r: {
```
tfrcarvalho: Sorry for another ask, but we do no need the objects now. Can be just a mapping of metrics to aggregators.

cdebourcy: What do you mean by objects?

cdebourcy: oh i see

app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
```diff
@@ -3,14 +3,18 @@ import { forbidExtraProps } from "airbnb-prop-types";
 import PrimaryButton from "../buttons/PrimaryButton";
 import PropTypes from "prop-types";
 import RemoveIcon from "../../icons/RemoveIcon";
+import {
```
tfrcarvalho: Can we keep this drop down generic, please? 
Use the props as before and pass the values from outside.

cdebourcy: Well, it is called "ThresholdFilterDropdown". Why is it inappropriate to use a constant called THRESHOLD_METRICS if this is the level of abstraction we're operating at?

cdebourcy: I added this constant here because the props in all places where it was used for duplicated. IMO, we can reevaluate if we ever start using a threshold filter for metrics that are not related to the current ones. Right now, this use case is not even on the horizon, so I feel using props for data that never changes is misleading

tfrcarvalho: Because you cannot ever customize the metrics you want to use it with for different pages.
ThresholdFilterDropdown only means that you are applying a group of logical comparisons to a group of metrics.
I made it generic on purpose for easy reusability. 

What you are doing here is simply passing props, thorugh constants.

If you really want to make a default one, use default props to set defaults.
Also, please consolidate the two imports by defining something that equates to `MetricTextAndValue(THRESHOLD_METRICS)` in the `Metrics.js`.

cdebourcy: ok

PR #1561, opened 2018-10-03
Fix adhoc due to family level taxa in taxon counts

tfrcarvalho:
The pipeline is generating taxon counts for taxa at the family rank. The report helper was only designed to handle taxa up to level 2.
Added tax_level filter.

### DIFF
--- a/app/lib/report_helper.rb
+++ b/app/lib/report_helper.rb
@@ -291,7 +291,7 @@ module ReportHelper
         pipeline_run_id = #{pipeline_run_id.to_i} AND
         taxon_counts.genus_taxid != #{TaxonLineage::BLACKLIST_GENUS_ID} AND
         taxon_counts.count_type IN #{count_types} AND
-        taxon_counts.tax_level < 3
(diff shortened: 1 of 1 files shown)


boris-dimitrov:
The problem here is that fake genera are being created for family level counts;  this just needs a simple conjunct to limit the behavior to species like so

```unless taxon_counts_2d[genus_taxid] and tax_level == species```

Adding a family filter at this level is potentially more disruptive and will require additional testing, so I think should be outside the scope of stabilization.

Also, the bug here deserves an explanation.   It has been with us since April, but never showed up on real experiments, because all real experiments have -200 genus count bucket already.   The family counts reference genus -200, and it already existed, so no new fake genus had to be created.

In the benchmark runs, there was not a -200 count to start with because we had classified all the reads to real genera.  So then for the family count that referenced -200 genus, a new fake genus had to be created.

The proof is that adding -200 count to the benchmark runs hides the bug (I confirmed that for pr 13800 sample 8845).

PR #1564, opened 2018-10-03
Cleanup deleted samples

jameshowardwang:
This script downloads all sample IDs from prod and also enumerates all sample paths under s3, then compares the two. For any sample data that is present only in s3, the script deletes all versions of that sample's data in s3.

Test plan:
Tested a whole bunch and ran under staging


bin/remove_deleted_samples_from_s3.py
```diff
@@ -0,0 +1,59 @@
...
+  
+def db_sample_dict():
+  return_dict = {}
+  db_samples_file = open(DB_SAMPLES_FILENAME, 'r')
+  line = db_samples_file.readline() # first line says 'id'
+  while True:
+    line = db_samples_file.readline().strip()
+    if not line:
+      break
+    return_dict[line] = 1
```
cdebourcy: nit: just use a set? :)

PR #1566, opened 2018-10-03
Fix error on Edge

jshoe:
- So silly..

Error:
screen shot 2018-10-03 at 3 02 42 pm

Solution:
screen shot 2018-10-03 at 2 04 22 pm


jameshowardwang:
might there be more of these lurking in our code?

jshoe:
I did a quick code scan and did not find any more

PR #1570, opened 2018-10-04
Taxon Tree: Bug fixes

tfrcarvalho:
# Description:

* Remove z metrics: they tend to expand too many nodes and are not useful for the tree;
* Fix bug where descendants with uncategorized parents were attributed the wrong ancestors;
* Reset tree parameters that might change when options like metric and categories change ([Jira-447](https://jira.czi.team/browse/IDSEQ-447) was one of the consequences of not cleaning the tree);
* Move assignment of click action to update, so that it applies to all nodes.
* Fix a bug where parameter that defines the minimum number of nodes to trigger siblings collapse was not respected (e.g. would cause 1 node to be represented as collapased node).

## Type of change

### DIFF
--- a/app/assets/src/components/views/TaxonTreeVis.jsx
+++ b/app/assets/src/components/views/TaxonTreeVis.jsx
@@ -35,10 +35,8 @@ class TaxonTreeVis extends React.Component {
       },
       nt_r: { label: "NT r", agg: arr => arr.reduce((a, b) => a + b, 0) },
       nt_rpm: { label: "NT rpm", agg: arr => arr.reduce((a, b) => a + b, 0) },
-      nt_zscore: { label: "NT Z-Score", agg: arr => Math.max(...arr) },
       nr_r: { label: "NR r", agg: arr => arr.reduce((a, b) => a + b, 0) },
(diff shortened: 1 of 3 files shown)


app/assets/src/components/views/TaxonTreeVis.jsx
```diff
@@ -177,6 +177,7 @@ class TaxonTreeVis extends React.Component {
       let nodeId = taxon.tax_id;
       nodes.push({
         id: nodeId,
+        taxId: nodeId,
```
jameshowardwang: why is taxId being set to nodeId here?

tfrcarvalho: **Short answer**: For these nodes (tax level 1 and 2, species and genus) node id is always equal to taxon id. I could just set id, and consider taxid to be equal to id when not set, but I believe it is cleaner this way.

**Long answer**:
Tax id and node can be different, but not always as in the case of some ancestors. They are different when the tax id in the data is a generic negative number. In that case the node id is composed of the generic tax id concatenated with a prefix with all the previous ancestors until the last known/unique one.

Tax id is going to be used to link other application places. 
Node id is going to be used for internal tree processing, and make sure that some nodes that have the same generic negative tax id, can be differentiated if they have different ancestry.

PR #1573, opened 2018-10-04
JIRA-432: disable phylo_tree download button if output is not available

cdebourcy:
New column snp_annotations will serve as (a) an indicator on whether the file is available, (b) as a permanent materialized record of the file's location. With pipeline_runs we sometimes run into issues because the location function changes, so the new column will keep track of phylo_tree location permanently for future reference.


app/assets/src/components/views/phylo_tree/PhyloTreeListView.jsx
```diff
@@ -88,10 +88,11 @@ class PhyloTreeListView extends React.Component {
         <div className="phylo-tree-list-view__narrow-container">
           <ViewHeader title="Phylogenetic Trees">
             <DownloadButtonDropdown
-              options={[{ text: "SNP annotations", value: "SNP_annotations" }]}
+              options={[{ text: "SNP annotations", value: "snp_annotations" }]}
               onClick={() => {
                 location.href = `/phylo_trees/${currentTree.id}/download_snps`;
               }}
+              disabled={currentTree.snp_annotations == null}
```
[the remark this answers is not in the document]
cdebourcy: That's the opposite logic... I will use `!currentTree.snp_annotations`. Thanks

app/controllers/phylo_trees_controller.rb
```diff
@@ -127,10 +127,8 @@ def retry
 
   def download_snps
     snp_file = Tempfile.new
-    s3_file = @phylo_tree.s3_outputs["SNP_annotations"]
-    cmd_status = Open3.capture3("aws", "s3", "cp", s3_file, snp_file.path)[2]
-    unless cmd_status.success?
-      snp_file.write("Not yet available.")
+    s3_file = @phylo_tree.snp_annotations
+    unless s3_file && Open3.capture3("aws", "s3", "cp", s3_file, snp_file.path)[2].success?
```
yunfangjuan: we have various download function from PipelineRun, reuse?

db/migrate/20181004180225_add_snp_annotations_to_phylo_trees.rb
```diff
@@ -0,0 +1,14 @@
+require 'open3'
+
+class AddSnpAnnotationsToPhyloTrees < ActiveRecord::Migration[5.1]
+  def change
+    add_column :phylo_trees, :snp_annotations, :string
+
+    PhyloTree.all.each do |pt|
+      s3_path = pt.s3_outputs["snp_annotations"]["s3_path"]
+      if Open3.capture3("aws", "s3", "ls", s3_path)[2].success?
```
yunfangjuan: we should put this Open3.capture3("aws", "s3", "ls", s3_path)[2].success? into a more robust function.

app/controllers/phylo_trees_controller.rb
cdebourcy: ok. The download functions in PipelineRun are a bit different in the arguments (destination directory instead of file name) and behavior (create a new file, whereas here I want to use the tempfile I created previously). I will move my code to PipelineRunsHelper so it can be reused in the future.

db/migrate/20181004180225_add_snp_annotations_to_phylo_trees.rb
cdebourcy: moved to PipelineRunsHelper for future iteration

PR #1588, opened 2018-10-05
Fix project and sample deletion to be efficient and complete

jameshowardwang:
Also fixed an issue with rails console

Test plan: did sample and project deletion under rails c --sandbox

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -146,4 +146,10 @@ class Project < ApplicationRecord
                 .group("project_id")
                 .pluck(:project_id))
   end
+
+  def destroy
+    samples = Sample.find(sample_ids)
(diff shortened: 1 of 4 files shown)


app/models/project.rb
```diff
@@ -146,4 +146,10 @@ def self.public_projects
                 .group("project_id")
                 .pluck(:project_id))
   end
+
+  def destroy
+    samples = Sample.find(sample_ids)
+    samples.each(&:destroy)
```
tfrcarvalho: Why not `.where(...).destroy_all`?

jameshowardwang: mostly because i was unaware of that way of doing it. i think they're equivalent though, no?

PR #1595, opened 2018-10-06
Speed up docker-compose builds

jameshowardwang:
Make builds go faster (and bin/setup as well) by only building one docker image for all the rails-dependent services.


jameshowardwang:
web still has "build ."

PR #1596, opened 2018-10-06
JIRA-463: Expand genera containing species with pathogen tags by default. Also remove dead PathogenSummary code.

cdebourcy:
This is one of the hairiest parts of our codebase. Hopefully we can refactor it soon.


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -153,6 +151,7 @@ class PipelineSampleReport extends React.Component {
     this.expandAll = false;
     this.expandedGenera = [];
     this.thresholded_taxons = [];
+    this.row_class = this.row_class.bind(this);
```
jshoe: row_class isn't actually used, right?

tfrcarvalho: camelcase, please :)

cdebourcy: ok done

PR #1603, opened 2018-10-08
Fix crash metadata

tfrcarvalho:
We seem to assume that we always receive strings as a metadata. This PR check if we receive a string before applying trim function.

The root cause of the problem might be on: 
- new mapping of fields to render functions is not correct
- weird metadata format sent on server side or new unsupported field.


jameshowardwang:
looks good, but how do we know this fixes the issue?

PR #1608, opened 2018-10-08
Fix user name display for phylo_trees

cdebourcy:
(a) Prevent the site from breaking if there is no user for a phylo_tree
(b) Fix an inefficiency in the way the usernames are retrieved

### DIFF
--- a/app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
+++ b/app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
@@ -248,7 +248,7 @@ class PhyloTreeCreation extends React.Component {
   parsePhyloTreeData(phyloTreeData) {
     return phyloTreeData.map(row => ({
       name: row.name,
-      user: row.user ? row.user.name : null,
+      user: (row.user || {}).name,
       last_update: <Moment fromNow date={row.updated_at} />,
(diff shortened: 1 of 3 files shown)


app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
```diff
@@ -248,7 +248,7 @@ class PhyloTreeCreation extends React.Component {
   parsePhyloTreeData(phyloTreeData) {
     return phyloTreeData.map(row => ({
       name: row.name,
-      user: row.user.name,
+      user: row.user ? row.user.name : null,
```
tfrcarvalho: This works but I like the short `(row.user || {}).name`

cdebourcy: cool

app/controllers/phylo_trees_controller.rb
```diff
@@ -53,10 +53,11 @@ def index
 
     @phylo_trees = @phylo_trees.as_json(include: { user: { only: [:id, :name] } })
 
-    # Augment tree data with sample attributes and number of pipeline_runs
+    # Augment tree data with sample attributes, number of pipeline_runs, user name
```
tfrcarvalho: nit `s/,/ and/`

cdebourcy: ok

PR #1609, opened 2018-10-08
Fix taxon count db loading

yunfangjuan:
# Description

Fixing constant db loading error. (caused when the records are already loaded.) 
https://jira.czi.team/browse/IDSEQ-471

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -337,9 +337,9 @@ class PipelineRun < ApplicationRecord
 
     # check if there's any record loaded into taxon_counts. If so, skip
     check_count_type = refined ? 'NT+' : 'NT'
-    records = TaxonCount.where(pipeline_run_id: id)
-                        .where(count_type: check_count_type).count
-    return if records > 0
+    loaded_records = TaxonCount.where(pipeline_run_id: id)
+                               .where(count_type: check_count_type).count
+    return if loaded_records > 0
 
     # TODO: remove the following. deprecated.
     # version_s3_path = "#{alignment_output_s3_path}/#{VERSION_JSON_NAME}"
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -335,6 +335,12 @@ class PipelineRun < ApplicationRecord
     pipeline_output_dict = json_dict['pipeline_output']
     pipeline_output_dict.slice!('taxon_counts_attributes')
 
+    # check if there's any record loaded into taxon_counts. If so, skip
+    check_count_type = refined ? 'NT+' : 'NT'
+    records = TaxonCount.where(pipeline_run_id: id)
+                        .where(count_type: check_count_type).count
+    return if records > 0
+
     # TODO: remove the following. deprecated.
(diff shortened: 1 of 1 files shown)


app/models/pipeline_run.rb
```diff
@@ -723,7 +729,7 @@ def generate_aggregate_counts(tax_level_name)
     uncategorizable_id = TaxonLineage::MISSING_LINEAGE_ID.fetch(tax_level_name.to_sym, -9999)
     uncategorizable_name = "Uncategorizable as a #{tax_level_name}"
     TaxonCount.connection.execute(
-      "INSERT INTO taxon_counts(pipeline_run_id, tax_id, name,
+      "REPLACE INTO taxon_counts(pipeline_run_id, tax_id, name,
```
jameshowardwang: do you want REPLACE INTO or INSERT IGNORE INTO? are these always the exact same rows or are they sometimes actually being updated?

cdebourcy: Trying to understand better... wouldn't the block you added above remove the need for "replace"?

app/models/pipeline_run.rb
```diff
@@ -335,6 +335,12 @@ def load_taxons(downloaded_json_path, refined = false)
     pipeline_output_dict = json_dict['pipeline_output']
     pipeline_output_dict.slice!('taxon_counts_attributes')
 
+    # check if there's any record loaded into taxon_counts. If so, skip
+    check_count_type = refined ? 'NT+' : 'NT'
+    loaded_records = TaxonCount.where(pipeline_run_id: id)
+                               .where(count_type: check_count_type).count
+    return if loaded_records > 0
```
jameshowardwang: is this the right thing to do? do all the records get loaded in one transaction? or is it possible that some taxons are loaded but not all of them?

app/models/pipeline_run.rb
yunfangjuan: yes. i think so because we have NT and NT+ now.

app/models/pipeline_run.rb
yunfangjuan: they are loaded in one transaction because how update_attributes work

app/models/pipeline_run.rb
yunfangjuan: because we have NT/NT+ we kinda need this.

cdebourcy: Isn't it handled by "count_type"?

yunfangjuan: no. because this function will be called twice and the second time NT/NR records are already there, this will cause an insert error.

PR #1615, opened 2018-10-08
Report page: Highlight pathogen name on hover to show it's clickable

jshoe:
oct-08-2018 15-49-50


jshoe:
@tfrcarvalho moved into (a) .report-table section. Verified that hover-wrapper and link-tag are only in the report table sections / still working.

PR #1622, opened 2018-10-10
sanitize project name

cdebourcy:


app/controllers/phylo_trees_controller.rb
```diff
@@ -51,9 +51,8 @@ def index
                  name: taxon_lineage.name }
     end
 
-    @phylo_trees = @phylo_trees.as_json(include: { user: { only: [:id, :name] } })
-
```
cdebourcy: This change is unrelated to the main diff, sorry for mixing in. It completes a change I had made previously (user data is filled in a few lines below, so no need for inefficient "include: user"). I had neglected to remove the now unnecessary code.

app/helpers/application_helper.rb
```diff
@@ -1,7 +1,21 @@
 require 'csv'
 require 'json'
+require 'open3'
 
 module ApplicationHelper
+  def sanitize(user_input_text)
+    # accept alpanumeric characters only
+    user_input_text.gsub(/[^A-Za-z0-9]/, '-')
```
jshoe: for the regex, we still want to accept spaces in the project names right? I think we can include space. As long as the params are separated Open3 will handle the spaces apparently:

```
irb(main):010:0> stdin, stdout, stderr = popen3("aws", "s3", "ls", "s3://idseq-samples-staging && echo 'HELLO'")
=> [#<IO:fd 18>, #<IO:fd 20>, #<IO:fd 22>, #<Process::Waiter:0x00007fa44f14e4c0 sleep>]
irb(main):011:0> puts stdout.read

=> nil
irb(main):012:0> stderr.read
=> "\nParameter validation failed:\nInvalid bucket name \"idseq-samples-staging && echo 'HELLO'\": Bucket name must match the regex \"^[a-zA-Z0-9.\\-_]{1,255}$\"\n"
irb(main):013:0>
```

cdebourcy: That's true. I will make the regex more permissive

cdebourcy: thanks

jshoe: "_" and "-" too probably

app/controllers/projects_controller.rb
```diff
@@ -254,6 +255,7 @@ def set_project
 
   # Never trust parameters from the scary internet, only allow the white list through.
   def project_params
+    params[:name] = sanitize(params[:name])
```
jshoe: :name is nested right?

cdebourcy: yes. thanks

cdebourcy: moved sanitize to after the require/permit, at which point name is no longer nested

app/helpers/application_helper.rb
```diff
@@ -1,7 +1,21 @@
 require 'csv'
 require 'json'
+require 'open3'
 
 module ApplicationHelper
+  def sanitize(user_input_text)
+    # accept alpanumeric characters only
```
jshoe: nit: alphanumeric. or 'limited' or something

cdebourcy: ok fixed

app/helpers/application_helper.rb
```diff
@@ -1,7 +1,21 @@
...
+    user_input_text.gsub(/[^A-Za-z0-9_\- ]/, '-')
+  end
+
+  def safe_s3_rm(s3_path)
```
tfrcarvalho: Do you mind putting these two commands in AWS dedicated helper ? 
I would like to centralize every AWS access into AWS access layer.

app/helpers/application_helper.rb
tfrcarvalho: typo: alpanumeric -> alphanumeric

Maybe change to "accept alphanumeric characters and symbols `_` and `-`"

app/helpers/application_helper.rb
cdebourcy: agreed

PR #1625, opened 2018-10-10
always rerun with most current alignment_config

cdebourcy:


cdebourcy:
Closed because it's safer to change alignment_config manually.

PR #1627, opened 2018-10-10
submit benchmark samples based on time or code change triggers

boris-dimitrov:
Submit benchmark samples with specified frequency based on config and input files in s3 generated by idseq-bench.

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -481,6 +481,10 @@ class Sample < ApplicationRecord
     end
   end
 
+  def self.pipeline_commit(branch)
+    `git ls-remote https://github.com/chanzuckerberg/idseq-dag.git | grep refs/heads/#{branch}`.split[0]
+  end
+
   def kickoff_pipeline
     # only kickoff pipeline when no active pipeline_run running
     return unless pipeline_runs.in_progress.empty?
@@ -492,7 +496,7 @@ class Sample < ApplicationRecord
     # but was made an integer type in case we want to allow users to enter the desired number
     # of reads to susbample to in the future
     pr.pipeline_branch = pipeline_branch.blank? ? "master" : pipeline_branch
-    pr.pipeline_commit = `git ls-remote https://github.com/chanzuckerberg/idseq-dag.git | grep refs/heads/#{pr.pipeline_branch}`.split[0]
+    pr.pipeline_commit = Sample.pipeline_commit(pr.pipeline_branch)
 
     pr.alignment_config = AlignmentConfig.find_by(name: alignment_config_name) if alignment_config_name
     pr.alignment_config ||= AlignmentConfig.find_by(name: AlignmentConfig::DEFAULT_NAME)
--- a/config/idseq-bench-config.json
+++ b/config/idseq-bench-config.json
@@ -1,0 +1,55 @@
+{
+  "README": [
+    "********************************************************************************",
+    " This file is under version control at idseq-web/config/idseq-bench-config.json ",
+    " It is consumed by pipeline_monitor.rake.                                       ",
+    "                                                                                ",
+    " You can test and deploy changes to it by running                               ",
+    "                                                                                ",
+    "    aws s3 cp idseq-bench-config.json s3://idseq-web/config.json                ",
+    "                                                                                ",
+    " This process allows changing webapp configuration without the cost             ",
+    " and disruption of a production push.                                           ",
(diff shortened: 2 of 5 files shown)


config/idseq-bench-config.json
```diff
@@ -0,0 +1,55 @@
...
+    "                                                                                ",
+    "    4. After a day or two, move any older entries to retired_benchmarks.        ",
+    "                                                                                ",
+    "********************************************************************************"
+  ],
+  "defaults": {
+    "how_to_use_these_defaults":
+      "These properties may be overridden in each entry of active_benchmarks below.",
+    "project_name": "IDSeq Bench",
+    "user_email": "bdimitrov@chanzuckerberg.com",
```
boris-dimitrov: Before the decision to move this config under version control, I was happy to put my user email here.   Now I am less sure.  Any thoughts/concerns?

app/models/sample.rb
```diff
@@ -481,6 +481,10 @@ def reload_old_pipeline_runs
     end
   end
 
+  def self.pipeline_commit(branch)
```
jshoe: maybe this would make more sense as a PipelineRun function?

config/idseq-bench-config.json
jshoe: Well, it does open it up to spammers more, but I don't really know if there's an alternative? maybe a special benchmark user?

config/idseq-bench-config.json
```diff
@@ -0,0 +1,55 @@
...
+    "    2. aws s3 cp --recursive <idseq-bench-output> s3://idseq-bench/<number>     ",
+    "                                                                                ",
+    "    3. Deploy s3://idseq-bench/<number> by prepending to active_benchmarks.     ",
+    "                                                                                ",
+    "    4. After a day or two, move any older entries to retired_benchmarks.        ",
+    "                                                                                ",
+    "********************************************************************************"
+  ],
+  "defaults": {
+    "how_to_use_these_defaults":
```
tfrcarvalho: What is this property used for?

config/idseq-bench-config.json
tfrcarvalho: What is the email used for? Seems to me that should be some team e-mail... specially given you're leaving.

lib/tasks/pipeline_monitor.rake
```diff
@@ -70,12 +80,172 @@ class CheckPipelineRuns
...
+        next
+      end
+      bm_frequency_hours = prop_get(bm_props, 'frequency_hours', defaults)
+      bm_frequency_seconds = bm_frequency_hours * 3600
+      unless bm_frequency_hours >= IDSEQ_BENCH_MIN_FREQUENCY_HOURS
+        Rails.logger.info("Benchmark frequency under #{IDSEQ_BENCH_MIN_FREQUENCY_HOURS} hour: #{s3path}")
+        next
+      end
+      bm_name_prefix = benchmark_sample_name_prefix(s3path)
+      sql_query = "
```
tfrcarvalho: Why not use active record given that is a single table query?

config/idseq-bench-config.json
boris-dimitrov: comment :)

lib/tasks/pipeline_monitor.rake
boris-dimitrov: Probably a good idea!

app/models/sample.rb
boris-dimitrov: yeah... since it's just a global function, i wish we could put it outside of any ActiveRecord class

config/idseq-bench-config.json
boris-dimitrov: since i'm not leaving CZI, this email will remain valid... but yeah we probably could create a special account, it's just a matter of config change to switch to it

the benchmarking projects will likely be public

jshoe:
@boris-dimitrov could we tone down the run frequency actually? does it run on every web commit?

boris-dimitrov:
@jshoe It would kickoff a run for every pipeline merge and every webapp deploy, but NOT for every webapp merge.  :-)  The frequency is configurable, you can change those params without having to deploy the webapp.   The benchmark is tiny, only 60,000 reads.  Should be alright to run on every webapp deploy.  But there is a flag you can change to "false" in the config if you want it to ignore webapp deploys.

PR #1629, opened 2018-10-10
Add new logo to the headers

jshoe:
screen shot 2018-10-10 at 3 23 41 pm


jshoe:
png is for the favicon :)

PR #1640, opened 2018-10-11
Improve background dropdown

jshoe:
- Won't reload the page if the value didn't change.

- Fixes an issue where the current "value" wasn't being passed down as an int, so the first option was always selected on menu open (and could lead to accidentally changing backgrounds):
 
Before (sorry, hard to see the highlighting on the gif):
oct-11-2018 16-54-08
screen shot 2018-10-11 at 4 45 11 pm

After:
screen shot 2018-10-11 at 4 53 19 pm

### DIFF
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -830,9 +830,15 @@ class PipelineSampleReport extends React.Component {
   }
 
   handleBackgroundModelChange(_, data) {
+    if (data.value === this.state.backgroundId) {
+      // Skip if no change
+      return;
+    }
+
     const backgroundName = data.options.find(function(option) {
       return option.text == data.value;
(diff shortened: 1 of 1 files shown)


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1760,7 +1766,7 @@ function BackgroundModelFilter({ parent }) {
   return (
     <OurDropdown
       options={backgroundOptions}
-      value={parent.state.backgroundId}
+      value={parseInt(parent.state.backgroundId)}
```
tfrcarvalho: Can we do this earlier? 
Do you know what is the source of non-int background id? If is cookies or even props we can parse it as soon as we get it.

jshoe:
Updated. Also background_name in cookies wasn't being set correctly

PR #1643, opened 2018-10-12
Show TaxonModal on Taxonomic Tree View.

MarkAZhang:
TaxonModal is now always open when rendered.
To hide the modal, don't render it in the JSX.
This change allows compatibility with D3, where providing a trigger element is difficult.

Also did a small fix to capitalize common names on the Taxon Tree.

Related to issue: IDSEQ-467

oct-12-2018 15-09-18

### DIFF
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -6,7 +6,7 @@ import Tipsy from "react-tipsy";
 import ReactAutocomplete from "react-autocomplete";
 import { Label, Menu, Icon, Popup } from "semantic-ui-react";
 import numberWithCommas from "../helpers/strings";
-import StringHelper from "../helpers/StringHelper";
+import { getTaxonName } from "../helpers/taxon";
 import ThresholdMap from "./utils/ThresholdMap";
 import Nanobar from "nanobar";
 import BasicPopup from "./BasicPopup";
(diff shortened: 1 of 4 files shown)


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1034,57 +1034,42 @@ class PipelineSampleReport extends React.Component {
     return category_lowercase;
   }
 
-  render_name(tax_info, report_details, parent) {
-    let tax_scientific_name = tax_info["name"];
+  render_name(tax_info, report_details, parent, openTaxonModal) {
     let tax_common_name = tax_info["common_name"];
-    let taxonName;
+    const taxonName = getTaxonName(tax_info, this.state.name_type);
     let taxonNameDisplay;
```
tfrcarvalho: Maybe add `= <span>{taxonName}</span>`; and remove all the elses below.
Allows you to merge he ifs too.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1034,57 +1034,42 @@ class PipelineSampleReport extends React.Component {
...
     } else {
-      taxonName = tax_scientific_name;
-      taxonNameDisplay = <span>{tax_scientific_name}</span>;
+      taxonNameDisplay = <span>{taxonName}</span>;
     }
 
+    const openTaxonModalHandler = () =>
+      openTaxonModal({
+        taxInfo: tax_info,
+        parent,
```
tfrcarvalho: Maybe we can avoid passing parent here? just needs the background info no?

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1572,98 +1564,154 @@ function DetailCells({ parent }) {
...
-                      parent.state.countType === "NR"
-                        ? "active column-switcher"
-                        : "column-switcher"
-                    }
-                    onClick={() => {
-                      parent.setState({ countType: "NR" });
-                    }}
-                  >
-                    NR
+class ReportTableHeader extends React.Component {
```
tfrcarvalho: Can you move this new component to a separate file `views/report`? Maybe we can move `PipelineSampleReport there too to start breaking it apart.

Or if there is to much work for that or circular dependencies we can do it later.

tfrcarvalho: Maybe we should update the name to be more accurate, as this includes the whole table.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1870,6 +1918,8 @@ class RenderMarkup extends React.Component {
           metric={parent.state.treeMetric}
           nameType={parent.state.name_type}
           onNodeTextClicked={this._nodeTextClicked}
+          backgroundId={parent.state.backgroundId}
```
tfrcarvalho: Any reason not to pass an object with `backgroundInfo = {name:..., value:...}` here?

app/assets/src/components/views/TaxonTreeVis.jsx
```diff
@@ -92,6 +98,18 @@ class TaxonTreeVis extends React.Component {
     this.setState({ nodeHover: node });
   }
 
+  onNodeLabelClick(node) {
```
tfrcarvalho: I have been trying to enforce a `handle...` prefix - like `handleNodeLabelClick` - for event handlers.

Same for function below.

app/assets/src/components/views/report/TaxonModal.jsx
```diff
@@ -140,82 +136,82 @@ class TaxonModal extends React.Component {
...
+            </div>
+          )}
+          <div className="taxon-info__subtitle">Links</div>
+          <div className="taxon-info__links-section">
+            <ul className="taxon-info__links-list">
+              <li
+                className="taxon-info__link"
+                onClick={() => this.linkTo("ncbi")}
+              >
+                NBCI
```
tfrcarvalho: You might run into conflicts here once you merge master back... `NBCI` -> `NCBI`

app/assets/src/components/views/report/TaxonModal.jsx
```diff
@@ -18,15 +17,12 @@ class TaxonModal extends React.Component {
       wikiUrl: null
     };
 
-    this.loadedBackground = this.handleOpen = this.handleOpen.bind(this);
-    this.handleClose = () => this.setState({ open: false });
     this.linkTo = this.linkTo.bind(this);
   }
 
-  handleOpen() {
+  componentWillMount() {
```
tfrcarvalho: This function is being deprecated. You should use functions in the new react lifecycle. 
You can probably just use `componentDidMount`.

app/assets/src/components/views/report/TaxonModal.jsx
```diff
@@ -140,82 +136,82 @@ class TaxonModal extends React.Component {
...
+    // <TaxonModal> is always open when rendered. To hide <TaxonModal>, we simply don't render it.
     return (
       <Modal
         title={this.props.taxonName}
-        trigger={<span onClick={this.handleOpen}>{this.props.trigger}</span>}
-        open={this.state.open}
-        onClose={this.handleClose}
+        trigger={<span />}
+        open
+        onClose={this.props.handleClose}
```
tfrcarvalho: Should we add a default behavior that closes the modal? Or is this incompatible with your solution?

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1572,98 +1564,154 @@ function DetailCells({ parent }) {
...
+    const { parent } = this.props;
+
+    return (
+      <div className="reports-main">
+        {this.renderTaxonModal()}
+        <table id="report-table" className="bordered report-table">
+          <thead>
+            <tr className="report-header">
+              <th>
+                <span className={`table-arrow ""`}>
```
jshoe: I know this was here before but the "" doesn't do anything right? in `table-arrow ""`

app/assets/src/components/PipelineSampleReport.jsx
MarkAZhang: Updated the name. Will move to separate file in follow-up PR.

app/assets/src/components/PipelineSampleReport.jsx
tfrcarvalho: sounds good

PR #1645, opened 2018-10-13
Move ReportTable component to separate file.

MarkAZhang:
Follow-up PR.


jshoe:
not sure about best practices but should we go with the index.jsx naming? @tfrcarvalho

app/assets/src/components/views/report/ReportTable/DetailCells.jsx
```diff
@@ -0,0 +1,58 @@
+import React from "react";
```
tfrcarvalho: Current convention is to keep folders in snake case, so: `ReportTable` -> `report_table`

app/assets/src/components/views/report/ReportTable/index.jsx
```diff
@@ -0,0 +1,153 @@
+import React from "react";
+import Tipsy from "react-tipsy";
+
```
tfrcarvalho: Remove blank lines between imports.

Is this order following  some logic of imports? Like third-party imports, idseq components and same package imports?

app/assets/src/components/views/report/ReportTable/DetailCells.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+function DetailCells({ parent, openTaxonModal }) {
```
tfrcarvalho: Our convention is to use the format: `const DetailCells = ({parent, openTaxonModal}) => {...}`

MarkAZhang:
I looked at what the Semantic UI React and Material UI React Github repos do.

The best practice seems to be to have `index.jsx` be a simple redirect file:
`export default from "./ReportTable"`

And have the meat of the code be in `ReportTable.jsx`.

This allows other components to import ReportTable with:
`import ReportTable from "./components/ReportTable"`

instead of:
`import ReportTable from "./components/ReportTable/Reportable"`.

I'll make this change in my file. Thanks for checking!

app/assets/src/components/views/report/ReportTable/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+                          ? "active column-switcher"
+                          : "column-switcher"
+                      }
+                      onClick={() => {
+                        parent.setState({ countType: "NT" });
+                      }}
+                    >
+                      NT
+                    </div>
+                    <div
```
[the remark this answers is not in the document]
MarkAZhang: Next commit.

app/assets/src/components/views/report/ReportTable/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+              {parent.render_column_header(
+                "%conc",
+                `${parent.state.countType}_percentconcordant`,
+                `Percentage of aligned reads belonging to a concordantly mappped pair (NCBI NT/NR)`,
+                parent.showConcordance
+              )}
+              <th>
+                <Tipsy content="Switch count type" placement="top">
+                  <div className="sort-controls center left">
+                    <div
```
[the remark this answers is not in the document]
MarkAZhang: Next commit.

app/assets/src/components/views/report/ReportTable/index.jsx
MarkAZhang: Discussed in person to keep blank lines.

PR #1651, opened 2018-10-16
Refactor getRowClass function.

MarkAZhang:
Remove unused scrollToTaxon function.
Add classnames package to help manage complex class-lists.


MarkAZhang:
This was an exploratory refactor that does change some logic. Won't be doing any more of these until I finish splitting up files and removing parent references.

Included is a pretty big yarn.lock diff...looks like my version of yarn (which is up-to-date) includes integrity SHAs. Should we check those in?

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -1,3 +1,4 @@
+import cx from "classnames";
```
tfrcarvalho: I have a personal preference for more self explanatory names (even if longer). Can we just keep it `classNames` instead of `cx`?

MarkAZhang: In most cases, I would definitely agree but `cx` is widely used for this particular package so I prefer to follow convention.
For example, see Line 1: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/views/Card/Card.js

However if you still feel strongly about it, I am okay with `classNames`.

tfrcarvalho: Sure, we can keep it. I will make a note of common packages that have very specific names.

PR #1653, opened 2018-10-16
Update the RFP banner on master

jshoe:


jshoe:
The RFP closes on Dec 5, 2018, but I don't think the current frequency of updating the banner requires putting in start date/end date logic. (And in the other direction adding more bulletins would require more styling changes)

tfrcarvalho:
@jshoe But should we state the data explicitly in the banner?

jshoe:
Oh sorry, is that what you meant.. I can add that @tfrcarvalho

cdebourcy:
Just trying to avoid a situation in which we forget about the landing page / banner and then end up with stale announcements. Would be easy to add an end date in rails, just like we added a start date

jshoe:
Ok, I'll add that stuff back in then @cdebourcy

PR #1655, opened 2018-10-16
Refactor PipelineSampleReport

MarkAZhang:
Remove the `parent` prop from ReportTable and DetailCells.

The plan is to move up the component tree and do this for every component.

This DOES hugely increase the number of props passed, but these props were all implicitly being used in these components anyways.
Will look into ways to reduce the props after this phase of the refactor (for example, by pushing certain logic and state down the component tree instead of keeping it all in the root component).

### DIFF
--- a/app/assets/src/components/utils/propTypes.js
+++ b/app/assets/src/components/utils/propTypes.js
@@ -1,0 +1,51 @@
+import PropTypes from "prop-types";
+
+const TaxonDatabaseStats = PropTypes.shape({
+  aggregatescore: PropTypes.number.isRequired,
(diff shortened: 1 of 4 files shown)


app/assets/src/components/utils/propTypes.js
```diff
@@ -0,0 +1,51 @@
+import PropTypes from "prop-types";
```
tfrcarvalho: I am not sure if this makes sense in utils for now.... 
I believe it is better to keep this under the "report" view for now, and as we refactor more pages and find commonalities, maybe move it to a different folder. 
We can discuss more about this organization offline.

PR #1660, opened 2018-10-17
Move utility functions out of PipelineSampleReport.

MarkAZhang:
Move some utility functions that don't depend on state in PipelineSampleReport in their own file.

### DIFF
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -7,6 +7,12 @@ import { Label, Menu, Icon, Popup } from "semantic-ui-react";
 import numberWithCommas from "../helpers/strings";
 import { getTaxonName } from "../helpers/taxon";
 import ThresholdMap from "./utils/ThresholdMap";
+import {
+  computeThresholdedTaxons,
+  isTaxonIncluded,
+  getTaxonSortComparator,
+  getCategoryAdjective
+} from "./views/report/utils";
 import Nanobar from "nanobar";
 import BasicPopup from "./BasicPopup";
 import ThresholdFilterDropdown from "./ui/controls/dropdowns/ThresholdFilterDropdown";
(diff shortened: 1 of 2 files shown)


app/assets/src/components/utils/report.js
```diff
@@ -0,0 +1,140 @@
...
+    case "archaea":
+      return "archaeal";
+    case "eukaryota":
+      return "eukaryotic";
+    case "viruses":
+      return "viral";
+    case "viroids":
+      return "viroidal";
+    case "uncategorized":
+      return "uncategorized";
```
tfrcarvalho: Add a default case: You can actually return the category name by default and you do not need the last case.

app/assets/src/components/utils/report.js
```diff
@@ -0,0 +1,140 @@
...
+    if (a.genus_taxid < b.genus_taxid) {
+      return -1;
+    } else if (a.genus_taxid > b.genus_taxid) {
+      return 1;
+    }
+  };
+};
+
+export const getCategoryAdjective = category => {
+  const category_lowercase = category.toLowerCase();
```
tfrcarvalho: Change to camelCase

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -8,6 +8,12 @@ import { Label, Menu, Icon, Popup } from "semantic-ui-react";
 import numberWithCommas from "../helpers/strings";
 import { getTaxonName } from "../helpers/taxon";
 import ThresholdMap from "./utils/ThresholdMap";
+import {
+  computeThresholdedTaxons,
+  isTaxonIncluded,
+  getTaxonSortComparator,
+  getCategoryAdjective
+} from "./utils/report";
```
tfrcarvalho: If theses are report specific, I believe it makes more sense to put this inside the report view, instead of utils.

tfrcarvalho: We also should move `PipelineSampleReport` to `views/report` and maybe rename it in one of your next PRs.

app/assets/src/components/utils/report.js
```diff
@@ -0,0 +1,140 @@
+import ThresholdMap from "./ThresholdMap";
```
tfrcarvalho: I had several more suggestions for the code below, but you are just moving code for now, right? 
If so, you can ignore the below comments.

app/assets/src/components/PipelineSampleReport.jsx
MarkAZhang: Makes sense. I will move the utils file to inside report view.

app/assets/src/components/utils/report.js
MarkAZhang: Yes, just moving code. I'll take care of the camelcase.

PR #1676, opened 2018-10-22
Use more efficient sourcemap for webpack dev build.

MarkAZhang:
Increases CSS rebuild speed by factor of 3 and JS by factor of 5.

The trade-off is the dev output file is much larger.
We keep prod the same.


jshoe:
devtool: "source-map" will still override this for prod right?

tfrcarvalho:
What info do we lose with this change?

MarkAZhang:
It's supposed to only increase the bundle size, but it ends up breaking CSS sourcemaps.

Closing for now. We can optimize once the build times feel too long.

PR #1677, opened 2018-10-23
Add CSS module capabilities and refactor cli_user_instructions.scss.

MarkAZhang:
This is a new proposal for using CSS modules to scope our CSS classes. I've refactored cli_user_instructions.scss to demonstrate this.

The output now looks like this. The CSS classes have a hash appended to them.

screen shot 2018-10-22 at 6 25 03 pm

The styling of the page is unchanged.
screen shot 2018-10-22 at 6 34 35 pm

This allows us to use simpler CSS classes going forward without worrying about name conflicts with 3rd party libraries, or other CSS classes.

Note that the new CSS modules are only applied to files **outside of the traditional style directory**.
This allows us to move the existing CSS to CSS modules incrementally. The existing CSS still works.

Finally, some people warn against using CSS modules with SCSS.
https://medium.com/jsdownunder/webpack-build-performance-pitfall-of-using-sass-with-css-modules-ba32f89efdcb

Specifically, the webpack start-up time is supposed to get really slow as the number of SCSS entry points goes up. It's not clear if these issues have been addressed, but `sass-loader` does seem to have improved recently.
I tested with 7 SCSS entry points and didn't notice any increase in initial build time, so I think we are okay for the time being. If the webpack start-up time starts going up, we can move to PostCSS (which should be more much more performant) with minimal refactoring.

### DIFF
--- a/app/assets/src/components/views/CliUserInstructions.jsx
+++ b/app/assets/src/components/views/CliUserInstructions.jsx
@@ -1,6 +1,8 @@
 import React from "react";
 import { Form, TextArea } from "semantic-ui-react";
 import PropTypes from "prop-types";
+import cx from "classnames";
+import cs from "./cli_user_instructions.scss";
 
 class CliUserInstructions extends React.Component {
   render() {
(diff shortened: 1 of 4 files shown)


app/assets/src/components/views/other/CliUserInstructions.jsx
```diff
@@ -1,6 +1,8 @@
 import React from "react";
```
tfrcarvalho: We have been putting simpler views directly in `components/views` (instead of other).

app/assets/src/components/views/other/cli_user_instructions.scss
```diff
@@ -0,0 +1,65 @@
...
+  color: $darkest-grey;
+  font-size: $font-size-small;
+  margin-top: 20px;
+  margin-bottom: 25px;
+
+  .link {
+    font-weight: $font-weight-semibold;
+
+    &:hover {
+      color: $primary-light !important;
```
tfrcarvalho: Do we still need the important? Is this due to materialize or semantic? Can you add a comment with the root cause?

MarkAZhang: Added comment. The direct cause is some of our own CSS in `_header.scss`.

PR #1680, opened 2018-10-25
Add drag-and-drop + local web uploader

jshoe:
- Visual experience has been approved by Jenn/Mark
- Also fixes problem where server errors weren't being surfaced back to the user
- Removed serverErrors because it seems simpler to put them into errorMessage
- The "create from S3 URL" methods were basically copy-moved into the new blocks
- Special features: % uploaded indicators, some resuming ability if wifi is temporarily interrupted, warning before closing tab while uploading. Still works even if they drag in 2 files at once.
- Max size set as 5GB. If they drag in a file greater than that they'll get an error pop up.

- Eventually will probably set localUploadMode = true by default so they see the new uploader first
- Future steps: consolidated universal/bulk upload experience.

Starting:
oct-24-2018 16-51-58

Finished:
oct-24-2018 16-52-04


jshoe:
Oh wait need to update the sample status to "uploaded" as well somehow

jshoe:
^ Updated

tfrcarvalho:
The "uploading" status might be take a while, no?

db/schema.rb
```diff
@@ -344,12 +344,12 @@
     t.index ["pipeline_run_id", "tax_id", "count_type", "tax_level"], name: "index_pr_tax_hit_level_tc", unique: true
   end
 
-  create_table "taxon_descriptions", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
+  create_table "taxon_descriptions", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci" do |t|
```
tfrcarvalho: I believe this change is correct... i wonder how it regressed. 
Can you confirm with @yunfangjuan to be sure?

app/assets/src/components/SampleUpload.jsx
```diff
@@ -5,6 +5,8 @@ import $ from "jquery";
 import Tipsy from "react-tipsy";
 import IconComponent from "./IconComponent";
 import ObjectHelper from "../helpers/ObjectHelper";
+import Dropzone from "react-dropzone";
+import { Icon, Menu } from "semantic-ui-react";
```
tfrcarvalho: Can we not use semantic directly? 
We have Icon elements at least. What is the menu for?

app/assets/src/components/SampleUpload.jsx
```diff
@@ -274,21 +292,25 @@ class SampleUpload extends React.Component {
...
           submitting: false,
-          serverErrors: error.response.data,
-          errorMessage:
-            "Something went wrong. Try checking if sample name already exists in project?"
+          errorMessage: this.joinServerError(error.response.data)
         });
       });
   }
 
+  joinServerError = function(response) {
```
tfrcarvalho: Why do you need this different function declaration?

tfrcarvalho:
A couple of visual comments (more for @happyimadesignr):
- I believe the boxes are a bit too bulky, with a lot of empty space on the right. I wonder if we readjust position of the on the inner elements. Progress could maybe be on the right;
- I would suggest coloring the border blue on hover.

db/schema.rb
jshoe: whoops didn't mean to include this. will remove.

app/assets/src/components/SampleUpload.jsx
jshoe: 'Menu' for the button switcher. will change to use our Icon and wrap the Menu.

app/assets/src/components/SampleUpload.jsx
jshoe: no reason, will change

jshoe:
@tfrcarvalho :

- Sorry what do you mean by "uploading status might take a while"? like after they move on from the page?
- Will move the dropzone into a separate component
- Will tighten up the dropzone boxes a bit and color the border blue on hover

jshoe:
@tfrcarvalho ready for a 2nd pass now :) 

Other changes:
- It'll say "Waiting to Start" on the next page
- Smaller upload box width (below)
- Also adds babel-polyfill since I think that was missing from package.json
- Split into an upload box component still uses an onDrop handler prop

screen shot 2018-10-25 at 10 52 57 am

PR #1681, opened 2018-10-25
Cross-browser support + prevent sign in non-supported browsers.

tfrcarvalho:
# Description

* Disable sign-in and add error message to non-supported browsers.
* Better support fo IE11 to allow submitting initial request.
* Fix proptype definition for show bulletin.

screen shot 2018-10-25 at 9 12 46 am

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,9 @@
 {
   "presets": ["env", "react"],
-  "plugins": ["transform-object-rest-spread", "transform-class-properties"]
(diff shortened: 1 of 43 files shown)


happyimadesignr:
- Can it be shorter like “IE” instead of Internet Explorer?
- I’d make the font gray to stand out less
- Can you center it vertically in the bar
- Maybe hide the whole line “Already have an account? Sign in."

PR #1684, opened 2018-10-26
Modify check all checkbox to select all samples.

MarkAZhang:
Also add warning message for too many samples in BackgroundModal.

Also rework the sample labels in the header.

Also move check all checkbox to the table header.

Also add sample number to loading message in SamplesHeatmap.

oct-25-2018 17-03-11

The sample labels have been reworked and the checkbox has been moved to the table header. Went over the change with Jennifer.

screen shot 2018-10-24 at 8 14 54 pm

The BackgroundModal is also slightly altered, partially because we might not have pre-fetched all the data for the selected samples yet:

screen shot 2018-10-25 at 5 04 35 pm

### DIFF
--- a/app/assets/src/components/Samples.jsx
+++ b/app/assets/src/components/Samples.jsx
@@ -2,8 +2,12 @@ import React from "react";
 import axios from "axios";
 import ReactDOM from "react-dom";
 import moment from "moment";
+import cx from "classnames";
 import $ from "jquery";
 import Materialize from "materialize-css";
+import { intersection, union, difference, map, keyBy, take } from "lodash";
+// TODO(mark): Refactor lodash/fp functions into a file of immutable utilities.
+import { merge } from "lodash/fp";
 import { Sidebar, Popup, Label, Icon, Modal, Form } from "semantic-ui-react";
 import Nanobar from "nanobar";
 import SortHelper from "./SortHelper";
@@ -18,6 +22,8 @@ import PrimaryButton from "./ui/controls/buttons/PrimaryButton";
 import SecondaryButton from "./ui/controls/buttons/SecondaryButton";
 import MultipleDropdown from "./ui/controls/dropdowns/MultipleDropdown";
 import PhyloTreeCreationModal from "./views/phylo_tree/PhyloTreeCreationModal";
+// TODO(mark): Convert styles/samples.scss to CSS modules.
+import cs from "./samples.scss";
 
 class Samples extends React.Component {
   constructor(props, context) {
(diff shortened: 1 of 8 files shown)


app/assets/src/components/Samples.jsx
```diff
@@ -2,8 +2,12 @@ import React from "react";
 import axios from "axios";
 import ReactDOM from "react-dom";
 import moment from "moment";
+import cx from "classnames";
 import $ from "jquery";
 import Materialize from "materialize-css";
+import { intersection, union, difference, map, keyBy, take } from "lodash";
```
tfrcarvalho: Do we really need to use lodash? I am asking because I have been pushing for not using too many extra libraries. I believe we are more bloated then we should.
I guess intersection might be hard to replace... but you probably do not need the merge below.

What do you mean with the comment below?

app/assets/src/components/Samples.jsx
```diff
@@ -626,18 +638,22 @@ class Samples extends React.Component {
       .then(res => {
         this.setState(prevState => ({
           isRequesting: false,
-          allSamples: [...prevState.allSamples, ...res.data.samples],
+          fetchedSamples: merge(
```
tfrcarvalho: from the lodash conversation: couldn't this be just `Object.assign`?

app/assets/src/components/samples.scss
```diff
@@ -0,0 +1,58 @@
+@import "~styles/themes/colors";
+@import "~styles/typography";
+
+/* BackgroundModal */
+.sampleList {
```
tfrcarvalho: Did the recommended notation for selectors changed because we are using css modules?

app/assets/src/components/Samples.jsx
```diff
@@ -2035,10 +2059,23 @@ function TableColumnHeaders({ sort, colMap, filterStatus, state, parent }) {
       <div className="samples-card white">
         <div className="flex-container">
           <ul className="flex-items">
-            <li className="table-header-name">
+            <li className={cs.nameHeader}>
+              {parent.state.displaySelectSamples &&
+                parent.state.readySampleIdsForFilter.length > 0 && (
+                  <div className={cs.checkAllContainer}>
+                    <input
```
tfrcarvalho: Is this new?
I wonder if it would be a good time to actually use the Checkbox component.

tfrcarvalho: could be a different pr

app/assets/src/components/Samples.jsx
MarkAZhang: In my experience, lodash has been very convenient and allows developers to write more concise code (though there is a bit of a learning curve).

I would prefer to keep it. There are ways to reduce the bundle size (at the expense of slightly longer build times): https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark

Is bundle size a serious issue? lodash is one of the few libraries I routinely use.

app/assets/src/components/Samples.jsx
MarkAZhang: It's true. We can do `Object.assign({}, a, b)` or `{...a, ...b}` for immutability. Some of the other `lodash` methods are harder to replace. I would prefer to keep `lodash` to speed up development time unless bundle size is a big issue, in which case we might consider writing our own small set of utility functions.

app/assets/src/components/samples.scss
MarkAZhang: Yeah. If you do `.sample-list`, in JSX you'd have to do `className={cs['sample-list']}` instead of `{cs.sampleList}`.

app/assets/src/components/Samples.jsx
MarkAZhang: That would be great! I'll do it in a follow-up PR. I did not enjoy using the semantic-ui checkbox.

app/assets/src/components/Samples.jsx
tfrcarvalho: I believe bundle size could become a factor in some sites and if we do not carefully evaluate before adding packages.
On the other hand, `lodash` does seem useful and does allow for custom builds if we ever need to optimize so I think it is fine.

I guess my next question why do we need `lodash/fp`?

app/assets/src/components/samples.scss
tfrcarvalho: Ok, i see you already updated the development guid.

tfrcarvalho:
Will let you decide on what to do with lodash/fp...

MarkAZhang:
My decision: I think it's important to be consistent across the code-base with how we do things, and Object.assign is already used, so I'll stick with that.

PR #1685, opened 2018-10-26
Protect some more admin fields

jshoe:
- During the upload work noticed some admin params are still unprotected. Didn't see an obvious exploit but for hygiene
- Move :s3_preload_result_path, :alignment_config_name, :sample_memory, :subsample into the admin params
- Tested locally that any default values needed are filled downstream and that upload submission still works


jshoe:
I followed them down pretty thoroughly

PR #1689, opened 2018-10-27
Update the header styling for the Sample Table.

MarkAZhang:
JIRA ticket: IDSEQ-553

screen shot 2018-10-26 at 5 04 45 pm

Add tooltip to download button:

screen shot 2018-10-26 at 5 02 00 pm


MarkAZhang:
Now that I think about it, I don't see a pressing need for wrapping semantic-ui elements in our own elements unless: 1) We don't like the semantic-ui element's API and we want to modify it or 2) We add custom styling to the semantic-ui element. Otherwise, we end up having a wrapper for each semantic-ui element that doesn't do anything...Thoughts?

PR #1698, opened 2018-11-01
scale-in

cdebourcy:
goes with:
https://github.com/chanzuckerberg/idseq-infra/pull/148
https://github.com/chanzuckerberg/idseq-dag/pull/91


app/jobs/autoscaling.py
```diff
@@ -23,14 +25,22 @@
...
 EXPIRATION_PERIOD_MINUTES = 60
 
-
 DEBUG = False
 
 
+## constants specific to draining / safe scale-down:
+DRAINING_TAG = "draining"
+MAX_REFRESH_INTERVAL = 900
+# MAX_REFRESH_INTERVAL must be the same as MAX_INTERVAL_BETWEEN_DESCRIBE_INSTANCES in idseq-dag.
```
boris-dimitrov: May be just slightly longer would be better than exactly the same.

app/jobs/autoscaling.py
```diff
@@ -23,14 +25,22 @@
...
+# After this interval, we are sure we are safe from race conditions related to batch jobs dispatching gsnap/rapsearch jobs.
+JOB_TAG_PREFIX = "RunningIDseqBatchJob_"
+ALIGNMENT_JOB_EXPIRATION_SECONDS = 120 * 60
```
boris-dimitrov: 2 hours?

app/jobs/autoscaling.py
boris-dimitrov: I think it is probably okay to shorten MAX_REFRESH_INTERVAL in idseq-dag to 10 minutes.    We could also make it a system parameter in the DAG json just like max_concurrency, so it would be available from the same source in both places.  There should probably be a special section for system (infra) parameters in the dag config as opposed to parameters than affect the results.

app/jobs/autoscaling.py
```diff
@@ -198,9 +208,133 @@ def set_desired_capacity(asg, compute_desired_instances, can_scale=True):
...
+            stop_draining(asg, num_desired - previous_desired)
         aws_command(cmd)
 
 
+def instances_in(asg):
+    return [item["InstanceId"] for item in asg["Instances"]]
+
+
+def iso_time_now():
+    return datetime.datetime.now().isoformat()
```
boris-dimitrov: could we please use unixtime instead?   the code to parse and unparse unixtime for tags already exists in this file and unixtime is used in the ASG tags, see for example the line above `tag_value = "{unixtime:d}::{value:d}".format(unixtime=int(time.time()), value=int(value))`

boris-dimitrov: if for debugging purposes you wish to convert from unixtime to human readable local time, the python function time.localtime(unixtime_integer) will do so.

app/jobs/autoscaling.py
```diff
@@ -198,9 +208,133 @@ def set_desired_capacity(asg, compute_desired_instances, can_scale=True):
...
+    return instance_ids_sorted_by_increasing_launch_time[:num_instances]
+
+
+def instances_to_rescue(asg, num_instances):
+    '''
+    Rescue the instances that have been draining for the least amount of time.
+    That way we can probably terminate the non-rescued instances sooner.
+    '''
+    draining_instances = get_draining_servers(asg)
+    draining_instances_sorted = [key for key, value in sorted(draining_instances.items(), key=itemgetter(1), reverse=True)]
```
boris-dimitrov: I think you could also write this as
```
draining_instances_sorted  = sorted(draining_instances.keys(), key=lambda k: draining_instances[k], reverse=True)
```
I hadn't seen itemgetter before.

app/jobs/autoscaling.py
```diff
@@ -198,9 +208,133 @@ def set_desired_capacity(asg, compute_desired_instances, can_scale=True):
...
+    return draining_instances_sorted[:num_instances]
+
+
+def get_draining_servers(asg):
+    ''' returns a map of draining instance IDs to the time they started draining '''
+    cmd = "aws ec2 describe-tags --filters 'Name=tag-key,Values={draining_tag}' 'Name=resource-id,Values={instance_ids}'"
+    cmd = cmd.format(draining_tag=DRAINING_TAG, instance_ids=','.join(instances_in(asg)))
+    tag_dict = json.loads(aws_command(cmd))
+    instance_dict = { item['ResourceId']: dateutil.parser.parse(item['Value']) for item in tag_dict['Tags'] if item['Key'] == DRAINING_TAG }
```
boris-dimitrov: Yeah let's not parse and unparse time formats, if we can use unixtime.

app/jobs/autoscaling.py
```diff
@@ -198,9 +208,133 @@ def set_desired_capacity(asg, compute_desired_instances, can_scale=True):
...
+
+
+def start_draining(asg, num_instances):
+    instance_ids = instances_to_drain(asg, num_instances)
+    add_draining_tag(instance_ids)
+
+
+def stop_draining(asg, num_instances):
+    instance_ids = instances_to_rescue(asg, num_instances)
+    add_termination_protection(instance_ids, asg)
```
boris-dimitrov: I haven't finished reading this pull request, but it seems odd to be adding a protection here.   At a high level, I would expect protection to be added at creation to all instances, and removed only after they are finished draining (past which point draining cannot be stopped).

app/jobs/autoscaling.py
```diff
@@ -198,9 +208,133 @@ def set_desired_capacity(asg, compute_desired_instances, can_scale=True):
...
+    remove_draining_tag(instance_ids)
+
+
+def instances_to_drain(asg, num_instances):
+    instance_ids = instances_in(asg)
+    cmd = "aws ec2 describe-instances --instance-ids {list_instances} --query 'Reservations[*].Instances[*].LaunchTime' --no-paginate --output text"
+    cmd = cmd.format(list_instances=' '.join(instance_ids))
+    launch_times = [dateutil.parser.parse(datestring) for datestring in aws_command(cmd).splitlines()]
+    instance_ids_sorted_by_increasing_launch_time = [item[0] for item in sorted(zip(instance_ids, launch_times), key=itemgetter(1))]
```
boris-dimitrov: Thanks for adding this optimization.

app/jobs/autoscaling.py
```diff
@@ -198,9 +208,133 @@ def set_desired_capacity(asg, compute_desired_instances, can_scale=True):
...
+    ''' returns a map of draining instance IDs to the time they started draining '''
+    cmd = "aws ec2 describe-tags --filters 'Name=tag-key,Values={draining_tag}' 'Name=resource-id,Values={instance_ids}'"
+    cmd = cmd.format(draining_tag=DRAINING_TAG, instance_ids=','.join(instances_in(asg)))
+    tag_dict = json.loads(aws_command(cmd))
+    instance_dict = { item['ResourceId']: dateutil.parser.parse(item['Value']) for item in tag_dict['Tags'] if item['Key'] == DRAINING_TAG }
+    return instance_dict
+
+
+def count_running_alignment_jobs(asg):
+    ''' returns a map of instance IDs to number of jobs running on the instance '''
```
boris-dimitrov: For this initial version, can we skip this altogether in favor of something much simpler, like checking the load average over last 5 minutes with `ssh <host> uptime | awk '{print $12}'`  to determine if that host is finished draining?

(Even thought checking tags was my suggestion at one point, I didn't intend it for this early stage, and I meant it more for enabling more complex protocols later.  I believe simplicity is something we should truly strive for, even to the point of removing optimizations.)

app/jobs/autoscaling.py
```diff
@@ -198,9 +208,133 @@ def set_desired_capacity(asg, compute_desired_instances, can_scale=True):
...
+        return
+    cmd = "aws autoscaling set-instance-protection --instance-ids {list_instances} --auto-scaling-group-name {asg_name} --no-protected-from-scale-in"
+    cmd = cmd.format(list_instances=' '.join(instance_ids), asg_name=asg['AutoScalingGroupName'])
+    aws_command(cmd)
+
+
+def add_termination_protection(instance_ids, asg):
+    if not instance_ids:
+        return
+    cmd = "aws autoscaling set-instance-protection --instance-ids {list_instances} --auto-scaling-group-name {asg_name} --protected-from-scale-in"
```
boris-dimitrov: Is it possible to make "protected from scale in" a default, so that the only transition executed by our code would be remove that protection when an instance is finished draining?

app/jobs/autoscaling.py
cdebourcy: Pretty arbitrary, what time period would you recommend based on your benchmarking?

app/jobs/autoscaling.py
cdebourcy: Yes. I was using iso time because that's how AWS gives you the LaunchTime. I will convert the AWS output to unixtime

app/jobs/autoscaling.py
cdebourcy: Yes. I choose itemgetter because it was more compact

app/jobs/autoscaling.py
cdebourcy: Yes. The one time that will need to be parsed is the AWS-provided LaunchTime. I will convert that to unixtime and remove parsing everywhere else

app/jobs/autoscaling.py
cdebourcy: Makes sense. But I really like the tags as a way of easily seeing what's happening in the gsnap/rapsearch servers via the console. I would vote to leave this feature in, since I've already tested it and it works and it can be useful later

app/jobs/autoscaling.py
cdebourcy: Yes, that's exactly right -- I have a diff in idseq-infra for that: https://github.com/chanzuckerberg/idseq-infra/pull/148
This function `add_termination_protection` is actually for when we decide to `stop_draining` because in the meantime we decided to scale back up. In that case we just add the termination protection back, and avoid killing instances only to then launch new ones.

app/jobs/autoscaling.py
cdebourcy: Yes, that's how it works, see: https://github.com/chanzuckerberg/idseq-infra/pull/148
It is possible to stop draining if the termination has not yet been initiated.
When we want to stop draining, we just need to add protection back and remove the draining tag.
Your comment made me realize I have to add an additional check to instances_to_rescue: filter for state "InService" to guarantee the instance is still usable. Thanks!

app/jobs/autoscaling.py
cdebourcy: Agree - I changed MAX_REFRESH_INTERVAL to 1200. This may seem like too big an increase but I want to be safe with respect to synchronization lag; we can always revise it downwards later. I suppose scaling down 5 minutes earlier or later doesn't make such a huge different in cost.

Reducing MAX_INTERVAL_BETWEEN_DESCRIBE_INSTANCES seems like a good idea, but I think it should be a separate change.

app/jobs/autoscaling.py
```diff
@@ -23,14 +25,22 @@
...
+## constants specific to draining / safe scale-down:
+DRAINING_TAG = "draining"
+MAX_REFRESH_INTERVAL = 1200
+# MAX_REFRESH_INTERVAL must be slightly bigger than MAX_INTERVAL_BETWEEN_DESCRIBE_INSTANCES in idseq-dag.
+# After this interval, we are sure we are safe from race conditions related to batch jobs dispatching gsnap/rapsearch jobs.
+JOB_TAG_PREFIX = "RunningIDseqBatchJob_"
+ALIGNMENT_JOB_EXPIRATION_SECONDS = 120 * 60
+
+
+
```
tfrcarvalho: remove one line...

app/jobs/autoscaling.py
cdebourcy: Only if there are crashed jobs on a server. Normally job tags are removed from servers when job succeeds. To handle the case where a crash prevented the tag from being removed, we clean up old tags every ALIGNMENT_JOB_EXPIRATION_SECONDS seconds.

So ALIGNMENT_JOB_EXPIRATION_SECONDS does need to be long enough that we are certain the job tag should be discarded. However, I agree, we should make it as short as possible while fulfilling this constraint. Will ask @boris-dimitrov what the distribution of job runtimes is.

app/jobs/autoscaling.py
boris-dimitrov: I don't think it's safe to add termination protection back.   Once it has been removed, the instance is past resurrection.

app/jobs/autoscaling.py
```diff
@@ -198,9 +208,146 @@ def set_desired_capacity(asg, compute_desired_instances, can_scale=True):
...
+def stop_draining(asg, num_instances):
+    instance_ids = instances_to_rescue(asg, num_instances)
+    if instance_ids:
+        print "Stopping drainage of the following instances: " + ",".join(instance_ids)
+        add_termination_protection(instance_ids, asg)
+        # Note: if autoscaling.py ran too frequently, there would be a risk that an instance enters Terminating state between the time
+        # of the health check in instances_to_rescue and the time of add_termination_protection, in which case add_termination_protection
+        # would throw an error. But because autoscaling.py runs only infrequently, we can be sure that there is no scaling decision in limbo
+        # from the last execution of autoscaling.py.
```
boris-dimitrov: Hmm, no, I think that won't happen, because we are taking care to never rescue from draining an instance whose termination protection has been removed.

boris-dimitrov: Also in that design as we discussed in person we don't have to re-add termination protection.

app/jobs/autoscaling.py
```diff
@@ -198,9 +208,146 @@ def set_desired_capacity(asg, compute_desired_instances, can_scale=True):
...
+    return instance_ids_sorted_by_increasing_launch_time[:num_instances]
+
+
+def instances_to_rescue(asg, num_instances):
+    '''
+    Rescue the instances that have been draining for the least amount of time.
+    That way we can probably terminate the non-rescued instances sooner.
+    '''
+    draining_instances = get_draining_servers(asg)
+    healthy_instance_ids = [inst["InstanceId"] for inst in asg["Instances"] if inst["HealthStatus"] == "Healthy" and inst["LifecycleState"] != "Terminating"]
```
boris-dimitrov: So here I think rather than collecting instances based on LifecycleState, you want instead to collect instances that still have their scale-in protection set.

boris-dimitrov: In fact I would change get_draining instances to filter to those.   Once protection has been removed, the instance is no longer draining.  It is drained.

app/jobs/autoscaling.py
```diff
@@ -198,9 +208,146 @@ def set_desired_capacity(asg, compute_desired_instances, can_scale=True):
...
+
+
+def instances_to_drain(asg, num_instances):
+    instance_ids = instances_in(asg)
+    cmd = "aws ec2 describe-instances --instance-ids {list_instances} --query 'Reservations[*].Instances[*].[InstanceId,LaunchTime]' --no-paginate"
+    cmd = cmd.format(list_instances=' '.join(instance_ids))
+    aws_response = json.loads(aws_command(cmd))
+    zipped_instance_ids_and_launch_times = [(instance[0], iso2unixtime(instance[1])) for nest in aws_response for instance in nest]
+    instance_ids_sorted_by_increasing_launch_time = [item[0] for item in sorted(zipped_instance_ids_and_launch_times, key=itemgetter(1))]
+    return instance_ids_sorted_by_increasing_launch_time[:num_instances]
```
boris-dimitrov: One more optimization is possible here.   The instances that would be easiest to drain are those that are still initializing.   You could handle those specially, but there is one catch...  They would race with you to declare that they are done initializing.  There is, of course, a solution that would make that race unlikely.  Perhaps this isn't worth exploring, since it's more complicated.

app/jobs/autoscaling.py
boris-dimitrov: I think having "protection" implies that thee instance is still usable;  so that place should just make sure protection has not been removed.

app/jobs/autoscaling.py
cdebourcy: Agreed - I will remove that

boris-dimitrov:
One important question is, what happens if this process gets killed in the middle of performing its duties -- as can happen, for example, during a deploy.   (We should take measures that it doesn't happen, but I don't think we've taken those measures, yet.)   Does the next invocation clean up any problems nicely?  Or is there a lasting damage to the ASG state?

app/jobs/autoscaling.py
boris-dimitrov: I guess what I meant here was more along the lines of changing the meaning of that constant, not the behavior of this code.

app/jobs/autoscaling.py
cdebourcy: makes sense, thanks

app/jobs/autoscaling.py
cdebourcy: makes sense, thanks

cdebourcy:
This can be remedied by changing “count_healthy_instances” so it doesn’t include draining instances

app/jobs/autoscaling.py
cdebourcy: ok. I'm changing ALIGNMENT_JOB_EXPIRATION_SECONDS to 30 * 60 and will add the alive signal in idseq-dag

app/jobs/autoscaling.py
```diff
@@ -23,14 +25,44 @@
...
+    V
+{ service: gsnap-prod, draining: T0, ProtectedFromScaleIn: True }
+Draining. Scale-in protection still present.
+Can still accept chunks for up to min_draining_wait seconds after entering this state.
+Can still finish running chunks for up to (min_draining_wait + job_tag_expiration) seconds after entering this state.
+    |
+    V
+{ service: gsnap-prod, draining: T0, ProtectedFromScaleIn: False }
+Discarded. Scale-in protection withdrawn on transition to this state.
+'''
```
boris-dimitrov: Thanks for documenting these states and transitions.    Two of the states in the middle seem identical, should be compressed to single state.   You are missing the back arrow for reclaiming instances from draining.  What is the meaning of T0?

In computer science the word "state" is fairly overloaded, but so far as state machines, it's pretty standard to define a state machine as a set of states and set of transitions;   you can label those states S1, S2, ... and list in a table explaining what they are (like you have above) but then list the transitions S1 -> S2, S2 -> S3, S3 -> S4, and S3 -> S2 in a separate table perhaps.  Each transition also needs a word of explanation.

app/jobs/autoscaling.py
```diff
@@ -156,13 +188,15 @@ def update_metric_values(asg, value, my_environment):
...
+    draining_instance_ids = get_draining_servers(asg, tag_list, draining_tag).keys()
+    num_draining = len(draining_instance_ids)
     num_healthy = 0
     for inst in asg["Instances"]:
-        if inst["HealthStatus"] == "Healthy" and inst["LifecycleState"] != "Terminating":
+        if inst["InstanceId"] not in draining_instance_ids and \
+            inst["ProtectedFromScaleIn"] and inst["HealthStatus"] == "Healthy" and inst["LifecycleState"] != "Terminating":
             num_healthy += 1
-    return num_healthy
+    return num_healthy, num_draining
```
boris-dimitrov: It's hard to track the invariants and progress properties of this scaling operation when there are so many flags and tags being checked.   One of the advantages of modeling the life of an instance with a state machine is that you can then use the state machine state definitions to describe the state of the distributed system -- see there I used "state" with two different meanings in the same sentence, so evil.   The point is, you can say there are 4 states of an instance, call them S1, S2, S3, S4 and you can count the instances in each state using a single counting function.   Then any decision you make ought to be based on s1_count, s2_count, s3_count, and s4_count.  I believe when you structure the code in this way, the reasoning for its correctness will also become simpler.  I defer to you if/when you would consider such restructuring.

app/jobs/autoscaling.py
```diff
@@ -181,26 +215,166 @@ def clamp_to_valid_range(asg, desired_capacity):
...
+    current_timestamp = unixtime_now()
+    for instance_id, drain_date in drain_date_by_instance_id.items():
+        draining_interval = current_timestamp - drain_date
+        num_jobs = num_jobs_by_instance_id[instance_id]
+        if num_jobs == 0 and draining_interval > min_draining_wait:
+            instance_ids_to_terminate.append(instance_id)
+    if instance_ids_to_terminate:
+        print "The following instances are ready to be terminated: " + ", ".join(instance_ids_to_terminate)
+        if can_scale:
+            remove_termination_protection(instance_ids_to_terminate, asg)
```
boris-dimitrov: Do development environments share an ASG nowadays?   I don't think this is a major problem, but perhaps can_scale should be used to bypass most of the work here, not just be deferred to the end -- so development environments don't get bogged down doing AWS operations that are mute.

app/jobs/autoscaling.py
```diff
@@ -181,26 +215,166 @@ def clamp_to_valid_range(asg, desired_capacity):
...
+        delete_tags_from_instances(instance_ids, expired_jobs)
+    return count_dict
+
+
+def delete_tags_from_instances(instance_ids, tag_keys):
+    if not instance_ids or not tag_keys:
+        return
+    cmd = "aws ec2 delete-tags --resources {list_instances} --tags {list_tags}"
+    cmd = cmd.format(list_instances=' '.join(instance_ids), list_tags=' '.join(['Key='+tag for tag in tag_keys]))
+    aws_command(cmd)
```
boris-dimitrov: If this operation fails, will it be repeated next time autoscaling is reincarnated?

app/jobs/autoscaling.py
```diff
@@ -260,9 +440,14 @@ def autoscaling_update(my_num_jobs, my_environment="development"):
...
             return
+    tag_list = get_instance_tags()
+    gsnap_tags = find_tags_for_instances_in(gsnap_asg, tag_list)
+    rapsearch_tags = find_tags_for_instances_in(rapsearch2_asg, tag_list)
+    check_draining_servers(gsnap_asg, gsnap_tags, draining_tag, job_tag_prefix, job_tag_expiration, min_draining_wait, can_scale)
+    check_draining_servers(rapsearch2_asg, rapsearch_tags, draining_tag, job_tag_prefix, job_tag_expiration, min_draining_wait, can_scale)
     if num_real_jobs == 0 and num_development_jobs == 0:
-        set_desired_capacity(gsnap_asg, exactly(0), can_scale)
-        set_desired_capacity(rapsearch2_asg, exactly(0), can_scale)
+        set_desired_capacity(gsnap_asg, gsnap_instance_name, gsnap_tags, draining_tag, exactly(0), can_scale)
```
boris-dimitrov: I'd consider creating an ASG class to track all this state, and make these functions methods of the class, so you don't have to pass all these params all the time.

app/jobs/autoscaling.py
```diff
@@ -260,9 +440,14 @@ def autoscaling_update(my_num_jobs, my_environment="development"):
         # when debugging is disabled and we can't scale we exit early here
         if not DEBUG:
             return
+    tag_list = get_instance_tags()
+    gsnap_tags = find_tags_for_instances_in(gsnap_asg, tag_list)
+    rapsearch_tags = find_tags_for_instances_in(rapsearch2_asg, tag_list)
+    check_draining_servers(gsnap_asg, gsnap_tags, draining_tag, job_tag_prefix, job_tag_expiration, min_draining_wait, can_scale)
```
boris-dimitrov: Is there a reason check_draining_servers runs before set_desired_capacity?   I would have thought in the interest of reclaiming optimization you would wish to set desired capacity first, to give yourself the ability to reclaim some draining servers, instead of removing their protection at the start which makes them unavailable for reclaiming later.

boris-dimitrov:
This needs to be restructured so that invariants and progress properties can be stated simply and verified at a glance.   The decision of the optimal number of instances desired should be completely separated from the mechanism that sets in motion state transitions on selected instances' state machines.  The instance state machines and transitions need to be documented clearly and the code needs to be organized more to follow those lines.

I feel a decision needs to be made either to cut functionality so that the current hacky code structure remains adequate for the smaller functionality;   or, if this nicer improved functionality is desired, then the code needs to be reorganized so it's clear how it's implementing it.

app/jobs/autoscaling.py
cdebourcy: Yes: if the the old tag still exists, it will be seen by describe-tags and cleaned up in the next incarnation

app/jobs/autoscaling.py
cdebourcy: Good point. I will move it to the end

PR #1700, opened 2018-11-01
Markzhang/metadata sidebar

MarkAZhang:
Added MetadataEditor and SampleDetailsSidebar.

The Sidebar is currently gated by "user.admin". The old save_metadata endpoint still works. If we land this commit, non-admin users should not see a difference.

The Sidebar renders `<Dropdown>`, `<Input type='text'>` or `<Input type='number'>` depending on `metadata.dataType`. This is not the final design, but a lot of the functionality is here.

Would like to land this commit since it is admin-gated, to keep commit sizes small. 

screen shot 2018-11-01 at 12 28 06 pm

screen shot 2018-11-01 at 12 28 09 pm

Other changes:

Added base Sidebar component.
Added backend endpoints for fetching and saving metadata and fetching metadata types.
Added KEY_TO_DISPLAY_NAME map in metadatum model.

Started separate file for all API calls.
Changed `onChange` behavior for `<Input>` and `<Dropdown>` so it just passes the value into `onChange`.
Add `onClick` functionality and styling to `<RemoveIcon>`
Change Dropdown so first value isn't displayed even when it's not selected.
Change Dropdown so Menu is 100% width by default.


app/assets/src/components/views/report/SampleDetailsSidebar/MetadataEditor.jsx
```diff
@@ -0,0 +1,70 @@
...
+          </div>
+        ))}
+      </div>
+    );
+  }
+}
+
+MetadataEditor.propTypes = {
+  metadata: PropTypes.shape({
+    collection_location: PropTypes.string
```
jshoe: wait is there something special about collection_location vs the other strings?

app/views/samples/show.html.erb
```diff
@@ -1,7 +1,7 @@
 <div id="PipelineSampleReads_page">
   <%= javascript_tag do %>
     react_component('PipelineSampleReads', {
-      admin: '<%= current_user.admin ? 1 : 0 %>',
+      admin: '<%= current_user.admin ? 1 : 0 %>' === '1',
```
jshoe: this can just be `admin: <%= current_user.admin == 1 %>` I think if you want the bool

app/assets/src/components/views/report/SampleDetailsSidebar/MetadataEditor.jsx
```diff
@@ -0,0 +1,70 @@
...
+            {this.renderInput(metadataType)}
+          </div>
+        ))}
+      </div>
+    );
+  }
+}
+
+MetadataEditor.propTypes = {
+  metadata: PropTypes.shape({
```
tfrcarvalho: Why not add these types for `utils/propTypes`?
We can defer it to when we want to use it again, but they seem very common.

tfrcarvalho: Actually you can accept here more than just collection_location, no?

app/assets/src/components/views/report/SampleDetailsSidebar/SampleDetailsSidebar.jsx
```diff
@@ -0,0 +1,94 @@
+import React from "react";
+import { keyBy, mapValues } from "lodash";
+import PropTypes from "../../../utils/propTypes";
+import Sidebar from "../../../ui/containers/Sidebar";
+import RemoveIcon from "../../../ui/icons/RemoveIcon";
+import {
+  getSampleMetadata,
+  saveSampleMetadata,
+  getMetadataTypes
+} from "../../../../api";
```
tfrcarvalho: Should we add `path.resolve(__dirname, "app/assets/src/")` to solve this? 
Not sure how many conflicts we would have though...

tfrcarvalho: https://webpack.js.org/configuration/resolve/#resolve-modules

app/assets/src/components/views/report/SampleDetailsSidebar/SampleDetailsSidebar.jsx
```diff
@@ -0,0 +1,94 @@
...
+  onMetadataChange = (key, value, shouldSave) => {
+    this.setState({
+      metadata: {
+        ...this.state.metadata,
+        [key]: value
+      }
+    });
+
+    if (shouldSave) {
+      this._saveHelper(key, value);
```
tfrcarvalho: Honestly, as `_saveHelper` signature is almost the same as `saveSampleMetadata` I would just call the later... less confusion.

app/assets/src/components/views/report/SampleDetailsSidebar/MetadataEditor.jsx
```diff
@@ -0,0 +1,70 @@
...
+          onChange={val => onMetadataChange(metadataType.key, val, true)}
+          value={metadata[metadataType.key]}
+        />
+      );
+    }
+
+    return (
+      <Input
+        onChange={val => onMetadataChange(metadataType.key, val)}
+        onBlur={() => onMetadataSave(metadataType.key)}
```
tfrcarvalho: Probably should check if metadata changed before calling save callback

app/assets/src/components/ui/controls/dropdowns/Dropdown.jsx
```diff
@@ -42,22 +40,28 @@ class Dropdown extends React.Component {
...
       <BaseDropdown
+        placeholder={this.props.placeholder}
+        scrolling={this.props.scrolling}
         disabled={this.props.disabled}
         fluid={this.props.fluid}
         options={this.props.options}
         value={this.state.value}
         floating
-        className="idseq-ui"
+        className={cx("idseq-ui", this.props.rounded && "rounded")}
```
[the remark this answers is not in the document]
MarkAZhang: Yes, it'll be completely dropped.

PR #1704, opened 2018-11-02
Use npm instead of yarn for everything

jameshowardwang:
Also upgrade merge to 1.2.1


.dockerignore
```diff
@@ -1,3 +1,4 @@
 tmp
 log
-.git
\ No newline at end of file
+.git
+node_modules
```
tfrcarvalho: was this slowing us down before?

Dockerfile
```diff
@@ -33,14 +29,16 @@ WORKDIR /app
...
+# Do the same for node packages, allowing them to be cached
 RUN npm update -g
+COPY package.json package-lock.json ./
+RUN npm install
 
 # Copy the main application.
 COPY . ./
 
-RUN yarn install
-RUN npm rebuild node-sass
```
tfrcarvalho: So we do not need rebuild were neither, right?

jameshowardwang: yeah the issue was that locally built node-sass was overwriting the one built in the docker container. adding node_modules to .dockerignore is meant to fix this.

.dockerignore
jameshowardwang: not slowing us down, but node_sass built locally in darwin was overwriting the one built in the docker image

PR #1706, opened 2018-11-03
Rework ViewHeader component and use for Phylo Tree and Sample Report.

MarkAZhang:
Also add extractChildren utility for extracting child components based on type.

New Phylo tree header:
screen shot 2018-11-02 at 6 42 08 pm

Sample Report header is unchanged. (was refactored)
screen shot 2018-11-02 at 6 42 39 pm


app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -906,8 +861,8 @@ class PipelineSampleReads extends React.Component {
             <div className={cs.pipelineInfo}>
               PIPELINE {version_display} {pipeline_version_blurb}
             </div>
-            <div className={cs.topRow}>
-              <div className={cs.breadcrumbs}>
+            <ViewHeader className={cs.viewHeader}>
```
tfrcarvalho: Why do we need a class here? 
What are the differences in report's header style?

app/assets/src/components/pipeline_sample_reads.scss
```diff
@@ -9,69 +9,33 @@
     margin-bottom: 10px;
   }
 
-  .topRow {
-    display: flex;
+  .viewHeader {
     margin-top: 20px;
```
tfrcarvalho: Seems to me this distances should be standard, no ?

app/assets/src/components/utils/extractChildren.js
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import { clone, difference, filter } from "lodash";
+
+// Separates the children based on type.
+// children - this.props.children.
+// componentTypes - custom strings that must be manually added to sub-components.
+//   ex: static type = "ViewHeader.Content"
+// Returns an object that maps types to the matching component, and returns unmatched compoents in 'other'.
```
jshoe: s/compoents/components

app/assets/src/components/utils/extractChildren.js
```diff
@@ -0,0 +1,35 @@
...
+const extractChildren = (children, componentTypes) => {
```
tfrcarvalho: I wonder if we can turn `extractChildren` into a cutomizable proptypes function checker?

app/assets/src/components/views/phylo_tree/PhyloTreeListView.jsx
```diff
@@ -76,43 +73,45 @@ class PhyloTreeListView extends React.Component {
...
     return (
-      <div className="phylo-tree-list-view">
-        <div className="phylo-tree-list-view__narrow-container">
-          <ViewHeader title="Phylogenetic Trees">
-            <PhyloTreeDownloadButton tree={currentTree} />
+      <div className={cs.phyloTreeListView}>
+        <div className={cs.narrowContainer}>
+          <ViewHeader title="Phylogenetic Trees" className={cs.viewHeader}>
+            <ViewHeader.Content>
+              <div className={cs.preTitle}>Phylogenetic Tree</div>
```
tfrcarvalho: This "pre-titles" should be part of the header too... optional but part of it.

app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -917,11 +872,20 @@ class PipelineSampleReads extends React.Component {
                   </a>
                   <span className={cs.rightArrow}>{">"}</span>
                 </div>
```
tfrcarvalho: This one should be the same component as the pre-title Phylogenetic Tree below..

app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -906,8 +861,8 @@ class PipelineSampleReads extends React.Component {
             <div className={cs.pipelineInfo}>
               PIPELINE {version_display} {pipeline_version_blurb}
             </div>
```
tfrcarvalho: I also believe this should be part of the header as an optional component. This should all be placeholders that different pages/views can use.

tfrcarvalho:
Also, @MarkAZhang should the title wrap to two lines ? Versus using `...` ?

app/assets/src/components/PipelineSampleReads.jsx
MarkAZhang: The difference is in the `margin-top` and `margin-bottom` amounts between the `<ViewHeader>` and the elements above and below. I thought it better to let the client handle that.

app/assets/src/components/pipeline_sample_reads.scss
MarkAZhang: The Samples Report Header has the "pipeline" text above it, and the tabs below it, while the Phylo Tree one doesn't, so they require different distances. 

We can definitely standardize it in the future if it turns out there's only two different cases, but it seems a little premature now with only two headers.

app/assets/src/components/utils/extractChildren.js
MarkAZhang: Holding off for now. Custom prop validators don't appear to be able to take custom arguments...Seems non-trivial to implement.

MarkAZhang:
Discussed comments with Tiago. Made Pretitle a ViewHeader component but kept pipeline-info string separate. Keeping the margin-top/margin-bottom custom for now.

Also made the ViewHeader Title one line long (no wrapping).

MarkAZhang:
Also show Popup on hover when the Title label overflows.

screen shot 2018-11-06 at 1 07 42 pm

PR #1710, opened 2018-11-06
only show report when assembly is done

yunfangjuan:
# Description

1. only show report when assembly is done for pipeline 3+ 
2. download the taxid fasta based on assembly results
3. change nonhost fasta download to use (refined_)taxid_annot.fa

## Type of change

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:


cdebourcy:
Is there a way we could make this more future-proof? From past experience, dealing with these name prefixes is really error-prone. How about we make a clean break and leave the file names the same while the meaning changes? Or how about we store these output paths in the pipeline_runs table when we create a pipeline_run (based on pipeline_version), and never change them again? So we can avoid dealing with this complexity in the code and just always call the same column in the table?

app/models/pipeline_run.rb
```diff
@@ -400,7 +400,7 @@ def db_load_taxon_counts
   end
 
   def db_load_refined_taxon_counts
-    output_json_s3_path = "#{postprocess_output_s3_path}/#{REFINED_TAXON_COUNTS_JSON_NAME}"
+    output_json_s3_path = s3_file_for("refined_taxon_counts")
```
MarkAZhang: refined_taxon_counts doesn't seem to exist in `s3_file_for` any more. Did you mean `taxon_counts`?

app/lib/report_helper.rb
```diff
@@ -238,15 +238,13 @@ def all_categories
     ALL_CATEGORIES
   end
 
-  def fetch_taxon_counts(pipeline_run_id, background_id, refined = false)
+  def fetch_taxon_counts(pipeline_run_id, background_id, _refined = false)
```
MarkAZhang: Is `_refined` used anywhere?

app/models/pipeline_run.rb
MarkAZhang: Ah I see. I guess this loader doesn't get called any more.

Can you add some comments with `TODOs` just so others don't get confused?

PR #1716, opened 2018-11-07
Add metrics POST code for Datadog

jshoe:
- It works!
`web_1                      | [2018-11-09T21:47:25] INFO  Rails : Sending data: {"series":[{"metric":"idseq.web.development.samples.uploaded","points":[[1541800045,1]]}]}`
- DATADOG_API_KEY is set through chamber
- Next steps are adding in more hooks for things and adding stuff to take advantage of tagging like tagging with user_ids


app/controllers/samples_controller.rb
```diff
@@ -156,6 +156,7 @@ def bulk_upload
 
     respond_to do |format|
       if @errors.empty? && !@samples.empty?
+        MetricUtil.put_metric_now("samples.uploaded", @samples.count)
         format.json { render json: { samples: @samples, sample_ids: @samples.pluck(:id) } }
       else
         format.json { render json: { samples: @samples, errors: @errors }, status: :unprocessable_entity }
```
tfrcarvalho: Can you add a metrics for error when uploading?

app/controllers/samples_controller.rb
```diff
@@ -156,6 +156,7 @@ def bulk_upload
 
     respond_to do |format|
       if @errors.empty? && !@samples.empty?
+        MetricUtil.put_metric_now("samples.uploaded", @samples.count)
```
tfrcarvalho: What is the difference between this upload and the one below?
Do metrics allow some kind of context or attributes to differentiate the two? Or would it have to be a different metric?

app/lib/metric_util.rb
```diff
@@ -0,0 +1,52 @@
...
+    put_metric_point_series(name, points)
+  end
+
+  def self.put_metric_point_series(name, points)
+    name = "idseq.web.#{Rails.env}.#{name}"
+    data = JSON.dump("series" => [{
+                       "metric" => name,
+                       "points" => points
+                     }])
+    post_to_datadog(data)
```
tfrcarvalho: We could make it post to google analytics too... easier integration with the dashboard

app/controllers/samples_controller.rb
jshoe: This was in the bulk upload so it could be like 10 samples uploaded at a time.

Yes, you can add tags to the metrics (under the hood it counts as multiple metrics) https://docs.datadoghq.com/api/?lang=python#post-timeseries-points I plan to add the tags gradually for things like user_id

app/lib/metric_util.rb
jshoe: good idea, I will add the authorization/Google integration later

PR #1725, opened 2018-11-09
No more backticks in idseq-web code

jameshowardwang:
Removing all use of backticks to run shell commands, adding a helper library to make things easier

## Type of change


tfrcarvalho:
Do we still want to move from sys calls to api?

jameshowardwang:
i don’t see any urgency around it. i commented in the JIRA task as well.

PR #1726, opened 2018-11-09
organize results folder

cdebourcy:
results_folder


app/models/sample.rb
```diff
@@ -191,6 +191,7 @@ def results_folder_files
     if pr.pipeline_version.to_f >= 2.0
       file_list = list_outputs(pr.output_s3_path_with_version)
       file_list += list_outputs(sample_output_s3_path)
+      file_list += list_outputs(pr.postprocess_output_s3_path)
       file_list += list_outputs(pr.postprocess_output_s3_path + '/' + ASSEMBLY_DIR, 2)
```
jameshowardwang: do we still need the "assembly" prefix if they're grouped together now? (display_prefix=2 could be removed)

app/models/pipeline_run.rb
```diff
@@ -1061,6 +1061,42 @@ def compare_ercc_counts
     ret
   end
 
+  def outputs_by_step
+    return unless pipeline_run_stages
```
jameshowardwang: is this line necessary? if there are no pipeline_run_stages then the block below just won't run anything right? when might there be files returned by sample.result_folder_files but no pipeline_run_stages? and in that case, would we  want to show nothing at all?

app/models/sample.rb
cdebourcy: Agreed

app/models/pipeline_run.rb
cdebourcy: agreed, I removed the line

yunfangjuan:
where do you put all the assembly stuff? 
Could we talk about this in today's meeting? My main concern is the pre-bowtie step the fqs include some human data.  do we really want to let people download them?

cdebourcy:
> where do you put all the assembly stuff?
> Could we talk about this in today's meeting? My main concern is the pre-bowtie step the fqs include some human data. do we really want to let people download them?

Assembly stuff will be grouped under the step names defined in the DAG. Eventually we should add descriptions for each of these steps too (should be part of your assembly web PR or you can send me some descriptions and I'll include them here).

Users can already download the pre-bowtie fqs in the current result folder so not sure it's a concern for this PR. There should be very little host left in there after STAR. But we can think about how to tweak access control more.

PR #1729, opened 2018-11-09
Report page: Refresh every 5 minutes while pipeline in progress

jshoe:
- People complained the state wouldn't change / they wouldn't know when their sample was done
- Tested locally


jshoe:
@tfrcarvalho this one doesn't do an axios call to fetch the data (and everything is in props) so it'd require a bit more of a rewrite or making another controller action. What do you think?

tfrcarvalho:
ok, i see. All props come from rails.
I just want to make sure this does not survive once we refactor.
Can you add a comment explaining why you do the refresh and note should not be kept once we refactor?

jshoe:
@tfrcarvalho updated!

PR #1732, opened 2018-11-13
Add warning message to alignment viz.

MarkAZhang:
Also fix warning message for sample report.
Also change sample report tooltip to be 'hover' instead of 'click'.

screen shot 2018-11-12 at 9 56 37 pm


app/assets/src/components/AlignmentViz.jsx
```diff
@@ -1,6 +1,9 @@
-import axios from "axios";
 import React from "react";
+import cx from "classnames";
+import { Popup } from "semantic-ui-react";
+import { getSampleMetadata, fetchAlignmentData } from "~/api";
```
tfrcarvalho: nit: can we may keep the prefixes consistent: always `get` ?

app/assets/src/components/AlignmentViz.jsx
```diff
@@ -1,6 +1,9 @@
-import axios from "axios";
 import React from "react";
```
tfrcarvalho: Can you move this file into its views?
Since you are adding a new css file below... i would prefer to have it organized even if it is not fully refactored.
(Unless you are planning to rewrite it from scratch)

app/assets/src/components/AlignmentViz.jsx
MarkAZhang: Definitely, thanks for catching that.

yunfangjuan:
Great! Thanks! I will let @tfrcarvalho approve~

PR #1735, opened 2018-11-13
Add download options to download button and details sidebar.

MarkAZhang:
Download options were added to Button:
screen shot 2018-11-13 at 3 27 47 pm

And also Sidebar:
screen shot 2018-11-13 at 3 27 50 pm


app/controllers/samples_controller.rb
```diff
@@ -189,6 +190,7 @@ def metadata
       ercc_comparison = pr.compare_ercc_counts
 
       job_stats_hash = job_stats_get(pr.id)
+      assembled_taxids = pr.assembled_taxids ? JSON.parse(pr.assembled_taxids) : []
```
jshoe: I think you can put assembled_taxids = [] up top in case pr is nil

app/assets/src/components/views/report/utils/download.js
```diff
@@ -0,0 +1,73 @@
...
+const getDownloadOptions = (pipelineRun, assembledTaxids) => {
+  let stageTwoComplete = pipelineRun && pipelineRun.adjusted_remaining_reads;
+  let nonhostAssemblyComplete =
+    assembledTaxids && assembledTaxids.indexOf("all") >= 0;
+  const assembled = pipelineRun && pipelineRun.assembled === 1;
+
+  return compact([
+    stageTwoComplete && NON_HOST_READS_LABEL,
+    assembled && NON_HOST_CONTIGS_LABEL,
+    nonhostAssemblyComplete && NON_HOST_ASSEMBLIES_LABEL,
```
jshoe: I'm assuming YF said, but what's the difference between assembly being complete and non-host assembly being complete?

app/controllers/samples_controller.rb
cdebourcy: Another way of writing it that's cute is `JSON.parse(pr.assembled_taxids || "[]")`

app/assets/src/components/views/report/utils/download.js
MarkAZhang: I will check with her. This logic is already used in the code base for the buttons on the Sample Details tab, so not going to block this PR on that.

app/controllers/samples_controller.rb
MarkAZhang: Thanks!

PR #1736, opened 2018-11-14
Assembly annotation

yunfangjuan:
# Description

1. add taxon annotation  (full lineage) to each contig 
2. output a csv for that particular purpose
3. remove contigs as a result monitor target. make it a part of the contig_counts 
# Type of change

# Test and Reproduce

1. 

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page

###Single Upload

###Batch Upload

###All Project Page
* Switch between projects 
* Search for a sample

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:


app/models/pipeline_run.rb
```diff
@@ -319,25 +320,32 @@ def db_load_ercc_counts
...
     contig_stats_json = JSON.parse(File.read(downloaded_contig_stats))
     return if contig_stats_json.empty?
+
     update(assembled: 1)
     contig_fasta = PipelineRun.download_file_with_retries(contig_s3_path, LOCAL_JSON_PATH, 3)
     contig_array = []
+    taxid_list = []
+    contig2taxid.values.each { |entry| taxid_list += entry.values }
+    taxon_lineage_map = {}
+    TaxonLineage.where(taxid: taxid_list.uniq).each { |t| taxon_lineage_map[t.taxid.to_i] = t.to_a }
```
cdebourcy: do you need to select only the most recent / the most appropriate?

app/models/taxon_lineage.rb
```diff
@@ -159,6 +159,21 @@ def self.fill_lineage_details(tax_map, pipeline_run_id)
     tax_map
   end
 
+  def to_a
+    [species_taxid, genus_taxid, family_taxid, order_taxid, class_taxid, phylum_taxid,
```
cdebourcy: Could the following help with maintainability?
```
def to_a
  self.names_a.map { |col| self[col] }
end
```
and
```
def self.null_array
  TaxonLineage.column_defaults.values_at(*self.names_a)
end
```
That way, the list of columns is only hardcoded in names_a. I think it would be preferable

app/models/pipeline_run.rb
yunfangjuan: great point. i can make it oder by id so the latest one will always be there.

app/models/taxon_lineage.rb
yunfangjuan: good point.

app/models/pipeline_run.rb
```diff
@@ -365,9 +419,15 @@ def db_load_contig_counts
                                  tax_level: tax_entry['tax_level'],
                                  contig_name: contig_name,
                                  count: count }
+        if tax_entry['tax_level'].to_i == 1 # species
```
cdebourcy: TaxonCount:: TAX_LEVEL_SPECIES?

app/models/pipeline_run.rb
```diff
@@ -346,9 +353,55 @@ def db_load_contigs
...
+      contig_array << { name: header, sequence: sequence, read_count: read_count, lineage_json: lineage_json.to_json }
     end
+    contigs.destroy_all
     update(contigs_attributes: contig_array) unless contig_array.empty?
+    generate_contig_mapping_table
+    update(assembled: 1)
+  end
+
+  def contigs_fasta_s3_path
+    return "#{postprocess_output_s3_path}/#{ASSEMBLED_CONTIGS_NAME}" if pipeline_version && pipeline_version.to_f >= 3.1
```
cdebourcy: should we define `ASSEMBLY_PIPELINE_VERSION = '3.1'.freeze`?

PR #1740, opened 2018-11-14
Restrict access to s3 buckets and remove bucket names from examples

jameshowardwang:
Test plan:
- did some testing in rails console
- logged in as non-admin user and confirmed inability to upload from restricted buckets
- logged in as user with "czbiohub.org" email and confirmed ability to upload from biohub buckets


cdebourcy:
Cool. Should we make the real-life examples admin-only instead? It's super useful for testing

PR #1744, opened 2018-11-15
Heatmap refactor 

tfrcarvalho:
# Description

Refactor heatmap in preparation to add metadata.

* Bug fixes:
  * Fix clustering:
    * null values count for the difference between rows and between columns (null values are replaced by zero);
    * use log-eucledian distance when log view is enabled
  * Fixed a bug on threshold filters where a second change to value on a filter would not have an effect
* Refactor code
  * Heatmap split into react and D3 components (`SamplesHeatmapVis` and `Heatmap.js`)
  * Refactored symlog to allow some more operations with log
  * Refactor how data is passed to heatmap to improve encapsulation
  * New generic `SequentialLegend` component
* UX changes:
  * Download as SVG
  * Taxon labels on the left
  * Curve edges in dendogram
  * Use common `DataToolTip` component
  * New color scheme
  * Header: Use common `ViewHeader`; Thinner divider in the menu;
  * Re-align filter dropdowns and add `ellipsis` (...) as overflow behavior
  * Links (from samples label and cell) open in the same page or, if user press CMD or CTRL while clicking, in a new tab
* Side changes:
  * New `NarrowContainer` component

## Todo
  * Improve legend: round limits and add intermediate levels (in particular, for log scale view, the legend is misleading)

## Mock
(slightly outdated since in the meantime i merged @MarkAZhang's sidebar)

heatmap


jameshowardwang:
looks awesome! who is jazzmes? tiago's alter-ego?

cdebourcy:
Cool! Would you mind documenting how null values are handled for the different metrics? I'm assuming for `r` you treat them as zeroes?

tfrcarvalho:
@jameshowardwang yeah... did not notice I was publishing as that user.

tfrcarvalho:
@cdebourcy Good question/comment. I am treating all of them as zero. 
I wonder if that makes sense for aggregate score, since there can be negative values... wdyt?

cdebourcy:
> @cdebourcy Good question/comment. I am treating all of them as zero.
> I wonder if that makes sense for aggregate score, since there can be negative values... wdyt?

Yeah I think it is not right for either z-scores or the aggregate score, because those are negative when the sample's rpm is less than the background's. So setting them to 0 distorts the information. Since z-scores and aggregate scores are computed from rpm, the correct way to handle this would be to set rpm to 0 when the data is prepared in the controller, and derive the appropriate z-score / aggregate score values from there.

app/assets/src/components/layout/divider.scss
```diff
@@ -2,5 +2,5 @@
 
 .divider {
   height: 0px;
-  border-bottom: 8px solid $off-white;
+  border-bottom: 2px solid $off-white;
```
MarkAZhang: I think this breaks the styling on the Sample Report tabs. Since the blue underline is still 8px

app/assets/src/components/views/compare/SamplesHeatmapView.jsx
```diff
@@ -67,45 +70,30 @@ class SamplesHeatmap extends React.Component {
...
-    this.getColumnLabel = this.getColumnLabel.bind(this);
-    this.getRowLabel = this.getRowLabel.bind(this);
-    this.getTaxonFor = this.getTaxonFor.bind(this);
-    this.getTooltip = this.getTooltip.bind(this);
-    this.onApplyClick = this.onApplyClick.bind(this);
-    this.onCellClick = this.onCellClick.bind(this);
-    this.onRemoveRow = this.onRemoveRow.bind(this);
+    this.handleRemoveTaxon = this.handleRemoveTaxon.bind(this);
+    this.handleSampleLabelClick = this.handleSampleLabelClick.bind(this);
+    this.handleDownloadClick = this.handleDownloadClick.bind(this);
```
MarkAZhang: Use arrow functions instead of `bind`

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -0,0 +1,155 @@
...
+      { key: "NT.aggregatescore", label: "Score" },
+      { key: "NT.zscore", label: "NT Z Score" },
+      { key: "NT.rpm", label: "NT rPM" },
+      { key: "NT.r", label: "NT r (total reads)" },
+      { key: "NR.zscore", label: "NR Z Score" },
+      { key: "NR.rpm", label: "NR rPM" },
+      { key: "NR.r", label: "NR r (total reads)" }
+    ];
+
+    this.handleCellClick = this.handleCellClick.bind(this);
```
MarkAZhang: Use arrow function

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -0,0 +1,155 @@
...
+
+  handleCellClick(cell) {
+    openUrl(`/samples/${this.props.sampleIds[cell.columnIndex]}`, event);
+  }
+
+  render() {
+    return (
+      <div className="samples-heatmap-vis">
+        <div
+          className="samples-heatmap-vis__container"
```
MarkAZhang: Could you convert the styling to CSS modules for this file? If not too much trouble.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -0,0 +1,155 @@
...
+          ref={container => {
+            this.heatmapContainer = container;
+          }}
+        />
+        <div
+          id="heatmap"
+          className="samples-heatmap-vis__tooltip"
+          ref={tooltip => {
```
MarkAZhang: Should we use `React.createRef` for consistency?

app/assets/src/components/views/compare/compound.js
```diff
@@ -0,0 +1,65 @@
+export default function compound() {
```
MarkAZhang: Could you add a comment here and maybe name the function better?  Hard to understand what this function does.

app/assets/src/components/views/compare/symlog.js
```diff
@@ -0,0 +1,98 @@
+import { default as compound } from "./compound";
```
MarkAZhang: `import compound from "./compound";`

app/assets/src/components/layout/divider.scss
tfrcarvalho: I confirmed that this is not affecting the report page.

app/assets/src/components/ui/controls/dropdowns/ThresholdFilterDropdown.jsx
```diff
@@ -235,17 +234,17 @@ const ThresholdFilter = ({
...
         />
       </Grid.Column>
       <Grid.Column>
         <Input
           fluid
           focus
           type="number"
-          onChange={handleValueChange}
-          value={threshold.value}
+          onChange={handleValueChange.bind(this)}
```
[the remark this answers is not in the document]
tfrcarvalho: Not sure why this ended up here. Deleting.

app/assets/src/components/utils/links.js
```diff
@@ -0,0 +1,16 @@
+const openUrl = (link, event) => {
```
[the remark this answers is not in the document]
tfrcarvalho: The whole purpose is to abstract event evaluation. I do not want the client to worry about the event and let the function decide how to do the link opening.
Event is also optional.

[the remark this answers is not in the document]
tfrcarvalho: Actually event is javascript global, so i can remove it.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -0,0 +1,155 @@
...
+      },
+      {
+        name: "Values",
+        data: nodeHasData ? values : [["", "Taxon not found in Sample"]]
+      }
+    ];
+  }
+
+  handleCellClick(cell) {
+    openUrl(`/samples/${this.props.sampleIds[cell.columnIndex]}`, event);
```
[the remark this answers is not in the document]
tfrcarvalho: Actually, it is global. Can remove it from here.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
tfrcarvalho: This is actually the consistent way.... and the way i like for d3 integration.
I might change the createref i used in the `SamplesHeatmapView`

app/assets/src/components/views/compare/compound.js
tfrcarvalho: Moving it to `utils/scales`. The name makes sense in the context of scale.

app/assets/src/components/views/compare/symlog.js
tfrcarvalho: done

app/assets/src/loader.scss
```diff
@@ -55,5 +56,6 @@
 @import "styles/views/taxon_tree_vis";
 
 // visualizations
+@import "styles/visualizations/heatmap/heatmap";
```
MarkAZhang: Would be great to not add more `.scss` files to this file. If you use CSS Modules, you don't need to import the file here any more.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
tfrcarvalho: Done.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
tfrcarvalho: Done
A note (I OCD about this... :) ): i'm not a fan of mixing both. Should we start enforcing only arrow function, event if this is not used?

app/assets/src/loader.scss
tfrcarvalho: this is for the d3 file. And i ran into issues applying to d3. Will fix in a follow up.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
tfrcarvalho: done

app/assets/src/components/views/compare/SamplesHeatmapView.jsx
tfrcarvalho: done

app/assets/src/components/utils/links.js
```diff
@@ -0,0 +1,16 @@
+const openUrl = link => {
+  let openInNewTab = false;
+
+  // metakey is CMD in mac
+  if (event && (event.metaKey || event.ctrlKey)) {
```
MarkAZhang: This method doesn't have `event` parameter any more.
It might be good for the function to receive `openInNewTab` as a parameter directly.

PR #1749, opened 2018-11-17
remove phylo_tree creation conditions

cdebourcy:


app/assets/src/components/views/phylo_tree/PhyloTreeChecks.js
```diff
@@ -1,4 +1,4 @@
-const MIN_READS = 100;
+const MIN_READS = 1;
```
tfrcarvalho: Just trying to understand the new limit: does it make sense to have it for 1 read?

app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
```diff
@@ -13,7 +13,7 @@ class PhyloTreeCreation extends React.Component {
   constructor(props) {
     super(props);
 
-    this.minNumberOfSamples = 4;
+    this.minNumberOfSamples = 2;
```
tfrcarvalho: Can you add some comments to the limits explaining the rationale for the choices.
In this case, do we complement with NCBI references (since a tree with only two branches might not make much sense)?

app/assets/src/components/views/phylo_tree/PhyloTreeChecks.js
cdebourcy: Setting it to 1 is equivalent to not having a limit at all (other than presence/absence). Everybody is confused by the limit because it is not stated explicitly -- looks like a bug -- so we should remove it.

In the future, once the workflow goes through a redesign, we should reinstate it, but as a warning / a grayed-out checkbox explaining what's going on.

app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
cdebourcy: You're right, 4 data points are needed for a tree. However, it seems people who only have a couple samples uploaded get confused when they can't make a tree. I chose 2 because typically we pull in 1 NCBI reference per sample, so the total would come to 4. I haven't completely decided whether to go through with this change, because the point you make is valid and users SHOULD recognize that...

The correct solution in the future will be to add an explicit NCBI reference selector in the workflow (which we postponed to later), and reinstate the limit of 4.

cdebourcy: Update: added comment for one of the limits, removed the change to the 4 threshold for now

PR #1750, opened 2018-11-17
refactor autoscaler

cdebourcy:
To be tested in staging.
Also, need to reduce variables affecting latency of scale-down -- it is significant (can be close to 30 minutes).


app/jobs/autoscaling.py
```diff
@@ -260,227 +61,313 @@ def get_instance_tags():
...
+        # will be respected and not scaled down until we exit this regime.
+        result = at_least(24)
+    return result
 
-def start_draining(asg_instance_name, draining_tag, num_instances):
-    instance_ids = instances_to_drain(asg_instance_name, draining_tag, num_instances)
-    if instance_ids:
-        print "Starting to drain the following instances: " + ",".join(instance_ids)
-        add_draining_tag(instance_ids, draining_tag)
+def autoscaling_update(my_num_jobs, my_environment="development",
```
boris-dimitrov: not a big deal but i'd refrain from providing a default for my_environment

app/jobs/autoscaling.py
```diff
@@ -260,227 +61,313 @@ def get_instance_tags():
...
+                       max_job_dispatch_lag_seconds=900, job_tag_prefix="RunningIDseqBatchJob_",
+                       job_tag_keep_alive_seconds=600, draining_tag="draining"):
```
boris-dimitrov: will these params ever be modified here?   if not, i'd just move them in the ASG constructor, so we don't have to pass them through

cdebourcy: Yes, they will be modified, in fact defaults are only for documentation. Maybe I should remove them. They're passed from the DAG template to (a) the autoscaler, (b) the idseq-dag code

app/jobs/autoscaling.py
cdebourcy: ok I can remove but it was there before...?

app/jobs/autoscaling.py
```diff
@@ -260,227 +61,313 @@ def get_instance_tags():
...
+    Note: Will not scale up for jobs originating in a development environment.
+          You can provision machines for development jobs by manually setting MinSize on the ASGs,
+          which is respected by this autoscaler.
+    '''
 
+    asg_list = get_asg_list()
+    tag_list = get_tag_list()
+    gsnap_ASG = ASG("gsnapl", my_environment, asg_list, tag_list, draining_tag, job_tag_prefix, max_job_dispatch_lag_seconds, job_tag_keep_alive_seconds)
+    rapsearch_ASG = ASG("rapsearch2", my_environment, asg_list, tag_list, draining_tag, job_tag_prefix, max_job_dispatch_lag_seconds, job_tag_keep_alive_seconds)
+    if not (gsnap_ASG.can_scale and rapsearch_ASG.can_scale):
```
boris-dimitrov: Is this still necessary?   If I understand correctly, the new rule is that development cannot scale anything;  and every other environment has exactly one autoscaler, which is permitted to scale it.   The reason this code was written originally was twofold:   development had to be prevented from scaling, but also there was the opportunity for a staging autoscaler to be allowed to scale the production ASGs in case the production autoscaler died.  This is too sophisticated and unnecessary now, I think.

boris-dimitrov: I mean this particular code can stay but the tags that previously determined can_scale should probably be retired.

cdebourcy: You're right -- I will remove

app/jobs/autoscaling.py
```diff
@@ -260,227 +61,313 @@ def get_instance_tags():
...
+        num_desired = service_ASG.num_desired_instances()
+        healthy_instances, draining_instances, terminating_instances, corrupt_instances = service_ASG.classify_instances()
+        num_healthy = len(healthy_instances)
+        num_draining = len(draining_instances)
+        print "CURRENTLY:"
+        print "The desired capacity is {num_desired}.".format(num_desired=num_desired)
+        print "There are {num_healthy} healthy instances: {healthy_instances}.".format(num_healthy=num_healthy, healthy_instances=healthy_instances)
+        print "There are {num_draining} draining instances: {draining_instances}.".format(num_draining=num_draining, draining_instances=draining_instances)
+        print "There are {num_terminating} terminating instances: {terminating_instances}.".format(num_terminating=len(terminating_instances), terminating_instances=terminating_instances)
+        assert not corrupt_instances, "There are {num_corrupt} corrupt instances: {corrupt_instances}.".format(num_corrupt=len(corrupt_instances), corrupt_instances=corrupt_instances)
```
boris-dimitrov: Are you sure you want to crash here?   Who would notice this?

app/jobs/autoscaling.py
```diff
@@ -260,227 +61,313 @@ def get_instance_tags():
...
+        corrupt_instances = []
+        tag_dict = self.tags_by_instance_id()
+        for inst in self.asg["Instances"]:
+            instance_id = inst["InstanceId"]
+            instance_tags = tag_dict.get(instance_id, {})
+            if DEBUG:
+                print "{instance_id}: {instance_tags}".format(instance_id=instance_id, instance_tags=instance_tags)
+            if not inst["ProtectedFromScaleIn"] and not inst["LifecycleState"] == "Pending":
+                terminating_instances.append(instance_id)
+            elif self.draining_tag in instance_tags:
```
boris-dimitrov: I suggest avoiding "elif" for something like this, because the conditions can pile up in difficult-to-comprehend ways.    Right now this entire statement is equivalent to the following;  are all those conjuncts correct?

```
if not inst["ProtectedFromScaleIn"] and not inst["LifecycleState"] == "Pending":
    terminating_instances.append(instance_id)

if (inst["ProtectedFromScaleIn"] or inst["LifecycleState"] == "Pending") and self.draining_tag in instance_tags:
    draining_instances.append(instance_id)

if (inst["ProtectedFromScaleIn"] or inst["LifecycleState"] == "Pending") and not self.draining_tag in instance_tags and inst["HealthStatus"] == "Healthy" and inst["LifecycleState"] != "Terminating":
    healthy_instances.append(instance_id)

if (inst["ProtectedFromScaleIn"] or inst["LifecycleState"] == "Pending") and not self.draining_tag in instance_tags and (inst["HealthStatus"] != "Healthy" or inst["LifecycleState"] == "Terminating")
    corrupt_instances.append(instance_id)
```

By that last one, it's difficult to follow exactly what condition makes an instance corrupt.   It would be best if that were simplified, avoiding the use of `elif` so these conjunct pileups don't lead to unintended consequences.

After that, you would wish to enforce mutual exclusion on the states.   If an instance occurs in more than one state, you can dump an error message, as it means logic error in this classifier.

boris-dimitrov: I'll say that all things considered it's very nice to have any such complexity encapsulate here in this function, so the rest of the program can deal in 4 simple mutually exclusive states.

app/jobs/autoscaling.py
```diff
@@ -260,227 +61,313 @@ def get_instance_tags():
...
+        new_num_desired = service_ASG.clamp_to_valid_range(raw_new_num_desired)
+        print "Enforcing the MinSize and MaxSize set by the operator, the desired capacity will be set to {new_num_desired}.".format(new_num_desired=new_num_desired)
+        service_ASG.set_desired_capacity(new_num_desired)
+
+        print "HEALTHY <--> DRAINING transitions:"
+        service_ASG.draining_instances = draining_instances
+        if new_num_desired < num_healthy:
+            num_to_drain = num_healthy - new_num_desired
+            instances_to_drain = service_ASG.start_draining(num_to_drain)
+            print "{num_to_drain} instances need to be drained: {instances_to_drain}.".format(num_to_drain=num_to_drain, instances_to_drain=instances_to_drain)
```
boris-dimitrov: I'd put the print before the function that does the work, just in case it crashes....

app/jobs/autoscaling.py
```diff
@@ -260,227 +61,313 @@ def get_instance_tags():
...
+        elif new_num_desired > num_healthy:
```
boris-dimitrov: It seems from the `else` print below, you need conjunct `... and num_draining > 0` here.

app/jobs/autoscaling.py
```diff
@@ -260,227 +61,313 @@ def get_instance_tags():
...
-        t.start()
-    for t in threads:
-        t.join()
-    if errors:
-        return None
-    return sum(len(job_list["jobSummaryList"]) for job_list in queues.itervalues())
+                            _________________
+                           |                 |   Discarded.
+                           |   TERMINATING   |   Scale-in protection removed.
+                           |_________________|   Instance will terminate due to discrepancy between DesiredCapacity and actual number of instances.
```
boris-dimitrov: This is beautiful, the part I am curious about is "Instance will terminate due to discrepancy between DesiredCapacity and actual number of instances."    I think you mean, if desired capacity is lower than the current capacity and if it stays lower, then the instances here will be terminated.   But what if it doesn't stay lower -- if you come around in the next round of autoscaling and raise it.   What happens to any instances still in this state?

app/jobs/autoscaling.py
```diff
@@ -260,227 +61,313 @@ def get_instance_tags():
...
-            elif crbj == 0:
-                print "No jobs are running in batch."
-                # in this case num_development_jobs are either zombies or jobs caught in transition
-                # between pipeline stages;  the non-zombies won't be hurt, only delayed by this
-                set_desired_capacity(gsnap_asg, gsnap_instance_name, gsnap_tags, draining_tag, exactly(0), can_scale)
-                set_desired_capacity(rapsearch2_asg, rapsearch_instance_name, rapsearch_tags, draining_tag, exactly(0), can_scale)
+    # If you are running into problems with this autoscaler, just delete this tag from
+    # any affected autoscaling group, and it will be left alone.  The value is a list
+    # of environments permitted to trigger scaling, usually just "prod".
+    scaling_permission_tag = 'IDSeqEnvsThatCanScale'
```
boris-dimitrov: Oh wait this comment explains why I had originally built "can_scale" this way.  I had thought there could be bugs in this script and we could just take over manually in that case by removing the tag from the ASG.   You know what... that was a good idea... :)   Because you don't want to have to force a production web push to deal with pipeline operational problems.   I think perhaps this is worth leaving in, after all.

app/jobs/autoscaling.py
cdebourcy: removed

app/jobs/autoscaling.py
cdebourcy: ok done

app/jobs/autoscaling.py
cdebourcy: based on subsequent discussion, leaving this in. It provides a mechanism to easily remove ASGs from the autoscaler's authority by deleting tags if things go wrong

app/jobs/autoscaling.py
cdebourcy: good point. I will just print a warning. I have not encountered this case in my testing and doubt it exists. In any case, do not want to crash.

app/jobs/autoscaling.py
cdebourcy: Yes, I've thought about this and commented on it. We enforce a certain time interval between incarnations of the autoscaler. This interval guarantees that AWS has time to initiate termination of instances before any new adjustments of DesiredCapacity are made through a new incarnation of the autoscaler

app/jobs/autoscaling.py
cdebourcy: The assumption (enforced in the rake code) is that incarnations of the autoscaler are spaced by enough time to prevent this case from occurring

app/jobs/autoscaling.py
cdebourcy: Yes!

app/jobs/autoscaling.py
cdebourcy: ok. I refactored to have the conditions for each state fully written out. I removed the "corrupt" state and replaced it with a check that each instance is in exactly 1 of the valid states.

app/jobs/autoscaling.py
cdebourcy: Ok following our discussion I have put a check for "discarded but not terminated" instances in place. Any increases in DesiredCapacity will be postponed until there are no such instances.

I have also moved some print statements before the functions that do the work.

app/jobs/autoscaling.py
cdebourcy: Update: put a check in place that should guarantee the case will not occur

app/jobs/autoscaling.py
cdebourcy: Ok you're right. It is actually required for the logic. Even though the effects are only cosmetic. Changed

app/jobs/autoscaling.py
```diff
@@ -260,227 +61,343 @@ def get_instance_tags():
...
+        terminating_instances = []
+        tag_dict = self.tags_by_instance_id()
+        for inst in self.asg["Instances"]:
+            instance_id = inst["InstanceId"]
+            instance_tags = tag_dict.get(instance_id, {})
+            if DEBUG:
+                print "{instance_id}: {instance_tags}".format(instance_id=instance_id, instance_tags=instance_tags)
+            is_terminating = (inst["LifecycleState"] == "Terminating")
+            is_discarded = (not inst["ProtectedFromScaleIn"] and not inst["LifecycleState"] in ["Pending", "Terminating"])
+            is_draining = ((inst["ProtectedFromScaleIn"] or inst["LifecycleState"] == "Pending") and
```
boris-dimitrov: Doesn't draining require the instance to be protected from scale in?    That would suggest inst["ProtectedFromScaleIn"] should be a conjunct above, not a disjunct.

app/jobs/autoscaling.py
```diff
@@ -260,227 +61,343 @@ def get_instance_tags():
...
+                           self.draining_tag in instance_tags)
+            is_healthy = ((inst["ProtectedFromScaleIn"] or inst["LifecycleState"] == "Pending") and
+                          not self.draining_tag in instance_tags and
+                          inst["HealthStatus"] == "Healthy" and inst["LifecycleState"] != "Terminating")
```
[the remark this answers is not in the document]
boris-dimitrov: Yes, I would recommend removing anything extraneous like that :)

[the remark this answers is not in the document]
boris-dimitrov: That extra stuff was implicit in a previous version of the code via the use of "elif".    The fact you can remove it means it was a good call to get rid of the "elif" chaining here.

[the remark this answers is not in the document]
cdebourcy: ok removed redundant conditions and changed 'healthy' to 'available', with an explanation of how 'available' includes both pending and ready

PR #1751, opened 2018-11-18
upgrade node to prevent npm error during build

cdebourcy:
Error was:
```
npm ERR! code ERR_STREAM_WRITE_AFTER_END
npm ERR! write after end
```
see https://github.com/npm/npm/issues/19989


tfrcarvalho:
I saw this happening last thursday at least.
Btw, why do we need to use a PPA version to install node?

PR #1753, opened 2018-11-19
add DAG json to result folder

cdebourcy:
image

image

### DIFF
--- a/app/assets/src/components/ResultsFolder.jsx
+++ b/app/assets/src/components/ResultsFolder.jsx
@@ -43,11 +43,11 @@ class ConfigFile extends React.Component {
 
   render() {
     return (
-      <tr>
-        <td
-          className="file-link"
-          onClick={this.download_string2file.bind(this, this.stageDagJson)}
-        >
+      <tr
+        className="file-link"
+        onClick={this.download_string2file.bind(this, this.stageDagJson)}
+      >
+        <td>
           <i className="fa fa-file" />
           config.json
         </td>
--- a/app/assets/src/components/ResultsFolder.jsx
+++ b/app/assets/src/components/ResultsFolder.jsx
@@ -1,12 +1,69 @@
 import React from "react";
 import Divider from "./layout/Divider";
 
+class ResultsFolderStepList extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.stepDict = props.stepDict;
+  }
+
+  download(url) {
+    if (url) {
+      location.href = `${url}`;
+    }
+  }
+
+  render() {
+    console.log(this.stepDict);
+    return Object.keys(this.stepDict).map((step_name, i) => {
+      let step = this.stepDict[step_name];
+      let description = step["step_description"];
+      let reads_after = step["reads_after"];
+      let fileList = step["file_list"];
+      return (
+        <tbody key={i}>
+          <tr key="first">
+            <td>
+              Step <b>{step_name}</b>: {description}{" "}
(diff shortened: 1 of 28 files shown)


app/assets/src/components/ResultsFolder.jsx
```diff
@@ -1,12 +1,123 @@
...
+
+class ConfigFile extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.stageDagJson = props.stageDagJson;
+  }
+
+  download_string2file(str) {
+    let file = new Blob([str], { type: "text/plain" });
+    let download_url = URL.createObjectURL(file);
```
cdebourcy: @MarkAZhang are there any risks to this?

app/assets/src/components/ResultsFolder.jsx
```diff
@@ -1,12 +1,123 @@
...
+    if (url) {
+      location.href = `${url}`;
+    }
+  }
+
+  render() {
+    return (
+      <tr
+        className={`${this.file.url ? "" : "disabled-"}file-link`}
+        onClick={this.download.bind(this, this.file.url)}
```
jshoe: For standardization this should use utils/links.js

app/assets/src/components/ResultsFolder.jsx
```diff
@@ -1,12 +1,123 @@
...
+  download(url) {
+    if (url) {
+      location.href = `${url}`;
+    }
+  }
+
+  render() {
+    return (
+      <tr
+        className={`${this.file.url ? "" : "disabled-"}file-link`}
```
jshoe: re className all new code should use CSS Modules and camelCase. See ui/controls/Tabs.jsx and tabs.scss

app/assets/src/components/ResultsFolder.jsx
```diff
@@ -1,12 +1,123 @@
...
+  }
+}
+
+class ConfigFile extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.stageDagJson = props.stageDagJson;
+  }
+
+  download_string2file(str) {
```
jshoe: same with these functions for camelCase and CSS Modules

jshoe: well, I guess the CSS is not really new

app/models/pipeline_run_stage.rb
```diff
@@ -58,6 +58,11 @@ def checked?
     job_status == STATUS_CHECKED
   end
 
+  def redacted_dag_json
+    # redact any s3 paths
+    dag_json.gsub(%r{(\"s3://).*(\")}, '"s3://..."')
```
yunfangjuan: awesome!

app/assets/src/components/ResultsFolder.jsx
cdebourcy: yeah the CSS is not new and the download function is also not new.  Just broke single ResultsFolder component into some smaller pieces.
I will rename `downloadString2file`. This whole page is going to be redesigned and refactored in the future

app/assets/src/components/ResultsFolder.jsx
cdebourcy: ok moved

app/assets/src/components/ResultsFolder.jsx
cdebourcy: I couldn't get the modules to work. Will bear in mind for new CSS in the future and seek clarification

PR #1756, opened 2018-11-20
Bugfix: default null value outside of the data limits inferred from data.

tfrcarvalho:


app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -132,8 +133,14 @@ export default class Heatmap {
     this.columnLabelsHeight += this.options.spacing;
 
     this.limits = {
-      min: d3.min(this.data.values, array => d3.min(array)),
-      max: d3.max(this.data.values, array => d3.max(array))
```
jshoe: so what were these returning without the nullValue?

tfrcarvalho: they were just computed based on the data. The problem was if zero was outside of the limits then our d3 scales would return NaN.

PR #1758, opened 2018-11-20
Make row hover actions always visible, with a disabled state.

MarkAZhang:
Also show tooltips when disabled.

Fix issue where phylo button had no tooltip.

Refactor PhyloTreeCreationModal to take open prop instead of trigger prop.

screen shot 2018-11-20 at 3 40 31 pm

screen shot 2018-11-20 at 3 40 38 pm

### DIFF
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -27,10 +27,10 @@ import MetricPicker from "./views/report/filters/MetricPicker";
 import SpecificityFilter from "./views/report/filters/SpecificityFilter";
 import NameTypeFilter from "./views/report/filters/NameTypeFilter";
 import ReportSearchBox from "./views/report/filters/ReportSearchBox";
-import PhyloTreeCreationModal from "./views/phylo_tree/PhyloTreeCreationModal";
 import PhyloTreeChecks from "./views/phylo_tree/PhyloTreeChecks";
 import TaxonTreeVis from "./views/TaxonTreeVis";
 import LoadingLabel from "./ui/labels/LoadingLabel";
+import HoverActions from "./views/report/ReportTable/HoverActions";
 
 class PipelineSampleReport extends React.Component {
   constructor(props) {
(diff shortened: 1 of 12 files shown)


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -702,87 +702,39 @@ class PipelineSampleReport extends React.Component {
...
+      this.state.contigTaxidList.indexOf(taxInfo.tax_id) >= 0;
+    const alignmentVizEnabled =
+      this.canSeeAlignViz && validTaxId && taxInfo.NT.r > 0;
+    const phyloTreeEnabled =
       this.allowPhyloTree &&
       taxInfo.tax_id > 0 &&
-      PhyloTreeChecks.passesCreateCondition(taxInfo.NT.r, taxInfo.NR.r)
-    )
-      phyloTreeDot = (
-        <PhyloTreeCreationModal
```
tfrcarvalho: Was this creating the Modal for every row?

app/assets/src/components/ui/containers/Modal.jsx
```diff
@@ -10,8 +10,11 @@ class Modal extends React.Component {
   }
 
   render() {
+    // The Semantic UI Modal component always expects a 'trigger', even though it always doesn't make sense for us.
```
tfrcarvalho: "always doesn't" => "never"
or did you meant "does not always"?

app/assets/src/components/ui/containers/Modal.jsx
```diff
@@ -31,7 +34,8 @@ Modal.propTypes = {
   ]).isRequired,
   onClose: PropTypes.func,
   title: PropTypes.string,
-  trigger: PropTypes.node.isRequired
+  trigger: PropTypes.node,
```
tfrcarvalho: If trigger makes no sense... should we just pass a dummy value and remove the prop from our component?

app/assets/src/components/PipelineSampleReport.jsx
MarkAZhang: Hmmm I guess so!

app/assets/src/components/ui/containers/Modal.jsx
MarkAZhang: I meant "doesn't always". Will fix.

app/assets/src/components/ui/containers/Modal.jsx
MarkAZhang: That sounds good.

PR #1759, opened 2018-11-21
Remove taxon confirmation feature from Report Table.

MarkAZhang:
Also made last column narrower.

screen shot 2018-11-20 at 5 32 39 pm


jshoe:
You could git reset schema.rb from master or db rollback and try generating a migration again. Or that could be a separate PR

MarkAZhang:
Okay I'll try to get it working. If I can't, then I'll remove the migration from this PR.

cdebourcy:
> Okay I'll try to get it working. If I can't, then I'll remove the migration from this PR.

It's possible staging db actually has some relics in it. Since you pulled staging db into your development, you would have those too.

I think dropping the taxon_confirmations table is not too urgent

MarkAZhang:
Okay, I've removed the db migration.

PR #1760, opened 2018-11-21
fine-grain autoscaler

cdebourcy:
- Calculate number of gsnap/rapsearch instances needed from number of non-host reads in the sample.
- Track number of chunks that have already completed so they can be subtracted.
- Refactor to define chunk size and concurrency parameters centrally and pass them to both the DAG and the autoscaler from there.


app/jobs/autoscaling.py
```diff
@@ -208,15 +188,24 @@ class ASG(object):
     job_dispatch_lag_grace_period_seconds = 60
     job_tag_keep_alive_grace_period_seconds = 60
 
-    def __init__(self, service, environment, asg_list, tag_list, draining_tag, job_tag_prefix, max_job_dispatch_lag_seconds, job_tag_keep_alive_seconds):
+    numa_partitions = 2 # i3.16xlarge or i3.metal
+    desired_queue_depth = 3 # initialization takes time: it's fine if some number of chunks have to run sequentially rather than in parallel
+    instance_limit = 24 # this autoscaler will never go beyond this number unless operator intervenes
```
boris-dimitrov: Isn't this what the "max_capacity" setting on the ASG is for?

app/jobs/autoscaling.py
```diff
@@ -73,21 +65,9 @@ def get_asg_list():
...
-    elif 1 < num_jobs <= 6:
-        result = exactly(8)
-    else:
-        # This means, as long as we're in this regime, a DesiredCapacity > 24 set manually by the operator
-        # will be respected and not scaled down until we exit this regime.
-        result = at_least(24)
-    return result
-
-def autoscaling_update(my_num_jobs, my_environment,
+def autoscaling_update(num_gsnap_chunks, num_rapsearch_chunks,
```
boris-dimitrov: Have you considered passing autoscaling_config through to the ASG constructor?   Then you wouldn't have to pass a list of parameters derived from it.   Long parameter lists are error prone.

app/jobs/autoscaling.py
```diff
@@ -269,6 +258,22 @@ def tags_by_instance_id(self):
             result[instance_id][key] = value
         return result
 
+    def required_capacity(self, current_desired_capacity):
+        best_concurrency = max(self.numa_partitions, self.max_concurrent - self.numa_partitions)
+        best_num_instances = int(math.ceil(float(self.num_chunks) / (self.desired_concurrency * self.desired_queue_depth)))
```
boris-dimitrov: Perfection.

app/jobs/autoscaling.py
```diff
@@ -406,11 +411,12 @@ def remove_scalein_protection(instance_ids):
...
 
 if __name__ == "__main__":
-    assert len(sys.argv) > 2
-    assert sys.argv[1] in ("update", "debug")
-    assert sys.argv[2].isdigit()
-    assert sys.argv[4].isdigit()
-    assert sys.argv[6].isdigit()
-    if sys.argv[1] == "debug":
+    _, autoscaling_config_json = sys.argv
+    autoscaling_config = json.loads(autoscaling_config_json)
```
boris-dimitrov: This is actually pretty cool;   I didn't think you could put an entire json dict on the command line ;)

app/models/pipeline_run.rb
```diff
@@ -186,6 +194,65 @@ def self.in_progress_at_stage_1_or_2
     in_progress.where("job_status NOT LIKE '3.%' AND job_status NOT LIKE '4.%'")
   end
 
+  def self.count_chunks(run_ids, known_num_reads, count_config, completed_chunks)
+    chunk_size = count_config[:chunk_size]
+    can_pair_chunks = count_config[:can_pair_chunks]
```
boris-dimitrov: subtle

boris-dimitrov: i would expect it to be the same as is_run_paired though...

app/models/pipeline_run.rb
```diff
@@ -726,6 +794,17 @@ def check_and_log_long_run
     end
   end
 
+  def load_chunk_stats
```
boris-dimitrov: This is (1) rather costly and (2) only a perf/cost optimization.   It may benefit from being run not more often than every 10 minutes.  Particularly if there are tons of jobs.  Perhaps it would be worthwhile doing a performance experiment with lots of jobs where the actual gsnap and rapsearch ssh is stubbed out to just "sleep 10 minutes and copy known outputs to result folder" which would give you a chance to measure the impact of these result pulls on job scheduling and other latencies.

app/jobs/autoscaling.py
cdebourcy: true

app/jobs/autoscaling.py
```diff
@@ -269,6 +258,22 @@ def tags_by_instance_id(self):
             result[instance_id][key] = value
         return result
 
+    def required_capacity(self, current_desired_capacity):
+        best_concurrency = max(self.numa_partitions, self.max_concurrent - self.numa_partitions)
+        best_num_instances = int(math.ceil(float(self.num_chunks) / (self.desired_concurrency * self.desired_queue_depth)))
+        if DEBUG:
+            msg = "num_chunks / (desired_concurrency * desired_queue_depth) = {num_chunks} / ({desired_concurrency} * {desired_queue_depth}) = {best_num_instances}"
+            print msg.format(num_chunks=self.num_chunks, desired_concurrency=self.desired_concurrency, desired_queue_depth=self.desired_queue_depth, best_num_instances=best_num_instances)
+        if best_num_instances < self.instance_limit:
```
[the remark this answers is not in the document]
cdebourcy: ok

app/jobs/autoscaling.py
```diff
@@ -208,15 +188,24 @@ class ASG(object):
...
         self.service = service
+        self.num_chunks = num_chunks
+        self.max_concurrent = max_concurrent
         self.environment = environment
         self.draining_tag = draining_tag
         self.job_tag_prefix = job_tag_prefix
         self.max_job_dispatch_lag_seconds = max_job_dispatch_lag_seconds
         self.job_tag_keep_alive_seconds = job_tag_keep_alive_seconds
 
+        self.desired_concurrency = max(self.numa_partitions, self.max_concurrent - self.numa_partitions)
```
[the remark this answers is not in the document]
cdebourcy: removed

app/jobs/autoscaling.py
cdebourcy: ok done

app/jobs/autoscaling.py
```diff
@@ -208,19 +192,28 @@ class ASG(object):
...
+        self.job_tag_prefix = config['job_tag_prefix']
+        self.max_job_dispatch_lag_seconds = config['max_job_dispatch_lag_seconds']
+        self.job_tag_keep_alive_seconds = config['job_tag_keep_alive_seconds']
 
         self.min_draining_wait = self.max_job_dispatch_lag_seconds + self.job_dispatch_lag_grace_period_seconds
         self.job_tag_expiration = self.job_tag_keep_alive_seconds + self.job_tag_keep_alive_grace_period_seconds
         self.instance_name = self.service + "-asg-" + self.environment
 
+        # Machine initialization takes time: it's fine if some number of chunks have to run sequentially rather than in parallel
+        self.desired_queue_depth = 3 if self.service == "rapsearch2" else 5
```
boris-dimitrov: Cool.   Though some day, move to config DB, like store as a tag.

lib/tasks/pipeline_monitor.rake
```diff
@@ -53,30 +53,39 @@ class CheckPipelineRuns
...
-    last_job_count = autoscaling_state[:job_count]
-    runs = PipelineRun.in_progress_at_stage_1_or_2
-    runs = runs.where("id > 10") if Rails.env == "development"
-    new_job_count = runs.count
-    if last_job_count.nil?
-      Rails.logger.info("Autoscaling update to #{new_job_count}.")
+    last_chunk_counts = autoscaling_state[:chunk_counts]
+    new_chunk_counts = PipelineRun.count_alignment_chunks_in_progress
+    if last_chunk_counts.nil?
+      Rails.logger.info("Autoscaling update to #{new_chunk_counts}.")
```
boris-dimitrov: I would add " chunks" to the print message both here and the line below.   :P)

cdebourcy: Yes!

PR #1774, opened 2018-11-27
Add Institution field to Users and admin pages

jshoe:
- Adds the institution string field to users and on the admin-pages (same pattern as on those pages). Nothing user-facing for now.

screen shot 2018-11-27 at 2 28 24 pm
screen shot 2018-11-27 at 2 28 06 pm

### DIFF
--- a/app/assets/src/components/CreateUser.jsx
+++ b/app/assets/src/components/CreateUser.jsx
@@ -18,6 +18,7 @@ class CreateUser extends React.Component {
     this.selectedUser = {
       email: this.user ? this.user.email : "",
       name: this.user ? this.user.name : "",
+      institution: this.user ? this.user.institution : "",
       password: "",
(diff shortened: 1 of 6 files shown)


app/assets/src/components/CreateUser.jsx
```diff
@@ -280,6 +283,20 @@ class CreateUser extends React.Component {
                 />
                 <label htmlFor="user_name">Name</label>
               </div>
+              <div className="input-field">
+                <i className="fa fa-building" aria-hidden="true" />
+                <input
+                  type="text"
+                  onChange={e =>
```
MarkAZhang: Nit: For consistency, can you make a `handleInstitutionChange` method?

jshoe: ty

PR #1775, opened 2018-11-28
Use new metadata in Phylo Tree tooltips.

MarkAZhang:
Fix word wrap issues in various locations.

When metadata updates in Sidebar, ensure Phylo Tree receives update.

------

screen shot 2018-11-27 at 7 49 25 pm

We show only a subset of the metadata since you can click on the sample to get all the metadata in the Sidebar.

We still show a tooltip because when you hover an NCBI sample it shows a tooltip, and we want to be consistent.


app/controllers/phylo_trees_controller.rb
```diff
@@ -54,8 +55,9 @@ def index
 
     # Augment tree data with sample attributes, number of pipeline_runs and user name
     @phylo_trees = @phylo_trees.as_json
+    sample_details_by_tree_id = sample_details_by_tree_id()
```
jshoe: maybe rename the var sample_details_by_tree_id? usually the function call is implicit so using it in the block could be confusing later

PR #1779, opened 2018-11-29
Add lineage_version to AlignmentConfig

jshoe:
- Next: refactoring references to TaxonLineage.started_at/ended_at..

### DIFF
--- a/db/migrate/20181128174938_add_lineage_version_to_alignment_config.rb
+++ b/db/migrate/20181128174938_add_lineage_version_to_alignment_config.rb
@@ -1,0 +1,20 @@
+class AddLineageVersionToAlignmentConfig < ActiveRecord::Migration[5.1]
+  def up
+    # Max 32767
+    add_column :alignment_config, :lineage_version, :integer, limit: 2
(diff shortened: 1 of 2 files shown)


db/migrate/20181128174938_add_lineage_version_to_alignment_config.rb
```diff
@@ -0,0 +1,21 @@
+class AddLineageVersionToAlignmentConfig < ActiveRecord::Migration[5.1]
+  def up
+    # Max 32767
+    add_column :alignment_configs, :lineage_version, :integer, limit: 2
+    AlignmentConfig.where(name: "2018-02-15").update(lineage_version: 1)
+    AlignmentConfig.where(name: "2018-04-01").update(lineage_version: 2)
+
+    # It will go from alignment lineage_version to "lineage records with a [version_start, version_end] range that includes that number."
```
MarkAZhang: Can you add a comment like:
"We use the lineage_version to figure out which lineage records to use in our pipeline run. Lineage records are highly duplicated between versions, so instead of storing all of them, we store a [version_start, version_end] range with each lineage record."

Mentioning that the lineage records are duplicated is really helpful for understanding the purpose of this I think.

PR #1781, opened 2018-11-29
[Lineage] Refactor references to use lineage_version

jshoe:
- Tested locally on a couple of samples
- Replace the logic of "date between started_at and ended_at" to be "alignment_config.lineage_version between version_start and version_end" in the lineage calls
- taxon_search_list is meant to be called manually / should be updated later. Or we can add column defaults to version_start/version_end as well.

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -962,6 +962,7 @@ class PipelineRun < ApplicationRecord
     # What is crucially important is the uncategorizable_id.
     uncategorizable_id = TaxonLineage::MISSING_LINEAGE_ID.fetch(tax_level_name.to_sym, -9999)
     uncategorizable_name = "Uncategorizable as a #{tax_level_name}"
+    lineage_version = alignment_config.lineage_version
     TaxonCount.connection.execute(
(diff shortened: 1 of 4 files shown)


cdebourcy:
It would be great if you could add some performance benchmarks, so we can evaluate and be aware of the improvement!

jshoe:
@cdebourcy 

# Benchmarks results
- Ran some speed tests locally. 
- Benchmark: Load report_info for all the samples in the range
- Results: Overall about 20-30% speedup with this switch

# Samples 900-950 (lots of skipped ones here)
OLD: 13.2179527 sec
NEW: 7.9781604 sec

# Samples 600-650
OLD: 76.796431 sec
NEW: 62.6806688 sec

# Samples 500-550
OLD: 48.5035036 sec
NEW: 34.7451729 sec

- Mixed up the orders of the tests to avoid caching issues
- Comparison code: https://github.com/chanzuckerberg/idseq-web/compare/jsheu/lineage-bench
- 
```
start_time = Time.now
(500..550).each do |i|
  puts "LOADING: " + i.to_s
  SamplesController.new.report_info(i)
end
end_time = Time.now
duration = end_time - start_time
puts "DONE"
puts "Time elapsed: #{duration}"
```

PR #1783, opened 2018-11-30
make ReadsPerGene a target in the dag

cdebourcy:
goes with https://github.com/chanzuckerberg/idseq-dag/pull/104/files


yunfangjuan:
you will need to change the dag code for this and you will need to test it carefully. i thiink changnig the web code seems to be easier?

PR #1787, opened 2018-11-30
Clean up PipelineSampleReads file.

MarkAZhang:
Rename to SamplePage.
Break up file into smaller components.
Remove materialize styling and jquery usage.
Directly use props instead of assigning to "this".
Modified our <Tabs> to make it more robust.
Add PropType annotations.
Remove unnecessary code.

-----

Main visual difference is the Tabs have changed, and since we removed the materialize styling, the horizontal alignment is slightly different.

screen shot 2018-11-29 at 7 46 01 pm


app/assets/src/api/index.js
```diff
@@ -46,11 +60,15 @@ const getAlignmentData = (sampleId, alignmentQuery, pipelineVersion) =>
...
 export {
   getSampleMetadata,
   saveSampleMetadata,
   getMetadataTypes,
   saveSampleName,
   saveSampleNotes,
-  getAlignmentData
+  getAlignmentData,
+  getURLParamString,
+  deleteSample
```
tfrcarvalho: Should we start splitting some of these (metadata & sample)?

MarkAZhang: The metadata is sample metadata, so not sure if that distinction makes sense. I think we can wait a little longer, but should definitely split this eventually.

app/assets/src/components/views/sample_page/SamplePage.jsx
```diff
@@ -0,0 +1,328 @@
...
+import ViewHeader from "~/components/layout/ViewHeader";
+import NarrowContainer from "~/components/layout/NarrowContainer";
+import Tabs from "~/components/ui/controls/Tabs";
+import SampleDetailsSidebar from "~/components/views/report/SampleDetailsSidebar";
+import SampleControls from "./SampleControls";
+import PipelineVersionSelect from "./PipelineVersionSelect";
+
+import cs from "./sample_view.scss";
+
+class SampleView extends React.Component {
```
tfrcarvalho: nit: Should it all be called SampleView in `views/SampleView` as a folder? A bit confused about the naming...

app/assets/src/components/views/sample_page/SamplePage.jsx
```diff
@@ -0,0 +1,328 @@
...
+    }
+    for (let stat of jobStats) {
+      if (stat["task"] === "gsnap_filter_out") {
+        return null;
+      }
+    }
+    return "gsnap filter on human/chimp genome was not run.";
+  };
+
+  refreshPage = overrides => {
```
tfrcarvalho: `overrides` -> `overrideUrlParams`?

app/assets/src/components/views/sample_page/SamplePage.jsx
```diff
@@ -0,0 +1,328 @@
...
+          trigger={
+            <i className={cx("fa fa-exclamation-circle", cs.warningIcon)} />
+          }
+          position="bottom left"
+          content={content}
+          wide="very"
+          horizontalOffset={15}
+        />
+      );
+    } else {
```
tfrcarvalho: Do need the else?

app/assets/src/components/views/sample_page/SamplePage.jsx
```diff
@@ -0,0 +1,328 @@
...
+      return null;
+    }
+  };
+
+  renderTab = () => {
+    if (this.state.currentTab === "Report") {
+      const waitingSpinner = (
+        <div className={cs.waitingContainer}>
```
tfrcarvalho: Should this be site-wide element?
I thought there was already one... but cannot find it, so wemight never have done it.

app/views/samples/show.html.erb
```diff
@@ -1,30 +1,28 @@
-<div id="PipelineSampleReads_page">
+<div id="sample_view">
   <%= javascript_tag do %>
-    react_component('PipelineSampleReads', {
+    react_component('SampleView', {
       admin: <%= current_user.admin %>,
       allowedFeatures: JSON.parse('<%= raw escape_json(current_user.allowed_feature_list) %>'),
```
tfrcarvalho: Just a question have we set on camelCase vs snake_case for the JSONs? 
Or did you just tried to make this uniform.

app/assets/src/components/views/sample_page/index.js
```diff
@@ -0,0 +1,3 @@
+import SamplePage from "./SamplePage";
+
+export default SamplePage;
```
tfrcarvalho: I guess the question should have been here.
I have been calling this "Views", i.e. SampleView.

I would also like to understand: when do you use `samples_page` folder vs `SamplePage`?

app/assets/src/api/index.js
tfrcarvalho: The logic sounds good. I would still put all in `api/sample.js` and leave the generic methods here.
Can be in a future follow up.

app/assets/src/components/PipelineSampleReads.jsx
```diff
@@ -1,447 +0,0 @@
-import React from "react";
```
tfrcarvalho: 👍

app/views/samples/show.html.erb
MarkAZhang: I tried to make it uniform. We haven't formally decided on it. Although...since these keys are passed into the components as props, and we generally want props to be camelCase, I think this is a reasonable convention. Let me know what you think/

app/assets/src/components/views/sample_page/index.js
MarkAZhang: I think of "views" as things that can be reused in different pages, while "pages" are things that are only used on a particular page (roughly corresponds to an `.erb` file I guess). 

But I will do `SampleView` here for now. Since we have `PhyloTreeListView` which is very similar.

Regarding `samples_page` vs `SamplePage`, I think I made a mistake here. I think it should be `SamplePage` here, since it is really one big "component" with several related sub-components, as opposed to several distinct components. But it is a bit of a gray area.

app/assets/src/components/views/sample_page/SamplePage.jsx
MarkAZhang: We don't have a <Spinner> yet. Would be good to do at some point!

app/assets/src/components/views/sample_page/index.js
tfrcarvalho: Yeah, we just need to sync on these concepts. `Views` were intended to be what you are calling `pages`.

app/views/samples/show.html.erb
tfrcarvalho: We cannot meet both sides. Ruby use snake_case and Javascript camelCase.
We just have to define a convention to follow in the future.
I would prefer camelCase just because it is easy to convert camelCase->snake_case in rails if we want to using [deep_transform_keys](https://apidock.com/rails/v4.0.2/Hash/deep_transform_keys) and [underscore](https://apidock.com/rails/ActiveSupport/Inflector/underscore).

PR #1793, opened 2018-12-01
Button that syncs samples in staging to production data in mysql and s3

jameshowardwang:
Only works in staging. Assumes a local parallel database "idseq_prod" exists that contains data to sync from (will be documented in process for taking a snapshot of prod for staging).


app/views/projects/show.html.erb
```diff
@@ -11,7 +11,71 @@
...
+        <tr class=<%= sample_status_display_for_hidden_page(sample)%>>
+          <td><%= link_to sample.name, sample %></td>
+          <td><%= sample.id %></td>
+          <td class='small'><%= sample.user.email if sample.user %></td>
+          <td><%= link_to sample.project.name, sample.project %></td>
+          <td><%= sample.status %>
+            <% if sample.status.to_s == 'created' and sample.created_at < 30.minutes.ago %>
+                <%= button_to 'reupload', reupload_source_sample_path(sample), method: :put %>
+            <% end %>
+            <% if Rails.env == 'staging' %>
```
jameshowardwang: this is the only new line. the rest was just moved out of app/views/samples/_samples.html.erb

PR #1797, opened 2018-12-05
Heatmap Metadata

tfrcarvalho:
* Adds metadata to heatmap columns;
* New component to select and search for metadata fields (selected fields are shown on top when component is first opened);
* Metadata selection component triggered on click no metadata control icon below the metadata rows;
* User can add multiple metadata rows;
* Metadata legend per row displayed when hovering over metadata labels.
dec-05-2018 10-31-28


tfrcarvalho:
@jshoe 
* **on the icon**: The icon is still being discussed (might be a cog). But agree that is a bit small. Waiting on @happyimadesignr feedback.
* **on the onHover**: not sure what you mean. Do you mean onHover for each metadata field?

jshoe:
@tfrcarvalho yeah, for each metadata field? like the background shading 
dec-05-2018 13-33-16

tfrcarvalho:
Oh yeah, i see. Makes sense

webpack.config.common.js
```diff
@@ -13,8 +13,11 @@ const config = {
   resolve: {
     extensions: [".js", ".jsx"],
     alias: {
-      styles: path.resolve(__dirname, "app/assets/src/styles"),
-      "~": path.resolve(__dirname, "app/assets/src")
+      "~": path.resolve(__dirname, "app/assets/src"),
+      api: path.resolve(__dirname, "app/assets/src/api"),
```
MarkAZhang: How do you feel about using `~ui` as the alias instead of `ui`? Makes it more clear that it's an alias.

`styles` doesn't have the `~` because it's primarily used in `scss`, which automatically appends the `~`.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -8,6 +8,18 @@ import SvgSaver from "svgsaver";
 import symlog from "../../utils/d3/scales/symlog.js";
 import cs from "./heatmap.scss";
 import cx from "classnames";
+import { CategoricalColormap } from "../../utils/colormaps/CategoricalColormap.js";
```
MarkAZhang: Would it be a lot of work to break this into small files? Similarly to how we talked about with React components. It'd be great to avoid these large files.

app/assets/src/components/views/report/SampleDetailsSidebar/SampleDetailsSidebar.jsx
```diff
@@ -20,7 +20,7 @@ import {
   processMetadataTypes,
   processPipelineInfo,
   processAdditionalInfo
-} from "./utils";
+} from "../../../utils/metadata";
```
MarkAZhang: Use an alias: `~/components/utils/metadata`

webpack.config.common.js
tfrcarvalho: I can do that. Might also avoid some conflicts later on.

app/assets/src/components/visualizations/heatmap/Heatmap.js
MarkAZhang: (Can do this as follow-up)

tfrcarvalho: Yes, I plan to do that eventually. But would prefer not to do it on this round. Needs bit of thought.

app/assets/src/components/ui/containers/ContextPlaceholder.jsx
```diff
@@ -0,0 +1,140 @@
...
+  getStyle() {
+    let style = { position: "absolute" };
+    if (this.props.context && this.placeholderRef) {
+      Object.assign(style, this.getPosition());
+    }
+    return style;
+  }
+
+  getPosition() {
+    const contextRect = this.props.context.getBoundingClientRect();
```
MarkAZhang: Could you move this logic to a pure utility function in a separate file?

webpack.config.common.js
MarkAZhang: Sounds good. Although...how are there other files if you just added the aliases?

tfrcarvalho: Fair point, i added this a long time ago... but i guess it was never included in a PR. So, I can update it.

app/assets/src/components/ui/controls/SearchBoxList.jsx
```diff
@@ -0,0 +1,103 @@
...
+    }
+    this.setState({ selected }, () => {
+      return this.props.onChange && this.props.onChange(this.state.selected);
+    });
+  };
+
+  handleFilterChange = filter => this.setState({ filter });
+
+  sortOptions() {
+    // TODO(tcarvalho): review data structures to simplify this function
```
MarkAZhang: You might be able to use `partition` from `lodash/fp`.

app/assets/src/components/ui/controls/SearchBoxList.jsx
```diff
@@ -0,0 +1,103 @@
...
+            className={cs.searchBox}
+            icon="search"
+            placeholder="Search"
+            onChange={this.handleFilterChange}
+          />
+        </div>
+        <div className={cs.listBox}>
+          {this.sortedOptions.map(
+            option =>
+              option.label.toLowerCase().includes(this.state.filter) && (
```
MarkAZhang: Was a little surprised to see the filtering happening here, but it does work.

If we were worried about performance issues, we might want to store the filteredOptions in a state variable instead (and compute them in the `handleFilterChange function).

app/assets/src/components/ui/controls/SearchBoxList.jsx
tfrcarvalho: Thanks for the tip. Will make it simpler.
Although the comment was more about the whole logic seeming weird.

app/assets/src/components/views/report/SampleDetailsSidebar/SampleDetailsSidebar.jsx
MarkAZhang: I'm not sure if all of these methods should be moved into `components/utils`. Originally, they were in `SampleDetailsSidebar/utils`. `processPipelineInfo` outputs stuff in a display format specific to `SampleDetailsSidebar`. I would just move `processMetadata` into `components/utils`, as that function seems a little more generic.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -57,18 +70,24 @@ class SamplesHeatmapVis extends React.Component {
...
+      };
+    });
   }
 
   extractTaxonLabels() {
-    return this.props.taxonIds.map(id => this.props.taxonDetails[id].name);
+    return this.props.taxonIds.map(id => {
+      return {
+        label: this.props.taxonDetails[id].name,
+        metadata: this.props.taxonDetails[id].metadata
```
MarkAZhang: Only `sampleIds` have metadata, right? Do you need this for `taxonIds`?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -754,4 +1045,14 @@ export default class Heatmap {
...
       });
   }
+
+  getAddMetadataTriggerRef() {
+    return this.addMetadataTrigger;
+  }
+
+  getColumnMetadataLegend(value) {
+    return Object.assign({}, this.metadataColors[value], {
+      Unkown: this.options.colorNoValue
```
MarkAZhang: s/Unkown/Unknown/

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -147,13 +255,49 @@ class SamplesHeatmapVis extends React.Component {
...
+              top: `${this.state.tooltipY + 20}px`
+            }}
+          >
+            {this.renderColumnMetadataLegend(
+              this.state.columnMetadataHoverNode
+            )}
+          </div>
+        )}
+        {this.state.addMetadataTrigger && (
+          <ContextPlaceholder
```
MarkAZhang: Could we make the styling of the `<ContextPlaceholder>` and columnMetadataLegend consistent? Specifically, the box-shadow and the border-radius? Right now, the `<ContextPlaceholder>` has no box-shadow, and no border-radius.

app/assets/src/components/ui/controls/search_box_list.scss
```diff
@@ -0,0 +1,54 @@
...
+      margin-right: 10px;
+      white-space: nowrap;
+    }
+
+    .listCheckmark {
+      display: inline-block;
+      width: 20px;
+    }
+
+    .listLabel {
```
MarkAZhang: I think it would be good to make sure the styling of the options here is consistent with the Dropdown items. Things like `letter-spacing: 0.4px` and proper top and bottom padding between items. So that all of our dropdowns look consistent.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -127,6 +158,82 @@ class SamplesHeatmapVis extends React.Component {
     openUrl(`/samples/${this.props.sampleIds[cell.columnIndex]}`, currentEvent);
   };
 
+  handleUpdateFinished = () => {
```
MarkAZhang: Is `handleUpdateFinished` ever used?

app/assets/src/components/ui/containers/ContextPlaceholder.jsx
```diff
@@ -0,0 +1,140 @@
+import React from "react";
```
MarkAZhang: The naming of this is a bit confusing. Why is it called a `ContextPlaceholder`? Sorry I think we talked about this but I can't remember. Can you add a comment at the top explaining why this is different from `Popup` (and why you needed to re-implement it)?

app/assets/src/components/ui/containers/ContextPlaceholder.jsx
MarkAZhang: The positioning logic here is quite complicated, and I'm not sure it will work for all cases (this stuff tends to be finicky). You might need to use a `<Portal>` to move `<ContextPlaceholder>` elements higher on the DOM tree so it's not the child of any elements that are `position: absolute` or `position: relative`. 

No change needed, but just a note that I think we should try to avoid writing this kind of complicated logic without a good reason. Awesome that it works though!

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -147,13 +255,49 @@ class SamplesHeatmapVis extends React.Component {
             <DataTooltip data={this.state.nodeHoverInfo} />
           </div>
         )}
+        {this.state.columnMetadataHoverNode && (
```
MarkAZhang: The D3/React communication here was handled really well!

app/assets/src/components/ui/containers/ContextPlaceholder.jsx
tfrcarvalho: Added a comment, but my thoughts on writing my own:
1. a popup is typically a very simple component with some text (that why the naming). Could use a better name than placeholder.
2. do now want a lot of the popup default styles (this is a invisible component for layout) and some positioning options were weird.
3. if we can implement simpler controls ourselves better (some semantic components are not that straightforward to use).
4. if a popup works on other use cases feel free to use it.

app/assets/src/components/ui/containers/ContextPlaceholder.jsx
tfrcarvalho: this logic gets element position but also sets the context. I do not think that is that generic.

MarkAZhang: For my first comment (moving into pure utility function), I meant that since `getPosition` is taking in parameters (such as contextRect and placeholderRect) and outputting a value (the position), it would be easy to factor it out into a separate file like `ContextPlaceholder/utils.js` so that all the business logic isn't cluttering up the file. This file is pretty short though, so not a big deal.

app/assets/src/components/ui/controls/search_box_list.scss
tfrcarvalho: You got a point, but actually i think the problem is larger than that.
For instance, `letter-spacing` is something that should be consistent everywhere, so probably should not be set at the dropdown level.
I can change padding, but I also think this is different component... not a dropdown. I think eventually, if we decided to go without the checkboxes... we might want to reduce the padding.
Anyway is a topic to sync with Jenn, and maybe for a larger change

app/assets/src/components/ui/containers/ContextPlaceholder.jsx
MarkAZhang: I see...While a `<Popup>` usually just contains text, we do use it in components like `<ThresholdFilterDropdown>` for more complex popups so I think it is capable of handling that.

Let's keep ContextPlaceholder, but please do include in your comment an example of a use case where it's easier to use than Popup (for example, a simpler use case where you don't care about X) so the team can understand when it's more appropriate. Alternatively, if it's very specific to SamplesHeatmap, maybe move the component out of `/ui` and into that folder. Thanks!

MarkAZhang: Edit: I see your point about using as an invisible component for layout only, as opposed to also providing styling. Didn't catch that earlier.

app/assets/src/components/ui/containers/ContextPlaceholder.jsx
tfrcarvalho: It takes at least three more parameters: `horizontalOffset`, `verticalOffset`, `position`, which makes it a bit more specific to this element. 
I also not sure if you ever should use it outside of this element, or use this element instead.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
tfrcarvalho: We will probably use taxonIds metadata too.... but not yet. This is when I was trying to make the code generic for that. But will remove.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
tfrcarvalho: Nope. Good catch!

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
tfrcarvalho: You meant the search box...
i guess for this we have to decide what kind of element it is: dropdown? tooltip? :)
I will sync with Jenn tomorrow on how she sees this component.

app/assets/src/components/views/report/SampleDetailsSidebar/SampleDetailsSidebar.jsx
tfrcarvalho: Humm... not sure I agree.
It seems like you do a lot of useful stuff there. Wherever you want to display/use this data, it will be useful.
Actually, I think this pipeline info can also eventually be useful for heatmap (compare by date, .
For instance we might want to compare samples of different pipelines.

But will refactor per need for now.

app/assets/src/components/ui/controls/SearchBoxList.jsx
tfrcarvalho: Actually no. Cannot use partition in this case, because I'm not just partitioning the array.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
MarkAZhang: Ah yes, since the ContextPlaceholder is invisible.

Both of these elements "overlap" another element, so I think the box-shadow is appropriate. And it looks like most of our elements do have a small border-radius, which seems to tbe missing on the search box. But feel free to talk to Jenn.

app/assets/src/components/ui/controls/search_box_list.scss
MarkAZhang: Yeah, talk to Jenn here too.

app/assets/src/components/visualizations/heatmap/Heatmap.js
tfrcarvalho: done.

app/assets/src/components/views/report/SampleDetailsSidebar/SampleDetailsSidebar.jsx
```diff
@@ -15,12 +15,8 @@ import {
...
 import PipelineTab from "./PipelineTab";
 import NotesTab from "./NotesTab";
-import {
-  processMetadata,
-  processMetadataTypes,
-  processPipelineInfo,
-  processAdditionalInfo
-} from "./utils";
+import { processMetadata, processMetadataTypes } from "~utils/metadata";
+import { processPipelineInfo, processAdditionalInfo } from "./info_utils";
```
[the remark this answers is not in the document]
tfrcarvalho: As we discussed, should use "non-capitalized camelCase" for files with functional utility functions or constants and "capitalized CamelCase" for files that export a class, e.g. "Heatmap"

PR #1798, opened 2018-12-05
count reads per month

cdebourcy:


app/helpers/dev_helper.rb
```diff
@@ -0,0 +1,33 @@
+module DevHelper
+  def _alignment_cost_per_million_reads(month_starting, ec2_cost)
+    # Computes the average cost per million reads of gsnap/rapsearch alignment servers over the course of 1 month.
```
boris-dimitrov: This is really cool.

Just an aside, for future perf work, it would be useful to expressly track CPU time and elapsed time for each gsnap and rapsearch chunk, and store that data in an analytics DB.   Running the gsnap/rapsearch command through "time" would provide that info.

PR #1801, opened 2018-12-06
Add support for mosquito and vector metadata.

MarkAZhang:
Metadata types now dependent on host genome.
Validation of metadata works properly.
Sidebar updates properly.
Also added missing human metadata fields (immunocomp, discharge_date)

screen shot 2018-12-06 at 1 25 10 am

### DIFF
--- a/app/assets/src/components/views/report/SampleDetailsSidebar/constants.js
+++ b/app/assets/src/components/views/report/SampleDetailsSidebar/constants.js
@@ -1,4 +1,4 @@
-export const METADATA_SECTIONS = [
+export const HUMAN_METADATA_SECTIONS = [
   {
     name: "Sample Info",
     keys: [
@@ -19,7 +19,10 @@ export const METADATA_SECTIONS = [
       "primary_diagnosis",
       "antibiotic_administered",
       "admission_date",
-      "admission_type"
+      "admission_type",
+      "discharge_date",
+      "discharge_type",
+      "immunocomp"
     ]
   },
   {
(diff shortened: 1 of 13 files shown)


app/assets/src/components/views/report/SampleDetailsSidebar/constants.js
```diff
@@ -19,7 +19,10 @@ export const METADATA_SECTIONS = [
       "primary_diagnosis",
       "antibiotic_administered",
       "admission_date",
-      "admission_type"
+      "admission_type",
+      "discharge_data",
```
cdebourcy: This is cool! 'discharge_datE' here?

app/assets/src/components/views/report/SampleDetailsSidebar/constants.js
```diff
@@ -44,6 +47,40 @@ export const METADATA_SECTIONS = [
...
 
+// Includes mosquito and other vectors.
+export const VECTOR_METADATA_SECTIONS = [
+  {
+    name: "Sample Info",
+    keys: [
+      "unique_id",
+      "sample_type",
+      "nucleotide_type",
+      "sample_unit",
```
tfrcarvalho: What is the "sample unit"? Asking just because it is not part of Sample Info in the human schema. 
I am wondering how this is going to evolve, and if we should try to keep these sections similar  and common. Then you would just enable sections to a human or vector. It would be easier to scale.
Just flagging.. It is not that relevant yet, but if/when we add more we should have the discussion.

tfrcarvalho: I am not fond of having the metadata to show defined on both frontend and backend, needing to have to update frontend everytime we add a new field. I fear this might change quite a lot.

app/assets/src/components/views/report/SampleDetailsSidebar/constants.js
MarkAZhang: Yes! Thank you.

app/assets/src/components/views/report/SampleDetailsSidebar/constants.js
MarkAZhang: That's a great point. I originally restrained from adding "sample info" and "donor info" to the back-end because they seemed like display-specific attributes (i.e. for front-end only), but now that I think about it, there's no reason not to.

I will re-think this in a follow-up commit.

MarkAZhang: Sample Unit seems to be whether the sample is comprised of a single organism or pooled from a bunch. Not 100% sure what it is. But it seemed like sample info.

PR #1805, opened 2018-12-07
Replace IconComponent with separate Icon files + Add Cat Genome

jshoe:
- SUCCESSFUL TEST RUN WITH CAT: https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logEventViewer:group=/aws/batch/job;stream=aegea_batch/default/a28d1a3a-85c4-42b4-bebd-539e89b975c8
- This PR replaces IconComponent with separate icon components (to get away from dangerouslySetInnerHTML). Copied and pasted the SVG stuff.
- Adds an icon for CatGenome. Icon picked by Design.
- Migration adds the cat HostGenome

screen shot 2018-12-07 at 12 58 43 pm


db/migrate/20181207202634_add_cat_genome.rb
```diff
@@ -0,0 +1,25 @@
+class AddCatGenome < ActiveRecord::Migration[5.1]
```
tfrcarvalho: Do you know what tables we use migrations to add new data to? 
No problem in starting without proper background?

PR #1806, opened 2018-12-08
Add metadata selector and color by metadata on phylo tree.

MarkAZhang:
I decided to add a search box to `<Dropdown>` instead of using `<SearchBoxList>`, because `<SearchBoxList>` currently only selects multiple options (so I would have had to change code either way). Also, having a search box on our Dropdowns will be very useful. 

Because of this, the styling is a little different from `<SearchBoxList>`.

-----

screen shot 2018-12-07 at 3 58 13 pm

screen shot 2018-12-07 at 3 58 30 pm

screen shot 2018-12-07 at 3 58 40 pm

screen shot 2018-12-07 at 3 58 46 pm


jshoe:
- Left margin missing? or is that intentional

screen shot 2018-12-10 at 10 56 29 am

- Minor thing that's hard to notice: It looks like "RELATIVE DISTANCE" gets redrawn (and builds up) when you switch around the Color By
screen shot 2018-12-10 at 10 57 45 am

- Would be nice to get "arrow keys" on the search box dropdown someday

- Basically ignoring the code-move changes in the diff

MarkAZhang:
screen shot 2018-12-20 at 6 56 09 pm

Update takes new dropdown styling into account.

Also added a search box and filtering functionality to the base BareDropdown component, so we can easily add a search box to any of our dropdowns.

Also fixed a bug in `metadatum.rb` that was causing all metadata values to be returned as `""`. The issue was that `string_validated_value`, `date-validated_value`, etc. were not instance variables of the model. Verified that the new solution worked.

app/models/metadatum.rb
```diff
@@ -352,7 +352,7 @@ def self.new_without_save(sample, key, value)
   end
 
   def validated_value
-    return instance_variable_get("#{data_type}_validated_value")
+    return self["#{data_type}_validated_value"]
```
jshoe: wait, was this change really necessary? isn't it basically doing the same thing (getting the value of the instance variable?)

MarkAZhang: (Posting this mainly for posterity)

I did some research. It's surprisingly confusing.

`instance_variable_get("@string_validated_value")` (you need the `@`, which was missing in the code) will get the instance variable `@string_validated_value`. Unfortunately, the database value for `string_validated_value` is not stored on `@string_validated_value`, but actually on `@attributes["string_validated_value"]` (Rails magic). And if you try to access `@attributes["string_validated_value"]` directly, you'll find that it's not actually the database value,  but an internal object of type `ActiveRecord::Attribute::FromDatabase`. I think Rails intends you to call the getter method `string_validated_value` it automatically generates, via self.string_validated_value, which gives you the correct value. 

NOTE:
Instance variables are always private by default. Within a class, instance variables can only be accessed by `@foo`. If you instead type `foo`, it will first try to resolve to a local variable, and then resolve to `self.foo`, which is the getter function for `foo`. This is why  `data_type` works within the class. It is actually resolving to `self.data_type` which is the Rails auto-generated getter method for the database attribute. `@data_type` would NOT work, because that's the instance variable, and the database value is not stored there.

PR #1811, opened 2018-12-11
Makes the bin/shell script compatible with bless

edulop91:
# Description
Removing the hardcoded `ec2-user` instead relies on the system ssh config. Also prefixing calls to docker with sudo so we don't assume membership to the docker group. This breaks systems with misconfigured ssh configs.

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix a feature that knowingly causes existing functionality to not work as expected)

# Test and Reproduce

1. ./shell <env>

### DIFF
--- a/bin/clam
+++ b/bin/clam
@@ -24,4 +24,4 @@ MACHINE=`bin/get_private_ip $ENV`
 
 CONTAINER=`ssh $MACHINE "sudo docker ps | grep ecs-idseq-$ENV-web | head -1" | awk '{print $NF}'`
 
-ssh ec2-user@$MACHINE 'sudo docker exec '$CONTAINER' bin/entrypoint.sh '"'$*'"
+ssh $MACHINE 'sudo docker exec '$CONTAINER' bin/entrypoint.sh '"'$*'"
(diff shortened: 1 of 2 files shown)


bin/clam
```diff
@@ -22,6 +22,6 @@ shift
 
 MACHINE=`bin/get_private_ip $ENV`
 
-CONTAINER=`ssh ec2-user@$MACHINE "docker ps | grep ecs-idseq-$ENV-web | head -1" | awk '{print $NF}'`
+CONTAINER=`ssh $MACHINE "sudo docker ps | grep ecs-idseq-$ENV-web | head -1" | awk '{print $NF}'`
 
-ssh ec2-user@$MACHINE 'docker exec '$CONTAINER' bin/entrypoint.sh '"'$*'"
+ssh ec2-user@$MACHINE 'sudo docker exec '$CONTAINER' bin/entrypoint.sh '"'$*'"
```
tfrcarvalho: missed one here.

edulop91: ty, fixed

PR #1812, opened 2018-12-11
Consolidate raw metadata values.

tfrcarvalho:
Consolidate raw metadata values: use raw_value field for both text number. Pre-change to adding a date-type metadata field.


app/models/metadatum.rb
```diff
@@ -174,6 +176,16 @@ class Metadatum < ApplicationRecord
     ]
   }.freeze
 
+  def raw_value_float_if_number
+    if data_type == "number"
+      begin
+        Float(raw_value)
```
jshoe: this should be self.number_validated_value = Float(raw_value), right?

PR #1813, opened 2018-12-12
Add Contig and Contig Reads columns to sample table.

MarkAZhang:
Also add min contig size filter.
Only show the new columns for Pipeline 3.1 and above.
Re-distribute the width of the columns to fit the new columns.
Enable sorting of the new columns.
Reload automatically when user sets a new min contig size.
Enable filtering by threshold for the new columns.

screen shot 2018-12-11 at 4 16 34 pm

### DIFF
--- a/app/assets/src/api/index.js
+++ b/app/assets/src/api/index.js
@@ -64,14 +64,22 @@ const getAlignmentData = (sampleId, alignmentQuery, pipelineVersion) =>
 
 const deleteSample = id => deleteWithCSRF(`/samples/${id}.json`);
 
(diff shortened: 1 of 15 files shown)


jshoe:
I was a little confused by the wording b/c I was still expecting it to act like the threshold filter and remove or gray out "contig" values below the number I put in.. but not sure how to explain the wording better..

MarkAZhang:
screen shot 2018-12-12 at 11 31 13 am

Tried to add an example, but I think it's too much text.

Will submit for now, but we can change the label later.

jshoe:
that looks good

PR #1816, opened 2018-12-14
download AMR results in bulk

cdebourcy:


lib/tasks/download_amr_for_project.rake
```diff
@@ -0,0 +1,13 @@
+# Generate a shall script to download SRST2 output from a project to a local destination.
```
jshoe: s/shall/shell

PR #1821, opened 2018-12-17
[Metadata Sidebar] Add pipelineVersion for version switcher

jshoe:
- Adds pipelineVersion prop
- Also a pipelineRun check so the Pipeline tab won't crash when a sample just starts

### DIFF
--- a/app/assets/src/components/views/SampleView/SampleView.jsx
+++ b/app/assets/src/components/views/SampleView/SampleView.jsx
@@ -286,6 +286,11 @@ class SampleView extends React.Component {
           visible={this.state.sampleDetailsSidebarVisible}
           onClose={this.toggleSampleDetailsSidebar}
           sampleId={sample.id}
+          pipelineVersion={
+            reportDetails &&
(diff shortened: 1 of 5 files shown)


app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -286,6 +286,11 @@ class SampleView extends React.Component {
           visible={this.state.sampleDetailsSidebarVisible}
           onClose={this.toggleSampleDetailsSidebar}
           sampleId={sample.id}
+          pipelineVersion={
+            reportDetails &&
```
MarkAZhang: You can also do `get('pipeline_info.pipeline_version', reportDetails)` and `import { get } from "lodash/fp"`. Slightly cleaner.

MarkAZhang: `get` will return undefined if any part of the chain fails.

jshoe: thanks that's nice

PR #1825, opened 2018-12-18
No downloads of homo sapiens contigs

jameshowardwang:
When human is not the host (e.g. mosquito samples), don't allow downloads of human contigs or alignment visualizations


app/assets/src/components/views/report/ReportTable/HoverActions.jsx
```diff
@@ -52,7 +52,7 @@ class HoverActions extends React.Component {
       icon: "fa-bars",
       handleClick: this.props.onAlignmentVizClick,
       enabled: this.props.alignmentVizEnabled,
-      disabledMessage: "Requires at least one read in NT",
+      disabledMessage: "Alignment Visualization Not Available",
```
MarkAZhang: We discussed keeping this message more specific, i.e. `Alignment visualization requires at least one read in NT` and `Alignment visualization is not enabled for humans.`

PR #1828, opened 2018-12-19
[Report Page] Simplify report_ready check to fix blank table when Alignment is running

jshoe:
- We actually did a good job of migrating data with PipelineRuns -> PipelineRunStages -> OutputStates, so every PipelineRun has non-nil results_finalized, and every successful sample has OutputStates for taxon_counts filled in (see below)
- So the mash of report checks we had here can finally be simplified to report_ready, which checks if the report ready output state is loaded (which is taxon_counts). Basically IFF taxon_counts is loaded, you can show the report table.
- This also fixes the bug where, if you click on a sample while it's running Alignment, you wouldn't see the spinner at all and would actually just get a blank table.
- This is better for people who are keeping the tab open and see the page refresh every 5 minutes and wonder what's going on after host filtering.
- pipeline_run_retriable was no longer used.

#### BEFORE
screen shot 2018-12-19 at 3 11 06 pm

#### AFTER
screen shot 2018-12-19 at 3 01 49 pm

- You can verify the output states above:

```
irb(main):072:0> PipelineRun.where(results_finalized: nil)
=> #<ActiveRecord::Relation []>
irb(main):073:0>

irb(main):044:0> Sample.all.each do |s|
irb(main):045:1*   pr = s.pipeline_runs[0]
irb(main):046:1>   if pr.results_finalized == PipelineRun::FINALIZED_SUCCESS
irb(main):047:2>     outputs = pr.output_states
irb(main):048:2>     if outputs.find_by(output: "taxon_counts")
irb(main):049:3>       if outputs.find_by(output: "taxon_counts").state != "LOADED"
irb(main):050:4>         puts s.id
irb(main):051:4>       end
irb(main):052:3>     end
irb(main):053:2>   end
irb(main):054:1> end
(returns nothing)
```

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -283,6 +283,8 @@ class SamplesController < ApplicationController
     background_id = get_background_id(@sample)
     @report_page_params = { pipeline_version: @pipeline_version, background_id: background_id } if background_id
     @report_page_params[:scoring_model] = params[:scoring_model] if params[:scoring_model]
+
+    # Check if the report table should actually show
     if @pipeline_run && (((@pipeline_run.adjusted_remaining_reads.to_i > 0 || @pipeline_run.results_finalized?) && !@pipeline_run.failed?) || @pipeline_run.report_ready?)
       if background_id
         @report_present = true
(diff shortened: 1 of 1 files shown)


MarkAZhang:
LGTM, thanks for the clear explanation!!

PR #1831, opened 2018-12-20
Ensure Heatmap doesn't crash when metric data is missing.

MarkAZhang:
This guards against a hypothetical crash where `this.state.data` has data for some metrics but not all.

Previously, if any metric had data, we would try to render the heatmap even if the currently selected metric had no data, causing a crash.
Now, we only render the heatmap if the currently selected metric has data. (using the same condition from line 311 which renders the legend)

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapView.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView.jsx
@@ -328,7 +328,7 @@ class SamplesHeatmapView extends React.Component {
     if (
       this.state.loading ||
       !this.state.data ||
-      Object.values(this.state.data).every(e => !e.length) ||
+      !(this.state.data[this.state.selectedOptions.metric] || []).length ||
       !this.state.metadataTypes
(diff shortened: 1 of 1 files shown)


tfrcarvalho:
A couple of comments/questions:
1. I actually think we should always render the heatmap, even if its all gray. Do you know where was the actual crash? This also is tied to actually showing all the samples even if they have no data (which I did not get to implement.... butt I thought I saw someone working on it?)
2. This should probably never actually happen for "good" data, right? Do you have an example?

jazzmes:
Independently of the comments, feel free to commit this. We can work on the other behavior later if necessary.

MarkAZhang:
It's true that this should never happen for good data, but since I noticed it and it's an easy fix, it still seems better to have the right conditional.

I'm totally down to render the heatmap even if it's all gray, but right now if you pass an empty array of data (for a given metric) into the Heatmap it crashes, so it'll be a bit more work to get it working properly.

PR #1834, opened 2018-12-21
Update more styles.

MarkAZhang:
Change some inputs to height 34px that were missed the first time.
Change the font-size back to 13px.
Change the dropdown item color to pure black (semantic ui has 0.87 opacity)

screen shot 2018-12-20 at 4 34 48 pm


MarkAZhang:
Also remove some styles that aren't being used any more.

PR #1837, opened 2018-12-21
Disabling phylogenetic tree creation on human tax ids. 

jameshowardwang:
Also touching up tooltips and removing unused route

### DIFF
--- a/app/assets/src/components/views/report/ReportTable/HoverActions.jsx
+++ b/app/assets/src/components/views/report/ReportTable/HoverActions.jsx
@@ -52,7 +52,8 @@ class HoverActions extends React.Component {
       icon: "fa-bars",
       handleClick: this.props.onAlignmentVizClick,
       enabled: this.props.alignmentVizEnabled,
-      disabledMessage: "Alignment Visualization Not Available",
+      disabledMessage:
+        "Alignment Visualization Not Available - requires reads in NT",
       extraProps: {
         "data-tax-level": this.props.taxLevel === 1 ? "species" : "genus"
       }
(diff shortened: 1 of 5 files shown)


app/assets/src/components/views/report/ReportTable/HoverActions.jsx
```diff
@@ -66,7 +67,8 @@ class HoverActions extends React.Component {
       icon: "fa-code-fork",
       handleClick: this.handlePhyloModalOpen,
       enabled: this.props.phyloTreeEnabled,
-      disabledMessage: "Phylogenetic Analysis Not Available"
+      disabledMessage:
+        "Phylogenetic Analysis Not Available - requires 100+ reads in NT/NR"
```
MarkAZhang: Yay!

config/routes.rb
```diff
@@ -40,7 +40,6 @@
   end
 
   get 'samples/:id/fasta/:tax_level/:taxid/:hit_type', to: 'samples#show_taxid_fasta'
-  get 'samples/:id/alignment/:taxon_info', to: 'samples#show_taxid_alignment'
```
MarkAZhang: Nit: Also remove `show_taxid_alignment` from READ_ACTIONS in samples_controller.

app/controllers/phylo_trees_controller.rb
```diff
@@ -37,6 +37,7 @@ def index
     @taxon = {}
 
     taxid = params[:taxId]
+    return if HUMAN_TAX_IDS.include? taxid.to_i
```
MarkAZhang: I'm a little concerned how the front-end will respond to these empty responses. It would be great if it displayed error messages or something. I just checked and `PhyloTreeListView` can crash if no trees are provided, but that's a separate bug I can address.

For this PR, can you return something like `render json: { status: :not_acceptable, message: "Human taxids are not allowed"}` instead of returning nothing.

config/routes.rb
jameshowardwang: it's already gone, i removed it in the last diff and forgot to include this change in the diff

app/controllers/phylo_trees_controller.rb
jameshowardwang: ok sounds good

PR #1843, opened 2019-01-02
Fix checking the checkbox when opening samples

jshoe:
- This is kind of hacky to work within the existing code, but basically before the htmlFor in <label> would bind the "sample name zone" to the checkbox and count that as a click. So it made it weird to open up lots of samples with SHIFT+CLICK and such because they'd also get checked. And when you open a sample it'll get checked too.
- This basically removes the htmlFor and adds data-checked/onClick/data-sample-id to the label. The label is/was basically resting on top of the checkbox in alignment so it'll get the click.

### NOW

jan-02-2019 10-30-10

### DIFF
--- a/app/assets/src/components/Samples.jsx
+++ b/app/assets/src/components/Samples.jsx
@@ -687,6 +687,7 @@ class Samples extends React.Component {
   }
 
   viewSample(id, currentEvent) {
+    currentEvent.stopPropagation();
     openUrl(`/samples/${id}`, currentEvent);
   }
 
@@ -791,6 +792,9 @@ class Samples extends React.Component {
       .length === 0;
 
   selectSample(e) {
+    console.log("time is 2:17pm");
+    e.stopPropagation();
+
     let sampleId = parseInt(e.target.getAttribute("data-sample-id"));
 
     const sampleList = e.target.checked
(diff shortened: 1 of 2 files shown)


app/assets/src/components/Samples.jsx
```diff
@@ -1870,12 +1873,16 @@ function SampleCardCheckboxes({
             key={`sample_${sample.db_sample.id}`}
```
MarkAZhang: Can you remove the `onClick` and `data-sample-id` props? They aren't needed, and it makes this code more confusing.

app/assets/src/components/Samples.jsx
```diff
@@ -1870,12 +1873,16 @@ function SampleCardCheckboxes({
...
+            checked={checked}
             disabled={report_ready != 1}
           />{" "}
-          <label htmlFor={i}>{sample_name_info}</label>
+          <label
+            data-checked={checked}
+            onClick={parent.selectSample}
+            data-sample-id={sample.db_sample.id}
+          >
+            {sample_name_info}
```
MarkAZhang: Can you wrap `{sample_name_info}` in `<div onClick={e => e.stopPropagation()} ></div>`?

I checked and this seems to work (though you should check as well). it prevents `selectSample` from being called when you click on the label.

The code works right now, but what's actually happening right now is that `parent.selectSample` IS getting called when you click the label. But sampleId resolves to `NaN` and `e.target.getAttribute("data-checked")` is `null`, so the code happens to do what you want. Better to not call `selectSample` though.

PR #1849, opened 2019-01-04
[Sample Uploads] Refactor sample creation API calls

jshoe:
- Cleans up the local and remote upload functions. Helps with bulk uploader calls.


jshoe:
ty

PR #1852, opened 2019-01-08
[DataTable] Changes to support updating data

jshoe:
- Changes from Tiago
- Rename prepareData to indexData and other changes

### DIFF
--- a/app/assets/src/components/visualizations/table/DataTable.jsx
+++ b/app/assets/src/components/visualizations/table/DataTable.jsx
@@ -7,11 +7,11 @@ class DataTable extends React.Component {
   constructor(props) {
     super(props);
 
-    this.originalData = this.prepareData(this.props.data);
-
     this.state = {
       filter: "",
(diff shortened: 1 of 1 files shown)


app/assets/src/components/visualizations/table/DataTable.jsx
```diff
@@ -7,24 +7,27 @@ class DataTable extends React.Component {
...
+    let newState = {};
     if (newFilter !== state.filter) {
-      return {
-        filter: newFilter
-      };
+      newState.filter = newFilter;
+    }
+    if (props.data !== state.originalData) {
+      newState.originalData = props.data;
+      newState.indexedData = DataTable.indexData(props.data);
```
tfrcarvalho: We probably need to reset `selectedRows` too. Maybe to `props.selectedRows` again. Wdyt?

jshoe: added

PR #1854, opened 2019-01-09
[Bulk Uploads] Add local file bulk uploading to the upload page

jshoe:
- A lot of the diff in BulkUploadImport is just from moving things around
- Decided to flag the bulk local uploader with admin-only for now so we can test and refine a bit more before release
- BulkUploadImport functions handle the file picking/uploading process and BulkSampleUploadTable is just used for display
- BulkSampleUploadTable is specific for this action and is a wrapper around DataTable
- Didn't think I could use CSS modules in BulkSampleUploadTable because it ended up looking like "bulkSampleUploadTable-26Aw4" -> "idseq-ui data-table" -> "data-table__header column-progress" and then I was unable to add the non-CSS module styles
- Limitations: Error handling is kind of crappy and just a consistent errorMessage string. Also only 1 project and 1 host for now. Eventual flow will include selectors for each.
- Hard to see in the gifs but it is indeed a striped table

#### Example flows
jan-08-2019 17-06-41

#### Another one
jan-08-2019 17-05-55

#### Example selection error
screen shot 2019-01-08 at 4 54 52 pm

#### Striped tables
screen shot 2019-01-08 at 4 55 02 pm

#### Example creation errors
screen shot 2019-01-08 at 4 55 35 pm

#### Another one
screen shot 2019-01-08 at 4 57 00 pm


app/assets/src/components/utils/links.js
```diff
@@ -21,4 +21,14 @@ const downloadStringToFile = str => {
   location.href = `${downloadUrl}`;
 };
 
-export { openUrl, downloadStringToFile };
+const goToPage = page => {
```
tfrcarvalho: Here you can just use openUrl without an event.

app/assets/src/components/BulkUploadImport.jsx
```diff
@@ -335,15 +346,19 @@ class BulkUploadImport extends React.Component {
     } else {
       errors.project = "Please select a project";
     }
-    if (this.refs.bulk_path) {
-      if (this.refs.bulk_path.value === "") {
+
+    if (!this.state.localUploadMode) {
+      if (this.refs.bulk_path) {
+        if (this.refs.bulk_path.value === "") {
```
tfrcarvalho: Do you need this many errors? You can just say that it is invalid and do
```
if (!this.refs.bulk_path
	 || this.refs.bulk_path.value === ""
	 || !this.filePathValid(this.refs.bulk_path.value))
```
Probably even put the second condition inside as part of the validation function.

app/assets/src/components/utils/links.js
jshoe: 👍

app/assets/src/components/BulkUploadImport.jsx
jshoe: 👍

PR #1856, opened 2019-01-09
Update mosquito metadata schema for Biohub users.

MarkAZhang:
Several fields have been split up.
Since we will be moving to custom schemas soon, this is fairly temporary and doesn't reflect the final set of default metadata fields.

### DIFF
--- a/app/assets/src/components/views/report/SampleDetailsSidebar/constants.js
+++ b/app/assets/src/components/views/report/SampleDetailsSidebar/constants.js
@@ -47,8 +47,7 @@ export const HUMAN_METADATA_SECTIONS = [
   }
 ];
 
-// Includes mosquito and other vectors.
-export const VECTOR_METADATA_SECTIONS = [
+export const MOSQUITO_METADATA_SECTIONS = [
   {
     name: "Sample Info",
     keys: [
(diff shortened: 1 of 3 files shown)


app/assets/src/components/views/report/SampleDetailsSidebar/constants.js
```diff
@@ -57,13 +56,57 @@ export const VECTOR_METADATA_SECTIONS = [
...
+      "collection_lat",
+      "collection_long",
+      "collected_by"
+    ]
+  },
+  {
+    name: "Specimen Info",
+    keys: [
+      "reported_sex",
+      "comp_sex",
```
tfrcarvalho: Any reason not use `computed` (assuming thats the meaning) ? After all is the same size as reported.

app/models/metadatum.rb
```diff
@@ -114,9 +123,17 @@ class Metadatum < ApplicationRecord
...
     sample_unit: "Sample Unit",
     life_stage: "Life Stage",
-    id_method: "ID Method",
-    genus_species: "Genus/Species",
-    blood_fed: "Blood Fed"
+    preservation_method: "Preservation Method",
+    trap_type: "Trap Type",
+    blood_fed: "Blood Fed",
+    reported_sex: "Reported Sex",
+    comp_sex: "Comp. ID Sex",
```
tfrcarvalho: I do not think it makes sense to shorten words on the description.

app/assets/src/components/views/report/SampleDetailsSidebar/constants.js
MarkAZhang: These are the keys biohub sent to me. In their template it's "reported_sex" and "comp_sex". I agree with you that it's arbitrary, but it'll make uploading their metadata a little easier.

When we move to customizable metadata, we can adjust our baseline fields to be the way we want them.

app/models/metadatum.rb
MarkAZhang: "Comp. ID Sex" does look weird. I'll do "Computed Sex". The full description is "Computationally Identified Sex", but that seems a little long for a label.

PR #1865, opened 2019-01-11
Add Bulk Metadata Upload for existing samples.

MarkAZhang:
Includes front end and back end endpoints.
Wrote comprehensive Rails controller tests for back-end.

Also:
Add Project Upload Menu in project page to upload samples and metadata.
Sample page can now take a project id and auto-populate that project in the dropdown.
Small change to Wizard to handle the Metadata Upload Modal.
  Needed to be able to enable/disable the continue button programmatically.
Styling tweaks to Wizard and Filepicker.

screen shot 2019-01-11 at 3 42 09 pm

screen shot 2019-01-11 at 3 42 14 pm

screen shot 2019-01-11 at 3 42 47 pm

screen shot 2019-01-11 at 3 42 51 pm

screen shot 2019-01-11 at 3 43 05 pm


test/controllers/projects_controller_test.rb
```diff
@@ -55,4 +56,334 @@ class ProjectsControllerTest < ActionDispatch::IntegrationTest
...
+
+    assert_response :success
+
+    assert_equal 7, @response.parsed_body['issues']['errors'].length
+    # Error should throw if invalid float is passed for float data type.
+    assert_match 'foobar is not a valid number (row 1)', @response.parsed_body['issues']['errors'][0]
+    # Error should throw if invalid date is passed for date data type.
+    assert_match 'foobar is not a valid date (row 1)', @response.parsed_body['issues']['errors'][1]
+    # Error should throw if metadata type is not supported for the sample's host genome.
+    assert_match 'blood_fed is not a supported metadatum for host genome Human (row 1)', @response.parsed_body['issues']['errors'][2]
```
cdebourcy: Slightly worried that it will be a pain to fix all the tests if we want to change the wording of any of these error messages... Maybe extract the boilerplate text out into functions?

test/controllers/projects_controller_test.rb
```diff
@@ -55,4 +56,334 @@ class ProjectsControllerTest < ActionDispatch::IntegrationTest
 
     assert_redirected_to projects_url
   end
+
+  # Tests for validate_metadata_csv
```
cdebourcy: Could consider extracting into a dedicated file as it's getting pretty long?

app/helpers/projects_helper.rb
```diff
@@ -1,2 +1,134 @@
...
+          errors.push("#{row[sample_name_index]} (row #{index}) does not match any sample names in this project.")
+          next
+        end
+      end
+
+      # Validate the metadatum values with the sample.
+      row.each_with_index do |value, row_index|
+        if row_index >= metadata["headers"].length
+          next
+        end
```
cdebourcy: FYI there's a nice way you can write `next if row_index >= metadata["headers"].length` to make the code a bit more concise

app/helpers/projects_helper.rb
```diff
@@ -1,2 +1,134 @@
...
+    warnings = []
+
+    if new_samples
+      # Require host_genome_name as a column.
+      unless metadata["headers"].include?("host_genome_name")
+        errors.push("host_genome_name column is required.")
+        return { "errors" => errors, "warnings" => warnings }
+      end
+    else
+      # Require sample_name as a column.
```
cdebourcy: So if new_samples is true, then sample_name is not required? How come?

app/helpers/projects_helper.rb
```diff
@@ -1,2 +1,134 @@
...
+          errors.push("Row #{index} is missing host_genome_name.")
+          next
+        end
+        host_genome = HostGenome.where(name: row[host_genome_name_index]).first
+        if host_genome.nil?
+          errors.push("#{row[host_genome_name_index]} (row #{index}) is an invalid host_genome_name.")
+          next
+        end
+
+        # Create temporary sample to run metadatum validation
```
cdebourcy: This seems a bit contrived :) Can't you just use class methods to achieve this rather than an instance method? I.e. define def self.metadatum_validate(...) and then use as Sample.metadatum_validate(...)

app/assets/src/components/ui/icons/AlertIcon.jsx
```diff
@@ -0,0 +1,33 @@
+// SVG generated with Sketch, from Jenn
+import React from "react";
+import PropTypes from "prop-types";
+
+const AlertIcon = props => {
+  return (
+    <svg className={props.className} viewBox="0 0 24 24">
+      <g id="Upload-Data" stroke="none" strokeWidth="1" fillRule="evenodd">
+        <g
+          id="3.3-Upload-Copy-10"
```
jshoe: nit: you could remove the Sketch ids for generality

app/helpers/projects_helper.rb
MarkAZhang: `host_genome_names` is required instead. For new_samples, assuming that the samples haven't been created yet. Later, will need both `project_id` and `host_genome_name` to figure out what metadata fields to use.

app/helpers/projects_helper.rb
MarkAZhang: I think it would be a little hard. We need to call `metadatum.valid?`. In order to do that, `metadatum.sample` must be set, so we need a sample instance. I think it's harder to figure out a different way than to just use a temporary sample instance.

test/controllers/projects_controller_test.rb
MarkAZhang: Done.

test/controllers/projects_controller_test.rb
MarkAZhang: Done.

PR #1882, opened 2019-01-24
Extend Search

cdebourcy:
1. Makes pathogen search work at all levels (species/genus/etc) in the phylo tree workflow.
2. Adds a typeahead to the search box on the samples page. It will suggest taxa/samples/tissues/hosts/uploaders that match your query. If you click one of them, it will filter the samples for it. If you hit enter instead, it will do the old-style search.
3. New behavior in point 2 is only for admins or users who have explicitly been given access to the feature.

Need to deploy https://github.com/chanzuckerberg/idseq-infra/pull/179 first.
Need to run `create_elasticsearch_indices.rake` to build indices.


app/assets/src/components/ui/controls/SearchBox.jsx
```diff
@@ -11,51 +12,54 @@ class SearchBox extends React.Component {
...
     }, this.delayCheckMatch);
-  }
+  };
 
   render() {
     const { isLoading, value, results } = this.state;
     return (
       <Search
-        className="idseq-ui input search"
+        className={`idseq-ui input search ${
```
cdebourcy: cx("idseq-ui input search", this.props.rounded && "rounded")

app/assets/src/components/Samples.jsx
```diff
@@ -103,7 +104,9 @@ class Samples extends React.Component {
       selectedProjectId: this.fetchParams("project_id") || null,
       filterParams: this.fetchParams("filter") || "",
       searchParams: this.fetchParams("search") || "",
-      sampleIdsParams: this.fetchParams("ids") || [],
+      sampleIdsParams: this.fetchParams("ids")
```
tfrcarvalho: Can you move this logic to `fecthParams`.
fetchParams should, for certain fields, split into an array.

app/assets/src/components/Samples.jsx
```diff
@@ -120,6 +123,16 @@ class Samples extends React.Component {
       readySampleIdsForFilter: [],
       selectedSampleIds: [],
       displayDropdown: false,
+
+      // For structured search suggestions
+      selectedTaxids: this.fetchParams("taxid")
```
tfrcarvalho: same as above

app/assets/src/components/Samples.jsx
```diff
@@ -120,6 +123,16 @@ class Samples extends React.Component {
...
+        ? this.fetchParams("taxid").split(",")
+        : [],
+      selectedUploaderIds: this.fetchParams("selectedUploaderIds")
```
tfrcarvalho: same as above

lib/tasks/update_lineagedb.rake
```diff
@@ -75,8 +75,11 @@ task update_lineage_db: :environment do
   TaxonLineage.where(started_at: current_date).update_all(version_start: current_lineage_version) # rubocop:disable Rails/SkipsModelValidations
   TaxonLineage.where(ended_at: TaxonLineage.column_defaults["ended_at"]).update_all(version_end: current_lineage_version) # rubocop:disable Rails/SkipsModelValidations
 
+  ## Update Elasticsearch index
+  ## TODO: test if this works
```
tfrcarvalho: Is this TODO done?

docker-compose.yml
```diff
@@ -49,6 +50,24 @@ x-honeycomb-variables: &honeycomb-variables
 
 version: '3.4'
 services:
+  elasticsearch:
```
tfrcarvalho: Should we update our deploy scripts to add ES?

config/initializers/elasticsearch.rb
```diff
@@ -0,0 +1,5 @@
+config = {
+  host: ENV['ES_ADDRESS'],
+  transport_options: { request: { timeout: 200 } }
+}
+Elasticsearch::Model.client = Elasticsearch::Client.new(config) unless Rails.env == 'test'
```
tfrcarvalho: what happens in `test`?

app/helpers/elasticsearch_helper.rb
```diff
@@ -0,0 +1,35 @@
+module ElasticsearchHelper
+  def prefix_match(model, field, prefix, condition)
+    search_params = { query: { query_string: { query: "#{prefix}*", fields: [field] } } }
+    results = if Rails.env == "test"
+                # Return all records. Tests can't use elasticsearch.
```
tfrcarvalho: Why cant test use?

tfrcarvalho: We need to find a way (can be done later) to get rid of all the 'if test' conditions.

app/assets/src/components/ui/controls/category_search_box.scss
```diff
@@ -0,0 +1,45 @@
+@import "~styles/themes/colors";
+
+:global(.prompt) {
```
tfrcarvalho: Can you encapsulate all this in a unique class that you assign to your main component?

app/assets/src/components/ui/controls/CategorySearchBox.jsx
```diff
@@ -0,0 +1,19 @@
+import PropTypes from "prop-types";
+import React from "react";
+import SearchBox from "./SearchBox";
+import cs from "./category_search_box.scss"; /* eslint-disable-line */
```
tfrcarvalho: Do the encapsulation as suggested below and then remove this comment.

app/assets/src/components/Samples.jsx
```diff
@@ -877,50 +907,97 @@ class Samples extends React.Component {
...
-      "selectedTissueFilters",
-      "Sample Type: "
-    );
+    let search_tag_list;
+    if (this.admin !== 0 || this.allowedFeatures.includes("structuredSearch")) {
+      search_tag_list = this.state.searchTags.map((entry, i) => {
+        return (
+          <Label className="label-tags" size="tiny" key={`search-tag-${i}`}>
+            {entry.display}
+            <Icon
```
tfrcarvalho: Use an icon from `ui/icons` or add one yourself.
I believe for this is the `RemoveIcon`.

app/assets/src/components/ui/controls/SearchBox.jsx
```diff
@@ -11,51 +12,64 @@ class SearchBox extends React.Component {
...
       isLoading: false,
       results: [],
       value: this.props.initialValue
     };
   }
 
-  resetComponent() {
-    this.setState(this.blankState);
-  }
+  onKeyDown = e => {
```
tfrcarvalho: Why do we need onEnter? Isn't search automatic?

app/assets/src/components/Samples.jsx
cdebourcy: good idea

app/assets/src/components/Samples.jsx

app/assets/src/components/ui/controls/SearchBox.jsx
cdebourcy: Search is automatic but only triggered onSearchChange. If the input didn't change, pressing Enter won't do anything by default. Jenn said when you hit enter (as opposed to clicking one of the suggestions), the search should work just like it does now

app/assets/src/components/ui/controls/category_search_box.scss
cdebourcy: yes, thanks

app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
```diff
@@ -480,16 +458,12 @@ class PhyloTreeCreation extends React.Component {
...
             <div className="wizard__page-2__searchbar__container">
-              {this.state.taxaLoaded ? (
-                <SearchBox
-                  source={this.state.taxonList}
-                  onResultSelect={this.handleSelectTaxon}
-                  initialValue={this.state.taxonName}
-                  placeholder="Genus name"
-                />
-              ) : (
-                <LoadingIcon />
```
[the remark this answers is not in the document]
cdebourcy: LoadingIcon shows up when the *search box* is not yet loaded... because we used to transfer all the taxa to the frontend, which meant it took a while for the search box to show up. Now, since we don't preload any of the data, the search box will always show up immediately, so no need for LoadingIcon. Other search boxes that don't preload data also don't have a loading icon

app/controllers/samples_controller.rb
```diff
@@ -119,6 +122,82 @@ def all
...
+    unless projects.empty?
+      results["Project"] = {
+        "name" => "Project",
+        "results" => projects.index_by(&:name).map do |_, p|
+          { "category" => "Project", "title" => p.name, "id" => p.id }
+        end
+      }
+    end
+    unless samples.empty?
+      results["Sample"] = {
```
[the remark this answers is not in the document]
cdebourcy: This is required for the category names to show up correctly in the semantic search box. This string will be the displayed category name (Jenn's mocks have it as a singular). Didn't seem worth changing the internals of the component

app/controllers/samples_controller.rb
```diff
@@ -119,6 +122,82 @@ def all
...
+        "name" => "Sample",
+        "results" => samples.index_by(&:name).map do |val, record|
+          { "category" => "Sample", "title" => val, "id" => val, "sample_ids" => [record.id], "project_id" => record.project_id }
+        end
+      }
+    end
+    unless locations.empty?
```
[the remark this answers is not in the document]
cdebourcy: The mocks didn't have empty categories, but happy to discuss

app/helpers/elasticsearch_helper.rb
cdebourcy: ok let's do it later. We only really use tests for access control

lib/tasks/update_lineagedb.rake
tfrcarvalho: Missed this question?

yarn.lock
```diff
@@ -0,0 +1,9006 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
tfrcarvalho: Remove this file?

docker-compose.yml
tfrcarvalho: You did not set it as a dependency of web... so it wont be started.

app/assets/src/components/ui/controls/SearchBox.jsx
tfrcarvalho: ok

app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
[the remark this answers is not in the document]
tfrcarvalho: understood

app/controllers/samples_controller.rb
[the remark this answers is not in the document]
tfrcarvalho: Here is more a case of thinking server vs. client. Server should not be designed to fit the client.
The client should adapt to server side formats.

docker-compose.yml
cdebourcy: This is for local development only... it does get started when do docker-compose up locally. For staging/prod, you just have to do ‘make apply’ in idseq-infra

cdebourcy: This is for local development only... it does get started when do docker-compose up locally. For staging/prod, you just have to do ‘make apply’ in idseq-infra

tfrcarvalho: Can you add the dependency? 
I only do `docker-compose up web` because most of the time i do not need the rescue workers. So it would be nice to be able to keep up doing that :)

app/assets/src/components/Samples.jsx
cdebourcy: done

lib/tasks/update_lineagedb.rake
cdebourcy: Resolved. Following further investigation, it seems the todo is unnecessary -- elasticsearch callbacks will handle it

docker-compose.yml
cdebourcy: makes sense. I added it :)

app/helpers/elasticsearch_helper.rb
cdebourcy: Agree, this is not great for code organization. Right now, I struggle to think of the best way to handle the issue without too much overhead. Will make a task to reevaluate this

cdebourcy:
> Can you create Jira tasks for any issues that we will address in the future?

Done

app/controllers/samples_controller.rb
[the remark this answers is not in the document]
cdebourcy: Conceptually it makes sense to me, but I struggle to think which format changes would be of concrete benefit here. Feel free to add thoughts to this Jira task: https://jira.czi.team/browse/IDSEQ-844

PR #1886, opened 2019-01-25
[MetadataFields] Get rid of data_type and KEY_TO_TYPE

jshoe:
- Basically data_type now comes from metadata_field.base_type and KEY_TO_TYPE is no longer needed.
- At this point every piece of metadata should have its metadata_field_id column from the migrations (covered all the keys).

screen shot 2019-01-25 at 1 08 38 pm


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20_190_116_005_251) do
+ActiveRecord::Schema.define(version: 20_190_124_230_435) do
```
tfrcarvalho: You could just remove this file.

jshoe:
^ Also added the "groups" to come from MetadataFields

PR #1888, opened 2019-01-28
Add confirmation modal to make project public + refactor project settings.

tfrcarvalho:


app/assets/src/components/views/samples/ProjectSettingsModal/ProjectSettingsModal.jsx
```diff
@@ -0,0 +1,119 @@
...
+import PublicProjectConfirmationModal from "./PublicProjectConfirmationModal";
+import cs from "./project_settings_modal.scss";
+import axios from "axios";
+
+class ProjectSettingsModal extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      modalOpen: true,
```
MarkAZhang: I feel like this should be `false` by default.

app/assets/src/components/views/samples/ProjectSettingsModal/UserManagementForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+      statusMessage: "",
+      statusClass: null
+    };
+  }
+
+  handleAddUser = () => {
+    const { csrf, onUserAdded, project } = this.props;
+    const { username, email } = this.state;
+
+    if (StringHelper.validateEmail(email) && StringHelper.validateName(name)) {
```
MarkAZhang: This should be `validateName(username)`

app/assets/src/components/views/samples/ProjectSettingsModal/UserManagementForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+              None
+            </div>
+          )}
+        </div>
+        <div className={cs.fillIn} />
+        <div className={cx(cs.statusMessage, this.state.statusClass)}>
+          {this.state.statusMessage}
+        </div>
+        <div className={cs.addMemberForm}>
+          <div className={cs.addMemberFormField}>
```
MarkAZhang: I think this UI needs another look.

At the moment the Inputs are unlabeled, so it's not clear what to put in them. Can add a label or a placeholder.

Also not clear why the user needs to enter both a name AND an email to add a user. When I looked at the back-end, I realized that the back-end adds the existing user if the e-mail already exists, and creates a new user if not.

We might consider splitting up the UI to handle "Add existing user" and "Add new user" separately? Maybe talk to Jenn

tfrcarvalho: Your concern on existing vs. new users is valid but is out of scope of this change. Existing user for instance should have a search field or something where you pick the user.

This PR does not change the current functionality on that. Eventually, we can add that, but not the focus of this change. Created JIRA ticket for that: https://jira.czi.team/browse/IDSEQ-834.

app/assets/src/components/Samples.jsx
```diff
@@ -1631,136 +1588,65 @@ class BackgroundModal extends React.Component {
...
+          <div className={cs.projectMenuItem}>
+            <LockIcon className={cs.smallIcon} /> Private project
+          </div>
+        ))}
+      {proj &&
+        parent.canEditProject(proj.id) && (
+          <div className={cs.projectMenuItem}>
+            <UserIcon className={cs.smallUserIcon} />{" "}
+            {proj_users_count
+              ? `${parent.state.project_users.length} member${parent.state
```
MarkAZhang: This will render "false" as a word if the length is 1.
`length > 1 ? "s" : ""`

PR #1892, opened 2019-01-29
[MetadataFields] Give new projects the current set of default fields

jshoe:


app/assets/src/components/views/report/SampleDetailsSidebar/MetadataTab.jsx
```diff
@@ -25,7 +25,8 @@ class MetadataTab extends React.Component {
 
   getMetadataSections = () => {
     // Group the MetadataFields by group name
-    let nameToFields = {};
+    // Include Sample Info by default for special cases in SAMPLE_ADDITIONAL_INFO
+    let nameToFields = { "Sample Info": [] };
```
tfrcarvalho: Not sure I understand why a section with no fields is required... 
In any case, shouldnt this be done on client side, i.e. SampleDetailsSidebar that uses this?

PR #1894, opened 2019-01-30
Add initial version of Metadata Dictionary, Metadata Template, and up…

MarkAZhang:
…load instructions.

Generate random values for metadata template for now.
Metadata template contains a row for each host genome.

screen shot 2019-01-29 at 5 19 41 pm

screen shot 2019-01-29 at 5 26 24 pm

screen shot 2019-01-29 at 5 27 22 pm

### DIFF
--- a/app/assets/src/components/ui/containers/Wizard.jsx
+++ b/app/assets/src/components/ui/containers/Wizard.jsx
@@ -15,7 +15,8 @@ class Wizard extends React.Component {
 
     this.state = {
       currentPage: 0,
-      continueEnabled: true
+      continueEnabled: true,
+      overlay: null
     };
 
     this.showPageInfo = this.props.showPageInfo || true;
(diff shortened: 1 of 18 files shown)


app/assets/src/components/ui/containers/Wizard.jsx
```diff
@@ -15,7 +15,8 @@ class Wizard extends React.Component {
 
     this.state = {
       currentPage: 0,
-      continueEnabled: true
+      continueEnabled: true,
+      overlay: null
```
tfrcarvalho: Would have been possible and simpler to do a modal ? Same size as the wizard?

app/controllers/metadata_controller.rb
```diff
@@ -0,0 +1,15 @@
+class MetadataController < ApplicationController
+  include MetadataHelper
+
+  def dictionary
+  end
+
+  # All users get the same fields.
+  def official_metadata_fields
+    render json: official_metadata_fields_helper()
```
jshoe: Rubocop will probably complain about the "()"s

app/helpers/metadata_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+    if field.base_type.zero?
+      if field.options.present?
+        options = JSON.parse(field.options)
+        return options[Random.new.rand(options.length)]
+      end
+
+      return "Example " + field.display_name
+    end
+
+    if field.base_type == 1
```
jshoe: should probably use the constants for 1, 2

PR #1895, opened 2019-01-30
Add scrollbar mixin and apply to all elements.

MarkAZhang:
screen shot 2019-01-29 at 6 20 40 pm


MarkAZhang:
mixin is recommended over extend.

https://csswizardry.com/2014/11/when-to-use-extend-when-to-use-a-mixin/

Also extend doesn't work when trying to extend `*`

tfrcarvalho:
Thank you for looking into mixin vs extend

PR #1899, opened 2019-01-30
Sort the metadata in the heatmap alphabetically

jameshowardwang:
See https://jira.czi.team/browse/IDSEQ-822

### DIFF
--- a/app/assets/src/components/ui/controls/SearchBoxList.jsx
+++ b/app/assets/src/components/ui/controls/SearchBoxList.jsx
@@ -49,8 +49,11 @@ class SearchBoxList extends React.Component {
         unselectedOptions.push(option);
       }
     });
+    unselectedOptions.sort((a, b) => (a.label > b.label ? 1 : -1));
+
     return Array.from(selected)
       .map(optionValue => selectedOptions[optionValue])
+      .sort((a, b) => (a.label > b.label ? 1 : -1))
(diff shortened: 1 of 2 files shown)


app/assets/src/components/ui/controls/SearchBoxList.jsx
```diff
@@ -49,8 +49,11 @@ class SearchBoxList extends React.Component {
         unselectedOptions.push(option);
       }
     });
+    unselectedOptions.sort((a, b) => (a.label > b.label ? 1 : -1));
```
jshoe: I think we've been consolidating on lodash/fp for these kind of things so you could do something like `import { sortBy } from "lodash/fp";` and `unselectedOptions = sortBy("label", unselectedOptions)` https://gist.github.com/jfmengels/6b973b69c491375117dc#_sortbyiteratees-collection

jameshowardwang: 🙏

MarkAZhang: I guess this is fine here since the same function is also used below. (normally we'd use `lodash/fp`, but then you'd have to re-write the chaining below)

Maybe do `const sortByLabel = (a, b) => (a.label > b.label ? 1 : -1)` and then `.sort(sortByLabel)` in the three places below.

PR #1907, opened 2019-02-05
[Analytics] Set up Segment API calls

jshoe:
- Add logAnalyticsEvent for sending to Segment

- It works!

screen shot 2019-02-04 at 4 42 37 pm

### DIFF
--- a/app/assets/src/api/constants.js
+++ b/app/assets/src/api/constants.js
@@ -1,0 +1,4 @@
+// Names may change as schema is developed
+export const ANALYTICS_EVENT_NAMES = {
+  SAMPLE_REPORT_VIEW: "SAMPLE_REPORT_VIEW"
+};
--- a/app/assets/src/api/constants.js
+++ b/app/assets/src/api/constants.js
@@ -1,4 +1,9 @@
-// Names may change as schema is developed
+// Event name guidelines:
+// Follow object_action convention with object being the name of the core model or component name
+// if it makes sense. Keep names meaningful, descriptive, and non-redundant (e.g. prefer sample_view
+// to sample_view_view).
+// ANALYTICS_EVENT_NAMES will have the camelCase key and snake_case value.
+
 export const ANALYTICS_EVENT_NAMES = {
-  SAMPLE_REPORT_VIEW: "SAMPLE_REPORT_VIEW"
+  sampleView: "sample_view"
 };
--- a/app/views/layouts/_segment_analytics.html.erb
+++ b/app/views/layouts/_segment_analytics.html.erb
@@ -1,19 +1,11 @@
 <!-- Global site tags for Segment analytics-->
-<script>
-  const SEGMENT_IDS = {
-    // TODO: Replace these with new accounts after they get set up
-    // These are designed to be public: https://community.segment.com/t/m26sng/writekey-accessible-by-anyone
-    prod: "7gUE8JMXEkhEo9ioD1uHke1RgpDFxxlG",
-    staging: "7gUE8JMXEkhEo9ioD1uHke1RgpDFxxlG",
-    development: "7gUE8JMXEkhEo9ioD1uHke1RgpDFxxlG"
-  };
-
-  // Snippet from https://segment.com/docs/sources/website/analytics.js/quickstart/
-  const matched_id = SEGMENT_IDS["<%= Rails.env %>"];
-  if (matched_id) {
+<% if ENV["SEGMENT_ID"] %>
+  <script>
+    // Snippet from https://segment.com/docs/sources/website/analytics.js/quickstart/
+    // Segment ID is designed to be public: https://community.segment.com/t/m26sng/writekey-accessible-by-anyone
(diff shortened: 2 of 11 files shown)


jshoe:
^ Going to rework the keys and protect them a bit more

app/views/layouts/_segment_analytics.html.erb
```diff
@@ -0,0 +1,19 @@
+<!-- Global site tags for Segment analytics-->
+<script>
+  const SEGMENT_IDS = {
+    // TODO: Replace these with new accounts after they get set up
+    // These are designed to be public: https://community.segment.com/t/m26sng/writekey-accessible-by-anyone
+    prod: "7gUE8JMXEkhEo9ioD1uHke1RgpDFxxlG",
```
tfrcarvalho: Do you want to load this from server as discussed offline?

tfrcarvalho: Just noticed the comment above.

app/views/layouts/application.html.erb
```diff
@@ -2,6 +2,7 @@
 <html>
   <head>
     <%= render "layouts/google_analytics" %>
```
gregdingle: What's the plan for google analytics once segment is live? Is there a task to replace it?

app/views/layouts/_segment_analytics.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  };
+
+  // Snippet from https://segment.com/docs/sources/website/analytics.js/quickstart/
+  const matched_id = SEGMENT_IDS["<%= Rails.env %>"];
+  if (matched_id) {
+    !function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t,e){var n=document.createElement("script");n.type="text/javascript";n.async=!0;n.src="https://cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(n,a);analytics._loadOptions=e};analytics.SNIPPET_VERSION="4.1.0";
+      analytics.load(matched_id);
+      analytics.page();
```
gregdingle: Did you review the optional properties here https://segment.com/docs/sources/website/analytics.js/#page and here https://segment.com/docs/sources/website/analytics.js/#load-options to see if any are relevant?

gregdingle: And to what extent should we use analytics.page vs analytics.track to record something like a SAMPLE_REPORT_VIEW ? Does the SAMPLE_REPORT_VIEW page have a unique URL?

app/assets/src/api/constants.js
```diff
@@ -0,0 +1,4 @@
+// Names may change as schema is developed
```
gregdingle: I recommend providing some more guidance on naming conventions here. For example, should we follow the convention of "noun_verb" as in SAMPLE_REPORT_VIEW? Should we elaborate on the source component names such as `SampleView.jsx` -> `SAMPLE_REPORT_VIEW` or keep consistent as in `SampleView.jsx` -> `SAMPLE_VIEW`?

app/views/layouts/_segment_analytics.html.erb
jshoe: 👍

app/assets/src/api/constants.js
```diff
@@ -0,0 +1,4 @@
...
+export const ANALYTICS_EVENT_NAMES = {
+  SAMPLE_REPORT_VIEW: "SAMPLE_REPORT_VIEW"
```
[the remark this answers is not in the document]
jshoe: 👍

app/views/layouts/application.html.erb
jshoe: I think we could get rid of Google Analytics eventually esp. if we forward Segment events there. Made a ticket for it https://jira.czi.team/browse/IDSEQ-841

app/views/layouts/_segment_analytics.html.erb
jshoe: - Yes, I don't find the optional fields that interesting yet but we could add some later.
- I would propose that our team relies on the "tracks" primarily because those will represent the most explicit actions (e.g. there could be a series like "sample_upload_selected", "sample_upload_started") but I think the "page" info could be useful as a secondary source.

app/assets/src/api/constants.js
jshoe: Thanks, added some guidelines. Up for discussion if desired

jshoe:
^ So now SEGMENT_ID will come from Systems Manager "Parameter store" and won't be active in development unless you set var SEGMENT_ID or use chamber in dev

app/views/layouts/_segment_analytics.html.erb
gregdingle: Sounds good 👍

gregdingle:
Although I haven’t used segment before, I would recommend not using
different keys for google analytics, so you can track the same users in a
logged in and logged out state.

jshoe:
@gregdingle Tiago had a concern about "key misuse" so we were considering the separate keys, but I would like to be able to track the logged-out -> logged-in users so I would propose keeping the 1 key and we can address if there are issues? @tfrcarvalho

jshoe:
^ Tiago is OK with the 1 key for now and we could split later once we understand Segment better

PR #1908, opened 2019-02-06
Initial projects view for data discovery

tfrcarvalho:
* Only an initial step in data discovery
* Includes:
	* discovery view scaffold with nav
	* projects view (with table based on react-virtualized)
	* header refactoring
	* header nav
* RequestContext: Add a request context using react context to hold information about the request, in order to be able to use the information across multiple components.


app/assets/src/components/views/projects/projects_view.scss
```diff
@@ -0,0 +1,54 @@
+@import "~styles/themes/colors";
+@import "~styles/themes/typography";
+
+.project {
+  text-align: left;
+  padding: 15px 0;
+
+  &Name {
```
MarkAZhang: Small nit just because I noticed it. I prefer to just use `cs.name` in the JSX and `.name` here, instead of `cs.projectName` and `&Name`. I think it's a little easier to read. And I can CTRL+F for the class and find it easily. If I tried to CTRL+F `projectName`, I wouldn't find it anywhere in the project. Let me know if there's a reason you prefer `&Name`.

tfrcarvalho: I like `projectName` in the html code because i find it easier to read in markup language (not a prob in css).
But I think your "searchable" argument is valid and more relevant. So I can use that.

app/assets/src/components/visualizations/table/Table.jsx
```diff
@@ -0,0 +1,151 @@
+import React from "react";
```
MarkAZhang: Hey Tiago! This class is beautifully written. However, I think one of the reasons we wanted to adopt `react-virtualized` is so we can have fixed headers and a fixed left column while we do both horizontal and vertical scrolling, as in this example: https://bvaughn.github.io/react-virtualized/#/components/ScrollSync

I'm not sure this kind of thing is possible with `<Table>`, and we might have to wrap `<Grid>` directly.

This example uses `<ScrollSync>` with four `<Grids>`: https://github.com/bvaughn/react-virtualized/blob/master/source/ScrollSync/ScrollSync.example.js

Unfortunately this is substantially more complicated, so will let you make the decision about prioritization.

tfrcarvalho: As discussed offline, I will still try to make it work with the table later within the data discovery scope and before adding further features to Table.jsx. If unsuccessful, I will convert it to grid.

app/assets/src/api/index.js
```diff
@@ -212,6 +212,24 @@ const getSampleTaxons = (params, cancelToken) =>
     cancelToken
   });
 
+// TODO: add remaining parameters: filterm search, page, sortBy
```
jshoe: is filterm a tyop?

app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -0,0 +1,101 @@
...
+      const [samples, projects] = await Promise.all([
+        getSamples({ onlyLibrary, excludeLibrary }),
+        getProjects({ onlyLibrary, excludeLibrary })
+      ]);
+      this.setState({
+        samples,
+        projects
+      });
+    } catch (error) {
+      // TODO: handle errors
```
jshoe: maybe console.log at least?

app/assets/src/components/views/projects/ProjectsView.jsx
```diff
@@ -0,0 +1,120 @@
...
+      </div>
+    );
+  };
+
+  renderProjectDetails = ({ cellData: project }) => {
+    return (
+      <div className={cs.project}>
+        <div className={cs.projectName}>{project.name}</div>
+        <div className={cs.projectDescription}>
+          {project.description || "No description (DELETE THIS)"}
```
jshoe: is the "DELETE THIS" for us? or is every project supposed to have a description / ""

app/assets/src/components/views/projects/ProjectsView.jsx
```diff
@@ -0,0 +1,120 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Table } from "~/components/visualizations/table";
+import { merge, pick } from "lodash";
```
jshoe: @MarkAZhang I've noticed some of these regular "lodash" ones slipping in.. should we still be preferring lodash/fp? Tbh I've been thrown off by lodash/fp args sometimes because most code snippets online use regular lodash

app/assets/src/components/views/projects/ProjectsView.jsx
```diff
@@ -0,0 +1,120 @@
...
+        disableSort: true,
+        cellRenderer: this.renderList
+      },
+      {
+        dataKey: "tissues",
+        width: 200,
+        disableSort: true,
+        cellRenderer: this.renderList
+      },
+      { dataKey: "number_of_samples", width: 140, label: "No. Of Samples" }
```
jshoe: nit: maybe "Number of Samples" or "No. of samples"

app/views/home/library.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if current_user.allowed_features.include?("data_discovery") %>
+ 
+<%= javascript_tag do %>
```
jshoe: indent these blocks?

app/assets/src/components/views/projects/ProjectsView.jsx
jshoe: nvm I see that these are basically placeholders

jshoe:
nit for when you do more styling: the Favorite Projects should get more space on top probably without the All Samples

screen shot 2019-02-07 at 11 03 56 am

jshoe:
Noticed a bug where the dropdown row didn't get the onClick, only the link text

feb-07-2019 11-13-07

app/assets/src/components/views/projects/ProjectsView.jsx
MarkAZhang: I think it would be good to be consistent and use `lodash/fp`. It looks like we use it in 90% of our files.

app/assets/src/api/index.js
tfrcarvalho: yes!

app/assets/src/components/views/projects/ProjectsView.jsx
tfrcarvalho: Yes, just placeholders guarded by flag.

app/assets/src/components/views/projects/ProjectsView.jsx
tfrcarvalho: To be precise is 32 / 15 import lines for `lodash/fp`. :)
I gave it a try and I really dislike the args opinions of `lodash/fp` and I think it is affecting the code readability for me (maybe just a personal issue). 

Would hope we can reconsider here... 
(I am normally all for consistency, and pushed for it in this case at the beginning but in this case I would prefer ditching `lodash/fp`)
We can discuss this offline...

app/assets/src/components/views/projects/ProjectsView.jsx
tfrcarvalho: This is design specification. Will bring it up with Jenn since this will not be the final version... but given that is a short column I would prefer the reduced width.

app/views/home/library.html.erb
tfrcarvalho: done.

app/assets/src/components/views/discovery/DiscoveryView.jsx
tfrcarvalho: done

config/routes.rb
```diff
@@ -50,6 +51,8 @@
   get 'select', to: 'home#index'
   get 'home', to: 'home#index'
   get 'taxon_descriptions', to: 'home#taxon_descriptions'
+  get 'public', to: 'home#public'
```
ovalenzuela19, line 54: How does this routing work here @tfrcarvalho ?

PR #1910, opened 2019-02-07
[Analytics] Add backend events and add more analytic event hooks

jshoe:
- Adds Segment gem for backend events.
- Decided to switch to object_action past tense based on this guide https://segment.com/academy/collecting-data/naming-conventions-for-clean-data/ + seems easier to tell sample_view component from sample_viewed event.
- Included all the form info from Landing.jsx so that Product can use the warehouse as a reference later (interest submissions not currently recorded in a table).
- Had to split to SEGMENT_RUBY_ID and SEGMENT_JS_ID because different sources get different keys (and generated by them).
- Later: sample upload events!

screen shot 2019-02-06 at 5 39 27 pm

screen shot 2019-02-06 at 5 48 08 pm


app/controllers/projects_controller.rb
```diff
@@ -180,8 +180,8 @@ def create
 
     respond_to do |format|
       if @project.save
-        tags = %W[project_id:#{@project.id} user_id:#{current_user.id}]
-        MetricUtil.put_metric_now("projects.created", 1, tags)
```
tfrcarvalho: Shouldn't we keep datadog metrics ? 
At least until we can have a dashboard to snowflake.

jshoe: @tfrcarvalho ah good point. Forgot about our dashboard for some reason. Will change

jshoe: @tfrcarvalho updated :)

PR #1919, opened 2019-02-09
Only show Delete Sample button if you can edit sample.

MarkAZhang:


app/assets/src/components/views/SampleView/Controls.jsx
```diff
@@ -69,8 +69,10 @@ class Controls extends React.Component {
           direction="left"
         />
       );
-    } else if (sample.status === "created" || !reportPresent) {
+    } else if (canEdit) {
```
cdebourcy: should we still check that there's no report present though? Right now, we only allow deletion if it's inconsequential because sample failed anyway

PR #1924, opened 2019-02-11
Consolidate bulk upload endpoints.

MarkAZhang:
Test strategy:
Wrote tests for the Rails endpoints.
Tested the new Sample Upload flow by hand for both local and remote. Should test again on staging to make sure the entire pipeline works (it's flaky locally).
Also tested the existing upload flow works.


MarkAZhang:
Thanks. Made a note of testing strategy in the description.

PR #1926, opened 2019-02-12
Enforce required metadata in sample bulk upload endpoint.

MarkAZhang:
screen shot 2019-02-11 at 5 34 35 pm

Also enforce required metadata in validation endpoint.

Test strategy:
Wrote Rails tests.
Verified that error correctly throws during sample upload flow.
Verified that error does NOT throw when uploading metadata to existing samples.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -320,6 +320,20 @@ class SamplesController < ApplicationController
       errors << SampleUploadErrors.invalid_project_id(sample)
     end
 
+    # After creation, if a sample is missing required metadata, destroy it.
+    # TODO(mark): Move this logic into a validator in the model in the future.
+    # Hard to do right now because this isn't launched yet, and also many existing samples don't have required metadata.
+    removed_samples = []
+    samples.each do |sample|
+      missing_required_metadata_fields = sample.missing_required_metadata_fields
+      unless missing_required_metadata_fields.empty?
+        errors << SampleUploadErrors.missing_required_metadata(sample, missing_required_metadata_fields)
+        sample.destroy
+        removed_samples << sample
+      end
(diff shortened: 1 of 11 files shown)


app/controllers/samples_controller.rb
```diff
@@ -320,6 +320,20 @@ def bulk_upload_with_metadata
...
 
+    # After creation, if a sample is missing required metadata, destroy it.
+    # TODO(mark): Move this logic into a validator in the model in the future.
+    # Hard to do right now because this isn't launched yet, and also many existing samples don't have required metadata.
+    removed_samples = []
+    samples.each do |sample|
+      missing_required_metadata_fields = sample.missing_required_metadata_fields
+      unless missing_required_metadata_fields.empty?
+        errors << SampleUploadErrors.missing_required_metadata(sample, missing_required_metadata_fields)
+        sample.destroy()
```
jshoe: RuboCop will probably cite the ()

app/models/sample.rb
```diff
@@ -112,6 +111,18 @@ def input_files_checks
     end
   end
 
+  def required_metadata_fields
+    host_genome.metadata_fields.where(is_required: 1).pluck(:name)
+  end
+
+  def missing_required_metadata_fields
+    required_metadata_fields - metadata.pluck(:key)
```
jshoe: since we want to remove the :key on Metadatum later could you do `required_metadata_fields - metadata.map(&:metadata_field).pluck(:key)`

app/models/sample.rb
```diff
@@ -112,6 +111,18 @@ def input_files_checks
...
+  end
+
+  def required_metadata?
```
jshoe: is this one used?

PR #1928, opened 2019-02-12
[Uploads] Add a _1, _2, etc if sample name already exists

jshoe:
- Removed "success: false" from clearError() so that the successful info messages can still show onChanging project/host.
- Not super elegant but seemed difficult to make a more general validator with name updates.
- resolveSampleNames called on samples imported and handleProjectChangeForSample.

### Example of names already existing
feb-12-2019 15-01-39

### Example of switching to a project with names already existing
feb-12-2019 15-03-59


app/assets/src/components/BulkUploadImport.jsx
```diff
@@ -90,6 +92,8 @@ class BulkUploadImport extends React.Component {
     );
     this.initializeSelectTag();
     this.initializeSelectAll();
+
+    if (this.state.imported) this.resolveSampleNames();
```
MarkAZhang: Do we need to call this here? We already call it at line 353, which is where `imported = true` gets set.

I worry that it'll make all these api calls every time the component re-renders (which can be often)

app/assets/src/components/BulkUploadImport.jsx
```diff
@@ -497,6 +502,29 @@ class BulkUploadImport extends React.Component {
     });
   }
 
+  // Update sample names if they already exist in the project
+  resolveSampleNames = async () => {
+    const updatedSamples = this.state.samples.map(async sample => {
+      // For each sample, ask the server for a validated name
+      const resp = await postWithCSRF(
```
MarkAZhang: Can you move `validateSampleName` into `api/index.js` as a function?

Also, in the new sample upload flow, you can only upload to one project at a time. I'll probably end up refactoring this to check all the sample names in batch (though I guess it doesn't make sense to do it right now).

app/assets/src/components/BulkUploadImport.jsx
jshoe: you're right it wasn't needed 👍

app/assets/src/components/BulkUploadImport.jsx
jshoe: 👍

PR #1930, opened 2019-02-13
Persist visualization params of sample report page in URL

gregdingle:
# Description

This adds URL state to the report page, same as the heatmap page, so that a user can go back to the same state. In a future PR, I plan to add a save button which which will save the state defined here to the database, again, same as the heatmap page. 

For example: http://localhost:3000/samples/12303?pipeline_version=3.3&background_id=26&search_taxon_id=0&searchKey=&rows_passing_filters=80&rows_total=104&pagesRendered=1&sort_by=nt_aggregatescore&name_type=Scientific%20name&rendering=false&loading=false&countType=NT&readSpecificity=1&treeMetric=aggregatescore&phyloTreeModalOpen=true&minContigSize=8&view=tree

image

The state covers all the controls above the visualization. It does not cover the controls inside the SVG.

There was existing code to persist `pipeline_version` in the URL. I did not touch this, but I wonder why not `history.pushState` to avoid the full page load?

Along the way, I tweaked the phylo tree page so that the ID in the URL persists. For example, http://localhost:3000/phylo_trees/index?treeId=115 . I have no idea why `resetUrl()` was there. I did not see anything break when I removed it. Please advise. 

I also change the phylo tree list to be ordered by most recent first. This was to help me in testing. I'm not sure why there was no order before.

## Type of change

# Test and Reproduce

1. Open a page such as  http://localhost:3000/samples/12303?pipeline_version=3.3&background_id=26&search_taxon_id=0&searchKey=&rows_passing_filters=80&rows_total=104&pagesRendered=1&sort_by=nt_aggregatescore&name_type=Scientific%20name&rendering=false&loading=false&countType=NT&readSpecificity=1&treeMetric=aggregatescore&phyloTreeModalOpen=true&minContigSize=8&view=tree
2. See non-default values loaded
3. Change all controls
4. See URL update accordingly

# Checklist:

- [X] I have run through the testing script to make sure current functionality is unchanged
- [X] I have done relevant tests that prove my fix is effective or that my feature works
- [X] I have spent time testing out edge cases for my feature
- [X] I have updated the test script or pull request template if necessary
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/app/assets/src/api/index.js
+++ b/app/assets/src/api/index.js
@@ -1,6 +1,6 @@
 // TODO(mark): Split this file up as more API methods get added.
 import axios from "axios";
-import { toPairs } from "lodash/fp";
+import { toPairs, pickBy } from "lodash/fp";
 import { cleanFilePath } from "~utils/sample";
 
 const postWithCSRF = async (url, params) => {
@@ -49,10 +49,12 @@ const deleteWithCSRF = url =>
     }
   });
 
-const getURLParamString = params =>
-  toPairs(params)
+const getURLParamString = params => {
+  const filtered = pickBy((v, k) => typeof v !== "object", params);
+  return toPairs(filtered)
     .map(pair => pair.join("="))
     .join("&");
+};
 
 const getSampleMetadata = (id, pipelineVersion) => {
   return get(
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -34,6 +34,7 @@ import LoadingLabel from "./ui/labels/LoadingLabel";
 import HoverActions from "./views/report/ReportTable/HoverActions";
 import { getSampleReportInfo, getSummaryContigCounts } from "~/api";
 import { pipelineVersionHasAssembly } from "./utils/sample";
+import queryString from "query-string";
 
 const DEFAULT_MIN_CONTIG_SIZE = 4;
 const HUMAN_TAX_IDS = [9605, 9606];
@@ -163,6 +164,12 @@ class PipelineSampleReport extends React.Component {
       minContigSize: cachedMinContigSize || DEFAULT_MIN_CONTIG_SIZE
     };
 
+    this.state = {
+      ...this.state,
+      // Override from the URL
+      ...this.parseUrlParams()
+    };
+
     this.expandAll = false;
     this.expandedGenera = [];
     this.thresholded_taxons = [];
@@ -179,6 +186,26 @@ class PipelineSampleReport extends React.Component {
 
   componentDidMount() {
     this.scrollDown();
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    // Set the state in the URL
+    this.props.refreshPage(this.state, false);
+  }
+
+  // See also parseUrlParams in SamplesHeatmapView
+  parseUrlParams() {
+    let urlParams = queryString.parse(location.search, {
+      arrayFormat: "bracket"
+    });
+    for (var key in urlParams) {
+      try {
+        urlParams[key] = JSON.parse(urlParams[key]);
+      } catch (e) {
+        // pass
+      }
+    }
+    return urlParams;
   }
 
   fetchSearchList = () => {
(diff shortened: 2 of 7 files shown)


jshoe:
I think the reason for the full-page reloads used to be because the page didn't have the async call for report data (since the report scores you see are affected by the pipeline version and background id). 

Also I think resetUrl() was there because someone wanted the url to be cleaner but I like it with an ID and I also like the ordering you put.

gregdingle:
> I think the reason for the full-page reloads ...
Ok, I will change to async loads.

> Also I think resetUrl() was there because someone wanted the url to be cleaner but I like it with an ID and I also like the ordering you put.
Ok good.

> Re params there are some comments above fillUrlParams but I believe we want URL params to overrule saved cookie settings

Thanks for @cdebourcy for bringing this up.

The way cookies are being used is not intuitive I believe, and it will conflict with saved visualizations. What would you expect to happen if 

1) you save a viz
2) you customize a second viz
3) you reload the first viz

Will the cookie state override the save state? 

Scenario #2:

1) you customize a viz
2) you save it
3) you create a new viz with the same samples

Does the new viz appear the same as the saved viz? Or does it start fresh?

Another issue with the cookie state is that it does not appear to be namespaced, or cover all params, which will lead to hard to debug problems down the road. 

I recommend removing the cookie state entirely because the URL and saved state addresses the same need for persistence. If there is an additional need for per-user or per-project defaults, I think those should be asked for explicitly. 

Thoughts?

jshoe:
^ The reason for the cookie states on the report page at least is so that a user can set them once and then browse quickly through / compare other reports with the same settings. I think @eggerdesigns was against removing them in the past, but I do agree they complicate the UI

gregdingle:
> set them once and then browse quickly through / compare other reports with the same settings

For this I recommend an explicit user preference, either inline "Save current settings as default" , or in profile "Default X for viewing reports".

jshoe:
cc @happyimadesignr

cdebourcy:
These are all really good points. I think the desired behavior is the following:
- saved state should always override cookie state
- params explicitly specified in shareable URL should always override cookie state
- default view of *sample report*, when no params are specified in a saved state or the url, should apply cookie state from last/current session. Note, I think this only applies to the sample report view (people might be upset if that current behavior goes away).
I don’t mean to overcomplicate development here - feel free to find a compromise that works - but the above is the ideal case behavior

gregdingle:
> * saved state should always override cookie state
> * default view of _sample report_, when no params are specified in a saved state or the url, should apply cookie state from last/current session

These two rules will lead to behavior where:
* create a new report
* customize the report 
* save the report 
* customize the report more
* save the report again
* open a new report
* see state matches first report when it was first saved

Is that what we want? I think most people would expect the new report state to either match the default (no cookie) or match the last known state, not something in between. (And this is ignoring the issue that the cookie state is a subset of the URL/save state.)

cdebourcy:
> > * saved state should always override cookie state
> > * default view of _sample report_, when no params are specified in a saved state or the url, should apply cookie state from last/current session
> 
> These two rules will lead to behavior where:
> 
> * create a new report
> * customize the report
> * save the report
> * customize the report more
> * save the report again
> * open a new report
> * see state matches first report when it was first saved
> 
> Is that what we want? I think most people would expect the new report state to either match the default (no cookie) or match the last known state, not something in between. (And this is ignoring the issue that the cookie state is a subset of the URL/save state.)

More good observations. At this point, I defer to Jenn and your judgment!

tfrcarvalho:
> > * saved state should always override cookie state
> > * default view of _sample report_, when no params are specified in a saved state or the url, should apply cookie state from last/current session
> 
> These two rules will lead to behavior where:
> 
> * create a new report
> * customize the report
> * save the report
> * customize the report more
> * save the report again
> * open a new report
> * see state matches first report when it was first saved
> 
> Is that what we want? I think most people would expect the new report state to either match the default (no cookie) or match the last known state, not something in between. (And this is ignoring the issue that the cookie state is a subset of the URL/save state.)

I believe, as I think you mentioned before, now that we will have ways to save reports, we need to make "save defaults" an explicit action as opposed to the "hidden" cookies we are using now. We should address this later.

gregdingle:
I'm merging as-is and we can decide on the desired interaction with save once the save button appears in #1936

PR #1931, opened 2019-02-13
Add new visualizations route, controller and active record

gregdingle:
# Description

image

This adds basic save and view functionality for heatmaps. It moves existing logic from the samples controller to a new helper. In a future PR, I will make the visualization controller handle phylo trees and taxon trees as well. 

And in another PR, I will add the heatmap data to the search index and search results so that the user can actually do something with the saved heatmap.

## Type of change

# Test and Reproduce

1. Go to a heatmap such as http://localhost:3000/samples/heatmap?background=26&dataScaleIdx=0&metric=NT.aggregatescore&readSpecificity=1&sampleIds[]=10399&sampleIds[]=4540&species=1&subcategories=%7B%7D&taxonsPerSample=30&thresholdFilters=%5B%5D
2. Click save
3. Open by visualization page by new ID. For example http://localhost:3000/visualizations/heatmap/3 . 
4. See same viz as in first page. 

# Checklist:

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapView.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView.jsx
@@ -12,8 +12,7 @@ import DetailsSidebar from "~/components/common/DetailsSidebar";
 import { Divider, NarrowContainer, ViewHeader } from "~/components/layout";
 import SequentialLegendVis from "~/components/visualizations/legends/SequentialLegendVis.jsx";
 import Slider from "~ui/controls/Slider";
-import Icon from "~ui/icons/Icon";
-import { PrimaryButton, SecondaryButton } from "~ui/controls/buttons";
+import { SaveButton, ShareButton } from "~ui/controls/buttons";
 import {
   Dropdown,
   DownloadButtonDropdown,
@@ -721,24 +720,18 @@ class SamplesHeatmapView extends React.Component {
             <ViewHeader.Controls className={cs.controls}>
               <Popup
                 trigger={
-                  <PrimaryButton
-                    text="Share"
+                  <ShareButton
                     onClick={this.onShareClick}
                     className={cs.controlElement}
-                    icon={
-                      <Icon size="large" className={"cloud share alternate"} />
-                    }
                   />
                 }
                 content="A shareable URL has been copied to your clipboard!"
                 on="click"
                 hideOnScroll
               />
-              <SecondaryButton
-                text="Save"
+              <SaveButton
                 onClick={this.onSaveClick}
                 className={cs.controlElement}
-                icon={<Icon size="large" className={"cloud save alternate"} />}
               />
               <DownloadButtonDropdown
                 className={cs.controlElement}
--- a/app/assets/src/components/views/compare/SamplesHeatmapView.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView.jsx
@@ -12,7 +12,8 @@ import DetailsSidebar from "~/components/common/DetailsSidebar";
 import { Divider, NarrowContainer, ViewHeader } from "~/components/layout";
 import SequentialLegendVis from "~/components/visualizations/legends/SequentialLegendVis.jsx";
 import Slider from "~ui/controls/Slider";
-import { PrimaryButton } from "~ui/controls/buttons";
+import Icon from "~ui/icons/Icon";
+import { PrimaryButton, SecondaryButton } from "~ui/controls/buttons";
 import {
   Dropdown,
   DownloadButtonDropdown,
@@ -20,7 +21,7 @@ import {
   MultipleNestedDropdown
 } from "~ui/controls/dropdowns";
 import { processMetadata } from "~utils/metadata";
-import { getSampleTaxons, getSampleMetadataFields } from "~/api";
+import { getSampleTaxons, getSampleMetadataFields, saveHeatmap } from "~/api";
 import cs from "./samples_heatmap_view.scss";
 import SamplesHeatmapVis from "./SamplesHeatmapVis";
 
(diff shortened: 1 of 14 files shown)


db/schema.rb
```diff
@@ -117,8 +114,8 @@
   create_table "host_genomes_metadata_fields", id: false, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
     t.bigint "host_genome_id", null: false
     t.bigint "metadata_field_id", null: false
-    t.index ["host_genome_id", "metadata_field_id"], name: "index_host_genomes_metadata_fields"
-    t.index ["metadata_field_id", "host_genome_id"], name: "index_metadata_fields_host_genomes"
+    t.index ["host_genome_id", "metadata_field_id"], name: "index_host_genomes_metadata_fields", unique: true
```
gregdingle: Not sure what happened here... what's the best practice for resetting to the master branch migration before applying my own?

app/assets/src/components/views/compare/SamplesHeatmapView.jsx
```diff
@@ -715,12 +725,21 @@ class SamplesHeatmapView extends React.Component {
...
+                    icon={
+                      <Icon size="large" className={"cloud share alternate"} />
+                    }
                   />
                 }
                 content="A shareable URL has been copied to your clipboard!"
                 on="click"
                 hideOnScroll
               />
+              <SecondaryButton
```
tfrcarvalho: Since we are going to have multiple of these buttons, can you create a `SaveButton` and `ShareButton` to encapsulate these?

db/schema.rb
tfrcarvalho: Sorry, no best practice yet :) This has been a recurring that we need to solve, probably with the initial db config.

You might have to remove it by hand... or reset and copy over.

app/helpers/heatmap_helper.rb
```diff
@@ -0,0 +1,66 @@
...
+    vis.sample_ids = @data[:sampleIds]
+
+    render json: {
+      status: "success",
+      message: "Heatmap saved successfully",
+      data: @data
+    }
+  rescue => err
+    render json: {
+      status: 'failed',
```
tfrcarvalho: nit: `'` -> `"`

app/helpers/heatmap_helper.rb
```diff
@@ -0,0 +1,66 @@
...
+    @sample_taxons_dict = sample_taxons_dict(params)
+    output_csv = generate_heatmap_csv(@sample_taxons_dict)
+    send_data output_csv, filename: 'heatmap.csv'
+  end
+
+  def save_heatmap
+    @data = params[:heatmapParams]
+    vis = Visualization.create(
+      user: current_user,
+      data: @data
```
tfrcarvalho: Should we remove the sampleIds from the data?

app/controllers/samples_controller.rb
```diff
@@ -952,7 +909,12 @@ def sample_taxons_dict(params)
                         else
                           JSON.parse(params[:thresholdFilters] || "[]")
                         end
-    include_phage = (JSON.parse(params[:subcategories]) || {}).fetch("Viruses", []).include?("Phage")
+    subcategories = if params[:subcategories].respond_to?(:to_h)
+                      params[:subcategories].permit!.to_h
```
tfrcarvalho: Sorry, do not know enough about this... what is this branch doing ?

tfrcarvalho: I got it... i guess now the parameter is not necessary a string, and we were allowing all keys already.

app/helpers/heatmap_helper.rb
```diff
@@ -0,0 +1,66 @@
...
+    )
+    vis.sample_ids = @data[:sampleIds]
```
cdebourcy: Does this actually save the vis<->samples association? I would have assumed you would need to do ‘vis.update(sample_ids: ...)’ (in which case, better to add it to the create operation), but I may be missing something :)

app/assets/src/components/views/compare/SamplesHeatmapView.jsx
cdebourcy: Seconded, especially given we’re going to have to change the look of these buttons in the upcoming design change

db/migrate/20190208233321_create_visualizations.rb
```diff
@@ -0,0 +1,19 @@
+class CreateVisualizations < ActiveRecord::Migration[5.1]
+  def change
+    create_table :visualizations do |t|
+      t.references :user, foreign_key: true
```
cdebourcy: In general, we have been avoiding foreign keys. We removed all of them a while ago because of database locking issues and are now just enforcing them in the code where necessary. I’d recommend the same here

gregdingle: ok good to know!

app/helpers/heatmap_helper.rb
gregdingle: Yes it does work... active record magic! ✨

app/controllers/samples_controller.rb
gregdingle: exactly

app/helpers/heatmap_helper.rb
gregdingle: The sampleIds and the data should always be replaced together, but to enforce this, I will add a check at read-time, as I would like keep the data object the same in the db and in javascript.

app/helpers/heatmap_helper.rb
gregdingle: is this in our lint rules anywhere?

db/schema.rb
gregdingle: To be clear, I should just copy over the lines related to my new tables and ignore any changes to other tables?

tfrcarvalho: Those are enough, yes. You can leave the timestamp on top too.

app/helpers/heatmap_helper.rb
tfrcarvalho: Yes: https://github.com/chanzuckerberg/idseq-web/blob/master/DEV_GUIDELINES.md) 
But in this case, it is more clear because it is not consistent with only a few lines above...

app/helpers/heatmap_helper.rb
tfrcarvalho: I would just do the conversion instead of keeping two copies...

PR #1932, opened 2019-02-13
[Metadata] Allow CSV submission by display_name (in addition to regular name)

jshoe:
- Also fix an issue where deleting cells from the template in Excel leaves a row of empty strings which causes errors like "•Missing sample_name (row 2), •Missing sample_name (row 3), •Missing sample_name (row 4)...".

#### Before
screen shot 2019-02-12 at 5 36 10 pm

#### After
feb-12-2019 17-33-08
feb-12-2019 17-33-31

### DIFF
--- a/app/helpers/metadata_helper.rb
+++ b/app/helpers/metadata_helper.rb
@@ -70,8 +70,6 @@ module MetadataHelper
     # Verify that the column names are supported.
     metadata["headers"].each_with_index do |header, index|
       # Check for matching MetadataField or the sample_name/host_genome_name
-      puts "foobar 3:59pm"
-      puts header
       unless header == "sample_name" || MetadataField.find_by(name: header) || MetadataField.find_by(display_name: header)
         errors.push(MetadataValidationErrors.column_not_supported(header, index + 1))
       end
@@ -114,7 +112,7 @@ module MetadataHelper
         field = metadata["headers"][col_index]
 
         # Ignore invalid columns.
-        if field != "sample_name" && MetadataField.find_by(name: field)
+        if field != "sample_name" && (MetadataField.find_by(name: field) || MetadataField.find_by(display_name: field))
           issues = sample.metadatum_validate(field, value)
 
           issues[:errors].each do |error|
--- a/app/helpers/metadata_helper.rb
+++ b/app/helpers/metadata_helper.rb
@@ -112,7 +112,7 @@ module MetadataHelper
         field = metadata["headers"][col_index]
 
         # Ignore invalid columns.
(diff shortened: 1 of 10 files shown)


jshoe:
^ need to fix tests

app/helpers/metadata_helper.rb
```diff
@@ -107,7 +110,7 @@ def validate_metadata_csv_for_samples(samples, metadata)
         field = metadata["headers"][row_index]
 
         # Ignore invalid columns.
-        if field != "sample_name" && MetadataField.find_by(name: field)
+        unless field == "sample_name" || MetadataField.find_by(name: field) || MetadataField.find_by(display_name: field)
```
MarkAZhang: Should this condition be `unless field == "sample_name" || (!MetadataField.find_by(name: field) && !MetadataField.find_by(display_name: field))`?

MarkAZhang:
It looks like you might still be working on this? (the last commit says WIP) Let me know when you're ready.

app/helpers/metadata_helper.rb
jshoe: thanksss

app/models/sample.rb
```diff
@@ -530,8 +530,15 @@ def metadatum_add_or_update(key, val)
...
-      m.metadata_field = MetadataField.find_by!(name: key.to_s)
+      if MetadataField.find_by(name: key.to_s)
+        m.metadata_field = MetadataField.find_by(name: key.to_s)
+      elsif MetadataField.find_by(display_name: key.to_s)
+        m.metadata_field = MetadataField.find_by(display_name: key.to_s)
+        # If matched based on display_name, set key to regular name
+        m.key = m.metadata_field.name
+      else
+        raise ActiveRecord::RecordNotFound("No matching field for #{key}")
+      end
```
MarkAZhang: There's some duplication between here and the code in `set_validated_values`. Would it make sense to just set the metadata_field in `set_validated_values` and do `errors.add(...)` if a `metadata_field` doesn't exist? And remove this part entirely.

I guess we eventually want to remove `key` from the Metadatum sample, so maybe it does make sense to do it this way. In which case, we should add a `TODO` to remove the code in `set_validated_values`. Eventually, `metadata_field` MUST be set for Metadatum prior to saving, since there will no longer be a key.

jshoe: @MarkAZhang I think it's good set metadata_field at the point Metadatum.new is called and remove key eventually, so I reworked this a little and removed that from set_validated_values. I checked in staging and prod that all current Metadata entries have a metadata_field.

PR #1933, opened 2019-02-13
[Uploads] Fix local uploads

jshoe:
- All on non-admin account

#### Local single upload
feb-13-2019 10-42-02

#### S3 single upload
feb-13-2019 10-42-22

#### Local batch upload
feb-13-2019 10-43-04

#### S3 batch upload
feb-13-2019 10-42-45


cdebourcy:
Awesome

PR #1936, opened 2019-02-14
Add saving of table or taxon from the report page

gregdingle:
# Description

This PR builds on #1930 and #1931 to add saving of table or taxon from the report page. (Sorry, #1930 
 changes are repeated here. Once that merges, they will go away.) 

image

In the next PR, I will add a way to load those saved visualizations. (It's not clear whether they should be loaded on the same multipurpose page or each on a new standalone /visualizations/type/ page.)

## Type of change

# Test and Reproduce

1. go to sample report page
2. hit save button
3. see "success"
4. make sure heatmaps save and load still work
5. try "share" button

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [X] Report Page
# Checklist:

- [X] I have done relevant tests that prove my fix is effective or that my feature works
- [X] I have spent time testing out edge cases for my feature
- [X] New and existing unit tests pass locally with my changes


tfrcarvalho:
This is not a PR to master...

app/assets/src/api/index.js
```diff
@@ -49,10 +50,27 @@ const deleteWithCSRF = url =>
...
-  toPairs(params)
+const getURLParamString = params => {
+  const filtered = pickBy((v, k) => typeof v !== "object", params);
+  return toPairs(filtered)
     .map(pair => pair.join("="))
     .join("&");
+};
+
+// See also parseUrlParams in SamplesHeatmapView
+const parseUrlParams = () => {
```
tfrcarvalho: This is not exactly an API function and is not used from within api. 
I see the `getURLParamString` was not either. Do you mind moving this to the `utils` folder?
Maybe in a `url.js` file or something.

app/assets/src/components/views/SampleView/Controls.jsx
```diff
@@ -85,7 +85,8 @@ Controls.propTypes = {
   reportDetails: PropTypes.ReportDetails,
   reportPageParams: PropTypes.shape({
     pipeline_version: PropTypes.string,
-    background_id: PropTypes.string
+    // TODO (gdingle): standardize on string or number
+    background_id: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
```
tfrcarvalho: number 👍

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -652,30 +632,27 @@ class PipelineSampleReport extends React.Component {
         }
       },
       () => {
+        // TODO (gdingle): do we really want to reload the page here?
```
tfrcarvalho: What is the alternative?

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -16,6 +21,9 @@ import Tabs from "~/components/ui/controls/Tabs";
 import DetailsSidebar from "~/components/common/DetailsSidebar";
 import Controls from "./Controls";
 import PipelineVersionSelect from "./PipelineVersionSelect";
+import { SaveButton, ShareButton } from "~ui/controls/buttons";
+import { Popup } from "semantic-ui-react";
```
tfrcarvalho: Use our Popup wrapper in ui. 
As a common practice, we wrap all elements with our own. This allows us to add custom styling and some common functionality to the elements.
The Popup from semantic ui has other issues as, because of the way it is used, always renders with the whole page (the impact depends on where you use it).

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -66,19 +74,30 @@ class SampleView extends React.Component {
...
     );
-    window.location =
+    const url =
       location.protocol +
       "//" +
       location.host +
       location.pathname +
       "?" +
       getURLParamString(newParams);
+    if (reload) {
```
tfrcarvalho: Seems like this snippet can be moved to a utils file too, no? 
You used it in multiple places, right?

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -268,6 +288,17 @@ class SampleView extends React.Component {
     return {};
   };
 
+  // TODO (gdingle): refactor with onShareClick in SamplesHeatmapView?
+  onShareClick = () => {
```
tfrcarvalho: The onshare will use the new id to a visualization in the future, right?

app/controllers/visualizations_controller.rb
```diff
@@ -3,20 +3,47 @@ class VisualizationsController < ApplicationController
...
     @heatmap_data = heatmap
 
     id = params[:id]
     if id
       vis = Visualization.find(id)
+
+      # TODO: (gdingle): does this makes sense for non-heatmap?
       sample_ids = vis.data['sampleIds']
-      if sample_ids != vis.sample_ids
+      if vis.visualization_type == "heatmap" && sample_ids != vis.sample_ids
```
tfrcarvalho: Are these array guaranteed to be sorted? 
Also, again, single source of data mean you do not worry about this.

app/controllers/visualizations_controller.rb
```diff
@@ -3,20 +3,47 @@ class VisualizationsController < ApplicationController
...
-    # type = params[:type]
-
+    # TODO: (gdingle): refactor off of heatmap
     @heatmap_data = heatmap
 
     id = params[:id]
     if id
       vis = Visualization.find(id)
+
+      # TODO: (gdingle): does this makes sense for non-heatmap?
```
tfrcarvalho: Yes, we plan to have more visualizations with multiple samples (e.g. world map).
Phylo tree visualizations (possibly with some refactoring) will also need customization and be related to multiple samples.

app/assets/src/components/PipelineSampleReport.jsx
gregdingle: The alternative is to change only the URL with history.pushState. 

Another question about this call is why only set the background_id? Why not keep all current param values?

app/assets/src/components/views/SampleView/SampleView.jsx
gregdingle: Ok good to know!

app/assets/src/components/views/SampleView/SampleView.jsx
gregdingle: This is the only place for now. But yes in general it would be good to decide how we want URL history to work everywhere and rely on a single function for it.

app/assets/src/components/views/SampleView/SampleView.jsx
gregdingle: I'm not sure. I assumed no, because you wouldn't normally expect "share" to have any side effects.

app/controllers/visualizations_controller.rb
gregdingle: 👍

app/controllers/visualizations_controller.rb
gregdingle: Good question. I assumed they would be sorted because they were assigned through ActiveRecord.

app/assets/src/components/views/SampleView/SampleView.jsx
tfrcarvalho: Not what I was thinking... but see is not very clear.
So, good for now and lets discuss with Jenn what is the behavior we want.

app/controllers/visualizations_controller.rb
tfrcarvalho: I think they normally are, but because one of them comes from the client... it will be hard to guarantee it completely.
Maybe sorting to be sure would not be a bad idea... We do not expect to have more than a few hundred samples.

Remind me again why do we keep two copies?

gregdingle: I changed this to keep only a single copy in the following PR. ☺️

PR #1938, opened 2019-02-14
[MetadataFields] Update our field dictionary

jshoe:
- Make updates resulting from more discussion. See https://czi.quip.com/a1VwAorpnu9x/Baseline-Metadata-Schema for the latest table

#### Example
screen shot 2019-02-14 at 2 24 46 pm

### DIFF
--- a/db/migrate/20190214210719_update_metadata_fields.rb
+++ b/db/migrate/20190214210719_update_metadata_fields.rb
@@ -66,6 +66,10 @@ class UpdateMetadataFields < ActiveRecord::Migration[5.1]
     #########################
     # Modify existing fields
     #########################
+
+    # Replace 'participant' with 'host' in the description
+    MetadataField.find_each { |field| field.update(description: field.description.gsub("participant", "host")) }
+
     mf = MetadataField.find_by(name: "participant_id")
     mf.update(name: "host_id", display_name: "Host ID") if mf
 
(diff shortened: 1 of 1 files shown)


db/migrate/20190214210719_update_metadata_fields.rb
```diff
@@ -0,0 +1,184 @@
...
+    MetadataField.find_each { |field| field.update(description: field.description.gsub("participant", "host")) }
+
+    mf = MetadataField.find_by(name: "participant_id")
+    mf.update(name: "host_id", display_name: "Host ID") if mf
+
+    mf = MetadataField.find_by(name: "blood_fed")
+    hg = HostGenome.find_by(name: "Tick")
+    mf.host_genomes << hg if mf && hg
+
+    mf = MetadataField.find_by(name: "life_stage")
```
MarkAZhang: We should rename "Race/Ethnicity" to "Host Race/Ethnicity"

db/migrate/20190214210719_update_metadata_fields.rb
```diff
@@ -0,0 +1,184 @@
...
+
+    mf = MetadataField.find_by(name: "sex")
+    mf.update(name: "host_sex", display_name: "Host Sex", description: "Sex of the host") if mf
+
+    mf = MetadataField.find_by(name: "genus_species")
+    mf.update(name: "host_genus_species", display_name: "Host Genus Species") if mf
+
+    # Assign all the fields on Cat to Pig (non-human defaults)
+    if HostGenome.find_by(name: "Pig") && HostGenome.find_by(name: "Cat")
+      HostGenome.find_by(name: "Pig").metadata_fields << HostGenome.find_by(name: "Cat").metadata_fields
```
MarkAZhang: Can you do `HostGenome.find_by(name: "Pig").metadata_fields << (HostGenome.find_by(name: "Cat").metadata_fields - HostGenome.find_by(name: "Pig").metadata_fields)` just to be safe?

If Pig already has some of Cat's metadata_fields, I think this will throw an error.

I actually already added Cat's metadata_fields to Pig in prod. Happy to remove (and happy to do a migration next time if we think that's best)

db/migrate/20190214210719_update_metadata_fields.rb
jshoe: 👍

db/migrate/20190214210719_update_metadata_fields.rb
jshoe: 👍

PR #1939, opened 2019-02-14
Make sample report page load last saved viz

gregdingle:
# Description

This closes the loop so that a user will see what they expect when they save a report and go back to it from the samples page. The help text is "The current view will be saved as the default for this sample".

image

* Url params will override the saved state
* The saved state will update the URL params so that the saved report can be shared as a URL
* Interaction with cookie state is untouched and unspecified -- we will need to sync on this as discussed
* We keep the whole history of saves but only the last one is ever loaded -- not sure if we want to do anything more about history -- or whether it should work the same for heatmaps
* the last saved report is per sample, so multiple users could overwrite each other's saves -- not sure what is required here

Also, I changed a few lines of the heatmap code to make it more generic. 

## Type of change

# Test and Reproduce

1. load a report page
2. click on taxon tree
3. save 
4. reload page without any url params
5. see saved state
6. click on table view
7. reload with url params
8. see table view

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [X] Report Page

# Checklist:

- [] I have run through the testing script to make sure current functionality is unchanged

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -483,8 +483,23 @@ class SamplesController < ApplicationController
       @ercc_comparison = @pipeline_run.compare_ercc_counts
     end
 
+    viz = last_saved_visualization
+    @saved_param_values = viz ? viz.data : {}
+
     tags = %W[sample_id:#{@sample.id} user_id:#{current_user.id}]
     MetricUtil.put_metric_now("samples.showed", 1, tags)
+  end
+
+  def last_saved_visualization
+    valid_viz_types = ['tree', 'table'] # See PipelineSampleReport.jsx
+    Sample
+      .includes(:visualizations)
+      .find(@sample.id)
+      .visualizations
+      .where(user: current_user)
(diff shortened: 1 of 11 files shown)


gregdingle:
Just added something else for saving heatmaps: update the URL on save, so that click save then click share works as expected.

app/controllers/samples_controller.rb
```diff
@@ -483,10 +483,25 @@ def show
...
+  def last_saved_visualization
+    valid_viz_types = ['tree', 'table'] # See PipelineSampleReport.jsx
+    Sample
+      .includes(:visualizations)
+      .find(@sample.id)
+      .visualizations
+      .where('visualizations.visualization_type IN (?)', valid_viz_types)
+      .order('visualizations.updated_at desc')
+      .limit(1)[0]
+    # TODO: (gdingle): show save per user? where user.id == user_id ?
```
tfrcarvalho: Didnt we decide to do it per user?

PR #1940, opened 2019-02-15
Rewrite Sample Upload Page with new design.

MarkAZhang:
Modify axios api to automatically unwrap errors (by doing error.response.data).

Add new Radio Button component.

Also add ProjectCreationForm and ProjectSelect common components.

Prioritize prefix matches when searching in Dropdown.

Move host genome selection into metadata upload.
Update validate endpoint to handle host genome inside metadata.

Refactor validate_sample_names to handle multiple names.

Add more validation errors to validate_metadata_csv_for_samples,
including duplicate sample rows, missing sample rows, and missing/invalid
host genomes.

Consolidate handling of remote and local samples in SampleUploadFlow, so we just
pass around samples instead of localSampleNameToFiles and remoteSamples.

------

Basic View:
screen shot 2019-02-14 at 5 05 16 pm

Project Select:
screen shot 2019-02-14 at 5 05 23 pm

Create Project Form:
screen shot 2019-02-14 at 5 05 30 pm

Upload from S3 Option:
screen shot 2019-02-14 at 5 05 40 pm

Local File Upload Table:
screen shot 2019-02-14 at 5 05 49 pm

Remote File Upload Table:
screen shot 2019-02-14 at 5 06 07 pm

Some new validation errors:
screen shot 2019-02-14 at 1 20 59 pm

Testing strategy:
Verified that new sample upload flow works.
Verified that existing single and bulk upload flows still work.
Verified that upload metadata to project still works.
All Rails test pass.

### DIFF
--- a/app/assets/src/components/ui/controls/BulkSampleUploadTable.jsx
+++ b/app/assets/src/components/ui/controls/BulkSampleUploadTable.jsx
@@ -1,7 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import DataTable from "../../visualizations/table/DataTable";
-import { isEmpty } from "lodash/fp";
+import { isEmpty, concat, size } from "lodash/fp";
 import RemoveIcon from "~/components/ui/icons/RemoveIcon";
 import LoadingIcon from "~/components/ui/icons/LoadingIcon";
 import CheckmarkIcon from "~/components/ui/icons/CheckmarkIcon";
@@ -12,7 +12,12 @@ import cs from "./bulk_sample_upload_table.scss";
 // bulk upload interactions
 class BulkSampleUploadTable extends React.Component {
   render() {
-    const { sampleNamesToFiles, fileNamesToProgress, onRemoved } = this.props;
+    const {
+      sampleNamesToFiles,
+      fileNamesToProgress,
+      onRemoved,
+      hideProgressColumn
+    } = this.props;
 
     if (isEmpty(sampleNamesToFiles)) return null;
 
@@ -56,7 +61,11 @@ class BulkSampleUploadTable extends React.Component {
         }
       }
 
-      const filesList = <div>{files.map(f => <div>{f.name}</div>)}</div>;
+      const filesList = (
+        <div>
+          {files.map(f => <div key={f.source}>{f.name || f.source}</div>)}
+        </div>
+      );
 
       const entry = {
         progress: progress,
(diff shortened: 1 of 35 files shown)


app/assets/src/components/ui/controls/BulkSampleUploadTable.jsx
```diff
@@ -69,18 +78,31 @@ class BulkSampleUploadTable extends React.Component {
 
     return (
       <div className={cs.bulkSampleUploadTable}>
-        <div className={cs.detectedMsg}>
-          These files were detected and matched. Remove files you do not want to
-          upload:
-        </div>
+        {this.props.showCount ? (
```
jshoe: it's weird that the GitHub highlighting broke around here but I don't see a bug..

app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx
```diff
@@ -37,34 +37,32 @@ class ReviewStep extends React.Component {
...
         // For example, some samples may have successfully saved, but not others. Should explain to user.
-        .catch(error => {
+        .catch(() => {
           this.setState({
             submitState: "review",
-            errorMessage:
-              get("data.status", error) ||
-              "Unable to process sample(s), " +
-                "ensure sample is not a duplicate in the selected project"
+            errorMessage: "There were some issues creating your samples"
```
jshoe: hmm I still think it'd be helpful to see the error? or at least console.log?

cdebourcy:
Cool! Please make sure the following are thoroughly tested:
- project creation still works
- uploads work for both admin and non-admin users
- uploads through the CLI still work, especially where they create a new project (since there's a change to projects_controller)

cdebourcy:
Also, please test both local uploads and S3 uploads for each flow :)

app/controllers/projects_controller.rb
```diff
@@ -42,6 +43,15 @@ def index
         @projects = current_power.projects
       end
       format.json do
+        only_updatable = ActiveModel::Type::Boolean.new.cast(params[:onlyUpdatable])
+
+        # TODO(mark): Reconcile this with the part below.
+        # These returned projects contain different fields than the ones below.
+        if only_updatable
+          render json: current_power.updatable_projects
+          return
```
cdebourcy: Should we apply the "Do One Thing and Do It Well" principle to these controller actions? Rather than having the early return here, it might make sense to have a separate index action for uploads

app/assets/src/components/ui/controls/BulkSampleUploadTable.jsx
MarkAZhang: Weird...

PR #1942, opened 2019-02-15
[Metadata] Create a URL for seeing UploadInstructions by itself

jshoe:
- Seemed cleaner for the CLI to provide a URL as with the instructions and CSV template.


MarkAZhang:
This is great! Couple of styling comments:
- Can you use `size="small"` on the NarrowContainer? The full width seems too wide.
- Can you add `margin-top: 50px` to the `cs.uploadInstructions` div? You'll need `cx(cs.uploadInstructions, this.props.standalone && cs.standalone)` so you only add margin-top on standalone.

We can do more work to standardize pages like this, for example making sure pages like Privacy Policy and Metadata Dictionary have the same header style, but this is good for now.

PR #1944, opened 2019-02-20
Save taxon collapse state

gregdingle:
# Description

image

This adds the collapsed state of the taxon viz to the URL so that the same-looking viz will re-appear on share or save. 

For example:
http://localhost:3000/samples/12302?pipeline_version=3.3&background_id=26&10239_-650=c&2_-650=c&view=tree

## Type of change

# Test and Reproduce

1. open a taxon viz
2. click on a node to collapse it
3. see the update of the url 
4. reload the page
5. see the same viz
6. un-collapse the node
7. reload and see the same viz

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [x] Report Page

### DIFF
--- a/app/assets/src/components/views/TaxonTreeVis.jsx
+++ b/app/assets/src/components/views/TaxonTreeVis.jsx
@@ -57,7 +57,8 @@ class TaxonTreeVis extends React.Component {
         onNodeHover: this.handleNodeHover,
         onNodeLabelClick: this.handleNodeLabelClick,
         onCreatedTree: this.fillNodeValues,
-        tooltipContainer: this.treeTooltip
+        tooltipContainer: this.treeTooltip,
+        persistCollapsedInUrl: true
       }
     );
     this.treeVis.update();
@@ -85,19 +86,13 @@ class TaxonTreeVis extends React.Component {
     }
   }
 
-  static createNode() {
-    return {
-      children: [],
-      collapsed: false
-    };
-  }
-
   handleNodeHover(node) {
     this.setState({ nodeHover: node });
   }
 
   handleNodeLabelClick(node) {
     if (!node.data.modalData) {
+      // TODO (gdingle): we should show a "no description" instead of no-op
       this.props.onTaxonClick(null);
       return;
     }
--- a/app/assets/src/components/views/TaxonTreeVis.jsx
+++ b/app/assets/src/components/views/TaxonTreeVis.jsx
@@ -58,10 +58,45 @@ class TaxonTreeVis extends React.Component {
         onNodeLabelClick: this.handleNodeLabelClick,
         onCreatedTree: this.fillNodeValues,
         tooltipContainer: this.treeTooltip,
-        persistCollapsedInUrl: true
+        onCollapsedStateChange: this.persistCollapsedInUrl,
+        collapsed: this.getCollapsedInUrl() || new Set()
       }
     );
     this.treeVis.update();
+  }
+
+  persistCollapsedInUrl(node) {
+    function hasAllChildrenCollapsed(node) {
+      return !!(!node.children && node.collapsedChildren);
+    }
+    try {
+      const href = new URL(window.location.href);
+      if (hasAllChildrenCollapsed(node)) {
+        href.searchParams.set(node.id, "c"); // 'c'ollapsed
+      } else {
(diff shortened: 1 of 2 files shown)


gregdingle:
@tfrcarvalho sorry, the commit histiory was messed up. It's cleaned up now.

app/assets/src/components/views/TaxonTreeVis.jsx
```diff
@@ -85,19 +86,13 @@ class TaxonTreeVis extends React.Component {
     }
   }
 
-  static createNode() {
```
gregdingle: I just don't think this was used anywhere

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -171,6 +179,31 @@ export default class TidyTree {
     });
   }
 
+  persistCollapsedInUrl(node) {
```
tfrcarvalho: Are we going to run into URL limits? There are a lot of nodes on this tree.

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -24,7 +24,8 @@ export default class TidyTree {
         useCommonName: false,
         minNonCollapsableChildren: 2,
         smallerFont: 8,
-        largerFont: 12
+        largerFont: 12,
+        persistCollapsedInUrl: false
```
tfrcarvalho: This visualization should be independent of the view outside,
(There are few fixes need to be done to keep it independent).
Use callbacks to handle state changes instead.

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -171,6 +179,31 @@ export default class TidyTree {
     });
   }
 
+  persistCollapsedInUrl(node) {
+    try {
```
tfrcarvalho: Here is where you want to use callbacks. This code should be on the visualization file `TaxonTreeViz`.
Here `TidyTree should only call `onCollapsedStateChange` for instance.

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -209,6 +242,9 @@ export default class TidyTree {
         y0: node.y
       })
     );
+    if (this.options.persistCollapsedInUrl) {
+      this.persistCollapsedInUrl(node);
```
tfrcarvalho: here would be `onCollapsedStateChange`...

app/assets/src/components/visualizations/TidyTree.js
tfrcarvalho: Also, the previous state should be passed in the data nodes...

app/assets/src/components/views/TaxonTreeVis.jsx
tfrcarvalho: Good catch...

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -171,6 +179,31 @@ export default class TidyTree {
...
+        href.searchParams.delete(node.id);
+      }
+      history.pushState(window.history.state, document.title, href);
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.error(e);
+    }
+  }
+
+  isCollapsedInUrl(node) {
```
tfrcarvalho: The information on collapsed nodes should come with the input node structure.

app/assets/src/components/visualizations/TidyTree.js
gregdingle: In the worst case, yes. I intentionally shortened the param names and values. I estimate we can store ~200 collapsed nodes in the URL.  I think it is acceptable to fail after that.

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -209,6 +214,7 @@ export default class TidyTree {
         y0: node.y
       })
     );
+    this.options.onCollapsedStateChange(node);
```
tfrcarvalho: Check if the function exists first: 
`this.options.onCollapsedStateChange && this.options.onCollapsedStateChange(node)`

PR #1947, opened 2019-02-20
Shorten share URLs

gregdingle:
# Description

This shortens the long URLs on the visualization pages. 

image

Most of the work is handled by the `shortener` gem. 

## Type of change

# Test and Reproduce

1. open a heatmap / report page
2. customize
3. hit share
5. paste the short url into a new tab
6. see the same vis state

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [X] Report Page

### DIFF
--- a/app/assets/src/helpers/url.js
+++ b/app/assets/src/helpers/url.js
@@ -1,5 +1,7 @@
 import QueryString from "query-string";
 import { toPairs, pickBy } from "lodash/fp";
+import { shortenUrl } from "~/api";
+import copy from "copy-to-clipboard";
 
 export const resetUrl = () => {
   // remove parameters from url
@@ -26,4 +28,10 @@ export const getURLParamString = params => {
   return toPairs(filtered)
     .map(pair => pair.join("="))
     .join("&");
+};
+
+export const copyShortUrlToClipboard = async url => {
+  url = url || window.location.href;
+  const shortUrl = await shortenUrl(url);
+  copy(window.location.origin + "/" + shortUrl.unique_key);
 };
--- a/db/migrate/20190220191436_create_shortened_urls_table.rb
+++ b/db/migrate/20190220191436_create_shortened_urls_table.rb
@@ -1,0 +1,34 @@
+class CreateShortenedUrlsTable < ActiveRecord::Migration[4.2]
+  def change
+    create_table :shortened_urls do |t|
+      # we can link this to a user for interesting things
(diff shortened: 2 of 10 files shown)


db/migrate/20190220191436_create_shortened_urls_table.rb
```diff
@@ -0,0 +1,34 @@
...
+      t.string :unique_key, limit: 10, null: false
+
+      # a category to help categorize shortened urls
+      t.string :category
+
+      # how many times the link has been clicked
+      t.integer :use_count, null: false, default: 0
+
+      # valid until date for expirable urls
+      t.datetime :expires_at
```
tfrcarvalho: Cool! How are expired urls cleaned? Is there a rake task or something for that?

app/assets/src/helpers/url.js
```diff
@@ -27,3 +29,9 @@ export const getURLParamString = params => {
     .map(pair => pair.join("="))
     .join("&");
 };
+
+export const copyShortUrlToClipboard = async url => {
+  url = url || window.location.href;
+  const shortUrl = await shortenUrl(url);
+  copy(window.location.origin + "/" + shortUrl.unique_key);
```
tfrcarvalho: Form the snapshot noticed this unique is a little short (4 chars). Can we increase the size a bit?

db/migrate/20190220191436_create_shortened_urls_table.rb
gregdingle: Expiry is optional on creation of a short url. Do you think we should add it? I didn't see any reason why.

app/assets/src/helpers/url.js
gregdingle: The default is 5 chars, which gives us 36^5 = 60M keys. I can configure it now, or we can wait until it is a problem and add more chars. What do you think?

tfrcarvalho: If the change is backwards-compatible, its fine...
But if we are not expiring I would add a 2/3 more characters.

db/migrate/20190220191436_create_shortened_urls_table.rb
tfrcarvalho: Just clean-up... I guess we can do it, if it becomes very popular.

PR #1950, opened 2019-02-20
Add default_for_new_host_genome metadata field.

MarkAZhang:
Auto-add default metadata fields when a new host genome is created.

Tested with `rake db:migrate` and `rake db:rollback` and by creating a new HostGenome and verifying that it has the default fields.

### DIFF
--- a/app/controllers/metadata_controller.rb
+++ b/app/controllers/metadata_controller.rb
@@ -14,6 +14,7 @@ class MetadataController < ApplicationController
   def instructions
   end
 
+  # TODO(mark): Factor this out into metadata_fields_controller, and add other CRUD endpoints.
   # All users get the same fields.
   def official_metadata_fields
     render json: official_metadata_fields_helper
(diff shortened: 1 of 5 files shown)


yunfangjuan:
Could you move the defaults to model. They seem to be useful in the futures too.

MarkAZhang:
You should be able to get the defaults by `MetadataField.where(default_for_new_host_genome: 1)` since they are saved in the database as a column. The hard-coded array is just to seed the values.

yunfangjuan:
What if the default changes in the future? It seems a bit hard for people to track down how the defaults are set.

MarkAZhang:
Talked to YF offline. She was concerned that MetadataFields is being managed manually (i.e. with commands in the Rails console) and this feels a bit dangerous. Having visibility into what fields are currently default_for_new_host_genomes (without needing to log into the database) would help with this, as would having a metadata field editor down the line (which uses well-tested REST endpoints).

For now, adding `default_for_new_host_genomes` to the `field_info` method so that you can view it at `/metadata/official_metadata_fields`. In a follow-up commit, will refactor so that metadata field has its own controller with create, edit, view, delete, and we no longer need to ssh into Rails console.

MarkAZhang:
Definitely.

PR #1953, opened 2019-02-21
[Metadata] Add default fields for project creation from CLI

jshoe:
- Web flow was already covered
- Might be good to additionally modify the CLI to make project creation more explicit

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -762,7 +762,8 @@ class SamplesController < ApplicationController
       project_name = params.delete(:project_name)
       project = Project.find_by(name: project_name)
       unless project
-        project = Project.create(name: project_name)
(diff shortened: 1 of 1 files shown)


app/controllers/samples_controller.rb
```diff
@@ -762,7 +762,8 @@ def create
       project_name = params.delete(:project_name)
       project = Project.find_by(name: project_name)
       unless project
-        project = Project.create(name: project_name)
+        # Web flow would already create the project so this block is for the CLI
+        project = Project.create(name: project_name, metadata_fields: MetadataField.where(is_default: 1))
```
MarkAZhang: Could we add this as a `before_create` hook in the project model instead?

jshoe: 👍 moving this to my other PR

PR #1955, opened 2019-02-22
[Metadata] Update web for CLI metadata uploads

jshoe:
- Unlock backend /samples/upload endpoints for non-admins (only if called directly)
- Allow CLI to list projects and create new projects (so that the project creation is more explicit and so we don't have to auto-create project names that don't exist)
- Accept host_genome_name from the CLI
- Add default metadata fields to new projects
- Bug fixes for "raise ActiveRecord::RecordNotFound" and "m.destroy" because it didn't seem to work for non-materialized records

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -44,7 +44,6 @@ class ProjectsController < ApplicationController
   # GET /projects
   # GET /projects.json
   def index
-    puts "index was called at least"
     respond_to do |format|
       format.html do
         # keep compatibility with old route
@@ -60,8 +59,6 @@ class ProjectsController < ApplicationController
           render json: current_power.updatable_projects
           return
         end
-
-        puts "got up to here at least"
 
         only_library = ActiveModel::Type::Boolean.new.cast(params[:onlyLibrary])
         exclude_library = ActiveModel::Type::Boolean.new.cast(params[:excludeLibrary])
@@ -390,14 +387,12 @@ class ProjectsController < ApplicationController
   end
 
   def set_project
-    puts "why is set_project being called"
     @project = projects_scope.find(params[:id])
     assert_access
   end
 
   # Never trust parameters from the scary internet, only allow the white list through.
   def project_params
-    puts "original params here #{params}"
     result = params.require(:project).permit(:name, :public_access, user_ids: [])
     result[:name] = sanitize(result[:name]) if result[:name]
     result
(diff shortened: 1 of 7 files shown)


app/controllers/projects_controller.rb
```diff
@@ -22,6 +22,14 @@ class ProjectsController < ApplicationController
   EDIT_ACTIONS = [:edit, :update, :destroy, :add_user, :all_users, :update_project_visibility].freeze
   OTHER_ACTIONS = [:create, :new, :index, :send_project_csv, :choose_project].freeze
 
+  # Required for token auth for CLI actions
+  skip_before_action :verify_authenticity_token, only: [:index, :create]
+  before_action :authenticate_user!, except: [:index, :create]
+  acts_as_token_authentication_handler_for User, only: [:index, :create], fallback: :devise
+  current_power do
```
MarkAZhang: Nit: this initialization of current_power is already in `ApplicationController`. Do we need it here again?

app/controllers/projects_controller.rb
```diff
@@ -63,7 +71,14 @@ def index
...
                    end
 
-        @projects = @samples.group(:project).count
+        @projects = if only_library || exclude_library
+                      @samples.group(:project).count
+                    else
+                      # This check is so that we still return projects without any samples in the
+                      # default case. Ex: Project listing used by the CLI.
+                      # TODO: Optimize perf if it is an issue.
+                      current_power.projects.map { |p| [p, Sample.where(project: p).count] }
```
MarkAZhang: Can you consolidate the above `only_updatable` logic with this?

Do `current_power.updatable_projects.map { |p| [p, Sample.where(project: p).count] }` if `only_updatable` is true.

This way we return a consistent object.

app/models/project.rb
```diff
@@ -140,4 +142,8 @@ def destroy
     samples.each(&:destroy)
     super
   end
+
+  def add_default_metadata_fields
```
MarkAZhang: There's a place in projects_controller `create` method where this is still being done. Can you remove that?

app/controllers/projects_controller.rb
jshoe: Weirdly enough it didn't work until I added this (and same was in samples_controller) as

```
  current_power do # Put this here for CLI
    Power.new(current_user)
  end
```

I guess it doesn't go through the same steps for API calls

app/controllers/projects_controller.rb
jshoe: 👍

app/models/project.rb
jshoe: 👍

PR #1957, opened 2019-02-22
Show visualizations list in new tabbed navigation

gregdingle:
# Description

This adds a list of saved visualizations that you can use to re-open a visualization with state. 

image

NOTE: like the other lists in the new tabs, this list still needs final visual design. For now, only admins can save a viz, and only users with the feature flag can see the lists. It's not possible in the UI yet to make a viz public.

## Type of change

# Test and Reproduce

1. open "my library"
2. see list of private saved vizs (manually assigned in rails console)
3. click link for each type of viz
4. see stateful viz

1. open "public"
2. see list of public enabled viz (manually assigned in rails console)
3. click link for each type of viz
4. see stateful viz


app/assets/src/components/views/projects/ProjectsView.jsx
```diff
@@ -110,7 +110,7 @@ class ProjectsView extends React.Component {
 }
 
 ProjectsView.defaultProps = {
-  project: []
+  projects: []
```
gregdingle: it looks like this was a bug, though harmless

app/controllers/samples_controller.rb
```diff
@@ -125,6 +125,7 @@ def index_v2
     @samples = if only_library
                  current_power.library_samples
                elsif exclude_library
+                 # TODO: (gdingle): does this actually exclude the user's library???
```
gregdingle: this is related to the Jira ticket I created for Tiago

app/assets/src/components/views/visualizations/VisualizationsView.jsx
```diff
@@ -0,0 +1,114 @@
...
+        className: cs.detailsCell,
+        cellRenderer: this.renderVisualizationDetails,
+        headerClassName: cs.detailsHeader,
+        sortFunction: p => p.created_at
+      }
+      // TODO: (gdingle): show associated samples and projects
+    ];
+  }
+
+  humanize(key) {
```
jshoe: maybe move this into a utils.js since it appeared in BaseTable as well?

db/schema.rb
```diff
@@ -344,6 +340,29 @@
     t.index ["user_id"], name: "index_samples_on_user_id"
   end
 
+  create_table "samples_visualizations", id: false, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
```
jshoe: sorry I know the staging db is messed up but could you edit the schema.rb to just have these two create_table blocks

app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -4,8 +4,9 @@ import NarrowContainer from "~/components/layout/NarrowContainer";
 import { Divider } from "~/components/layout";
 import DiscoveryHeader from "../discovery/DiscoveryHeader";
 import ProjectsView from "../projects/ProjectsView";
+import VisualizationsView from "../visualizations/VisualizationsView";
```
jshoe: for imports you may want to stick with the ~ aliases in webpack.config.common.js or define some more

gregdingle: instead of paths such as `"../visualizations/VisualizationsView"`?

jshoe: yes for ease of moving/not having relative paths, but minor/up for debate

gregdingle: ok I'll leave it for now

app/assets/src/components/views/visualizations/VisualizationsView.jsx
gregdingle: done

PR #1959, opened 2019-02-25
[Metadata] Fix current_power for CLI

jshoe:
- Left this off the other PR accidentally. Without it you still get `undefined method `admin?' for nil:NilClass` because current_power is not set properly in the control flow apparently.


jshoe:
ty

PR #1964, opened 2019-02-26
[Analytics] Only load Segment for logged in users

jshoe:
- This just loads Segment for our logged in users to "protect" the writekey a little more.
- Reasons: We're seeing "spammy hits" from Catchpoint from shared-infra which adds thousands to our "Monthly Tracked Users" quotas. And from other potential sources on the open web.
- Not really worth keeping it on the landing page / logged-in is good enough.

screen shot 2019-02-25 at 4 49 20 pm


gregdingle:
Agree we can track logged out users when we are trying to attract people to our public pages. 

It generally takes some work to filter out all the noise from public pages. If you have a list of known bots now, put it into whatever downstream filtering segment has.

PR #1966, opened 2019-02-27
Add Manual Input table for metadata upload.

MarkAZhang:
screen shot 2019-02-26 at 5 42 49 pm

screen shot 2019-02-26 at 5 42 30 pm

Testing strategy:
* Verified that the new Manual Input works for existing samples and also new local and remote samples.
* Verified that new Manual Input will properly block with an error if required metadata was not submitted.
* Verified that you can still edit metadata via the inputs in the Sidebar.
* Verified that you can still upload CSV properly for all cases.
* Verify that PhyloTree Create Wizard still looks fine despite our tweaks to Wizard.

Caveats:
* The fact that MetadataCSVUpload automatically validates when the user drops in the file, but MetadataManualInput doesn't validate until the user clicks "Continue" is the cause of a fair amount of complexity in the PR, since we have to handle both flows.
 * The fact that we used to key metadata by "key", but are now using "name" in the new code also caused some confusion. We should do a pass and make sure we are consistent. Note that a metadatum only contains metadata_field_id, so you need to almost always fetch metadataTypes along with metadata. Might want to rethink this in the future.
* The dropdowns are stuck inside the ManualInputTable, which is not ideal. Will fix soon.

screen shot 2019-02-26 at 5 42 03 pm


app/assets/src/components/common/MetadataInput.jsx
```diff
@@ -0,0 +1,65 @@
...
+MetadataInput.propTypes = {
+  className: PropTypes.string,
+  value: PropTypes.any,
+  metadataType: PropTypes.shape({
+    key: PropTypes.string,
+    dataType: PropTypes.oneOf(["number", "text", "date"]),
+    options: PropTypes.arrayOf(PropTypes.string)
+  }),
+  // First parameter is the metadataType NAME
+  // Third optional parameter is whether to immediately save. false means "wait for onSave to fire".
```
jshoe: I'm confused by the `Third optional parameter` comment? which one is that?

app/controllers/projects_controller.rb
```diff
@@ -321,6 +321,12 @@ def upload_metadata
     }
   end
 
+  def metadata_fields
+    project_ids = (params[:projectIds] || []).map(&:to_i)
+
+    render json: Project.where(id: project_ids).map(&:metadata_fields).flatten.map(&:field_info)
```
jshoe: you should still do current_power.projects, right?

app/assets/src/components/common/MetadataInput.jsx
MarkAZhang: Let me fix the comment.

We needed some way to tell the parent when to save.
For `<Inputs>`, we want the parent to save when the user blurs, not while the user is typing. But we still need to tell the parent about changes when the user is typing.

MarkAZhang: This is referring to the `true` in `onChange(metadataType.name, val ,true)`

app/controllers/projects_controller.rb
MarkAZhang: Yes, thanks!

PR #1969, opened 2019-02-27
Improve phylo_trees#index performance

cdebourcy:
# Description
- This PR fixes performance regressions due to inefficient `TaxonLineage` and `MetadataField` queries. 
- A follow-up PR should fix performance problems due to the fact that all metadata for all phylo_trees are being loaded, just to display the tree list. We should have separate `index` and `show` actions, where `index` just gets tree names and such, whereas `show` gets all the info for a single tree, only if the user actually visits that tree.
- A review of the performance of all metadata-related actions across the entire app will be necessary in the future, since a number of inefficient queries have crept in.

## Type of change

## Before
Excerpts from `log/development.log` of 1 single `phylo_trees#index` action:
screen shot 2019-02-27 at 2 32 52 pm
...
screen shot 2019-02-27 at 2 33 01 pm
...
screen shot 2019-02-27 at 2 33 24 pm
...

## After
only O(1) query per table (`log/development.log`)
screen shot 2019-02-27 at 2 28 26 pm

### DIFF
--- a/app/models/metadatum.rb
+++ b/app/models/metadatum.rb
@@ -235,6 +235,21 @@ class Metadatum < ApplicationRecord
     ""
   end
 
+  def self.validated_value_multiget(metadata)
+    metadata_fields = MetadataField.where(id: metadata.pluck(:metadata_field_id)).index_by(&:id)
+    validated_values = {}
+    metadata.each do |md|
+      mdf = metadata_fields[md.metadata_field_id]
+      if mdf
+        base = convert_type_to_string(mdf.base_type)
+        validated_values[md.id] = md["#{base}_validated_value"]
+      else
+        validated_values[md.id] = ""
+      end
+    end
+    validated_values
+  end
+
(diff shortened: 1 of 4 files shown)


app/models/metadatum.rb
```diff
@@ -235,6 +235,21 @@ def validated_value
     ""
   end
 
+  def self.validated_value_multiget(metadata)
+    metadata_fields = MetadataField.where(id: metadata.pluck(:metadata_field_id)).index_by(&:id)
```
jshoe: you could also do .map(&:metadata_field) inside

cdebourcy: You mean like `metadata_fields = metadata.map(&:metadata_field).index_by(&:id)`? That would incur a query for each metadata record.
image
image

jshoe: Ah ok, never mind then

cdebourcy: cool

PR #1970, opened 2019-02-27
Show search results stats in new sidebar

gregdingle:
# Description

This adds a sidebar to the new library and public pages which shows stats on the current search results. 

image

The metadata values are clickable. The effect in future will be to filter the current results. (Question: will it add a filter to the current filter set or replace the current set?)

The sidebar shows for the projects view as well, but not visualizations, because it not clear how or why we would summarize their metadata.

image 

The data for both views currently comes from the same api calls that the `DiscoveryView` makes. That means they only summarize what is currently loaded, not all possible results! (@tfrcarvalho and I have discussed other data approaches)

See @happyimadesignr mocks https://chanzuckerberg.invisionapp.com/share/EAQPA523SD9#/screens/348888814 . 

The main differences are that 

* Some of the stats are missing such as "non-host reads". It is not clear how to get these both efficiently for projects and samples, because they are not returned in the initial fetches. 
* There is no date graphic. This we can do in a later PR. (Note: I think treating date as a text field is still useful.)
* "Sample info" host and tissue are not shown. I'm not sure what these mean given that metadata also appears below. @happyimadesignr ?
* The "summary info / similar data" tabs do not exist. This will come in a later PR when similar data is ready. 
* I renamed "Sample info" to "Overall" to better reflect its contents. 
* I added a count of projects.
* The whole page width is different, pending addition of the left hand filters.
* I sorted the metadata values alphabetically. (Question: should this be by frequency?)

# Test and Reproduce

1. go to /library
2. wait for samples
3. see sidebar filled in with data
4. click on project tab
5. see sidebar filled in
6. go to /public
7. repeat 2 thru 5 above


app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -0,0 +1,213 @@
...
+  async componentDidMount() {
+    this.genomes = await getAllHostGenomes();
+  }
+
+  componentDidUpdate() {
+    const { currentTab, projects } = this.props;
+
+    // TODO (gdingle): this is not working when flipping back from visualization tab
+    if (this.state._computed && this.state._computed == currentTab) {
+      // Only update once because of
```
cdebourcy: @MarkAZhang Any ideas on best practice for avoiding this issue?

gregdingle: Yea, I wanted to ask in person someone because the update behavior is beyond me. `componentDidUpdate` is firing thousands of times.

happyimadesignr:
Thanks for sharing! So neat to see all the different tissue types already. Thanks for listing out things you’re still working on.
* To answer the filter question, it should add a filter to the current filter set rather than replace it.
* You’re right, the host and tissue aren’t needed in the overview as they are in the section below.
* Keep in mind for later, this panel will also show in Browse and inside a project.
* Why does it say host is unknown, when it's listed in the table? 
* Why are the dates created only one, where you can see on the left there were uploads from different times?
* We’ll also need to toggle to open and close this panel.
* I owe you colors for the bar charts, they will be our categorical colors. Maybe @tfrcarvalho can point you to those.
* I owe you some specs too for details of spacing and sizing

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
cdebourcy: I've encountered this issue before -- in my case it was when I had inadvertently *called* a function in a prop, instead of just *passing* it as the prop. So that's why I thought maybe a similar hard-to-spot mistake is at work here. Hope Mark knows more.

MarkAZhang: When you call `setState`, it causes the component to re-render again, which causes `componentDidUpdate` to fire again. If you call `setState` again in `componentDidUpdate`, this will cause the loop to happen again, creating an infinite loop.

A common approach is to use the `prevProps` or `prevState` parameters that are passed into `componentDidUpdate` to conditionally call `setState` only when the relevant thing has changed.

An example from `BareDropdown`:

```
  componentDidUpdate(prevProps) {
    if (
      prevProps.items !== this.props.items ||
      prevProps.options !== this.props.options ||
      prevProps.value !== this.props.value
    ) {
      this.setState({
        filteredItems: this.getFilteredItems(this.state.filterString)
      });
    }
  }
```

In your case, I think you can do `prevProps.currentTab !== this.props.currentTab`

MarkAZhang: See https://reactjs.org/docs/react-component.html#componentdidupdate for more info.

gregdingle: Thanks for the explanation. It’s been too long since I used react! I found the clear solution is to use getDerivedState method.

gregdingle:
@happyimadesignr 

> Keep in mind for later, this panel will also show in Browse and inside a project.

I hadn't thought of that. Can you point me to any mocks for those? 

> Why does it say host is unknown, when it's listed in the table?

I'm converting blank values from the database into "unknown". I have no idea where they are coming from. 

> Why are the dates created only one, where you can see on the left there were uploads from different times?

This was a bug. Good catch!

> We’ll also need to toggle to open and close this panel.

Ok I will add this now.

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -0,0 +1,209 @@
...
+        created_at: {}
+        // TODO (gdingle):
+        // location: {},
+      },
+      _computed: null
+    };
+  }
+
+  async componentDidMount() {
+    // TODO (gdingle): why not include genome name in payload from getSamples?
```
jshoe: that's a good idea / should probably be done at some point

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -0,0 +1,209 @@
...
+      console.error("Not supported: " + currentTab);
+      return prevState;
+    }
+  }
+
+  static processSamples(newSamples, genomes) {
+    const samples = newSamples.map(sample => {
+      const genome = genomes[sample.host_genome_id];
+      // TODO (gdingle): best description for blanks?
+      sample.host_genome = genome ? genome.name : "unknown";
```
jshoe: Host genome should always be known but for others like sample_tissue I'd be fine with empty string

cdebourcy: Yeah host genome is always present, so maybe a bug to be worked out?

gregdingle: I actually didn't see any unknown genomes so no worries. 

@jshoe empty string doesn't work here because we need a value to click on. The root cause of empty strings should be addressed somewhere and if "unknown" is a valid state then we should require it on db write.

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
gregdingle: Should I add this as a Jira ticket or somewhere else? Specific label?

jshoe: sure. Could file it as a frontend task with the description

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -0,0 +1,223 @@
...
+  static formatDate(created_at) {
+    return moment(created_at).format("YYYY-MM-DD");
+  }
+
+  static processSamples(newSamples, genomes) {
+    const samples = newSamples.map(sample => {
+      const genome = genomes[sample.host_genome_id];
+      // TODO (gdingle): best description for blanks?
+      sample.host_genome = genome ? genome.name : "unknown";
+      sample.sample_tissue = sample.sample_tissue || "unknown";
```
cdebourcy: The `sample_tissue` column (and other similar "metadata" columns) in the sample model is deprecated. We are now using the `Metadatum` and `MetadataField` models to handle metadata more flexibly. There is still some old data in the sample_tissue column, but all new data is in the MetadataField `sample_type`. Would you mind working with @jshoe  / @MarkAZhang  to switch over to the new metadata models? Once that change has been made everywhere, we can drop the corresponding columns from the sample model. The duplication is confusing for new developers :)

gregdingle: Got it. I'll clean this up in my next PR or a dedicated metadata migration PR.

PR #1972, opened 2019-02-28
Change taxon search box in report to the new one based on elasticsearch

cdebourcy:
Much time on the report page is spent loading all the taxon info into the search box. This PR changes the search box to the same one we have in the phylo_tree workflow, which searches as you type rather than loading all data upfront.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -174,7 +174,7 @@ class SamplesController < ApplicationController
     query = params[:query]
 
     # Not permission-dependent
-    taxon_list = taxon_search(query)
+    taxon_list = taxon_search(query, ["species", "genus"])
     hosts = HostGenome.where("name LIKE :search", search: "#{query}%")
 
     # Admin-only for now: needs permissions scoping
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -26,7 +26,7 @@ class SamplesController < ApplicationController
 
   OTHER_ACTIONS = [:create, :bulk_new, :bulk_upload, :bulk_upload_with_metadata, :bulk_import, :new, :index, :index_v2, :details, :all,
                    :show_sample_names, :cli_user_instructions, :metadata_types_by_host_genome_name, :metadata_fields, :samples_going_public,
-                   :search_suggestions, :upload].freeze
+                   :search_suggestions, :choose_taxon_all_levels, :upload].freeze
 
   before_action :authenticate_user!, except: [:create, :update, :bulk_upload, :bulk_upload_with_metadata]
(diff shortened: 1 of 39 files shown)


app/controllers/samples_controller.rb
```diff
@@ -926,6 +914,11 @@ def cli_user_instructions
     render template: "samples/cli_user_instructions"
   end
 
+  def choose_taxon_all_levels
```
tfrcarvalho: Isnt this actually limiting to 2 levels?

app/helpers/elasticsearch_helper.rb
```diff
@@ -14,13 +14,14 @@ def prefix_match(model, field, prefix, condition)
     results
   end
 
-  def taxon_search(prefix)
+  def taxon_search(prefix, tax_levels = %w[species genus], label_levels = false)
```
tfrcarvalho: Shouldnt the default be to search all levels?

app/helpers/elasticsearch_helper.rb
```diff
@@ -14,13 +14,14 @@ def prefix_match(model, field, prefix, condition)
...
     return {} if Rails.env == "test"
     matching_taxa = {}
-    %w[species genus].each do |level|
+    tax_levels.each do |level|
       search_params = { query: { query_string: { query: "#{prefix}*", fields: ["#{level}_name"] } } }
       TaxonLineage.__elasticsearch__.search(search_params).records.each do |record|
         name = record["#{level}_name"]
+        name += " (#{level})" if label_levels
```
tfrcarvalho: Can we return more fields?
And return the level separately, i.e. `level=`? 
The client should handle any kind of concatenation themselves.

app/helpers/elasticsearch_helper.rb
cdebourcy: Good question. I think default should be just species & genus because most of the app focuses only on species & genus. People should not be tempted to use a default that's a lot more work than necessary. I just added all levels to backfill for existing functionality in the report search box, even though it's probably overkill

app/helpers/elasticsearch_helper.rb
cdebourcy: Thanks, I'll make that change

config/routes.rb
```diff
@@ -85,6 +84,7 @@
   post 'phylo_trees/retry', to: 'phylo_trees#retry'
   get 'phylo_trees/:id/download', to: 'phylo_trees#download'
   get 'choose_taxon', to: 'phylo_trees#choose_taxon'
+  get 'choose_taxon_all_levels', to: 'samples#choose_taxon_all_levels'
```
tfrcarvalho: As said before, from an API perspective, makes more sense a route to get taxons, and then add a param that can constrain to certain levels, no?

PR #1974, opened 2019-03-05
[Metadata] Move metadata dictionary line up + Fix typo 'converage'

jshoe:
- So that they can see "See Metadata Dictionary" in both tabs for metadata uploading
screen shot 2019-03-05 at 3 23 26 pm

- Also s/converage/coverage

### DIFF
--- a/app/assets/src/components/views/SampleUploadFlow/UploadMetadataStep.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/UploadMetadataStep.jsx
@@ -78,6 +78,15 @@ class UploadMetadataStep extends React.Component {
         <div className={cx(this.state.showInstructions && cs.hide)}>
           <div>
             <div className={cs.title}>Upload Metadata</div>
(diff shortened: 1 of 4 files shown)


app/assets/src/components/views/SampleUploadFlow/UploadMetadataStep.jsx
```diff
@@ -78,6 +78,15 @@ class UploadMetadataStep extends React.Component {
         <div className={cx(this.state.showInstructions && cs.hide)}>
           <div>
             <div className={cs.title}>Upload Metadata</div>
+            <div className={cs.dictionary}>
```
MarkAZhang: Could you add the link to the Metadata Upload Modal as well? (I think it's <UploadPage>)

jshoe: 👍 
screen shot 2019-03-05 at 3 59 37 pm

PR #1976, opened 2019-03-06
improve projects/samples view performance

cdebourcy:
Removes inefficient SQL queries.
TO DO:
- fix tests
- test every part thoroughly

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -63,6 +63,26 @@ class ProjectsController < ApplicationController
                      current_power.samples
                    end
 
+        ############ OLD ####################
+        @projects = if only_library || exclude_library
+                      @samples.group(:project).count
+                    else
+                      # Make sure you still return projects without any samples. Ex: Project listing
+                      # used by the CLI.
+                      (only_updatable ? current_power.updatable_projects : current_power.projects).map do |p|
+                        [p, Sample.where(project: p).count]
+                      end
+                    end
+        extended_projects = @projects.map do |project, sample_count|
+          project.as_json(only: [:id, :name, :created_at, :public_access]).merge(
+            number_of_samples: sample_count,
+            hosts: @samples.where(project_id: project.id).includes(:host_genome).distinct.pluck("host_genomes.name").compact,
+            tissues: @samples.where(project_id: project.id).distinct.pluck(:sample_tissue).compact
+          )
+        end
+        old_extended_projects = extended_projects
+
+        ############ NEW #######################
         # Retrieve a json of projects associated with samples;
         # augment with number_of_samples, hosts, tissues.
         # TODO: Why are we passing host / tissue for each individual sample, with a lot of duplication?
@@ -76,17 +96,24 @@ class ProjectsController < ApplicationController
         host_genome_names_by_project_id = {}
         tissues_by_project_id = {}
         @samples.includes(:host_genome).each do |s|
-          (host_genome_names_by_project_id[s.project_id] ||= []) << s.host_genome_name
+          host_genome_names_by_project_id[s.project_id] ||= []
+          host_genome_names_by_project_id[s.project_id] |= [s.host_genome_name]
           # TODO: sample_tissue column is deprecated, retrieve sample_type from Metadatum model instead
-          (tissues_by_project_id[s.project_id] ||= []) << s.sample_tissue
+          tissues_by_project_id[s.project_id] ||= []
+          tissues_by_project_id[s.project_id] |= [s.sample_tissue]
         end
         extended_projects = @projects.map do |project|
           project.as_json(only: [:id, :name, :created_at, :public_access]).merge(
             number_of_samples: sample_count_by_project_id[project.id] || 0,
-            hosts: host_genome_names_by_project_id[project.id] || [],
-            tissues: tissues_by_project_id[project.id] || []
+            hosts: (host_genome_names_by_project_id[project.id] || []).compact,
+            tissues: (tissues_by_project_id[project.id] || []).compact
           )
         end
+        Rails.logger.warn("CHARLES:")
+        Rails.logger.warn("IDENTICAL? #{old_extended_projects == extended_projects}")
+        Rails.logger.warn("OLD: #{old_extended_projects}")
+        Rails.logger.warn("NEW: #{extended_projects}")
+
         render json: extended_projects
       end
     end
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -63,20 +63,28 @@ class ProjectsController < ApplicationController
                      current_power.samples
                    end
 
-        @projects = if only_library || exclude_library
-                      @samples.group(:project).count
+        # Retrieve a json of projects associated with samples;
+        # augment with number_of_samples, hosts, tissues.
+        # TODO: Why are we passing host / tissue for each individual sample, with a lot of duplication?
+        #       Can't we tally the hosts and tissues up on the server side, since only summary counts are shown on the page?
+        @projects = if only_updatable
+                      current_power.updatable_projects
                     else
-                      # Make sure you still return projects without any samples. Ex: Project listing
-                      # used by the CLI.
-                      (only_updatable ? current_power.updatable_projects : current_power.projects).map do |p|
-                        [p, Sample.where(project: p).count]
-                      end
+                      current_power.projects
                     end
-        extended_projects = @projects.map do |project, sample_count|
+        sample_count_by_project_id = Hash[@samples.group_by(&:project_id).map { |k, v| [k, v.count] }]
+        host_genome_names_by_project_id = {}
+        tissues_by_project_id = {}
+        @samples.includes(:host_genome).each do |s|
+          (host_genome_names_by_project_id[s.project_id] ||= []) << s.host_genome_name
+          # TODO: sample_tissue column is deprecated, retrieve sample_type from Metadatum model instead
+          (tissues_by_project_id[s.project_id] ||= []) << s.sample_tissue
+        end
+        extended_projects = @projects.map do |project|
           project.as_json(only: [:id, :name, :created_at, :public_access]).merge(
-            number_of_samples: sample_count,
-            hosts: @samples.where(project_id: project.id).includes(:host_genome).distinct.pluck("host_genomes.name").compact,
-            tissues: @samples.where(project_id: project.id).distinct.pluck(:sample_tissue).compact
+            number_of_samples: sample_count_by_project_id[project.id] || 0,
+            hosts: host_genome_names_by_project_id[project.id] || [],
+            tissues: tissues_by_project_id[project.id] || []
           )
         end
         render json: extended_projects
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -141,6 +141,8 @@ class SamplesController < ApplicationController
     respond_to do |format|
       format.json do
         render json: @samples.offset(offset).limit(limit).as_json(
+          # TODO: sample_tissue column is deprecated, use Metadatum model instead
+          # TODO: add: input_files, host_genome_name, private_until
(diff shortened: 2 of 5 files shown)


app/controllers/projects_controller.rb
```diff
@@ -63,20 +63,28 @@ def index
...
+        tissues_by_project_id = {}
+        @samples.includes(:host_genome).each do |s|
+          (host_genome_names_by_project_id[s.project_id] ||= []) << s.host_genome_name
+          # TODO: sample_tissue column is deprecated, retrieve sample_type from Metadatum model instead
+          (tissues_by_project_id[s.project_id] ||= []) << s.sample_tissue
+        end
+        extended_projects = @projects.map do |project|
           project.as_json(only: [:id, :name, :created_at, :public_access]).merge(
-            number_of_samples: sample_count,
-            hosts: @samples.where(project_id: project.id).includes(:host_genome).distinct.pluck("host_genomes.name").compact,
```
cdebourcy: This loop was doing a query on samples for each project

app/controllers/projects_controller.rb
```diff
@@ -63,20 +63,28 @@ def index
...
+        # augment with number_of_samples, hosts, tissues.
+        # TODO: Why are we passing host / tissue for each individual sample, with a lot of duplication?
+        #       Can't we tally the hosts and tissues up on the server side, since only summary counts are shown on the page?
+        @projects = if only_updatable
+                      current_power.updatable_projects
                     else
-                      # Make sure you still return projects without any samples. Ex: Project listing
-                      # used by the CLI.
-                      (only_updatable ? current_power.updatable_projects : current_power.projects).map do |p|
-                        [p, Sample.where(project: p).count]
```
cdebourcy: this was doing a query per project. Also the behavior of including empty projects is probably desired for all cases, not just in the else clause

app/controllers/samples_controller.rb
```diff
@@ -138,14 +139,19 @@ def index_v2
                  current_power.samples
                end
 
-    respond_to do |format|
-      format.json do
-        render json: @samples.offset(offset).limit(limit).as_json(
-          only: [:id, :name, :sample_tissue, :host_genome_id, :project_id, :created_at],
-          methods: []
```
cdebourcy: this was calling the as_json method of Sample, which had been overridden with a method that makes a sql query for each record in @samples (see below)

app/controllers/samples_controller.rb
```diff
@@ -533,7 +539,16 @@ def samples_going_public
       params[:userId] ? User.find(params[:userId]) : current_user,
       params[:projectId] ? Project.find(params[:projectId]) : nil
     )
-    render json: samples.to_json(include: [{ project: { only: [:id, :name] } }])
```
cdebourcy: @tfrcarvalho FYI: `samples.to_json(include: :project)` will make a SQL query for each record in samples. A `samples.includes(:project).each` loop will not.

app/controllers/samples_controller.rb
```diff
@@ -825,7 +840,15 @@ def update
     respond_to do |format|
       if @sample.update(sample_params)
         format.html { redirect_to @sample, notice: 'Sample was successfully updated.' }
-        format.json { render :show, status: :ok, location: @sample }
```
cdebourcy: Tricky find: this was directing to the `show` action for `@sample`, which implicitly called the `jbuilder` template below. I want to get rid of the implicit behavior of the jbuilder, so I wrote the code out explicitly here.

app/controllers/projects_controller.rb
```diff
@@ -63,20 +63,28 @@ def index
                      current_power.samples
                    end
 
-        @projects = if only_library || exclude_library
-                      @samples.group(:project).count
+        # Retrieve a json of projects associated with samples;
+        # augment with number_of_samples, hosts, tissues.
+        # TODO: Why are we passing host / tissue for each individual sample, with a lot of duplication?
```
tfrcarvalho: We are not passing per sample... it is grouped below.
This was made more complex due to the inability to access samples from projects... something we should revisit as we review our access control logic and/or methodology.

app/controllers/projects_controller.rb
tfrcarvalho: Indeed, was on my todo list. Thanks for grouping.
`includes` might trigger more queries than a `joins`... have you looked into that?

cdebourcy: Thanks, will look into it!

app/controllers/samples_controller.rb
```diff
@@ -138,14 +139,19 @@ def index_v2
...
+    @samples = @samples.offset(offset).limit(limit)
+
+    # Efficiently fetch auxiliary information from associated models
+    samples_resp = []
+    @samples.includes(:project, :host_genome, :input_files).each do |s|
+      # TODO: sample_tissue column is deprecated, use Metadatum model instead
+      entry = s.as_json(only: [:id, :name, :sample_tissue, :host_genome_id, :project_id, :created_at])
+      entry["host_genome_name"] = s.host_genome_name # this method uses s.host_genome under the hood
+      entry["private_until"] = s.private_until # this method uses s.project under the hood
+      entry["input_files"] = s.input_files.as_json # TODO: can probably be removed, but I ported it from the legacy code just in case
```
tfrcarvalho: As you note, we do not need `input_files` here, specially since index_v2 is currently only use in discovery view.
As we migrate other calls to index_v2, we can do another request for input files from other specific endpoints.

app/controllers/samples_controller.rb
```diff
@@ -138,14 +139,19 @@ def index_v2
...
-        )
-      end
+    @samples = @samples.offset(offset).limit(limit)
+
+    # Efficiently fetch auxiliary information from associated models
+    samples_resp = []
+    @samples.includes(:project, :host_genome, :input_files).each do |s|
+      # TODO: sample_tissue column is deprecated, use Metadatum model instead
+      entry = s.as_json(only: [:id, :name, :sample_tissue, :host_genome_id, :project_id, :created_at])
+      entry["host_genome_name"] = s.host_genome_name # this method uses s.host_genome under the hood
```
tfrcarvalho: Since we include/join host_genome above, we should not need to call the `host_genome_name`, right?

app/controllers/samples_controller.rb
```diff
@@ -533,7 +539,16 @@ def samples_going_public
       params[:userId] ? User.find(params[:userId]) : current_user,
       params[:projectId] ? Project.find(params[:projectId]) : nil
     )
-    render json: samples.to_json(include: [{ project: { only: [:id, :name] } }])
+
+    # Efficiently fetch auxiliary information from Project model
+    samples_resp = []
+    samples.includes(:project).each do |s|
```
tfrcarvalho: Should this be moved to a helper method to retrieve `project properties`? 
Since it is used above too.

app/controllers/samples_controller.rb
tfrcarvalho: I am not sure what you mean... but this does not work exactly like that.
That `as_json` method only adds methods if you do not pass the parameter `methods` in. In this case is clearing it.

We might not need those methods by default, but that is another question... I am afraid that removing that method could affect something else on the website...

app/models/sample.rb
```diff
@@ -334,11 +334,6 @@ def default_background_id
     host_genome && host_genome.default_background ? host_genome.default_background.id : Background.find_by(public_access: 1).id
   end
 
-  def as_json(options = {})
-    options[:methods] ||= [:input_files, :host_genome_name, :private_until]
```
tfrcarvalho: As said above, this only adds these methods if no methods are passed.
I am not agains removing this... But be careful with possible side effects.

app/controllers/samples_controller.rb
tfrcarvalho: Cool find.
These convention behaviors can be a nice help sometimes, but as you say it can be tricky to find/track or event notice.

On the other hand, we need to standardize formats that are returned by our app, and constant custom returns are hard to maintain. Anyway, that would be an issue for an API revamp.

app/controllers/samples_controller.rb
tfrcarvalho: The problem here is not the to_json, but the no inclusion/join of project before as in `samples.joins(:project).to_json(...`

app/controllers/samples_controller.rb
cdebourcy: Got it!

app/controllers/projects_controller.rb
cdebourcy: But it passes something like `{tissues: [plasma, plasma, plasma, csf, serum, plasma, plasma,...]}`, right? (Correct me if I'm wrong.) Just curious if we could pass something like `{tissues: {plasma: 5, csf: 1, serum: 1}}` instead. Anyway, that would be for future PR...

cdebourcy: Oh never mind -- I was missing the `distinct`. Thanks, something I need to fix in my change

cdebourcy: I think the way I will test my change is that I'll explicitly write a function to compare the old and the new output for each action, and then I'll post the results here -- so I'm not missing any more things like this

app/controllers/samples_controller.rb
cdebourcy: I see. So I can remove `:input_files, :host_genome_name, :private_until` from my code here actually, since it was never present, due to `methods: []`.

cdebourcy: I'm inclined to take the risk of removing the `as_json` method from app/models/sample.rb because it could cause so many hidden problems. I will do a thorough audit to catch all the occurrences I can. If any issues come up with the removal after deployoment, I'll just fix them immediately

app/controllers/samples_controller.rb
cdebourcy: So the issue is that `host_genome_name` is actually a function, not a column. That function uses `host_genome.name`. So if we don't have `includes(:host_genome)`, the loop will query the host_genomes table in every iteration. If we do `includes(:host_genome)`, then all the records are loaded in one go, so there will be only one query on host_genomes, whose results we access multiple times

app/controllers/samples_controller.rb
cdebourcy: Thanks for confirming, I will remove

app/controllers/samples_controller.rb
cdebourcy: got it

app/models/sample.rb
cdebourcy: Yeah it's a little bit scary... I will test and be prepared to fight any fires that result :D

app/controllers/samples_controller.rb
tfrcarvalho: Agree, i change the `as_json` recently(?) to allow at least overriding the methods arg and maintaining backward compatibility... but would be ideal to delete it and change any calls that require it.

app/controllers/samples_controller.rb
tfrcarvalho: Yes, i am just saying that since we are doing includes here now we can just call `s.host_genome.name` directly.
Eventually, the `host_genome_name` function can be deleted.

app/controllers/projects_controller.rb
```diff
@@ -63,20 +63,28 @@ def index
...
+        #       Can't we tally the hosts and tissues up on the server side, since only summary counts are shown on the page?
+        @projects = if only_updatable
```
tfrcarvalho: Just notice this ignores `only_library` and `exclude_library`, which means that if any of those are set will return the wrong list of projects.

As a side note, In my current PR these variables change to domain that can have the values "library", "public" and "updatable". Default, for now, is all the projects user can see.

cdebourcy: Actually further down it only keeps the intersection of `@projects` (all projects the user has access to) and `@samples` (only the samples that are in the specified option: `only_library` or `exclude_library`)

app/controllers/samples_controller.rb
cdebourcy: That's a good idea

PR #1980, opened 2019-03-06
Add features to DD right sidebar 

gregdingle:
# Description

This adds

* _italic_ unknown
* reads per sample
* non-host reads per sample
* host genomes from backend
* locations
* accordions

image

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoverySidebar.jsx
+++ b/app/assets/src/components/views/discovery/DiscoverySidebar.jsx
@@ -14,17 +14,17 @@ export default class DiscoverySidebar extends React.Component {
 
     this.state = {
       stats: {
-        samples: 0,
-        projects: 0
-        // TODO (gdingle):
-        // avg_reads_per_sample: 0,
+        samples: "",
+        projects: "",
+        total_reads: "",
+        adjusted_remaining_reads: ""
       },
       metadata: {
         host: {},
         tissue: {},
-        createdAt: {}
-        // TODO (gdingle):
-        // location: {},
+        createdAt: {},
+        // TODO (gdingle): is location good?
+        location: {},
       }
     };
   }
@@ -41,15 +41,17 @@ export default class DiscoverySidebar extends React.Component {
         stats: {
           samples: samples.length,
           projects: uniqBy("project", samples).length
-          // TODO (gdingle): reads not in samples data yet
-          // avg_reads_per_sample: 0,
+          // TODO (gdingle): this good?
+          // projects: uniqBy("project_id", samples).length,
+          total_reads: sumBy("total_reads", samples),
+          adjusted_remaining_reads: sumBy("adjusted_remaining_reads", samples)
         },
         metadata: {
           host: countBy("hostGenome", samples),
           tissue: countBy("sampleTissue", samples),
           createdAt: countBy("createdAt", samples)
-          // TODO (gdingle): location not in samples data yet
-          // location: {},
+          // TODO (gdingle): this good?
+          location: countBy("sample_location", samples)
         },
         _computed: currentTab
       };
@@ -60,6 +62,7 @@ export default class DiscoverySidebar extends React.Component {
 
       const hosts = flatten(map("hosts", projects));
       const tissues = flatten(map("tissues", projects));
+      const locations = flatten(map("sample_locations", projects));
 
       const createdAts = map(
         project => DiscoverySidebar.formatDate(project.created_at),
@@ -69,18 +72,16 @@ export default class DiscoverySidebar extends React.Component {
       return {
         stats: {
           samples: sumBy("number_of_samples", projects),
-          projects: projects.length
-          // TODO (gdingle): reads not in projects data yet
-          // avg_reads_per_sample: 0,
+          projects: projects.length,
+          total_reads: sumBy("total_reads", projects),
+          adjusted_remaining_reads: sumBy("adjusted_remaining_reads", projects)
         },
         metadata: {
           host: countBy(null, hosts),
           tissue: countBy(null, tissues),
           createdAt: countBy(null, createdAts)
-          // TODO (gdingle): location not in projects data yet
-          // location: {},
+          location: countBy(null, locations)
         },
-        _computed: currentTab
       };
     } else {
       // eslint-disable-next-line no-console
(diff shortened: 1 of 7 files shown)


app/controllers/samples_controller.rb
```diff
@@ -138,11 +138,25 @@ def index_v2
                  current_power.samples
                end
 
+    # Add total_reads
+    # TODO: (gdingle): this appears to be making n+1 queries
```
gregdingle: @tfrcarvalho told me to leave the data queries alone until his refactor is in.

app/controllers/projects_controller.rb
```diff
@@ -72,11 +72,15 @@ def index
                         [p, Sample.where(project: p).count]
                       end
                     end
+        # TODO: (gdingle): this seems inefficient... 5 extra queries * projects.length
```
gregdingle: @tfrcarvalho told me to leave the data queries alone until his refactor is in.

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -114,12 +105,20 @@ export default class DiscoverySidebar extends React.Component {
     return moment(created_at).format("YYYY-MM-DD");
   }
 
-  static processSamples(newSamples, genomes) {
+  // TODO (gdingle): humanize better into M such as 34M
```
gregdingle: is there an existing library for this? I'd like to be consistent

app/controllers/projects_controller.rb
cdebourcy: Thanks for flagging @gregdingle ! We may have more such patterns in the codebase and it's good to get rid of them. For this particular place, I already have a PR open, still to be tested: https://github.com/chanzuckerberg/idseq-web/pull/1976
Feel free to just incorporate anything you find useful into your PR because I may not have time to get to the testing immediately.

@tfrcarvalho Is there a branch we can look at for the refactor you mentioned? Or would it actually help alleviate some of your workload if one of us fixed these inefficiencies right now? I know you have taken on a ton of work for data discovery, which is amazing, it's such a huge project. In general I dislike leaving critical fixes such as this one for later! So I would strongly suggest either I merge my PR in or Greg incorporates the changes into his. People will copy and paste this code pattern and it will proliferate like bunny rabbits ;)

app/controllers/samples_controller.rb
cdebourcy: same comment as above.

app/controllers/projects_controller.rb
cdebourcy: It's quite worrying to me that just in this PR, we're already adding 3 more inefficient lines just because we want to leave the refactor for later

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
MarkAZhang: We use the helper function `numberWithCommas` in some places (in `helper/string`), but the `toLocaleString` that you use seems like a better choice.

I would recommend changing `numberWithCommas` to use `toLocalString()`, renaming `numberWithCommas` to something like `formatNumber`, and writing a comment to explain when it's appropriate to use that function. And then use that function here. I don't think we need to humanize `100,000,000` to `100M` at the moment, and I think we're unlikely to have numbers bigger than that.

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -24,52 +14,42 @@ export default class DiscoverySidebar extends React.Component {
...
-      );
+      const samples = DiscoverySidebar.processSamples(newProps.samples);
       if (!samples || !samples.length) {
         return prevState;
       }
       return {
         stats: {
           samples: samples.length,
-          projects: uniqBy("project_id", samples).length
+          projects: uniqBy("project_id", samples).length,
```
MarkAZhang: Rename `samples` to `numSamples`, so it's more clear that it's a count instead of an array of samples. Same with project.

Also, I know this is kind of annoying, but can you use camelCase for javascript variables? We should write something to automatically convert snake_case to camelCase at the API fetching layer so we only deal with camelCase on the front-end...

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -114,12 +96,20 @@ export default class DiscoverySidebar extends React.Component {
     return moment(created_at).format("YYYY-MM-DD");
   }
 
-  static processSamples(newSamples, genomes) {
+  // TODO (gdingle): humanize better into M such as 34M
+  formatNumber(number) {
```
MarkAZhang: Could you name this function something like `getAverageOverSamples`? To make it more clear what the function is doing.

MarkAZhang:
The label here isn't vertically-aligned.

screen shot 2019-03-06 at 5 51 12 pm

I notice that this sidebar is very similar to the details sidebar, but the styling is different. (for example, the font-size and whitespace). Is that intentional? @happyimadesignr 

screen shot 2019-03-06 at 5 52 21 pm

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
gregdingle: Thanks for the background. I noted the `34M` style numbers because that's what was in the mock by @happyimadesignr . Not sure whether it's worth breaking consistency. 

As for `toLocaleString`, it does seem better in the long run as we serve a global audience. (In French for example, numbers are written such as `1 000 000,1`.

Do you want to see that change in this PR or a new one or file it away in Jira ?

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
gregdingle: No problem. I agree those are improvements. 

> automatically convert snake_case to camelCase at the API fetching layer 

I'd love to know whether this has been done without issue before. Personally I hate all the time I spend switching cases 😠

happyimadesignr:
@MarkAZhang @gregdingle Yes, thanks for pointing out, let's keep it consistent with the samples details style. Though for both it would be nice to tighten up. The sample details has a bit too much height. Can we change the height of both to 34px?

Given these are the same style, is there a reason why they can't have shared styles or be a component? The padding and margins are currently done differently in both too.

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
MarkAZhang: Yeah it's definitely doable. The only question is whether to do it on the front-end or back-end. If we do it once the data is received on the front-end, then the data will be inconsistent with what you see in the "Network" tab on Chrome Developer tools, so back-end seems preferred. Something to look into in the future.

app/controllers/projects_controller.rb
tfrcarvalho: One option: since @gregdingle  is splitting the bug fix from this PR... lets leave it like this until my PR is in. Then I can submit a change to this one correcting the endpoints and connecting the data.
Second option: submit the frontend code only and I can connect it in a separate PR. 

I am already working on providing the data for the old code and will work on this soon.
Lets not copy more refactors from each other... that just confusing and hard to merge later.

gregdingle:
@MarkAZhang @happyimadesignr 

I adjusted the styles to match the details sidebar. As for extracting common component, I'd like to discuss how best to do this given existing patterns. 

image

gregdingle:
@tfrcarvalho could you review?

gregdingle:
@MarkAZhang @cdebourcy could I get a review?

PR #1981, opened 2019-03-06
Create viz on phylo tree create

gregdingle:
# Description

See comments in code. 

# Test and Reproduce

TODO: I'm not actually sure how best to test this. I'm having trouble creating a phylo tree in my localhost and I didn't see any examples of unit tests. Please advise.

### DIFF
--- a/app/models/phylo_tree.rb
+++ b/app/models/phylo_tree.rb
@@ -13,6 +13,20 @@ class PhyloTree < ApplicationRecord
   STATUS_FAILED = 2
   STATUS_IN_PROGRESS = 3
 
+  after_create :create_visualization
+
+  # We need to create a visualization object here to register the new phylo_tree
+  # as a generic visualization. Other types of visualizations are saved on-demand only.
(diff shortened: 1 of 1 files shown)


app/models/phylo_tree.rb
```diff
@@ -13,6 +13,20 @@ class PhyloTree < ApplicationRecord
   STATUS_FAILED = 2
   STATUS_IN_PROGRESS = 3
 
+  after_create :create_visualization
+
+  # We need to create a visualization object here to register the new phylo_tree
+  # as a generic visualization. Other types of visualizations are saved on-demand only.
+  # Because phylo_trees are created explictly, the user will expect them to persist.
+  # TODO: (gdingle): destroy visualization objects when phylo_tree is destroyed.
```
gregdingle: I'm not sure if it is possible to destroy a phylo tree currently. Is it worth the effort of adding this other callback then?

cdebourcy: It is currently impossible to destroy a tree :D

PR #1982, opened 2019-03-06
DD Date histogram

gregdingle:
# Description

This summarizes the dates of the current samples in a histogram. 

image

See original mock by @happyimadesignr 
https://chanzuckerberg.invisionapp.com/share/EAQPA523SD9#/screens/348888814 .

# Test and Reproduce

1. load public and library, for each...
2. see height of bars of dates in proportion to frequency
3. see first and last date correct
4. click on bar, see update in URL (for future filtering)

### DIFF
--- a/app/assets/src/components/layout/Accordion.jsx
+++ b/app/assets/src/components/layout/Accordion.jsx
@@ -4,14 +4,10 @@ import cx from "classnames";
 import cs from "./accordion.scss";
 
 class Accordion extends React.Component {
-  state = {
-    open: false
-  };
-
-  static getDerivedStateFromProps(newProps, prevState) {
-    return {
-      ...prevState,
-      open: newProps.open
+  constructor(props) {
+    super(props);
+    this.state = {
+      open: this.props.open
     };
   }
 
@@ -63,7 +59,8 @@ Accordion.propTypes = {
 };
 
 Accordion.defaultProps = {
-  toggleable: true
+  toggleable: true,
+  open: false
 };
 
 export default Accordion;
--- a/app/assets/src/components/layout/Accordion.jsx
+++ b/app/assets/src/components/layout/Accordion.jsx
@@ -8,10 +8,11 @@ class Accordion extends React.Component {
     open: false
   };
 
-  componentDidMount() {
-    this.setState({
-      open: this.props.open
-    });
+  static getDerivedStateFromProps(newProps, prevState) {
+    return {
+      ...prevState,
+      open: newProps.open
+    };
   }
 
   onToggle = () => {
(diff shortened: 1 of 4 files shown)


app/controllers/samples_controller.rb
```diff
@@ -138,11 +138,25 @@ def index_v2
...
+            :name,
+            :sample_tissue,
+            :host_genome_id,
+            :project_id,
+            :created_at,
+            :sample_location,
+            :total_reads,
+            :adjusted_remaining_reads
+          ],
+          methods: [:host_genome_name]
```
cdebourcy: flagging this as potentially inefficient (unless `includes(:host_genome)` is added), because it calls the `host_genome` model

cdebourcy: Please make sure it doesn't do a query on host_genomes for each sample

app/assets/src/components/layout/Accordion.jsx
```diff
@@ -8,10 +8,11 @@ class Accordion extends React.Component {
...
   };
 
-  componentDidMount() {
-    this.setState({
-      open: this.props.open
-    });
+  static getDerivedStateFromProps(newProps, prevState) {
+    return {
+      ...prevState,
+      open: newProps.open
```
MarkAZhang: I think this is going to cause problems for the uncontrolled case. (where `<Accordion>` is keeping track of its own `open` state and you don't pass in `props.open`)

If you don't pass in `props.open`, then whenever any props change (including `props.children`), this function is going to overwrite `this.state.open` with `undefined`, even if it had been `true`. This will cause the accordion to unintentionally close.

I think you made a change in `d1eef6320` that broke this component and that the code before was correct. It looks like you were trying to set the `<Accordion>` to be open at the start instead of closed. In this case, you should add a separate `defaultOpen` prop, and do

```
constructor(props) {
  super(props);
  this.state = {
    open: this.props.defaultOpen || false,
  };
}
```
to initialize the internal `open` state to the value you want.

Let me know if this doesn't make sense.

gregdingle: Very true, this is wrong. I'll fix it. Thanks for the explanation.

app/controllers/samples_controller.rb
gregdingle: Thanks for the improvement. I agree this needs to be at least as fast as before.

app/assets/src/components/layout/Accordion.jsx
gregdingle: I figured out a solution with the special `key` prop of react.

app/controllers/samples_controller.rb
gregdingle: Fixed with help from @cdebourcy

gregdingle:
@MarkAZhang this is ready for review again.

PR #1983, opened 2019-03-07
[Upload Flow] Start SampleUploadFlowHeader and add additional details

jshoe:
- This PR moves the current upload flow titles to Headers and adds some additional details like a help link.
- Going to put these up as they're ready. NEXT UP: the tab selectors and footer.
- One big change is moving getProjectMetadataFields and getAllHostGenomes up a level to MetadataUpload so that it can be called once and use those fields for host genomes and required fields listing + MetadataManualInput. So MetadataManualInput is no longer async.
- Also moves the title details for each page to SampleUploadFlowHeader

screen shot 2019-03-07 at 10 17 47 pm
screen shot 2019-03-07 at 10 18 00 pm
screen shot 2019-03-07 at 10 18 20 pm
screen shot 2019-03-07 at 10 18 30 pm
screen shot 2019-03-07 at 10 19 06 pm

### DIFF
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -279,7 +279,6 @@ class MetadataManualInput extends React.Component {
             )
           ) {
             return (
-<<<<<<< HEAD
               <div>
                 <MetadataInput
                   key={column}
@@ -293,18 +292,6 @@ class MetadataManualInput extends React.Component {
                 />
                 {this.renderApplyToAll(sample, column)}
               </div>
-=======
-              <MetadataInput
-                key={column}
-                className={cs.input}
-                value={this.getMetadataValue(sample, column)}
-                metadataType={this.props.projectMetadataFields[column]}
-                onChange={(key, value) =>
-                  this.updateMetadataField(key, value, sample)
-                }
-                withinModal={this.props.withinModal}
-              />
->>>>>>> Fix bug
             );
           }
           return (
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -40,12 +40,8 @@ class MetadataManualInput extends React.Component {
     }
   };
 
-  componentDidUpdate() {
-    const { projectMetadataFields, hostGenomes } = this.props;
-    // Only update if fields not set already
-    if (this.state.projectMetadataFields && this.state.hostGenomes.length > 0) {
-      return;
-    }
+  componentDidMount() {
+    const { projectMetadataFields, hostGenomes, samplesAreNew } = this.props;
 
     this.setState({
       projectMetadataFields: projectMetadataFields,
@@ -57,7 +53,7 @@ class MetadataManualInput extends React.Component {
       hostGenomes,
       headers: {
         "Sample Name": "Sample Name",
-        ...(this.props.samplesAreNew
+        ...(samplesAreNew
           ? {
               "Host Genome": (
                 <div>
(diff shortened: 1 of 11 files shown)


jshoe:
^ updated screenshots based on design requests

app/assets/src/components/common/MetadataManualInput.jsx
```diff
@@ -90,12 +86,6 @@ class MetadataManualInput extends React.Component {
     ];
   };
 
-  handleHostGenomeChange = id => {
```
MarkAZhang: I'm assuming you're going to remove this in https://github.com/chanzuckerberg/idseq-web/pull/1988

app/assets/src/components/views/SampleUploadFlow/UploadMetadataStep.jsx
```diff
@@ -116,7 +118,8 @@ UploadMetadataStep.propTypes = {
   project: PropTypes.shape({
     id: PropTypes.number,
     name: PropTypes.string
-  })
+  }),
+  handleRequiredMetadataFields: PropTypes.func
```
MarkAZhang: Doesn't look like `handleRequiredMetadataFields` is used.

MarkAZhang:
On a wide screen, this looks a little off.

Screen Shot 2019-03-08 at 11 50 12 AM

Can we wrap the Header in a NarrowContainer as well?

MarkAZhang:
I really don't like the multiple lines of text visually. For "Required Fields" and "Host Genomes", can we hide them under a "View/Hide Required Fields and Host Genomes" button, like we do with "More Info" here? So when the user clicks on that, they'll show up in a grey box. You can put this button on the same line as "See Metadata Dictionary" and maybe separate it with a "|", so there's only one line of text.

Screen Shot 2019-03-08 at 12 01 19 PM

MarkAZhang:
Can you make the `margin-bottom` on the title 5px instead of 15? There's a little too much whitespace for elements that should be grouped together visually.

MarkAZhang:
I also don't like these multiple lines, but can't figure out what to do with the "Need help? Message us". I kind of want to remove it, or just show it when we display errors, but let's leave it for now.

Screen Shot 2019-03-08 at 12 21 05 PM

app/assets/src/components/views/SampleUploadFlow/UploadMetadataStep.jsx
jshoe: oops I think I merged it wrong

jshoe:
@MarkAZhang 👍 updated with all the changes

Screen Shot 2019-03-11 at 9 18 45 AM
Screen Shot 2019-03-11 at 9 18 38 AM
Screen Shot 2019-03-11 at 9 18 24 AM

app/assets/src/components/common/MetadataManualInput.jsx
```diff
@@ -81,8 +77,17 @@ class MetadataManualInput extends React.Component {
       }
     });
 
-    // Default to the first host genome (Human)
-    this.props.samples.map(sample => this.updateHostGenome(1, sample));
+    // Set all to Human by default
```
MarkAZhang: Maybe add a comment saying "we can't use updateMetadataField because we are updating a bunch of fields in bulk."

app/assets/src/components/common/MetadataManualInput.jsx
```diff
@@ -81,8 +77,17 @@ class MetadataManualInput extends React.Component {
...
+    const newHeaders = union(["Host Genome"], this.state.headersToEdit);
+    let newFields = this.state.metadataFieldsToEdit;
+    this.props.samples.map(sample => {
```
MarkAZhang: forEach might be better than map here.

PR #1985, opened 2019-03-07
Add filters to discovery views.

tfrcarvalho:
* Add filters (time, visibility, taxon, host, tissue) to discovery views
  - Taxon search is an async dropdown (search box are server driven)
* Incorporate with sidebar
* Toggles for left and right pane
* All data fetching was moved to the root component
* Incorporated or adapted most ActiveRecord query improvements suggested

TODO (follow-up): Header search box is not currently connected to filters. Will do that on a follow up PR.


app/controllers/samples_controller.rb
```diff
@@ -124,40 +124,111 @@ def index
...
+    samples_visibility = visibility(limited_samples)
 
+    details_json = format_samples(limited_samples).as_json()
+    limited_samples_json.zip(details_json, samples_visibility).map do |sample, details, visibility|
+      sample[:public] = visibility
+      sample[:details] = details
+    end
+
+    results = { samples: limited_samples_json }
+    results[:all_samples_ids] = samples.pluck(:id) if include_ids
```
gregdingle: so I guess this will cause two queries to be run
* all samples
* limited samples

and I guess you found that running two queries and returning less data to the client is more efficient than running one query and limiting in the client

gregdingle: By the way, I find `include_ids` misleading because it implies that when it is false there are no ids. why not `include_all_samples_ids` as a flag?

app/helpers/samples_helper.rb
```diff
@@ -367,6 +420,18 @@ def format_samples(samples)
...
 
+  def visibility(samples)
+    # When in conjunction with some filters, the query below was not returning the public property,
+    # thus we need to get ids and redo the query independently
+    sample_ids = samples.pluck(:id)
+    return current_power
+           .samples
+           .where(id: sample_ids)
+           .joins(:project)
+           .select("samples.*", "IF(projects.public_access = 1 OR DATE_ADD(samples.created_at, INTERVAL projects.days_to_keep_sample_private DAY) < '#{Time.current.strftime('%y-%m-%d')}', true, false) AS public")
```
gregdingle: hmmm... this is getting complicated... it seems like long-run we should simplify, centralize and-or write test code so that we don't ever make mistakes with visibility and it doesn't cost in perf

app/assets/src/components/ui/icons/info_icon.scss
```diff
@@ -0,0 +1,22 @@
+// TODO(tiago): find a way to get rid of these styles
+.cls1 {
```
jshoe: hm interesting. you didn't want to just leave them in the svgs?

app/helpers/samples_helper.rb
```diff
@@ -196,6 +215,40 @@ def get_total_runtime(pipeline_run, run_stages)
...
+  def filter_by_time(samples, start_date, end_date)
+    samples.where("samples.created_at >= ? AND samples.created_at <= ?", start_date, end_date)
+  end
+
+  def filter_by_visibility(samples, visibility)
+    if visibility
+      public = visibility.include?("public")
+      private = visibility.include?("private")
+
+      if public ^ private
```
gregdingle: why XOR here? why not `elsif private` ? flat is better than nested

app/controllers/samples_controller.rb
```diff
@@ -124,40 +124,111 @@ def index
   end
 
   def index_v2
-    only_library = ActiveModel::Type::Boolean.new.cast(params[:onlyLibrary])
-    exclude_library = ActiveModel::Type::Boolean.new.cast(params[:excludeLibrary])
+    domain = params[:domain]
```
jshoe: I'm assuming some of these controller changes are implementing Charles's suggestions?

app/assets/src/components/ui/icons/info_icon.scss
tfrcarvalho: Had syntax error in React. Not sure what is the best way to incorporate it, so ended up with a quick fix. With more time I can turn them into properties, probably.
Let me know if you know of a better way...

app/controllers/projects_controller.rb
```diff
@@ -51,39 +51,128 @@ def index
...
+                .group(:string_validated_value)
+                .distinct
+                .count(:project_id)
+
+    # for metadata fields we need to send both value and text
+    locations = locations.map do |location, count|
+      { value: location, text: location, count: count }
+    end
+
+    tissues = Metadatum
```
gregdingle: I'd like to see all the types of metadata queries split into their own private methods because they are logically independent. Then we could maybe run them in parallel. Does rails support that?

app/controllers/visualizations_controller.rb
```diff
@@ -7,19 +7,16 @@ class VisualizationsController < ApplicationController
...
       visualizations = current_user.visualizations
-    elsif exclude_library
+    elsif domain == "public"
       visualizations = Visualization
                        .where(public_access: 1)
-    # TODO: See https://jira.czi.team/browse/IDSEQ-866
-    # .where.not(user: current_user)
     else
-      raise 'Visualizations must be for either "my library" or "public"'
+      Visualizations.where(public_access: 1, user: current_user)
```
gregdingle: will this ever make sense? public visualizations by the current user but not private ones?

app/helpers/samples_helper.rb
```diff
@@ -168,6 +168,25 @@ def parsed_samples_for_s3_path(s3_path, project_id, host_genome_id)
...
 
+  def filter_samples(samples, params)
+    host = params[:host]
+    location = params[:location]
+    taxon = params[:taxon]
+    time = params[:time]
+    tissue = params[:tissue]
+    visibility = params[:visibility]
+
+    # # TODO(tiago): check performance
```
gregdingle: in this PR?

app/helpers/samples_helper.rb
```diff
@@ -196,6 +215,40 @@ def get_total_runtime(pipeline_run, run_stages)
     end
   end
 
+  def filter_by_time(samples, start_date, end_date)
```
gregdingle: make all the `filter_by_` methods private ?

app/helpers/samples_helper.rb
```diff
@@ -352,8 +404,9 @@ def format_samples(samples)
     metadata_by_sample_id = metadata_multiget(sample_ids)
 
     # Massage data into the right format
-    samples.each_with_index do |sample|
-      job_info = {}
+    # samples.includes(:pipeline_runs, :host_genome, :project, :input_files).each_with_index do |sample|
```
gregdingle: why commented out?

app/models/sample.rb
```diff
@@ -437,11 +437,18 @@ def deletable?(user)
...
-      .where("(projects.public_access = 1 or
+      .where("(projects.public_access = 1 OR
               DATE_ADD(samples.created_at, INTERVAL projects.days_to_keep_sample_private DAY) < ?)",
              Time.current)
   end
 
+  def self.private_samples
+    joins(:project)
+      .where("(projects.public_access = 0 AND
+              DATE_ADD(samples.created_at, INTERVAL projects.days_to_keep_sample_private DAY) >= ?)",
```
gregdingle: This implies that once a sample is old enough it *leaves* the library and becomes public. Users may be surprised when they can't find something in their library that was there yesterday. It also seems strange that you may not be able to see all samples in a project because they could be divided between public and private.

app/controllers/samples_controller.rb
tfrcarvalho: + some extra improvements.

app/assets/src/components/views/discovery/discovery_api.js
```diff
@@ -0,0 +1,130 @@
...
+    ),
+    totalRuntime: get("run_info.total_runtime", sample.details)
+  };
+  return row;
+};
+
+const getDiscoverySamples = async ({
+  domain,
+  filters,
+  limit = 100,
```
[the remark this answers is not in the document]
tfrcarvalho: These get samples only and ask for certain ranges.
We do not want to ask for all samples with all data at a time due to the volume of data.

For statistics we should use different endpoints, that returns aggregated data for us, and leave the heavy weight to the database.
`/samples/dimensions` endpoint was built to fit my filtering nodes. We can expand/adpat it to fit other aggregation needs.

app/assets/src/components/views/discovery/discovery_api.js
```diff
@@ -0,0 +1,130 @@
...
+        sample.details
+      ),
+      percent: get(
+        "derived_sample_output.summary_stats.percent_remaining",
+        sample.details
+      )
+    },
+    notes: sample.sample_notes,
+    nucleotideType: get("metadata.nucleotide_type", sample.details),
+    passedQC: get(
```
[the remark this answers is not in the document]
tfrcarvalho: Whenever I process data on client side I convert to camelCase... 

I see your point about changing the actual name as it helps tracking variables and code readability (was trying to match the title in the designs). I will actually change the name back to `qcPercent`.

app/controllers/samples_controller.rb
```diff
@@ -124,40 +124,111 @@ def index
   end
 
   def index_v2
```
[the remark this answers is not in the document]
tfrcarvalho: Indeed. Done

app/controllers/samples_controller.rb
tfrcarvalho: 1st comment: This is done for paginated loaders and as you see is fully configurable (I might constrain the limit value in the future). Client can load all ids or not, or ask for all data.

Sending all ids is actually a legacy behavior, which we might change in the future. It is used for the select all logic for infinite tables

2nd comment: Agree it is misleading. Will rename.

app/helpers/samples_helper.rb
tfrcarvalho: Testing question is a trade-off between speed of development and correctness.
I agree and we discussed before that our bar is currently low and we need more tests for correctness. Just need time/resources.

This is not a critical path since it works on a set of samples and access control should be done through current_power/consul.

app/controllers/projects_controller.rb
tfrcarvalho: The problem here is redundancy since parallelization is orthogonal. Apparently there is promise gem we can explore. 

I will move the method to the model.

app/helpers/samples_helper.rb
tfrcarvalho: Because that block only applies if they exclusively selected (not simultaneously).

app/controllers/visualizations_controller.rb
tfrcarvalho: I believe so, but not necessarily right now (short term for an admin option). Changed it because I prefer not raising errors unless something is strictly wrong.

I can put it back and change when necessary if you have a strong feeling about it. Wdyt?

app/helpers/samples_helper.rb
tfrcarvalho: I actually checked already...
Still wanted to revisit once i have more time.

app/helpers/samples_helper.rb
tfrcarvalho: You might need to refresh. Was doing a test at the timeY

app/helpers/samples_helper.rb
tfrcarvalho: Nothing wrong with that. But that could apply to most methods on this file.

app/models/sample.rb
tfrcarvalho: * Users are notified in app when samples go public
* There is a project/task to send emails on visibility changes
* It is part of our data policy

* The public samples of private projects are used to display on 
heatmaps, maps, phylo trees etc, so it is not weird that samples and projects can have different visibilities. Having said that, our access control mechanism  has some caveats both logic and UX, so eventually we will have to revisit it.

app/assets/src/components/views/discovery/discovery_api.js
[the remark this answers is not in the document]
tfrcarvalho: For now, the sidebar statistics will be computed as we scroll... but we will have to address that in the future (by using the dimensions endpoint or some other solution)

app/controllers/visualizations_controller.rb
gregdingle: >  I prefer not raising errors unless something is strictly wrong

That's what I'm wondering... is there a domain other than library or public that is valid?

app/helpers/samples_helper.rb
gregdingle: I'm of the school that methods should be private if they are not intended to be called outside of the class. Minimize the size of the interface.

app/controllers/visualizations_controller.rb
tfrcarvalho: Yes, the union of both (all that the user has access to). Which is our default for when we loading samples. The question would be if the same default makes sense for visualizations.

I see the library and public as filters over the default.

gregdingle: Ok so wouldn't `Visualizations.where(public_access: 1, user: current_user)` be the intersection? Don't we need an `or` in there somehow?

app/helpers/samples_helper.rb
tfrcarvalho: I guess I do not see a problem to force developers to go through the aggregated  `filter_samples` since it is pretty configurable.

app/controllers/visualizations_controller.rb
tfrcarvalho: Yes, we are correct. That is a bug.

PR #1988, opened 2019-03-08
[Upload Flow] UI adjustments

jshoe:
Items from the todo list:
- When you create a new project, don't select private or public for the user.
- Add "More Info" to local uploads so they can see the file formats.
- Select Human as the default host genome to get all the fields enabled visually.

mar-07-2019 11-55-32
screen shot 2019-03-07 at 9 36 58 pm

### DIFF
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -93,12 +93,6 @@ class MetadataManualInput extends React.Component {
     ];
   };
 
-  handleHostGenomeChange = id => {
-    this.setState({
-      currentHostGenome: id
-    });
-  };
-
   // Update metadata field based on user's manual input.
   updateMetadataField = (key, value, sample) => {
     const newHeaders = union([key], this.state.headersToEdit);
@@ -257,22 +251,6 @@ class MetadataManualInput extends React.Component {
               cs.extraPadding
           );
 
-          if (column === "Host Genome") {
-            return (
-              <div>
-                <Dropdown
-                  className={inputClasses}
-                  options={this.getHostGenomeOptions()}
-                  value={this.state.currentHostGenome}
-                  onChange={id => this.updateHostGenome(id, sample)}
-                  usePortal
-                  withinModal={this.props.withinModal}
-                />
-                {this.renderApplyToAll(sample, column)}
-              </div>
-            );
-          }
-
           const sampleHostGenomeId = this.props.samplesAreNew
             ? get(
                 "id",
@@ -282,6 +260,22 @@ class MetadataManualInput extends React.Component {
                 )
               )
             : sample.host_genome_id;
+
+          if (column === "Host Genome") {
+            return (
+              <div>
+                <Dropdown
+                  className={inputClasses}
+                  options={this.getHostGenomeOptions()}
+                  value={sampleHostGenomeId}
+                  onChange={id => this.updateHostGenome(id, sample)}
+                  usePortal
+                  withinModal={this.props.withinModal}
+                />
+                {this.renderApplyToAll(sample, column)}
+              </div>
+            );
+          }
 
           // Only show a MetadataInput if this metadata field matches the sample's host genome.
           if (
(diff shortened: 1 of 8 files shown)


app/assets/src/components/common/ProjectCreationForm.jsx
```diff
@@ -9,7 +9,7 @@ import cs from "./project_creation_form.scss";
 class ProjectCreationForm extends React.Component {
   state = {
     name: "",
-    publicAccess: 1,
+    publicAccess: -1, // No selection by default
```
MarkAZhang: I would disable the button until the user picks one of the options.

app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
```diff
@@ -54,6 +58,12 @@ class LocalSampleFileUpload extends React.Component {
         )} cannot be uploaded. Size must be under 5GB for local uploads. For larger files, please try our CLI.`
     );
 
+  toggleInfo = () => {
```
MarkAZhang: Can you move "Upload Your Input Files" out of the Filepicker so it looks the same as "Path to S3 Bucket"?

screen shot 2019-03-08 at 1 34 45 am

When files are uploaded, I would set the filePickerTitle to "X files uploaded," but make it empty-string when there are no files.

I would also remove "These samples were detected and matched. Remove samples you do not want to upload". It makes sense in the other tab, but it feels unnecessary here.

screen shot 2019-03-08 at 1 37 11 am

app/assets/src/components/common/MetadataManualInput.jsx
```diff
@@ -80,6 +80,9 @@ class MetadataManualInput extends React.Component {
         )
       }
     });
+
+    // Default to the first host genome (Human)
+    this.props.samples.map(sample => this.updateHostGenome(1, sample));
```
MarkAZhang: Looks like I had a bug in my code from before. `this.state.currentHostGenome` should be removed (and also `handleHostGenomeChange`). 

`<Dropdown value={this.state.currentHostGenome} />` should be changed to `<Dropdown value={sampleHostGenomeId} />`

As it is right now, "Human" will not show up in the host genome dropdowns at the start. They'll still be empty.

app/assets/src/components/common/ProjectCreationForm.jsx
jshoe: 👍

app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
jshoe: 👍 

Screen Shot 2019-03-08 at 12 23 32 PM
Screen Shot 2019-03-08 at 12 23 24 PM

app/assets/src/components/common/MetadataManualInput.jsx
jshoe: 👍

app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
MarkAZhang: Could we make the height of the filepicker smaller when there's no title?

app/assets/src/components/common/ProjectCreationForm.jsx
```diff
@@ -96,7 +96,9 @@ class ProjectCreationForm extends React.Component {
           <div
             className={cx(
               cs.createButton,
-              this.state.name === "" && cs.disabled
+              this.state.name === "" &&
+                this.state.publicAccess === -1 &&
+                cs.disabled
             )}
             onClick={this.handleCreateProject}
```
MarkAZhang: You'll also need to check `publicAccess` in `handleCreateProject`.

We should eventually turn this into a different style of "button", so you can just set disabled once, and it'll handle both the styling and the behavior.

app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
jshoe: 👍 

Screen Shot 2019-03-08 at 1 01 35 PM

app/assets/src/components/common/ProjectCreationForm.jsx
jshoe: 👍

PR #1991, opened 2019-03-10
JIRA 837

cdebourcy:
Filter human regardless of host choice. 
Goes with https://github.com/chanzuckerberg/idseq-dag/pull/127/files.

Put the new steps it at the end of host filtering to minimize the performance impact. 
Note: if the chosen host is not human, there will now be two "STAR_genome" and "bowtie2_genome" files used in the same AWS Batch job. To prevent collisions, I'm renaming the index files to include the host name.
TO DO:
- test if it works


tfrcarvalho:
We might want to save the date you push this. Would be interesting to see the impact on the statistics on pipeline time you collected a couple of weeks ago.

cdebourcy:
Yes! 100%

PR #1992, opened 2019-03-11
[Data Discovery] Visual fixes, fix projects stats, add owner to projects

tfrcarvalho:
* Visual fixes: some table styles
* Fix projects stats: add counter per project
* Add owner to projects: load project owner from project's lowest id sample


app/assets/src/components/common/Header.jsx
```diff
@@ -184,7 +184,7 @@ const MainMenu = () => {
         className={cx(cs.item, isSelected("library") && cs.selected)}
         href="/library"
       >
-        My Library
+        My Data
```
gregdingle: interesting change... where did this come from? are we still going to refer to it as the user library internally?

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -74,10 +74,9 @@ export default class DiscoverySidebar extends React.Component {
           // avg_reads_per_sample: 0,
         },
         metadata: {
-          // TODO (gdingle): these freq counts per project, not per sample
```
gregdingle: is this not a problem for the UI? are we expecting to show stats per project in the project view, not weighted per sample?

app/assets/src/components/views/projects/projects_view.scss
```diff
@@ -16,15 +16,20 @@
   .projectDescription {
     color: $dark-grey;
     font-size: $font-size-h6;
-    margin-bottom: 8px;
     line-height: 16px;
-    height: 32px;
+    // remove description for now:
+    height: 0;
+    overflow: hidden;
+    // height: 32px;
```
gregdingle: commented out because...? please add a comment in the code

app/controllers/projects_controller.rb
```diff
@@ -69,16 +69,23 @@ def index
...
         tissues_by_project_id = {}
-        samples.includes(:host_genome).each do |s|
-          (host_genome_names_by_project_id[s.project_id] ||= Set.new) << s.host_genome.name if s.host_genome && s.host_genome.name
+        min_sample_by_project_id = {}
+        owner_by_project_id = {}
+        samples.includes(:host_genome, :user).each do |s|
+          (host_genome_names_by_project_id[s.project_id] ||= Set.new) << s.host_genome.name if s.host_genome.name
           # TODO: sample_tissue column is deprecated, retrieve sample_type from Metadatum model instead
           (tissues_by_project_id[s.project_id] ||= Set.new) << s.sample_tissue if s.sample_tissue
+          if !min_sample_by_project_id[s.project_id] || min_sample_by_project_id[s.project_id] < s.id
```
gregdingle: why is it important to get the min sample? are the sample owners not always the same within a project? does the first sample have a special status?

app/assets/src/components/common/Header.jsx
tfrcarvalho: Joe/Vida complained. Library is used as a term for sample apparently.
I would like to change internally later, once the terms are fixed...

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
tfrcarvalho: I thought it would be per project, when in the projects page.

app/controllers/projects_controller.rb
tfrcarvalho: No. Anyone associated with the project can upload.

app/assets/src/components/views/projects/projects_view.scss
tfrcarvalho: Comment above, temporary removal.
Just while descriptions do not make it to the database. I can add that to the comment.

app/controllers/projects_controller.rb
tfrcarvalho: Btw, we should have project owner explicitly set... but that is not the case right now.

cdebourcy: Might be worth extracting the assignment of an owner into a function and/or commenting very explicitly, since we might want to change the ownership function in the future?

tfrcarvalho: I will go with comment for now.
I think what you say is actually valid for tissues and host_genomes. All three queries need to be refactored.

PR #2000, opened 2019-03-12
Add PNG download and white background for heatmaps

gregdingle:
# Description

image

As title. I learned that PNG is supported by the existing lib, but not jpg, which I think is fine. 

Also, I think I fixed the root cause of the SVG "compatibility" issues. Before they were downloaded with a transparent background which may appear as gray in some applications. I changed the background to white for the heatmap globally which I think is appropriate. If we ever want to change the page color background, it will be obvious what to do. 

# Test and Reproduce

1. click "download png"
2. see download
3. open download and see good png on white background
4. repeat for svg download

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapView.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView.jsx
@@ -690,7 +690,12 @@ class SamplesHeatmapView extends React.Component {
   handleDownloadClick = fileType => {
     switch (fileType) {
       case "svg":
+        // TODO (gdingle): pass in filename per sample?
         this.heatmapVis.download();
+        break;
+      case "png":
+        // TODO (gdingle): pass in filename per sample?
+        this.heatmapVis.downloadAsPng();
         break;
       case "csv":
         this.downloadCurrentViewDataURL();
@@ -717,7 +722,8 @@ class SamplesHeatmapView extends React.Component {
   render() {
     let downloadOptions = [
       { text: "Download CSV", value: "csv" },
-      { text: "Download SVG", value: "svg" }
+      { text: "Download SVG", value: "svg" },
+      { text: "Download PNG", value: "png" }
     ];
 
     return (
--- a/app/assets/src/components/visualizations/heatmap/Heatmap.js
+++ b/app/assets/src/components/visualizations/heatmap/Heatmap.js
@@ -56,7 +56,9 @@ export default class Heatmap {
         columnMetadata: [],
         metadataColorScale: new CategoricalColormap(),
         enableColumnMetadata: false,
-        iconPath: "/assets/icons"
+        iconPath: "/assets/icons",
+        // This is needed for downloading PNG and SVG on solid background
+        backgroundColor: "white"
       },
       options
     );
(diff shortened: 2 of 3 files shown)


app/assets/src/components/views/compare/SamplesHeatmapView.jsx
```diff
@@ -690,8 +690,13 @@ class SamplesHeatmapView extends React.Component {
   handleDownloadClick = fileType => {
     switch (fileType) {
       case "svg":
+        // TODO (gdingle): pass in filename per sample?
         this.heatmapVis.download();
         break;
+      case "png":
+        // TODO (gdingle): pass in filename per sample?
```
tfrcarvalho: What do you mean by these?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -56,7 +56,9 @@ export default class Heatmap {
         columnMetadata: [],
         metadataColorScale: new CategoricalColormap(),
         enableColumnMetadata: false,
-        iconPath: "/assets/icons"
+        iconPath: "/assets/icons",
+        // This is needed for downloading PNG and SVG on solid background
+        backgroundColor: "white"
```
tfrcarvalho: Can you call this `svgBackgroundColor` ?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -204,7 +206,10 @@ export default class Heatmap {
       .append("svg")
       .attr("class", cs.heatmap)
       .attr("id", "visualization")
-      .attr("xmlns", "http://www.w3.org/2000/svg");
+      .attr("xmlns", "http://www.w3.org/2000/svg")
+      // Not standard but it works for downloads and svgsaver. See:
+      // https://stackoverflow.com/questions/11293026/default-background-color-of-svg-root-element
+      .attr("style", "background-color: " + this.options.backgroundColor);
```
tfrcarvalho: Use interpolation here: `background-color: ${this.options.backgroundColor}`

jshoe:
PNG looks great although it still has the black background in my Illustrator. You could ask IT for a Creative Cloud account perhaps

Screen Shot 2019-03-12 at 2 06 12 PM

app/assets/src/components/views/compare/SamplesHeatmapView.jsx
gregdingle: Make the filename unique for the set of sample selected.

gregdingle:
> PNG looks great although it still has the black background in my Illustrator. You could ask IT for a Creative Cloud account perhaps. not blocking

@jshoe thanks for checking. Are you sure it was the PNG? It says "svg" in your screenshot.

jshoe:
@gregdingle sorry, meant to say "the SVG still has the black background".

app/assets/src/components/views/compare/SamplesHeatmapView.jsx
tfrcarvalho: I would not add comments to the code for things that we *might* consider doing... a JIRA task might be better for that... 
If you still keep the comment, at least change to "per sample set".

gregdingle: Yes, I put some of these comments in for discussion in review. I'll make this one a low pri JIRA task.

PR #2006, opened 2019-03-12
[Search] Sanitize queries to Elasticsearch

jshoe:
- Filter everything that isn't `a-zA-Z0-9 _|`

### DIFF
--- a/app/helpers/elasticsearch_helper.rb
+++ b/app/helpers/elasticsearch_helper.rb
@@ -38,6 +38,6 @@ module ElasticsearchHelper
   private
 
   def sanitize(prefix)
-    prefix.gsub(/[^a-zA-Z0-9 _|]/, "")
+    prefix.gsub(/[^a-zA-Z0-9 _|]/, "") if prefix
   end
 end
--- a/app/helpers/elasticsearch_helper.rb
+++ b/app/helpers/elasticsearch_helper.rb
@@ -38,6 +38,7 @@ module ElasticsearchHelper
   private
 
   def sanitize(prefix)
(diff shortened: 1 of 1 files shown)


cdebourcy:
Is there a more native way of doing this? Has anyone ever encountered this problem before when using elasticsearch-rails?

jshoe:
Looks like it's not present in that library and it's recommended to sanitize ourselves https://discuss.elastic.co/t/ruby-elasticsearch-ruby-special-characters-not-escaped-by-the-library/15187/3

cdebourcy:
great

app/helpers/elasticsearch_helper.rb
```diff
@@ -32,4 +34,10 @@ def taxon_search(prefix, tax_levels = TaxonCount::NAME_2_LEVEL.keys)
     end
     matching_taxa.values
   end
+
+  private
+
+  def sanitize(prefix)
+    prefix.gsub(/[^a-zA-Z0-9 _|]/, "") if prefix
```
tfrcarvalho: You should add at least `.` to the mix.
Even then, I think the best solution is to escape the special characters: http://lucene.apache.org/core/3_0_3/queryparsersyntax.html#Escaping%20Special%20Characters

jshoe: 👍 Ok modified this to add backslash in front of special characters. Tested that it works for ones like idseq-bench-4|

Side note, posts like this one point out that query_string is the most raw field http://www.recursion.org/query-parser/#user_input_may_contain_special_characters

PR #2008, opened 2019-03-13
Use relevant sample names in Metadata CSV template when possible.

MarkAZhang:
When uploading new samples, use the names of the new samples.
When uploading to existing samples, use the names of existing samples.

Screen Shot 2019-03-12 at 6 50 08 PM

Verified that we are making peformant queries.

Here is 325 samples from Mosquito project. Endpoint takes about 1 second (most of it seems to be generating the CSV file)
Screen Shot 2019-03-12 at 6 52 19 PM

On other pages such as the CSV upload instructions, we still return example samples based on the host genomes, like before.

Screen Shot 2019-03-12 at 6 55 22 PM

----

This relates to stabilization because we need to generate large test files to test our metadata upload and validation, which this PR enables.

### DIFF
--- a/app/assets/src/components/common/MetadataUpload.jsx
+++ b/app/assets/src/components/common/MetadataUpload.jsx
@@ -11,6 +11,7 @@ import { getProjectMetadataFields, getAllHostGenomes } from "~/api";
 import cs from "./metadata_upload.scss";
 import MetadataManualInput from "./MetadataManualInput";
 import IssueGroup from "./IssueGroup";
+import { getURLParamString } from "~/helpers/url";
 
 const map = _fp.map.convert({ cap: false });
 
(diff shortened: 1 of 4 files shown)


app/assets/src/components/common/MetadataUpload.jsx
```diff
@@ -60,6 +61,17 @@ class MetadataUpload extends React.Component {
     this.props.onMetadataChange({ metadata, wasManual: true });
   };
 
+  getCSVUrl = () => {
```
tfrcarvalho: I wonder if building these urls should also be part of the `api`.
Maybe on `api/urls.js` ? Wdyt?

MarkAZhang: I think we can hold off for now, as these dynamic `<a>` urls are pretty rare and we already have a re-usable `getURLParamString` function. Maybe in the future.

PR #2012, opened 2019-03-13
Fix perf of report table render

gregdingle:
# Description

See https://jira.czi.team/projects/IDSEQ/issues/IDSEQ-723.

This reduces render time by 2/3 (or 300ms) in my testing by eliminating rendering of hidden elements. The showing is now handled by JS. 

This might also fix large tables not rendering at all. Please point me to a good example. @MarkAZhang 

image

# Test and Reproduce

1. Using chrome dev tools, record time switching views in report page
2. See time decrease 300ms before and after

1. hover over row
2. see links appear as before

### DIFF
--- a/app/assets/src/components/views/report/ReportTable/DetailCells.jsx
+++ b/app/assets/src/components/views/report/ReportTable/DetailCells.jsx
@@ -4,67 +4,79 @@ import { get } from "lodash/fp";
 // TODO(mark): Use alias instead, to avoid dots.
 import PropTypes from "../../../utils/propTypes";
 
-const DetailCells = ({
-  taxons,
-  taxonRowRefs,
-  renderName,
-  renderNumber,
-  showConcordance,
-  getRowClass,
-  onTaxonClick,
-  reportDetails,
-  backgroundData,
-  showAssemblyColumns
-}) => {
-  return taxons.map(taxInfo => (
-    <tr
-      key={taxInfo.tax_id}
-      id={`taxon-${taxInfo.tax_id}`}
-      ref={elm => {
-        taxonRowRefs[taxInfo.tax_id] = elm;
-      }}
-      className={getRowClass(taxInfo)}
-    >
-      <td>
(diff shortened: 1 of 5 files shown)


app/assets/src/components/views/report/ReportTable/DetailCells.jsx
```diff
@@ -4,67 +4,79 @@ import { get } from "lodash/fp";
 // TODO(mark): Use alias instead, to avoid dots.
 import PropTypes from "../../../utils/propTypes";
 
-const DetailCells = ({
```
gregdingle: These changes are all auto-format induced.

app/assets/src/components/views/report/ReportTable/DetailCells.jsx
```diff
@@ -4,67 +4,79 @@ import { get } from "lodash/fp";
...
+    return taxons.map(taxInfo => (
+      <tr
+        key={taxInfo.tax_id}
+        id={`taxon-${taxInfo.tax_id}`}
+        ref={elm => {
+          taxonRowRefs[taxInfo.tax_id] = elm;
+        }}
+        className={getRowClass(taxInfo)}
+        onMouseEnter={() => handleMouseEnter(taxInfo.tax_id)}
+        onMouseLeave={() => handleMouseLeave(taxInfo.tax_id)}
```
MarkAZhang: Can `taxInfo` ever be `undefined`? We have special rows like "All taxa with neither family nor genus classification" and "Non-genus-specific reads in family Sinobacteraceae" which don't correspond to a single taxon. Also, can `taxInfo.tax_id` be `null` or `undefined` for multiple rows? (I don't know the answer, but it would be good to check.) If so, that might cause multiple rows to show their `<HoverActions>` when you hover on any single one of them.

cdebourcy: “Non-specific” entries correspond to negative taxids. I don’t think they should be undefined

PR #2014, opened 2019-03-13
Fix accordion yet again

gregdingle:
# Description

image

Embarrassing! This fixes the accordion yet again, and was a lesson in React. I think the biggest lesson was not to make components again that can have the same behavior be both controlled either by props or by internal state.  It is too hard to think about. 

# Test and Reproduce

1. Try accordion on sample details sidebar
2. try accordion on DD sidebar

### DIFF
--- a/app/assets/src/components/layout/Accordion.jsx
+++ b/app/assets/src/components/layout/Accordion.jsx
@@ -4,23 +4,19 @@ import cx from "classnames";
 import cs from "./accordion.scss";
 
 class Accordion extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      open: this.props.open
-    };
-  }
+  state = {};
 
   onToggle = () => {
     this.setState({
-      open: !this.state.open
+      open: !this.state.open,
+      wasToggled: true
     });
   };
 
   render() {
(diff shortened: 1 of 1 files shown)


app/assets/src/components/layout/Accordion.jsx
```diff
@@ -4,23 +4,19 @@ import cx from "classnames";
...
+      open: !this.state.open,
+      wasToggled: true
     });
   };
 
   render() {
     const { header, children, toggleable, className } = this.props;
 
-    const open = this.state.open;
+    const open = this.state.wasToggled ? this.state.open : this.props.open;
```
gregdingle: still not happy with this fix but it works

MarkAZhang: I think a cleaner approach would be to:

Add a `defaultOpen` prop that allows you to set the `open` state on the initial component mount, but gives control over the component afterwards.

In `<DiscoverySidebar>`, don't render (or mount) the Accordions until the data has been loaded. Show a loading message when the data is still being fetched. This way, you can set the `defaultOpen` prop properly once the data is loaded (since you only want the Accordion to be open if there is data is show).

PR #2027, opened 2019-03-15
[Taxon Byteranges] Fix getter functions for fasta downloads

jshoe:
- Going to put up another PR later with more fixes but this is the safest
- First change is changing a `pipeline_run.taxon_byteranges.find_by(taxid: taxid, hit_type: hit_type)` to `pipeline_run.taxon_byteranges.where(taxid: taxid, hit_type: hit_type).last` because it appears we had pipeline runs inserting byteranges from the previous run and then the correct one being inserted later. This enables a "quick fix" by calling db_load_byteranges on all the pipeline runs again and then fetching the last correct one. Then we will clean up all the duplicates (saving the last one) and correct the unique index on (pipeline_run_id, taxid, hit_type) since there's a unique index (pipeline_run_id, taxid, hit_type, tax_level) that's useless.
- Second is fixing an assumption in s3_file_for that was saying `if pipeline_version and assembly-enabled, load the new path; else load this old file` so the old file was being loaded when pipeline_version was not yet set in a new pipeline run. Add a log line in ResultMonitorLoader for pipeline version. Migrate s3_file_for to just use the new assembly paths to deprecate running < 3.1.
- Will need some more test jobs in staging.

### How to reproduce
- Go to a sample on this broken list: https://jira.czi.team/browse/IDSEQ-881
- Go to the console and run pipeline_run.db_load_byteranges
- Call `pipeline_run.taxon_byteranges.where(taxid: taxid, hit_type: hit_type)` and notice that the last entry is the correct one. Example: `=> #<ActiveRecord::AssociationRelation [#<TaxonByterange id: 132270753, taxid: 12059, first_byte: 10670019, last_byte: 10900167, created_at: nil, updated_at: nil, hit_type: "NT", tax_level: nil, pipeline_run_id: 23682>, #<TaxonByterange id: 134558333, taxid: 12059, first_byte: 4675807, last_byte: 4839836, created_at: nil, updated_at: nil, hit_type: "NT", tax_level: nil, pipeline_run_id: 23682>]>`
- Grabbing the last byterange will result in a valid fasta download (or deleting all but the last one)

### DIFF
--- a/app/jobs/result_monitor_loader.rb
+++ b/app/jobs/result_monitor_loader.rb
@@ -3,12 +3,8 @@ class ResultMonitorLoader
   @queue = :q03_pipeline_run
 
   def self.perform(pipeline_run_id, output)
-    Rails.logger.info("Loading #{output} for pipeline run #{pipeline_run_id}")
     pr = PipelineRun.find(pipeline_run_id)
-
-    # Just return if version is not yet set (likely a run just starting)
-    return if pr.pipeline_version.blank?
-
+    Rails.logger.info("Loading #{output} for pipeline run #{pipeline_run_id} (v#{pr.pipeline_version})")
     output_state = pr.output_states.find_by(output: output)
     begin
       output_state.update(state: PipelineRun::STATUS_LOADING)
--- a/app/jobs/result_monitor_loader.rb
+++ b/app/jobs/result_monitor_loader.rb
@@ -5,6 +5,10 @@ class ResultMonitorLoader
   def self.perform(pipeline_run_id, output)
     Rails.logger.info("Loading #{output} for pipeline run #{pipeline_run_id}")
     pr = PipelineRun.find(pipeline_run_id)
+
+    # Just return if version is not yet set (likely a run just starting)
+    return if pr.pipeline_version.blank?
+
     output_state = pr.output_states.find_by(output: output)
     begin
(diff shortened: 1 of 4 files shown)


app/jobs/result_monitor_loader.rb
```diff
@@ -5,6 +5,10 @@ class ResultMonitorLoader
   def self.perform(pipeline_run_id, output)
     Rails.logger.info("Loading #{output} for pipeline run #{pipeline_run_id}")
     pr = PipelineRun.find(pipeline_run_id)
+
+    # Just return if version is not yet set (likely a run just starting)
+    return if pr.pipeline_version.blank?
```
cdebourcy: Why does this line not take care of it?
https://github.com/chanzuckerberg/idseq-web/blob/dbbc042c6a60958ee163f705262d01b3cb827c7b/app/models/pipeline_run.rb#L746

jshoe: Hmm yeah maybe it isn't the real issue. Let me replace this with a log line to monitor instead

jshoe: updated 👍

MarkAZhang:
We discussed adding a comment about our assumption that pipeline_version > 3.1 in get_s3_file

PR #2028, opened 2019-03-15
Add validation for file type on step one of Sample Upload.

MarkAZhang:
Screen Shot 2019-03-15 at 12 10 24 PM
Screen Shot 2019-03-15 at 12 10 29 PM

---

Testing strategy:

Verified that local and remote upload still work. 
Verified that sample name validation works regardless of whether you pick a project before or after.


MarkAZhang:
I tried to avoid copying the FILE_REGEX to the frontend just so we wouldn't have the same variable in two places.

Your idea about a set of small global constants that we automatically send from Rails to JS with every request sounds great.

Yeah, Upload from S3 already rejects invalid files before it returns the list.

PR #2029, opened 2019-03-15
Restrict phylo tree taxons to those present in selected project

gregdingle:
# Description

This fixes the problem of trees on organisms that have no samples. 

image

Perf was a concern so I made sure it was fast enough for a big project ("medical detectives") at ~250ms.

Note: I'm not sure if the SQL query *exactly* defines what we want. Please advise.

# Test and Reproduce

1. Try endpoint with and without project_id, see different set
http://localhost:3000/choose_taxon?query=rub&args=species,genus
http://localhost:3000/choose_taxon?query=rub&args=species,genus&project_id=70

2. Try autocomplete with project, see results

3. Try autocomplete without project, see results

### DIFF
--- a/app/assets/src/components/ui/controls/SearchBox.jsx
+++ b/app/assets/src/components/ui/controls/SearchBox.jsx
@@ -129,7 +129,7 @@ SearchBox.propTypes = {
   // If serverSearchAction is provided, query matching will happen on the server side (use for large data).
   clientSearchSource: PropTypes.array,
   serverSearchAction: PropTypes.string,
-  serverSearchActionArgs: PropTypes.string,
+  serverSearchActionArgs: PropTypes.object,
   rounded: PropTypes.bool,
   category: PropTypes.bool,
   levelLabel: PropTypes.bool,
--- a/app/assets/src/components/ui/controls/SearchBox.jsx
+++ b/app/assets/src/components/ui/controls/SearchBox.jsx
@@ -1,10 +1,12 @@
 import React from "react";
 import cx from "classnames";
-import cs from "./search_box.scss";
+import PropTypes from "prop-types";
 import { Search } from "semantic-ui-react";
 import { escapeRegExp, debounce } from "lodash";
-import PropTypes from "prop-types";
+
+import cs from "./search_box.scss";
 import { get } from "../../../api";
+import { getURLParamString } from "~/helpers/url";
 
 class SearchBox extends React.Component {
   constructor(props) {
@@ -70,10 +72,9 @@ class SearchBox extends React.Component {
         const isMatch = result => re.test(result.title);
         searchResults = this.props.clientSearchSource.filter(isMatch);
       } else if (this.props.serverSearchAction) {
-        // TODO: (gdingle): pass in project filter here... append to CSV?
         let url = `/${this.props.serverSearchAction}?query=${this.state.value}`;
         if (this.props.serverSearchActionArgs) {
-          url += `&args=${this.props.serverSearchActionArgs}`;
+          url += `&${getURLParamString(this.props.serverSearchActionArgs)}`;
         }
         searchResults = await get(url);
         if (this.props.levelLabel) {
--- a/app/assets/src/components/ui/controls/SearchBox.jsx
+++ b/app/assets/src/components/ui/controls/SearchBox.jsx
@@ -70,6 +70,7 @@ class SearchBox extends React.Component {
         const isMatch = result => re.test(result.title);
         searchResults = this.props.clientSearchSource.filter(isMatch);
       } else if (this.props.serverSearchAction) {
+        // TODO: (gdingle): pass in project filter here... append to CSV?
         let url = `/${this.props.serverSearchAction}?query=${this.state.value}`;
         if (this.props.serverSearchActionArgs) {
           url += `&args=${this.props.serverSearchActionArgs}`;
--- a/app/helpers/elasticsearch_helper.rb
+++ b/app/helpers/elasticsearch_helper.rb
@@ -45,7 +45,6 @@ module ElasticsearchHelper
 
   # Took 250ms in local testing on real data
   def filter_by_project(matching_taxa, project_id)
-    # TODO: (gdingle): do we want both species and genus? only genus appears to be returned by TaxonCount
     project_tax_ids = Set.new(TaxonCount
       .joins(pipeline_run: { sample: :project })
       .where(samples: { project_id: project_id })
--- a/app/helpers/elasticsearch_helper.rb
+++ b/app/helpers/elasticsearch_helper.rb
@@ -15,15 +15,13 @@ module ElasticsearchHelper
     results
   end
 
-  def taxon_search(prefix, tax_levels = TaxonCount::NAME_2_LEVEL.keys)
+  def taxon_search(prefix, tax_levels = TaxonCount::NAME_2_LEVEL.keys, project_id = nil)
     return {} if Rails.env == "test"
     prefix = sanitize(prefix)
     matching_taxa = {}
 
     tax_levels.each do |level|
-      # TODO: (gdingle): pass in project filter here
       search_params = { query: { query_string: { query: "#{prefix}*", fields: ["#{level}_name"] } } }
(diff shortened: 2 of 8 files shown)


app/helpers/elasticsearch_helper.rb
```diff
@@ -32,11 +33,32 @@ def taxon_search(prefix, tax_levels = TaxonCount::NAME_2_LEVEL.keys)
...
   private
 
+  # Took 250ms in local testing on real data
+  def filter_by_project(matching_taxa, project_id)
+    # TODO: (gdingle): do we want both species and genus? only genus appears to be returned by TaxonCount
+    project_tax_ids = Set.new(TaxonCount
+      .joins(pipeline_run: { sample: :project })
+      .where(samples: { project_id: project_id })
+      .where(tax_id: [matching_taxa.map { |taxa| taxa["taxid"] }])
+      .where("tax_id > 0") # negative numbers mean unknown... see TaxonLineage
```
gregdingle: should we also apply this to elasticsearch above?

app/assets/src/components/ui/controls/SearchBox.jsx
```diff
@@ -72,7 +74,7 @@ class SearchBox extends React.Component {
       } else if (this.props.serverSearchAction) {
         let url = `/${this.props.serverSearchAction}?query=${this.state.value}`;
         if (this.props.serverSearchActionArgs) {
-          url += `&args=${this.props.serverSearchActionArgs}`;
+          url += `&${getURLParamString(this.props.serverSearchActionArgs)}`;
```
gregdingle: there were no other callsites that used `serverSearchActionArgs`

app/helpers/elasticsearch_helper.rb
cdebourcy: I think it's unnecessary... if searches by name, so if it has a name for a level than it also has a positive taxid

app/helpers/elasticsearch_helper.rb
```diff
@@ -32,11 +33,31 @@ def taxon_search(prefix, tax_levels = TaxonCount::NAME_2_LEVEL.keys)
...
 
+  # Took 250ms in local testing on real data
+  def filter_by_project(matching_taxa, project_id)
+    project_tax_ids = Set.new(TaxonCount
+      .joins(pipeline_run: { sample: :project })
+      .where(samples: { project_id: project_id })
+      .where(tax_id: [matching_taxa.map { |taxa| taxa["taxid"] }])
+      .where("tax_id > 0") # negative numbers mean unknown... see TaxonLineage
+      .where(count_type: ["NT", "NR"])
+      .where("count > #{ReportHelper::MINIMUM_READ_THRESHOLD}")
```
cdebourcy: prefer `.where("count > ?", ReportHelper::MINIMUM_READ_THRESHOLD)` as a general practice :D

cdebourcy: for security reasons

app/helpers/elasticsearch_helper.rb
```diff
@@ -32,11 +33,31 @@ def taxon_search(prefix, tax_levels = TaxonCount::NAME_2_LEVEL.keys)
...
+    if project_id.present?
+      filter_by_project(matching_taxa.values, project_id)
+    else
+      matching_taxa.values
+    end
   end
 
   private
 
+  # Took 250ms in local testing on real data
```
cdebourcy: Local testing taxon_counts table will be orders of magnitude smaller than prod though, just FYI... still needs some testing on actual prod size data. Even staging only has a small subset of prod's taxon_counts

cdebourcy: If it holds up in prod, amazing!

gregdingle: ok sorry I didn't realize.... I'll test this query in the rails console in prod.

gregdingle: The explain results look good. 

```=> EXPLAIN for: SELECT DISTINCT `taxon_counts`.* FROM `taxon_counts` INNER JOIN `pipeline_runs` ON `pipeline_runs`.`id` = `taxon_counts`.`pipeline_run_id` INNER JOIN `samples` ON `samples`.`id` = `pipeline_runs`.`sample_id` INNER JOIN `projects` ON `projects`.`id` = `samples`.`project_id` WHERE `samples`.`project_id` = 70 AND `taxon_counts`.`tax_id` IN (59512, 11040, 1649490, 182402, 28067) AND (tax_id > 0) AND `taxon_counts`.`count_type` IN ('NT', 'NR') AND (count > 5)
+----+-------------+---------------+-------+------------------------------------------+----------------------------------+---------+-----------------------------+------+------------------------------------+
| id | select_type | table         | type  | possible_keys                            | key                              | key_len | ref                         | rows | Extra                              |
+----+-------------+---------------+-------+------------------------------------------+----------------------------------+---------+-----------------------------+------+------------------------------------+
|  1 | SIMPLE      | projects      | const | PRIMARY                                  | PRIMARY                          | 8       | const                       |    1 | Using index; Using temporary       |
|  1 | SIMPLE      | samples       | ref   | PRIMARY,index_samples_name_project_id    | index_samples_name_project_id    | 9       | const                       |   15 | Using index                        |
|  1 | SIMPLE      | pipeline_runs | ref   | PRIMARY,index_pipeline_runs_on_sample_id | index_pipeline_runs_on_sample_id | 9       | idseq_prod.samples.id       |    1 | Using index                        |
|  1 | SIMPLE      | taxon_counts  | ref   | index_pr_tax_hit_level_tc                | index_pr_tax_hit_level_tc        | 9       | idseq_prod.pipeline_runs.id |  216 | Using index condition; Using where |
+----+-------------+---------------+-------+------------------------------------------+----------------------------------+---------+-----------------------------+------+------------------------------------+
```

cdebourcy: yay!

cdebourcy:
Btw this might be for a different PR but this same search box is used on the report page. Obviously, on the report page, you would want to restrict the suggestions to only that sample

gregdingle:
> Btw this might be for a different PR but this same search box is used on the report page. Obviously, on the report page, you would want to restrict the suggestions to only that sample

Good point. I'll look now.

gregdingle:
So it's trivial to add the filter by project to the report page which I did because it is better than the current. I need to go now so I'd prefer to leave the sample filter until later.

app/helpers/elasticsearch_helper.rb
```diff
@@ -46,12 +46,12 @@ def taxon_search(prefix, tax_levels = TaxonCount::NAME_2_LEVEL.keys, project_id
   # Took 250ms in local testing on real data
   def filter_by_project(matching_taxa, project_id)
     project_tax_ids = Set.new(TaxonCount
-      .joins(pipeline_run: { sample: :project })
+      .joins(pipeline_run: :sample)
       .where(samples: { project_id: project_id })
       .where(tax_id: [matching_taxa.map { |taxa| taxa["taxid"] }])
       .where("tax_id > 0") # negative numbers mean unknown... see TaxonLineage
       .where(count_type: ["NT", "NR"])
-      .where("count > ?", ReportHelper::MINIMUM_READ_THRESHOLD)
```
tfrcarvalho: Why change to zero?

app/helpers/elasticsearch_helper.rb
```diff
@@ -32,11 +33,31 @@ def taxon_search(prefix, tax_levels = TaxonCount::NAME_2_LEVEL.keys)
         }
       end
     end
-    matching_taxa.values
+
+    if project_id.present?
+      filter_by_project(matching_taxa.values, project_id)
+    else
+      matching_taxa.values
+    end
```
tfrcarvalho: Do you know how many results elastic search returns?
What would happen if none of the results were in the project?

tfrcarvalho: I guess it would just have to wait for the next character for more suggestions?

gregdingle:
> Any clue of the consequences of changing it to a list of projects (or samples) on performance?

The query is indexed so the perf impact should be log N projects.

gregdingle:
@cdebourcy could I get a review?

PR #2036, opened 2019-03-18
Make samples selectable and add actions.

tfrcarvalho:
* Adds the ability to select samples and executes action
* New icon buttons (save, heatmap, phylo tree, download)
* Added for both Table and InfiniteTable although DD only uses the latter
* New components ported from Samples.jsx:
	- CollectionModal is a refactored BackgroundModal
	- ReportsDownloader: new component that aggregates download functions

Mar-18-2019 16-07-12


app/assets/src/api/index.js
```diff
@@ -15,8 +15,8 @@ const postWithCSRF = async (url, params) => {
     // Just return the data.
     // resp also contains headers, status, etc. that we might use later.
     return resp.data;
-  } catch (e) {
-    return Promise.reject(e.response.data);
+  } catch (error) {
+    return Promise.reject(error);
```
jshoe: I know the new upload flow uses this so you might want to check downstream to make sure they're OK

tfrcarvalho: Reverted.
Do not really need the change.

PR #2037, opened 2019-03-18
Configure airbrake for new "performance dashboard"

gregdingle:
# Description

Airbrake just released this new feature. Given our interest in perf, I though it would be a good addition. 

# Test

docker-compose build web


gregdingle:
No, the feature is actually pre release but the airbrake rep said we would get it once we added the gem.

Best, Greg

PR #2038, opened 2019-03-18
Fix conflict of airbrake and airbrake-ruby.

tfrcarvalho:
* Bundle update to fix conflict between gems add/updated in #2029


gregdingle:
Sorry about that. Looks good to me.

tfrcarvalho:
#2037 fixed it.

PR #2039, opened 2019-03-19
Make panes scroll independently

tfrcarvalho:
* Left and right panes scroll independently
* Center pane / table matches size of view port

Mar-19-2019 10-08-38


jshoe:
mentioned removing the error.response.data change

PR #2044, opened 2019-03-19
[Local Uploads] Add axios-retry for sample uploading

jshoe:
- Add this package for axios retries (https://github.com/softonic/axios-retry). Other package retry-axios was out of date with current axios version.
- Swap in an uploadFileToUrlWithRetries function with more resilient settings (experimented locally).

### Test/reproduce

- Used Network Link Conditioner to cut off connections until they would time out and tested with bogus URLs.

#### Example showing it retried 5 times for network error (via blocked URL)
Screen Shot 2019-03-19 at 3 20 16 PM
<br>

#### Example with 5 retries (2 files) for network error (via bogus URL)
Screen Shot 2019-03-19 at 3 45 33 PM
<br>

#### Example on the new sample flow page
Screen Shot 2019-03-19 at 4 08 27 PM


jshoe:
tyty

PR #2047, opened 2019-03-20
[S3 Copy] Ignore broken pipe errors as a side effect of piping to head

jshoe:
- See comments in changes

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -21,7 +21,6 @@ class Sample < ApplicationRecord
 
   TOTAL_READS_JSON = "total_reads.json".freeze
   LOG_BASENAME = 'log.txt'.freeze
-  S3_CP_PIPE_ERROR = '[Errno 32] Broken pipe'.freeze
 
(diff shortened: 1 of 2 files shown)


app/models/sample.rb
```diff
@@ -256,7 +256,16 @@ def initiate_s3_cp
...
                 end
       _stdout, stderr, status = Open3.capture3(command)
-      stderr_array << stderr unless status.exitstatus.zero? && stderr.empty?
+      unless status.exitstatus.zero? && stderr.empty?
+        # HACK: Ignore 'Broken pipe' error because it will always be written to stderr unless you
+        # 'head' the entire file. Exitstatus would still be 0 because it is the status of the last
+        # command and the intermediate statuses are not checked. We still want to check stderr for
+        # errs like HeadObject Forbidden.
+        # TODO: Consider refactoring this streaming copy to avoid hacks.
+        unless stderr.include?(InputFile::S3_CP_PIPE_ERROR)
```
yunfangjuan: add an airbrake error so we can do some internal tracking?

jshoe: these get logged and airbraked a few lines down

PR #2049, opened 2019-03-20
Add access control tests for metadata endpoints.

MarkAZhang:
Also modify 'validate' endpoints to be edit actions, since they are tied to a write action ('upload')

Verified that uploading to new samples and to existing samples still works. Verify that performance via SQL queries was not impacted.

**Added tests**:

bulkUploadWithMetadata

- Test that users can’t upload to private project.
- Test that users can’t upload to public project.
- Test that users can upload to their own project.

getOfficialMetadataFields

- Always the same for all users. No custom metadata is displayed, so no access control tests needed.

getSampleMetadata

- Users can see metadata of their own sample.
- User can see metadata for public sample.
- User can see metadata for expired sample.
- User cannot see metadata of private sample.

getSampleMetadataFields

- Users can see metadata field of their own sample.
- User can see metadata field for public sample.
- User can see metadata field for expired sample.
- User cannot see metadata  field of private sample.
- If multiple samples, just show visible ones.

getProjectMetadataFIelds

- Users can see metadata field of their own project.
- User can see metadata field for public project.
- User cannot see metadata field of private project.
- If multiple project, just show visible ones.

saveSampleMetadata

- Users can save metadata field of their own sample.
- User cannot save to anything else (expired, public, private).

uploadMetadataForProject

- Can upload to own project.
- Cannot upload to public or private project.

validateManualMetadataForProject

- Can validate against own project
- Cannot validate against public or private project.

validateManualMetadataForNewSamples

- Can validate against own project
- Cannot validate against public or private project.

### DIFF
--- a/test/controllers/samples_controller_test.rb
+++ b/test/controllers/samples_controller_test.rb
@@ -4,12 +4,18 @@ class SamplesControllerTest < ActionDispatch::IntegrationTest
   setup do
     @background = backgrounds(:real_background)
     @sample = samples(:one)
+    @sample_human_existing_metadata_public = samples(:sample_human_existing_metadata_public)
+    @metadata_validation_sample_human_existing_metadata = samples(:metadata_validation_sample_human_existing_metadata)
+    @sample_human_existing_metadata_joe_project = samples(:sample_human_existing_metadata_joe_project)
+    @sample_human_existing_metadata_expired = samples(:sample_human_existing_metadata_expired)
     @deletable_sample = samples(:deletable_sample)
     @project = projects(:one)
     @user = users(:one)
     @user.authentication_token = 'sdfsdfsdff'
     @user.save
     @user_params = { 'user[email]' => @user.email, 'user[password]' => 'password' }
+    @user_nonadmin = users(:joe)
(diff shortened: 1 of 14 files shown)


test/controllers/samples_controller_test.rb
```diff
@@ -178,4 +184,153 @@ class SamplesControllerTest < ActionDispatch::IntegrationTest
...
+      field: "sample_type",
+      value: "Foobar Sample Type"
+    }
+
+    assert_response :success
+
+    assert_equal "Foobar Sample Type", @sample_human_existing_metadata_joe_project.metadata.find_by(key: "sample_type").raw_value
+  end
+
+  test 'save_metadata_fields non-admin user private sample' do
```
cdebourcy: could we stick with our name convention like "joe cannot save metadata fields to private sample"? :) It makes it a lot easier to see what's going on and what the implications are if a particular test fails

MarkAZhang: Sure.

MarkAZhang: Fixed. It does look much better!

PR #2055, opened 2019-03-21
Show project and viz name in viz list, overwrite on save

gregdingle:
# Description

image

This adds the info to the backend needed for the ultimate design, and it changes the save behavior from append to overwrite.

NOTE: the table render design still needs work. 

# Test and Reproduce

1. try endpoint, http://localhost:3000/visualizations.json?domain=library
2. see data for both heatmap and phylo tree
3. see correct titles for  heatmap and phylo tree
4. try same in viz list UI


gregdingle:
I changed the PR so that the name of a visualization is stored in the database, eliminating the N+1 queries at read time. Likewise, the sample ids of a phylo tree are stored.

To make sense of the class, I refactored phylo tree methods such as `job_command` to be all private when not called outside the class  . Please sanity check.

app/models/visualization.rb
```diff
@@ -3,4 +3,20 @@ class Visualization < ApplicationRecord
...
   has_and_belongs_to_many :samples
   belongs_to :user
+  validates :name, presence: true
+  validates :data, presence: true
+
+  delegate :count, to: :samples, prefix: true
+
+  # In the common case, a visualization will come from a single project.
+  def project_name
+    if samples.length == 1
```
tfrcarvalho: Again, this if branch is not necessary

app/controllers/visualizations_controller.rb
```diff
@@ -50,24 +52,33 @@ def visualization
     end
   end
 
+  # TODO: (gdingle): overwrite on save
   def save
     @type = params[:type]
```
jshoe: I just realized this controller should probably use strong params (like visualizations_params with permitted whitelist) instead of using params directly

app/models/phylo_tree.rb
```diff
@@ -125,6 +131,50 @@ def monitor_job(throttle = true)
...
+  # Because phylo_trees are created explictly, the user will expect them to persist.
+  # TODO: (gdingle): destroy visualization objects when phylo_tree is destroyed,
+  # and rename when renamed.
+  def create_visualization
+    Visualization.create(
+      user: user,
+      visualization_type: "phylo_tree",
+      data: { treeId: id },
+      name: name,
+      # Samples of a PhyloTree are all samples of the project at the time of creation.
```
jshoe: I thought there was a step in the phylo tree wizard for adding additional samples that could be from other projects?

app/controllers/visualizations_controller.rb
gregdingle: Yea, good point. I only learned about strong param after I wrote this controller.

app/models/phylo_tree.rb
gregdingle: I didn't know about this. I'll check.

app/models/visualization.rb
gregdingle: because to_sentence will return the same result on an array of length 1? I wasn't sure so I thought this was safer

app/models/phylo_tree.rb
tfrcarvalho: I can confirm that @jshoe is correct.

gregdingle: Thanks, I'll change the PR to use the selected samples for the phylo tree.

app/models/visualization.rb
```diff
@@ -3,4 +3,68 @@ class Visualization < ApplicationRecord
...
+    elsif samples.length == 1
+      samples[0].name
+    elsif samples.length > 1
+      names = samples.map(&:name)
+      # make a string such as "Patient 016 (CSF) and 015 (CSF)"
+      prefix = longest_common_prefix(names)
+      if prefix.include?(' ')
+        # Use whole words only
+        # TODO: (gdingle): add other delimiters than space
+        prefix = prefix.split(' ')[0..-2].join(' ')
```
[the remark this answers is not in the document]
gregdingle: The problem here is that the prefix may include partial words such "patient 00" making "patient 001 and 2" when I think it's more natural to write "patient 001 and 002" . I think I will change `longest_common_prefix` to fix the root cause.

app/models/visualization.rb
```diff
@@ -3,4 +3,68 @@ class Visualization < ApplicationRecord
   serialize :data, JSON
   has_and_belongs_to_many :samples
   belongs_to :user
+
+  def samples_count
+    if visualization_type == "phylo_tree"
+      # WARNING: this could be an N+1 query
+      # TODO: (gdingle): store at write time if it is immutable
+      Sample.where(project: phylo_tree.project).count
```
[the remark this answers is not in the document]
gregdingle: this is fixed now

app/models/visualization.rb
```diff
@@ -3,4 +3,68 @@ class Visualization < ApplicationRecord
...
+    end
+  end
+
+  # In the common case, a visualization will come from a single project.
+  def project_name
+    # WARNING: this could be an N+1 query
+    # TODO: (gdingle): store name at write time if it is immutable
+    if visualization_type == "phylo_tree"
+      phylo_tree.project.name
+    elsif samples.length == 1
```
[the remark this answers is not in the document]
gregdingle: this is fixed now

gregdingle:
> Did you discuss this naming strategy with Jenn?

No, that's a good idea. @happyimadesignr: I came up with an auto naming algorithm for showing the heatmaps in the visualization list.   It will makes a string such as "Patient 016 (CSF) and 015 (CSF)" from the two samples "Patient 016 (CSF)", "Patient 015 (CSF)". What do you think? 

> Did you run a test phylo tree?

I ran it but it never seems to get past:  "Computation in progress. Please check back later!". Same in master. Can I test in staging or is the a better way?

app/models/visualization.rb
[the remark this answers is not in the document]
tfrcarvalho: > > Did you discuss this naming strategy with Jenn?
> 
> No, that's a good idea. @happyimadesignr: I came up with an auto naming algorithm for showing the heatmaps in the visualization list. It will makes a string such as "Patient 016 (CSF) and 015 (CSF)" from the two samples "Patient 016 (CSF)", "Patient 015 (CSF)". What do you think?
> 
> > Did you run a test phylo tree?
> 
> I ran it but it never seems to get past: "Computation in progress. Please check back later!". Same in master. Can I test in staging or is the a better way?

Yeah, staging might be better.

app/models/visualization.rb
tfrcarvalho: yes, since they are both array `to_sentence` handles it.

happyimadesignr:
Hey @gregdingle and @tfrcarvalho 

1. I think it's nice to include the project name, but I would put it in the table column instead of the name. I don't thinking including the samples names in the title will be scalable for a lot of samples.
2. For Phylo trees, they already have a name, so it should display the name the user made.
3. For Heatmaps, I was thinking that it could just be called "Heatmap" or "Untitled Heatmap" and the user can name it. Then the table details will help differentiate the heatmaps from each other if the user doesn't explicitly name it. Or we can also have users name it when they save for the first time. For the minimum version, why don't we called it Heatmap and complete the table columns? What do you think?

Screen Shot 2019-03-27 at 11 44 03 AM
https://chanzuckerberg.invisionapp.com/d/main#/console/16827472/350949082/preview

gregdingle:
@happyimadesignr 

> I think it's nice to include the project name, but I would put it in the table column instead of the name. 

Yea, that was always my intention. I didn't have time to add the columns when I first wrote this PR.

> For Heatmaps, I was thinking that it could just be called "Heatmap" or "Untitled Heatmap" and the user can name it. 

Ok, that will work, and is simpler. We can keep the auto name idea in mind for we add the ability to name heatmaps.

About the project name, how do you want to handle the case when there multiple projects for one visualization?

app/controllers/visualizations_controller.rb
```diff
@@ -63,11 +63,14 @@ def save
...
-      visualization_type: @type,
-      samples: { id: [sample_ids] }
-    ).order(created_at: :desc)
+    vis = Visualization.joins(:samples)
+                       .where(
+                         user: current_user,
+                         visualization_type: @type,
+                         samples: { id: [sample_ids] }
+                       )
+                       .where.not(visualization_type: [nil, 'undefined'], name: nil)
```
cdebourcy: for my own education, when would each of the scenarios `visualization_type: [nil, 'undefined']` arise?

gregdingle: I intended this only for filtering out early test data which did not have those values set. After this PR, they should always be set. 

I was going to add this as a comment but ruby does not seem to accept a # comment on a line in between method calls 🙁 

In any case, an alternative would be to delete those rows from the staging and prod DBs. How does the team feel about direct DB mutations?

cdebourcy: I think we're ok with direct mutations, particularly if they only involve derived or non-critical data. However, we want them to be always documented, reviewable and reproducible as explicit migrations :D

PR #2057, opened 2019-03-22
[Local Uploads] Add an in-browser heartbeat so we know if client is still uploading to S3

jshoe:
- Main issue we had is that, when a client uploads a local file directly to S3 and is in "Waiting", we don't know if their upload was interrupted or is just taking a long time. This is the simplest approach I could think of and just pings the server to update sample.client_updated_at every minute. So a dev can tell if their browser is still active or disconnected when requested to investigate.
- Caveat is that I think someone can sleep their laptop and then resume later and the put request to upload the file might still resume. So a lack of activity doesn't automatically mean we should mark the upload as failed.
- Next steps can refine this and surface it in the UI and also break down the InputFile progress.
- Explored alternative was S3 object notifications on the backend: https://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html But I believe you would only get one event on ObjectCreated and it just wouldn't appear in S3 if interrupted.

### Testing
- Happy path: Tested local and bulk upload paths and saw the browser send the updates. Verified the datetimes were getting updated continuously in the database. Page transitioned normally.
- Sad path: Saw helpful console logs that it failed to reach the server.


app/assets/src/api/index.js
```diff
@@ -141,12 +141,9 @@ const uploadFileToUrlWithRetries = async (
     onUploadProgress
   };
 
-  // Set a 10s response timeout
-  const client = axios.create();
-  client.defaults.timeout = 10000;
```
jshoe: I experimented a little more and found that just leaving this timeout off was better. Otherwise I think it forced a timeout after 10 sec.

app/assets/src/api/upload.js
```diff
@@ -124,6 +125,7 @@ export const bulkUploadLocal = ({
       .then(response => {
         // After successful sample creation, upload to the presigned URLs returned
         const sampleId = response.id;
+        startUploadHeartbeat(sampleId);
```
tfrcarvalho: Can the client report some kind of progress metric?

jshoe: @tfrcarvalho I think we could break it down by input file and do an upload progress per file, but I think it'd be inelegant with the current "3 pages" (old single, old bulk, new consolidated) so I'd prefer to refactor it when we switch over to the new flow? I think the state would need a structure of sampleIds => files => %progress which gets updated onUploadProgress

tfrcarvalho: OK, i think it will be good to have that info for multiple reasons, but can definitively be a follow up.

PR #2058, opened 2019-03-22
only uploader can access intermediate files

cdebourcy:
goes with https://github.com/chanzuckerberg/idseq-dag/pull/130/files


app/models/pipeline_run.rb
```diff
@@ -1330,15 +1330,13 @@ def outputs_by_step(can_see_stage1_results = false)
...
-      targets.each_with_index do |(target_name, output_list), step_idx|
+      targets.each do |target_name, output_list|
         next if given_targets.keys.include?(target_name)
         file_info = []
         output_list.each do |output|
           file_info_for_output = filename_to_info["#{output_dir_s3_key}/#{pipeline_version}/#{output}"]
           next unless file_info_for_output
-          if !can_see_stage1_results && stage_idx.zero? && step_idx < num_steps - 1
-            # Delete URLs for all host-filtering outputs but the last, unless user uploaded the sample.
+          if !can_see_stage1_results && stage_idx.zero?
```
tfrcarvalho: The difference is just the last step? 
Should that one be safe?

cdebourcy: We used to think that, but I am restricting the last step too now, because I plan to do some residual cleanup downstream. "Annotated non-host fasta" will be the first sequence file that's available and it will have been stripped of any residual reads that seemingly matched human during gsnap/rapsearch (downstream from host filtering stage)

tfrcarvalho: Whats our plan some users upload samples for others to analyze? Should we review our access control logic to include e.g admin or power users for some projects/samples beside uploader?
Or do you think these files are barely accessed anyway?

cdebourcy: We should err on the side of caution. All information anyone should need is in non-host fasta/fastq which we still expose

tfrcarvalho: That makes sense.
My only concern is to make sure we eventually stop doing manual downloads. Users either have access or they do not, but should be our API/UI who takes care of that decision.

Is this ready for final review?

cdebourcy: yes

PR #2072, opened 2019-03-26
Add tooltip on save button click

gregdingle:
# Description

image

The UEX was broken because there was no feedback on save. This adds a tooltip same as the "share" button. 

This also fixes a bug that was introduced when switching to `BasicPopup`, which ignored the `on=` attribute of the tooltip. 

I also changed the `pushState` updating of the URL to `replaceState`, following @jshoe 's fix on the report page. We'd need to do some work to get the back button working within visualizations. 

NOTE: the experience could still be better because it shows optimistically, unconditional on success, and the color stays the same while focussed, keeping with the current site wide color scheme for `Button`. 

# Test

1. Click save on heatmap, phylo tree and taxon tree
2. See tooltip message
3. repeat for share click


app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -356,12 +356,19 @@ class SampleView extends React.Component {
...
             <BasicPopup
               trigger={<ShareButton onClick={this.onShareClick} />}
-              content="A shareable URL will be copied to your clipboard!"
+              content="A shareable URL was copied to your clipboard!"
               on="click"
               hideOnScroll
             />{" "}
             {/* TODO: (gdingle): this is admin-only until we have a way of browsing visualizations */}
-            {this.props.admin && <SaveButton onClick={this.onSaveClick} />}{" "}
+            {this.props.admin && (
```
tfrcarvalho: why not put the popup inside `SaveButton` ?

gregdingle: Because one of instance requires a custom CSS class on the inner `Button`. I could add a prop `buttonClassName` and pass it down. Is that what you would do?

tfrcarvalho: You can just pass `className` and should work since props are passed downstream.
(We are actually creating a button, despite the fact that tooltip encloses the button, so the generic name className should apply to button).

PR #2074, opened 2019-03-26
Add basic option to projects index endpoint.

MarkAZhang:
The projects index endpoint was taking 30s or more to load. The basic version takes about 200ms.

Tested that admin and non-admin can fetch projects in the new upload flow and samples in the upload modal properly.

### DIFF
--- a/app/assets/src/api/index.js
+++ b/app/assets/src/api/index.js
@@ -205,10 +205,11 @@ const getProjectDimensions = ({ domain }) =>
 
 const getSamplesV1 = params => get("/samples.json", { params });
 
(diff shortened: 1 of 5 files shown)


MarkAZhang:
Thanks for the suggestion! I added a comment/TODO. Ideally, the basic version of the endpoint would be the default, and we should include additional data only if the client asks for it (like you suggested). I was talking with Tiago though, and it sounds like data discovery calls this endpoint in several places so changing the default for the endpoint is a more involved change. Tiago will be refactoring these endpoints soon.

PR #2078, opened 2019-03-27
Warn about unsaved changes

gregdingle:
# Description

Following @happyimadesignr's request, this adds a standard warning about unsaved changes to the heatmap page. 

image

It's gated on being `admin` like the save button. 

@happyimadesignr did you expect this to prompt on the first save only or all unsaved changes? As it stands, it works on all unsaved changes.

# Test

1. open a heatmap
2. click back, see no warning
3. open a heatmap
4. change taxon level
5. click back, see warning
6. open a heatmap
7. change taxon level
8. save
9. click back, see no warning
10. create a new heatmap
11. click back, see no warning


happyimadesignr:
Hey @gregdingle, I think all unsaved changes works.

1. Is this only for the back button or does it show up when users click outside the page, like on the logo or somewhere else in the nav dropdown?

2. I think for browser back button you can only use the default message right? But do we have any other control over how the content is displayed? Like can we change the "Leave Site?" wording?

3. For exiting the page from within the page, can we use a custom message. 
@tfrcarvalho Created this for the warning for when someone is making their project public:

Screen Shot 2019-03-27 at 11 30 11 AM

@MarkAZhang and I had talked about having this when exiting the upload flow, but I'm not currently seeing it.

Screen Shot 2019-03-27 at 11 29 02 AM

gregdingle:
Replies to @happyimadesignr 

> Is this only for the back button or does it show up when users click outside the page, like on the logo or somewhere else in the nav dropdown?

It will show up whenever someone navigates to a new page or closes the tab. 

> I think for browser back button you can only use the default message right? But do we have any other control over how the content is displayed? Like can we change the "Leave Site?" wording?

No, we do not. This actually changed in the last few years because of abuse by sites. Now the browsers only display a standard message, like the Chrome one here. 

> For exiting the page from within the page, can we use a custom message.

No unfortunately for the reason above.

PR #2079, opened 2019-03-27
[Data Discovery] Use new metadata fields instead of sample model fields.

tfrcarvalho:
* Use only metadata fields (in particular, for location and sample tissue/type) in projects and samples routes.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -77,21 +77,24 @@ class ProjectsController < ApplicationController
         locations_by_project_id = {}
         total_reads = 0
         adjusted_remaining_reads = 0
+
+        metadata = metadata_multiget(samples.pluck(:id))
+        Rails.logger.debug("metadata: #{metadata}")
         samples.includes(:host_genome, :user, :pipeline_runs).each do |s|
           (host_genome_names_by_project_id[s.project_id] ||= Set.new) << s.host_genome.name if s.host_genome && s.host_genome.name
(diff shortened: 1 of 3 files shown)


cdebourcy:
Cool!
(1) What is the test plan for this PR?
(2) Will https://github.com/chanzuckerberg/idseq-web/blob/2f0026b12800246af9c9cbe81bc3db6076486d39/app/models/sample.rb#L147 be edited/deprecated as well?

tfrcarvalho:
> Cool!
> (1) What is the test plan for this PR?
> (2) Will
> 
> [idseq-web/app/models/sample.rb](https://github.com/chanzuckerberg/idseq-web/blob/2f0026b12800246af9c9cbe81bc3db6076486d39/app/models/sample.rb#L147)
> 
> Line 147 in [2f0026b](/chanzuckerberg/idseq-web/commit/2f0026b12800246af9c9cbe81bc3db6076486d39)
> 
>          OR samples.sample_tissue LIKE :search 
> be edited/deprecated as well?

No new tests, because nothing changed regarding access control.
This PR changes endpoints used in DD. Added a prefix on title to make that more clear.
So, test plan was just checking projects and samples list return the new fields.

There are two other places where sample_tissue appears: that sample search endpoint which not being used right now by DD, and the phylo tree creation flow (for extra info). These two will eventually be fixed after the DD sprint.

tfrcarvalho:
> For my education, is sample type intended to be more general than sample tissue?

Yeah, I believe so. Definitely is a superset of sample tissue.

PR #2080, opened 2019-03-27
[Data Discovery] Fix sidebar 'accordion needs double-click to toggle' and 'doesn't update when samples filter to 0' with new empty state

jshoe:
- In the Accordion there was a wasToggled flag that wasn't checked onToggle (so this.state.open starts as null and then gets set to open and then needed another click)
- Previous state check would return prevState if the results filter down to 0, so the sidebar would still have results. New change adds a loadingSamples state so the sidebar will show an empty state if not loading and results went to 0.

Mar-27-2019 10-15-51

### Testing
- Go to My Data and Public and click through Projects and Samples tabs. Restrict filters so that you get 0 matching results.

### DIFF
--- a/app/assets/src/components/layout/Accordion.jsx
+++ b/app/assets/src/components/layout/Accordion.jsx
@@ -8,7 +8,7 @@ class Accordion extends React.Component {
 
   onToggle = () => {
     this.setState({
-      open: !this.state.open,
+      open: this.state.wasToggled ? !this.state.open : !this.props.open,
       wasToggled: true
     });
   };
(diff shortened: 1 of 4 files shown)


app/assets/src/components/layout/Accordion.jsx
```diff
@@ -8,7 +8,7 @@ class Accordion extends React.Component {
 
   onToggle = () => {
     this.setState({
-      open: !this.state.open,
+      open: this.state.wasToggled ? !this.state.open : !this.props.open,
```
gregdingle: Good fix. @MarkAZhang also had an idea how to make the accordion logic clearer: https://github.com/chanzuckerberg/idseq-web/pull/2014#discussion_r265372527

PR #2082, opened 2019-03-27
Remove json from prettier and keep default package.json formatting

jshoe:
- I've been seeing this issue where restarting npm in local development causes it to lag and reformat the entire package-lock.json file. I think npm's default format is to keep the `resolved` and `integrity` as one line without a break.
- So I propose we just remove 'json' from the prettier command in package.json (assuming that's what's causing the back and forth) and keep package/package-lock with their default formatting. (Alternatively could use a .prettierignore)


tfrcarvalho:
Were you able to confirm that this solves it?

jshoe:
@tfrcarvalho In my local config it would no longer try to reformat the file on precommit (as long as they're on one line). So unless people's setup includes another hook to reformat, they will not have the reformatting either

PR #2083, opened 2019-03-27
[Discovery Sidebar] Sort metadata rows by value descending and add show/hide for extra rows

jshoe:
- Metadata rows will be sorted by value descending and then by the label alphabetically.
- 4 rows will be shown by default with a show/hide toggle.

Mar-27-2019 13-30-47

### Testing
- Go to Projects/Samples, look at the sidebar, verify that the values are sorted and the toggle works as expected.

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoverySidebar.jsx
+++ b/app/assets/src/components/views/discovery/DiscoverySidebar.jsx
@@ -169,21 +169,20 @@ export default class DiscoverySidebar extends React.Component {
 
   buildMetadataRows(field) {
     const { metadata, expandedMetadataGroups } = this.state;
-
-    // TODO (gdingle): put "unknowns" last?
     const fieldData = metadata[field];
+
     // Sort by the value desc and then by the label alphabetically
     const sorted = orderBy(
       [k => k[1], k => k[0]],
       ["desc", "asc"],
       Object.entries(fieldData)
     );
-    const total = sum(Object.values(fieldData));
 
     // Display N fields and show/hide the rest
     const defaultN = 4;
     const defaultRows = sorted.slice(0, defaultN);
     const extraRows = sorted.slice(defaultN);
+    const total = sum(Object.values(fieldData));
     return (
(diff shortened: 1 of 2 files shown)


app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -167,34 +168,68 @@ export default class DiscoverySidebar extends React.Component {
...
+
+    // Sort by the value desc and then by the label alphabetically
+    const sorted = orderBy(
+      [k => k[1], k => k[0]],
+      ["desc", "asc"],
+      Object.entries(fieldData)
+    );
+
+    // Display N fields and show/hide the rest
+    const defaultN = 4;
```
tfrcarvalho: Maybe move this to props with a default value?

app/assets/src/components/views/discovery/discovery_sidebar.scss
```diff
@@ -94,4 +94,14 @@
     text-align: right;
     color: white;
   }
+
+  .showHide {
+    text-transform: uppercase;
+    margin-top: 10px;
+    margin-bottom: 10px;
+    color: $primary-light;
+    font-size: 10px;
```
tfrcarvalho: Can you find a mixin in typography to use here instead of setting font-size by yourself?

jshoe:
@tfrcarvalho updated with your suggestions 👍

app/assets/src/components/views/discovery/discovery_sidebar.scss
```diff
@@ -94,4 +94,14 @@
     text-align: right;
     color: white;
   }
+
+  .showHide {
+    text-transform: uppercase;
+    margin-top: 10px;
+    margin-bottom: 10px;
+    color: $primary-light;
+    font-size: $font-size-smallest;
```
tfrcarvalho: I guess there is no mixin for 10px.
Next time we should define one.

PR #2085, opened 2019-03-27
[Discovery Sidebar] Update visual style + add color to the bars

jshoe:
- Updated style from this week: https://chanzuckerberg.invisionapp.com/d/main#/console/17022287/353927762/inspect
- Visual adjustments to look like the mocks
- Sorting + show/hide of the bars is in the other PR
- Updating the behavior of sizing the histogram bars will be in a future PR

Screen Shot 2019-03-27 at 2 17 01 PM

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoverySidebar.jsx
+++ b/app/assets/src/components/views/discovery/DiscoverySidebar.jsx
@@ -259,36 +259,44 @@ export default class DiscoverySidebar extends React.Component {
           >
             <div className={cs.hasBackground}>
               <dl className={cx(cs.dataList)}>
-                <dt>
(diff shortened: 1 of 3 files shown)


tfrcarvalho:
Can we give more width to the labels on the first box? Avoid avg non host reads from spanning to three rows? Or just add Overall per sample?
@happyimadesignr for advising.

jshoe:
@tfrcarvalho I changed it to 50/50 and it looks pretty good to me? "Overall per sample" is still a little inaccurate for the Samples and Projects rows. But maybe "Avg. reads/sample" "Avg. non-host reads per sample"

tfrcarvalho:
I'll approve since we can do a overall review with Jenn after.

PR #2087, opened 2019-03-27
Log more context to segment

gregdingle:
# Description

This makes full use of segment in the client and server given the data we have at call-time. 

# Test

Load a page and see log data such as 
```
{
  "email": "gdingle@chanzuckerberg.com",
  "name": "Greg Dingle",
  "created_at": "2018-06-08T13:20:09.000-07:00",
  "updated_at": "2019-02-21T11:44:53.000-08:00",
  "role": 1,
  "allowed_features": [
    "data_discovery"
  ],
  "institution": null,
  "admin": true,
  "demo_user": false,
  "biohub_user": false,
  "czi_user": true,
  "createdAt": "2018-06-08T13:20:09-07:00",
  "firstName": "Greg",
  "lastName": "Dingle",
  "sign_in_count": 3,
  "current_sign_in_at": "2019-02-21T11:44:53.000-08:00",
  "last_sign_in_at": "2019-02-08T15:08:14.000-08:00",
  "current_sign_in_ip": "172.19.0.1",
  "last_sign_in_ip": "172.19.0.1"
}
```

### DIFF
--- a/app/lib/metric_util.rb
+++ b/app/lib/metric_util.rb
@@ -1,6 +1,7 @@
 require "uri"
 require "net/http"
 
+# TODO: (gdingle): update description
 # MetricUtil is currently used for posting metrics to Datadog's metrics endpoints.
 class MetricUtil
   SEGMENT_ANALYTICS = if ENV["SEGMENT_RUBY_ID"]
@@ -24,16 +25,19 @@ class MetricUtil
     sample_upload_batch_created: "sample_upload_batch_created"
   }.freeze
 
+  # TODO: (gdingle): mark as deprecated and all callers
   def self.put_metric_now(name, value, tags = [], type = "count")
     put_metric(name, value, Time.now.to_i, tags, type)
   end
 
+  # TODO: (gdingle): make private
   def self.put_metric(name, value, time, tags = [], type = "count")
     # Time = POSIX time with just seconds
     points = [[time, value]]
     put_metric_point_series(name, points, tags, type)
   end
 
+  # TODO: (gdingle): make private
   def self.put_metric_point_series(name, points, tags = [], type = "count")
     # Tags look like: ["environment:test", "type:bulk"]
     name = "idseq.web.#{Rails.env}.#{name}"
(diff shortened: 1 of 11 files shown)


app/lib/metric_util.rb
```diff
@@ -24,76 +27,116 @@ class MetricUtil
...
         user,
-        count: entries.count, project_id: project_id, host_genome_id: host_genome_id, client: client
+        { count: entries.count, project_id: project_id, host_genome_id: host_genome_id, client: client },
+        request
       )
     end
   end
+
+  # Start private class methods
+  class << self
```
tfrcarvalho: Why do you need this versus just using the private keyword?

app/models/user.rb
```diff
@@ -69,11 +69,62 @@ def can_upload(s3_path)
...
-      unless ["czbiohub.org", "ucsf.edu"].include?(email.split("@").last)
+      unless biohub_user?
         return false
       end
     end
 
     true
   end
+
+  def biohub_user?
```
jshoe: The string checking was a little hacky to begin with, so in the future we could lean on 'institution' and backfill it for Biohub/UCSF/CZI

app/views/layouts/_page_header.html.erb
```diff
@@ -11,9 +11,12 @@
         userSignedIn: true
       }, 'page_header', JSON.parse('<%= raw escape_json(request_context)%>'))
 
+
       <% if ENV["SEGMENT_JS_ID"] %>
         <!-- Associate user ID with Segment -->
-        analytics.identify(<%= current_user.id %>);
+        analytics.identify(
+          <%= current_user.id %>,
+          <%= raw escape_json(current_user.traits_for_segment) %>);
```
jshoe: out of curiosity do you see this triggering another db request for every request in your local logs?

gregdingle: Yes, that's why I did https://github.com/chanzuckerberg/idseq-web/pull/2088 .

app/models/user.rb
gregdingle: Yes, institution seems valuable in the long run. Is it currently required and standardized?

jshoe: it is required on the "interest form" but account creation either happens through 1) Product creating the user at /users/new or 2) many accounts created through the Add User To Project feature which auto-creates and invites users. I think we could update both forms to require institution

PR #2088, opened 2019-03-27
Add cached counts to user model for analytics

gregdingle:
# Description

Request for comment. Based on https://github.com/chanzuckerberg/idseq-web/pull/2087 . 

We need to define user groups by the level of usage of the app. For example, users that have uploaded a sample. See https://czi.quip.com/hgbzAVVqAbC6/Metrics-Design-and-Segment-Breakdown . And joining the user table with a large set of other tables on any tracking call seems like a bad idea. 

An easy way to get the counts efficiently is with cached counts in the user model of the numbers of associated objects. See https://blog.appsignal.com/2018/06/19/activerecords-counter-cache.html for more background. 

This means every `INSERT` to the affected assocs would add an `UPDATE` to the user table. Given the low expected write rate, I think this is an acceptable cost, and it would be useful in the app anywhere we want a count of a user's objects. 

The alternative would be to compute the same counts downstream by SQL query in the warehouse. This puts more burden on the analyst, and it also wouldn't work for Google analytics dimensions or other analytics tools we may want to use. 

# Test

Run migrations
Check user in rails console for non zero counts
Check for correct SQL queries in console

### DIFF
--- a/app/models/favorite_project.rb
+++ b/app/models/favorite_project.rb
@@ -1,4 +1,4 @@
 class FavoriteProject < ApplicationRecord
   belongs_to :project
-  belongs_to :user, counter_cache: true # count for analytics
+  belongs_to :user, counter_cache: true # use .size for cache, use .count to force COUNT query
 end
--- a/app/models/favorite_project.rb
+++ b/app/models/favorite_project.rb
@@ -1,4 +1,4 @@
 class FavoriteProject < ApplicationRecord
   belongs_to :project
-  belongs_to :user
+  belongs_to :user, counter_cache: true # count for analytics
 end
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -4,6 +4,7 @@ class Project < ApplicationRecord
     include Elasticsearch::Model::Callbacks
   end
 
+  # TODO: (gdingle): add counter_cache if we ever migrate user assoc to "has_many :through"
   has_and_belongs_to_many :users
   has_many :samples
   has_many :favorite_projects
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -4,7 +4,6 @@ class Project < ApplicationRecord
     include Elasticsearch::Model::Callbacks
   end
 
-  # TODO: (gdingle): add counter_cache if we ever migrate user assoc to "has_many :through"
   has_and_belongs_to_many :users
   has_many :samples
   has_many :favorite_projects
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,6 +11,7 @@ class User < ApplicationRecord
   devise :database_authenticatable, :recoverable,
          :rememberable, :trackable, :validatable
 
+  # NOTE: counter_cache is not supported for has_and_belongs_to_many.
   has_and_belongs_to_many :projects
   # All one-to-many assocs are counter cached for per-user analytics.
   # See traits_for_segment.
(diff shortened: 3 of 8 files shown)


cdebourcy:
Seems reasonable. I wonder if @davidrissato has any experience or thoughts on this?

davidrissato:
I don't have experience with ActiveRecord, but I read the blog post mentioned above and I can understand the technique, and I needed to use similar approach in the past. 
I don't know how well ActiveRecord handles it, but the potential issues with this approach might be counters getting out-of-sync and showing wrong values. How critical are these values?

tfrcarvalho:
This is cool. If if some of the concerns @davidrissato raised are satisfied, this could also be a good solution for our statistics in DD.

gregdingle:
Thanks for all the thoughtful comments.

> How often do we need to read these counters?
For the purpose of analytics, we would read them anytime 1) a full page loads 2) the backend logs to analytics (not sure how often this actually needs to happen). I don't actually think there would be a perf issue right now but I would like this API to be cost-less (ie constant time) so anybody who wants to add analytics logging doesn't need to think about perf. 

> How critical are these counts and what happens if they get out of sync?

For analytics, not critical. 

> Do we need to keep counters for different timespans like: "last 2 days", "last week", "last month"?
Good question, but I think all-time per user is good enough for now.

gregdingle:
I added the schema and data migrations. Please review.

tfrcarvalho:
> > Do we need to keep counters for different timespans like: "last 2 days", "last week", "last month"?
> > Good question, but I think all-time per user is good enough for now.

There seems to ways to override default behaviour and even add some extra

app/models/user.rb
```diff
@@ -113,6 +118,21 @@ def traits_for_segment
...
+      # Counts (should be cached in the users table for perf)
+      projects: projects.size, # projects counter is NOT cached because has_and_belongs_to_many
+      samples: samples.size,
+      favorite_projects: favorite_projects.size,
+      favorites: favorites.size,
+      visualizations: visualizations.size,
+      phylo_trees: phylo_trees.size,
+      # Has-some (this is important for Google Custom Dimensions)
+      # See https://segment.com/docs/destinations/google-analytics/#custom-dimensions
+      has_projects: !projects.empty?,
```
tfrcarvalho: Do these dimensions have to be boolean or have a specific prefix? 
Sorry, browsed the link but could not find the constraints for google analytics.

Can you add a comment why the sizes above do not suffice for google analytics?

app/models/favorite_project.rb
```diff
@@ -1,4 +1,4 @@
 class FavoriteProject < ApplicationRecord
   belongs_to :project
-  belongs_to :user
+  belongs_to :user, counter_cache: true # count for analytics
```
tfrcarvalho: nit: Not sure you need the comment. We can actually use this counters for more things... 
But up to you

app/models/project.rb
```diff
@@ -4,6 +4,7 @@ class Project < ApplicationRecord
     include Elasticsearch::Model::Callbacks
   end
 
+  # TODO: (gdingle): add counter_cache if we ever migrate user assoc to "has_many :through"
```
tfrcarvalho: I think you should change the comment to explain the limitation of `counter_cache` with `has_and_belongs_to_many`.
It is not a TODO per se...
Better keep the TODO tag for things we should do soon.

What was the best reference you found for this? I was looking and saw people complaining about has_many :through either.

tfrcarvalho: Another question, what would be the limitation of changing to has_many+through? Could we just change it now?

app/models/favorite_project.rb
gregdingle: Updated

app/models/project.rb
gregdingle: I spent about 30min researching this issue and I found that `has_and_belongs_to_many` is not supported for counters and many people recommend against ever using `has_and_belongs_to_many` for other reasons. 

I thought about migrating to has_many+through but I felt it was too much for this PR and I'm not sure enough of other consequences.

app/models/favorite_project.rb
gregdingle: # use .size for cache, use .count to force COUNT query

app/models/project.rb
tfrcarvalho: I understand. This is where test would give some comfort.
The "cheap" on the other hand is relative... since the number of analytics calls might increase rapidly. 
But I guess we can move on for now, and just keep an eye.

PR #2089, opened 2019-03-28
change non-host reads to passed filters

cdebourcy:
Renames a metric which had been confusing for users. Checked new naming with Katrina


cdebourcy:
Correct. They say: "well I know there wasn't any host in there, why is non-host percentage so low". It's because we do quality filtering, complexity filtering, and deduplication

cdebourcy:
I think result folder conveys the correct idea. Also it provides visibility into the actual steps so people will understand

cdebourcy:
After merging, I will update wiki!

PR #2091, opened 2019-03-28
[Data Discovery] Complete search actions when selecting project, sample and query string

tfrcarvalho:
This PR add some missing actions when user searches using the search box in the header:
* if user searches and selects a project, it will be directed to the corresponding project view
* if user searches and selects a sample, it will be directed to the corresponding sample report view
* if user searches for a string and presses enter, the current view is filtered by that string. The string will remain in the search box.
* removal of search by uploader

This PR has a few minor additional fixes:
* Bug fix: where filter tags where not being displayed correctly
* UI style tweaks

Mar-28-2019 16-04-22

Testing:
* In DD, use the search box to search for projects, samples and see if the results on the corresponding table are filtered
* For the bug fix: add filters using the filter dropdowns and verify that the corresponding tag is added to the filters


app/models/sample.rb
```diff
@@ -144,7 +144,6 @@ def self.search(search, eligible_pr_ids = [])
     if search
       search = search.strip
       results = where('samples.name LIKE :search
-        OR samples.sample_tissue LIKE :search
         OR samples.sample_location LIKE :search
```
jshoe: did your other PR take care of sample_location too?

tfrcarvalho: I still need to add the search on metadata... but that is valid for both fields.
I should remove location here too.

PR #2092, opened 2019-03-28
Add columns to DD visualization tab

gregdingle:
# Description

This brings the tab closer to the mock https://chanzuckerberg.invisionapp.com/d/main#/console/16827472/350949082/preview . 

image

I also changed to use `updated_at` instead of `created_at` because the saves now overwrite. 

# Test 

Load my data visualizations and see screen above.


app/assets/src/components/views/projects/ProjectsView.jsx
```diff
@@ -67,7 +67,7 @@ class ProjectsView extends React.Component {
         <div className={cs.projectDescription}>{project.description}</div>
         <div className={cs.projectDetails}>
           <span className={cs.projectCreationDate}>
-            {moment(project.created_at).fromNow()}
+            {moment(project.updated_at).fromNow()}
```
tfrcarvalho: Indeed `updated_at` seems closer to what @jenn initially intended.
But, for `updated_at` to be changed on new samples, we need to add `touch: true` to `belongs_to :project` in the samples model.

@happyimadesignr in projects view you have "uploaded x time ago". At the time we went we creation date. But might make sense to be date last time a sample was added/removed from that project? Or event user changes?

app/assets/src/components/views/visualizations/VisualizationsView.jsx
```diff
@@ -38,9 +50,9 @@ class VisualizationsView extends React.Component {
     return (
       <div>
         {publicAccess ? (
-          <GlobeIcon className={cx(cs.icon, cs.iconGlobe)} />
+          <PublicProjectIcon className={cx(cs.icon, cs.iconPublic)} />
```
tfrcarvalho: @happyimadesignr we need two new icons here for visualizations...

app/assets/src/components/views/projects/ProjectsView.jsx
gregdingle: @tfrcarvalho my understanding is that all types of visualizations are immutable with respect to samples. That is, they only use the samples specified at creation. 

My thinking for `updated_at` was to show when the view state of a vis has been changed.

app/assets/src/components/views/visualizations/VisualizationsView.jsx
gregdingle: is there some way to extract them from https://chanzuckerberg.invisionapp.com/d/main#/console/16827472/350949082/preview ?

app/assets/src/components/views/projects/ProjectsView.jsx
tfrcarvalho: This changes `ProjectsView` reference date, which is unrelated with visualizations...

gregdingle: Sorry, this change was totally accidental then! But I think some your points still stand about whether to order projects by created or updated.

app/assets/src/components/views/visualizations/VisualizationsView.jsx
tfrcarvalho: There could be, but Jenn has to explicitly make it available as an asset (which she did not do yet). 
Just ask her directly, and she will send you the svg file.

gregdingle:
I added the SVG icons 😓 and now it looks like this: 
image . 

I tested public and private though we only support private at the moment.

app/assets/src/components/ui/icons/HeatmapPrivate.jsx
```diff
@@ -0,0 +1,359 @@
...
+import cx from "classnames";
+
+import cs from "./heatmap_private.scss";
+
+/**
+ * This class was created by the following process:
+ * Extracted CSS to heatmap_private.scss.
+ * From Adobe generated SVG to JSX using https://magic.reactjs.net/htmltojsx.htm.
+ * Transformed CSS classnames.
+ */
```
tfrcarvalho: These SVGs can be simplified significantly.
I can do that in a follow up PR.

app/assets/src/components/ui/icons/HeatmapPrivate.jsx
```diff
@@ -0,0 +1,359 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import cx from "classnames";
```
tfrcarvalho: You do need cx for single class.

app/assets/src/components/views/visualizations/VisualizationsView.jsx
```diff
@@ -6,19 +6,23 @@ import cx from "classnames";
...
-import LockIcon from "~ui/icons/LockIcon";
+import { openUrl } from "~utils/links";
 
 import cs from "./visualizations_view.scss";
+import HeatmapPublic from "~ui/icons/HeatmapPublic";
+import HeatmapPrivate from "~ui/icons/HeatmapPrivate";
+import PhyloTreePublic from "~ui/icons/PhyloTreePublic";
+import PhyloTreePrivate from "~ui/icons/PhyloTreePrivate";
 
+// See also ProjectsView which is very similar
```
tfrcarvalho: What is the purpose of this comment?

app/assets/src/components/views/visualizations/VisualizationsView.jsx
```diff
@@ -29,40 +33,62 @@ class VisualizationsView extends React.Component {
...
-          <LockIcon className={cx(cs.icon, cs.iconLock)} />
-        )}
-      </div>
-    );
+  renderAccess = ({ cellData: visualization }) => {
+    if (visualization.visualization_type == "heatmap") {
+      return (
+        <div>
+          {visualization.publicAccess ? (
+            <HeatmapPublic className={cx(cs.icon)} />
```
tfrcarvalho: Again, no need for `classnames` here.

app/assets/src/components/views/visualizations/VisualizationsView.jsx
```diff
@@ -72,6 +98,11 @@ class VisualizationsView extends React.Component {
     );
   };
 
+  handleRowClick = ({ event, index, rowData }) => {
```
tfrcarvalho: Its good not the variable `event` as it a global variable (aka `window.event`).
To avoid possible future bugs, rename it to something else like `currentEvent`.

app/assets/src/components/ui/icons/HeatmapPrivate.jsx
tfrcarvalho: As an aside here is a PR that simplified some icons (https://github.com/chanzuckerberg/idseq-web/pull/2040).

gregdingle: Ok thanks for pointing me to that. It's not obvious to me how to know what to remove without knowing the semantics of SVG. 

Is it worth the effort in general to hand-edit SVG? There must be a standard way to export SVGs that doesn't require that much manual labor.

app/assets/src/components/ui/icons/HeatmapPrivate.jsx
gregdingle: Ah that makes sense!

app/assets/src/components/ui/icons/HeatmapPrivate.jsx
happyimadesignr: Yay for the icons!

@gregdingle Are you still working on updating the titles?

app/assets/src/components/views/visualizations/VisualizationsView.jsx
gregdingle: Yes, the purpose is to point out that there may be some implicit coupling. That is, if something changes in `ProjectsView` something else may need to change in `VisView`. Not having been a part of the design discussions or most of the coding, I just wanted to make note and not a write a `TODO`.

app/assets/src/components/views/visualizations/VisualizationsView.jsx
gregdingle: Good point. FYI, it is also named `event` in `ProjectsView` and `SamplesView`.

app/assets/src/components/ui/icons/HeatmapPrivate.jsx
gregdingle: @happyimadesignr The titles are now simply the name of the phylo tree or "heatmap", as we discussed, though that's not obvious from the screen I posted.

app/assets/src/components/views/visualizations/VisualizationsView.jsx
tfrcarvalho: Thanks for pointing that out. Should not.

app/assets/src/components/ui/icons/HeatmapPrivate.jsx
happyimadesignr: Cool, got it!

PR #2098, opened 2019-03-30
[Data Discovery] Add a bunch of Popups to the UI

jshoe:
- Also fixed the header style alignment on BaseTable a little

### Testing
- Go to data discovery and hover over the actions and names to test

Mar-29-2019 16-58-20


app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -204,13 +208,18 @@ class SamplesView extends React.Component {
...
     const { selectedSampleIds } = this.state;
     return (
-      <a
-        href={`/visualizations/heatmap?sampleIds=${Array.from(
-          selectedSampleIds
-        )}`}
-      >
-        <HeatmapIcon className={cs.icon} />
-      </a>
+      <BasicPopup
```
tfrcarvalho: Should we add this popups inside the icons?

app/assets/src/components/visualizations/table/base_table.scss
```diff
@@ -33,6 +34,15 @@
   outline: 0;
   text-align: center;
   text-transform: none;
+
+  .text {
```
tfrcarvalho: Can you call it something more clear? e.g. `columnHeaderTooltipText`?

app/assets/src/components/visualizations/table/base_table.scss
```diff
@@ -33,6 +34,15 @@
...
+    display: inline-block;
+    max-width: 100%;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    overflow: hidden;
+    margin-top: 7px;
```
tfrcarvalho: Did you just copy from current style?
I wonder if we should vertically align it.

app/assets/src/components/views/samples/SamplesView.jsx
jshoe: sure. looks like the icons were only used here, so I moved them in

app/assets/src/components/visualizations/table/base_table.scss
jshoe: sure, although this was already in .header

app/assets/src/components/visualizations/table/base_table.scss
jshoe: yes those lines were here https://github.com/bvaughn/react-virtualized/blob/8738097a6ead422f74590d6a631bd5fd83ebc831/source/styles.css#L38 besides the margin-top. Switched to the vertical-align and it looks the same 👍

jshoe:
@tfrcarvalho updated 👍

app/assets/src/components/views/discovery/discovery_sidebar.scss
```diff
@@ -120,7 +120,7 @@
     height: 12px;
     margin-left: 5px;
     margin-right: 7px;
-    padding: 0 5px;
+    padding-left: 1px;
```
jshoe: I threw in this change to reduce padding on the bars because it was giving it a min width of 10px which was too off scale

Screen Shot 2019-04-01 at 10 25 35 AM

Reduced:
Screen Shot 2019-04-01 at 10 26 16 AM

app/assets/src/components/visualizations/table/base_table.scss
tfrcarvalho: oh, i did not notice that. Could be just `tooltipText` then

jshoe:
@gregdingle 

- Added one to the viz titles (and the table headers were included), although after your pr goes in I could add another for the project column
Apr-01-2019 10-54-06

- I think the main purpose for the sample title and table header tooltips is for seeing the whole title even when the line overflows, so it makes sense to be the same text. For the action buttons it's for describing the button so those could be more specific. Do you think we should have something like "View Sample blah"? cc @happyimadesignr

PR #2101, opened 2019-04-01
[Data Discovery] Add time bins dimensions and connect dimensions to sidebar

tfrcarvalho:
Major:
* Dimensions also returns samples and projects grouped by N (max N = 34) bins by creation dated
* Apply current filters to dimensions endpoint
* Sidebar uses filtered dimensions results to display stats about all samples and projects
* Given that dimensions now include not set values for metadata, the filtering by metadata now considers "not_set" as a possible filter.

Minor:
* Some simplification of DiscoverySidebar code
* Removal of TimeFilter component. Given that all info is sent from server, it does not currently justify its own component.

Apr-01-2019 07-31-36


app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -30,138 +21,100 @@ export default class DiscoverySidebar extends React.Component {
...
+          newProps.samples,
+          "totalReads"
+        ),
+        avgNonHostReads: DiscoverySidebar.meanByAndFormat(
+          newProps.samples,
+          "nonHostReads.value"
+        )
+      },
+      metadata: {
+        host: (find({ dimension: "host" }, dimensions) || {}).values || [],
```
jshoe: could make a util function out of this extractor

app/helpers/samples_helper.rb
```diff
@@ -545,14 +545,20 @@ def filter_by_visibility(samples, visibility)
...
-      .where(
-        metadata: {
-          metadata_field_id: metadata_field.metadata_field_id,
-          metadata_field.validated_field => query
-        }
-      )
+
+    samples_with_metadata = samples
+                            .includes(metadata: :metadata_field)
+                            .where(metadata: { metadata_field_id: metadata_field.metadata_field_id })
```
cdebourcy: This is not from your diff, but... interesting choice to have a column `metadata_field_id` in the table `metadata_fields`... curious what the rationale was for doing that instead of using normal auto-incrementing id. Maybe @MarkAZhang can educate me

app/controllers/samples_controller.rb
```diff
@@ -178,23 +179,30 @@ def dimensions
...
-
     samples = filter_samples(samples, params)
 
+    sample_ids = samples.pluck(:id)
+
     locations = samples_by_metadata_field(sample_ids, "collection_location").count
     locations = locations.map do |location, count|
       { value: location, text: location, count: count }
     end
+    not_set_count = samples.count - locations.map { |l| l[:count] }.reduce(:+)
```
jshoe: could also use .sum() right? or would that trigger a query

app/helpers/samples_helper.rb
jshoe: There actually isn't a metadata_fields.metadata_field_id in the schema but there is a metadata.metadata_field_id from the association. is this line working?

cdebourcy: https://github.com/chanzuckerberg/idseq-web/blob/master/db/schema.rb#L119

cdebourcy: oh I see. It's a join table

cdebourcy: thx

app/controllers/samples_controller.rb
tfrcarvalho: Indeed. No new query, at this point it is just ruby array.

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
tfrcarvalho: Done.

PR #2102, opened 2019-04-01
[Data Discovery] Add stats endpoint to collect statistics on a group of samples

tfrcarvalho:
* New `samples/stats.json` endpoint that returns averages for total reads and adjusted remaining reads.
* Sidebar reads data from that endpoint directly
* Added filter to avoid getting dimensions twice per page (with and without filters)
Screen Shot 2019-04-01 at 12 00 48 PM


app/controllers/samples_controller.rb
```diff
@@ -277,6 +277,31 @@ def dimensions
     end
   end
 
+  def stats
+    domain = params[:domain]
```
jshoe: I see params used directly a lot in this controller but should probably whitelist and use sample_params?

tfrcarvalho: Discussed offline.
Apparently, the strong parameters just filter out parameters and log that, letting the action proceed as normal, but without the parameters.
It looks as if this is more useful for write actions, where we might want to pass the parameters directly to create and/or update actions.
If misconfigured, the effect of strong parameters might actually be worse than attempts to tamper it.

PR #2108, opened 2019-04-03
add alignment length to report

cdebourcy:
Helps users remove spurious hits by filtering on alignment length


happyimadesignr:
@cdebourcy does affect the interface? Can you send screenshot?

PR #2114, opened 2019-04-03
[Location] Start geosearch endpoint

jshoe:
### Notes
- Adds locations/external_search endpoint for geosearch/geocoding queries (only logged in users)
- Location IQ had a Ruby library https://locationiq.com/docs#official-client-libraries but they say it's autogenerated and untested so I thought urls are fine. They also have an autocomplete-specific endpoint that could provide additional results (but it expects you to keep typing so it would keep suggesting "UCSF Medical Center" if you typed in "UCSF").
- Lots of potential fields returned by the API https://locationiq.com/docs-html/index.html#forward_response I found display_name/country/state/county/city/lat/lon to be important for us.
- Rate limiting can be implemented on the backend if we end up using up our usage.
- Formatting behavior like "allow just the plain text input" (like coordinates without resolving) can be done on the frontend.

### Demo
Screen Shot 2019-04-03 at 3 58 23 PM
Screen Shot 2019-04-03 at 3 48 02 PM
Screen Shot 2019-04-03 at 3 47 53 PM
Screen Shot 2019-04-03 at 3 47 17 PM

### Test and Reproduce
- To get the API key locally, either edit your entrypoint.sh to start chamber in production, or retrieve the key from Parameter Store and do `LOCATION_IQ_API_KEY=key docker-compose up web`
- Pull the branch, migrate, and go to pages like http://localhost:3000/locations/external_search?query=bay%20area or http://localhost:3000/locations/external_search?query=UCSF
- See JSON results

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -1,0 +1,39 @@
+class LocationsController < ApplicationController
+  def external_search
+    query = location_params[:query]
+    render(json: []) && return if query.blank?
+
+    results = []
+    resp = Location.geosearch(query)
+
+    if resp.is_a?(Net::HTTPSuccess)
+      candidates = JSON.parse(resp.body)
+      candidates.each do |c|
+        name_parts = c["display_name"].partition(", ")
+        results << {
+          title: name_parts[0],
+          description: name_parts[-1],
+          country: c["address"]["country"] || "",
+          state: c["address"]["state"] || "",
+          county: c["address"]["county"] || "",
+          city: c["address"]["city"] || "",
+          lat: c["lat"] || "",
+          lon: c["lon"] || ""
+        }
+      end
+    end
+    render json: results
+  rescue => err
+    render json: {
+      status: "failed",
+      message: "Unable to perform geosearch",
+      errors: [err]
+    }, status: :internal_server_error
+  end
+
+  private
+
+  def location_params
+    params.permit(:query)
+  end
+end
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -20,6 +20,8 @@ class LocationsController < ApplicationController
         end
       end
     end
+    event = MetricUtil::ANALYTICS_EVENT_NAMES[:location_geosearched]
+    MetricUtil.log_analytics_event(event, current_user, query: query, request: request)
     render json: results
   rescue => err
     render json: {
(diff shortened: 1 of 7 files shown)


db/migrate/20190403210056_add_location_table.rb
```diff
@@ -0,0 +1,5 @@
+class AddLocationTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :locations
```
jshoe: I only added this for the sake of getting the mock to work. Future migration will add for real

app/controllers/locations_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+            county: c["address"]["county"] || "",
+            city: c["address"]["city"] || "",
+            lat: c["lat"] || "",
+            lon: c["lon"] || ""
+          }
+        end
+      end
+    end
+    render json: results
+  rescue => err
```
tfrcarvalho: What happens in case of error? 
Can the user insert any value? Would it be blocked?

test/controllers/locations_controller_test.rb
```diff
@@ -0,0 +1,62 @@
...
+      assert_equal "[]", @response.body
+    end
+  end
+
+  test "user can geosearch and see an error" do
+    post user_session_path, params: @user_params
+    get external_search_locations_path, params: { query: "UCSF" }
+    assert_response :error
+    assert_equal JSON.dump(@api_key_error), @response.body
+  end
```
tfrcarvalho: Maybe an extra test for empty query string?

db/migrate/20190403210056_add_location_table.rb
tfrcarvalho: What will be the purpose of the table? Are we saving locations here (vs. metadata fields)?

test/controllers/locations_controller_test.rb
jshoe: 👍

db/migrate/20190403210056_add_location_table.rb
jshoe: I think we probably should treat location separately, even though it breaks our old MetadataField mold. In the MetadataField approach we would have country/state/county/city/lat/lon as Metadatum on the sample, but then it'd be hard to query "what are the coordinates for this place", right? Maybe we can have hooks that update collection_location to be a display_name for a simple value, and then an associated record in Locations table for more info?

jshoe: Actually just deleted this. Don't need it "just for the stub"

PR #2118, opened 2019-04-04
Add generic analytics logging on active record create, update or destroy

gregdingle:
# Description

This will turn on analytics tracking of all current and future CRUD operations at the ActiveRecord level. 

I believe it is better to do this generically at the ActiveRecord level instead of at a higher level in ActiveController or in React because we can get automatic, reliable coverage and consistent meaning, all derived from the Rails framework. I plan on adding more granular events at higher levels where needed, as well as all read-only events, such as viewing a taxon tree. 

QUESTION: Are there any tables which are updated so frequently as to cause too much data to be logged? 

## Test

1. Create and update a heatmap.
2. See log message such as 
```
visualization_updated logged with properties {:id=>81, :name=>"Heatmap", :user_id=>94}
```

In staging
1. View new event traffic in segment

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,7 @@ class ApplicationController < ActionController::Base
   before_action :authenticate_user!
   before_action :check_rack_mini_profiler
   before_action :check_browser
+  before_action :set_current_user_for_logging!
 
   include Consul::Controller
 
@@ -26,12 +27,6 @@ class ApplicationController < ActionController::Base
   def no_demo_user
     login_required
     redirect_to root_path if current_user.demo_user?
-  end
-
-  def check_rack_mini_profiler
-    if current_user && current_user.admin?
-      Rack::MiniProfiler.authorize_request
-    end
   end
 
   # Rails method for adding to logging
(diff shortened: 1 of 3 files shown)


jshoe:
@gregdingle Sounds good to me, but thinking about it I'm actually concerned that the Segment analytics library will drop events at this level. I tried doing many events at once for "one event per sample when multiple samples uploaded" and many events didn't make it to the Segment debugger (but maybe needs more tuning). So for the create/update/destroy, if you're inserting a million taxon_counts when loading data, will this try to fire a million separate events for each record? Maybe you could push to staging to test a pipeline run sometime

gregdingle:
@jshoe I did some research. It seems that segment should be fine at less than 500 r/s (https://segment.com/docs/sources/server/http/#rate-limits) which should cover upload a sheet of samples or anything I can think of. In addition, segment says the server lib batches automatically (https://segment.com/docs/sources/server/http/#batch). 

With the taxon counts, I think the updates happen outside of ActiveRecord anyway. See `generate_aggregate_counts` and `store_summary`. Batch ActiveRecord operations such as `update_all ` and `delete_all` do not fire callbacks so we are safe there as well. If we do see an unreasonable number of events for some model, I would update the code to log 1% for that model to avoid the problem.

jshoe:
Ok that sounds good then. You may want to test locally though b/c I just didn't find the Ruby lib working as advertised. I would fire "1", "2" "3"... "10" and almost always lose a number in Segment debugger

PR #2119, opened 2019-04-05
[Bug fix] Fix feature guarding in heatmap.

tfrcarvalho:
Features were being loaded into a string, instead of an array leading to crashes in the frontend when user had no flags of any kind enabled.

Tested with users:
- With `allowed_features = nil`
- With some features (but not data discovery) enabled
- With at least data discovery enabled


jshoe:
any other places not using allowed_feature_list?

tfrcarvalho:
> any other places not using allowed_feature_list?

Not for reading.

PR #2120, opened 2019-04-05
Convention for analytics logging in React

gregdingle:
# Description

This establishes a convention for making analytics logging calls in React where user events are handled. It breaks the existing convention in `ANALYTICS_EVENT_NAMES`. 

Design goals:
* convention that can quickly and consistently applied to ~250 onClick, onChange, onClose and onSubmit callsites currently in the codebase
* make frontend and backend events easily distinguishable by name
* make event names findable in the codebase by simple string match
* make names machine parseable (i.e. splittable on delimiter)
* make meaning obvious in the terms of anybody who uses the app (*note* this is one reason why we can't simply automate the name generation completely)
* names are highly likely to be globally unique
* include important user groups automatically
* convention can easily be followed by any future React implementors

I put in a few callsites as examples.

# Test

1. Go to bulk upload form, create form and accession viz and do the following:
2. click around and enter text in inputs and submit
3. see no errors in console
4. see segement events fired with chrome extension

### DIFF
--- a/app/assets/src/api/index.js
+++ b/app/assets/src/api/index.js
@@ -248,9 +248,9 @@ const createProject = params =>
   });
 
 // See https://czi.quip.com/bKDnAITc6CbE/How-to-start-instrumenting-analytics-2019-03-06
+// See also documentation for withAnalytics below.
 const logAnalyticsEvent = (eventName, eventData = {}) => {
   // Wrapper around Segment analytics so we can add things later
-  // eventData should have keys in snake_case for the database
   if (window.analytics) {
     // Include high value user groups in event properties to avoid JOINs downstream.
     if (window.analytics.user) {
@@ -281,19 +281,22 @@ const logAnalyticsEvent = (eventName, eventData = {}) => {
  * context and meaning of an analytics event in a report and locate it in the
  * codebase.
  *
+ * The eventData should be used for extra data that would be useful for an
+ * analysis specific to the event. The keys in should be named the same as in
+ * the calling context for easy interpretation. Only scalars should be passed to
+ * keep things simple downstream. Arrays should be replaced by their lengths.
+ *
  * For example:
  *
  *    withAnalytics(
  *      this.renderMoreReads,
  *      "AccessionViz_more_reads_link_clicked",
- *      { reads: this.state.reads.length, allReads: this.allReads.length }
+ *      { projectId: this.state.projectId, reads: this.state.reads.length }
  *    )
  *
  * React events should have have a single callsite, so there is no need to put
  * them in ANALYTICS_EVENT_NAMES.
  *
- * The eventData should be used for extra data that would be useful for an
- * analysis specific to the event.
  **/
 const withAnalytics = (handleEvent, eventName, eventData = {}) => {
   return (...args) => {
--- a/app/assets/src/api/index.js
+++ b/app/assets/src/api/index.js
@@ -291,9 +291,9 @@ const logAnalyticsEvent = (eventName, eventData = {}) => {
  * The eventData should be used for extra data that would be useful for an
  * analysis specific to the event.
  **/
-const withAnalytics = (func, eventName, eventData = {}) => {
+const withAnalytics = (handleEvent, eventName, eventData = {}) => {
   return (...args) => {
-    const ret = func.apply(null, args);
+    const ret = handleEvent.apply(null, args);
     logAnalyticsEvent(eventName, eventData);
     return ret;
   };
--- a/app/assets/src/api/index.js
+++ b/app/assets/src/api/index.js
@@ -297,7 +297,7 @@ const logAnalyticsEvent = (eventName, eventData = {}) => {
  **/
 const withAnalytics = (handleEvent, eventName, eventData = {}) => {
   return (...args) => {
-    const ret = handleEvent.apply(null, args);
+    const ret = handleEvent(...args);
     logAnalyticsEvent(eventName, eventData);
     return ret;
   };
--- a/app/assets/src/api/index.js
+++ b/app/assets/src/api/index.js
@@ -262,6 +262,7 @@ const logAnalyticsEvent = (eventName, eventData = {}) => {
         czi_user: traits.biohub_user,
         demo_user: traits.demo_user,
         has_samples: traits.has_samples,
+        // caller can override above traits if they know what they are doing
         ...eventData
       };
     }
@@ -286,10 +287,13 @@ const logAnalyticsEvent = (eventName, eventData = {}) => {
  *
  * React events should have have a single callsite, so there is no need to put
  * them in ANALYTICS_EVENT_NAMES.
+ *
+ * The eventData should be used for extra data that would be useful for an
+ * analysis specific to the event.
  **/
 const withAnalytics = (func, eventName, eventData = {}) => {
-  return args => {
-    const ret = func(args);
+  return (...args) => {
+    const ret = func.apply(null, args);
     logAnalyticsEvent(eventName, eventData);
     return ret;
   };
--- a/app/assets/src/api/index.js
+++ b/app/assets/src/api/index.js
@@ -266,21 +266,7 @@ const logAnalyticsEvent = (eventName, eventData = {}) => {
         ...eventData
       };
     }
-
-    // Evaluate lazy values, giving some time for React to set state
-    // See https://reactjs.org/docs/react-component.html#setstate
-    let shouldWait = false;
-    for (var k in eventData) {
-      if (typeof eventData[k] === "function") {
-        shouldWait = true;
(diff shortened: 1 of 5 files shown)


app/assets/src/api/index.js
```diff
@@ -251,7 +251,56 @@ const createProject = params =>
...
+ *
+ * React events should have have a single callsite, so there is no need to put
+ * them in ANALYTICS_EVENT_NAMES.
+ *
+ * The eventData should be used for extra data that would be useful for an
+ * analysis specific to the event.
+ **/
+const withAnalytics = (handleEvent, eventName, eventData = {}) => {
+  return (...args) => {
+    const ret = handleEvent.apply(null, args);
```
MarkAZhang: Can you just do `handleEvent(...args)` here? 

`apply` will also set `this` to `null`, which I don't think is intended?

MarkAZhang:
LGTM!

app/assets/src/api/index.js
gregdingle: good catch!

app/assets/src/components/Samples.jsx
```diff
@@ -1657,7 +1660,10 @@ class BackgroundModal extends React.Component {
           <div className="button-container">
             <PrimaryButton
               text="Create Collection"
-              onClick={this.handleOpen}
+              onClick={withAnalytics(
+                this.handleOpen,
+                "Samples_collection_create_button_clicked"
```
jshoe: So with this scheme, I'm assuming to find number of collection creations attempted, an analyst would want to join all the related "collection create" events? or should there be another collection create event that is always called

app/assets/src/api/index.js
```diff
@@ -251,7 +251,56 @@ const createProject = params =>
...
+    window.analytics.track(eventName, eventData);
+  }
+};
+
+/**
+ * For wrapping event handlers in React. The first part of the name should be
+ * the React component. The middle part should be user-friendly UI name. The
+ * last part should be the action taken.
+ *
+ * <ReactComponent>_<friendly_name_of_ui_element>_<past_tense_action>
```
tfrcarvalho: Actually, I would use a different separator inside and outside of the components if possible: 
`<ReactComponent>_<friendly-name-of-ui-element>_<past_tense_action>` and
`AccessionViz_more-reads-link_clicked`

tfrcarvalho: Or alternatively just use camelCase for each component: `AccessionViz_moreReadsLink_clicked`...

app/assets/src/api/index.js
```diff
@@ -251,7 +251,56 @@ const createProject = params =>
...
+ * React events should have have a single callsite, so there is no need to put
+ * them in ANALYTICS_EVENT_NAMES.
+ *
+ * The eventData should be used for extra data that would be useful for an
+ * analysis specific to the event.
+ **/
+const withAnalytics = (handleEvent, eventName, eventData = {}) => {
+  return (...args) => {
+    const ret = handleEvent(...args);
+    logAnalyticsEvent(eventName, eventData);
```
tfrcarvalho: Shouldnt this be an async call? Or `withAnalytics` and async function?

app/assets/src/api/index.js
```diff
@@ -251,7 +251,56 @@ const createProject = params =>
 const logAnalyticsEvent = (eventName, eventData = {}) => {
```
tfrcarvalho: Shouldnt this be an async call?

app/assets/src/components/BulkUploadImport.jsx
```diff
@@ -527,7 +535,15 @@ class BulkUploadImport extends React.Component {
               <form
                 className="bulkSubmitForm"
                 ref="form"
-                onSubmit={this.handleRemoteUploadSubmit}
+                onSubmit={withAnalytics(
+                  this.handleRemoteUploadSubmit,
+                  "BulkUploadImport_upload_form_submitted",
+                  {
+                    samples: this.state.samples.length,
+                    project_id: this.state.projectId,
```
tfrcarvalho: Shouldnt we impose camelCase for arguments too ? Or is this constrained by thee server? 
I see some other messages above that use camelCase.

app/assets/src/components/Samples.jsx
gregdingle: An analyst could join all collection events to get stats such as collection success rate. 

There should be a backend collections event as well, `collection_created` for example, which I expect will be more commonly queried. 

It's good to have the diversity though for answering different questions about the user experience.

app/assets/src/api/index.js
gregdingle: Yes, using the dash (-) as a separator in the middle part would make it more parseable downstream, but not much, because snowflake and other systems I can imagine us using have a slice function. See https://docs.snowflake.net/manuals/sql-reference/functions/array_slice.html . 

CamelCase on the other hand is not easily parseable and downstream systems may be case insensitive. 

So I'm ambivalent between format such as `AccessionViz_more-reads-link_clicked` and `AccessionViz_more_reads_link_clicked`. I'm happy to go with the most popular.

app/assets/src/api/index.js
gregdingle: The segment API is synchronous and returns immediately. I believe events are put into a queue for dispatch by http request.

app/assets/src/api/index.js
gregdingle: Comment above.

app/assets/src/components/BulkUploadImport.jsx
gregdingle: Good catch. I'll make them all consistent with the names in the calling context.

app/assets/src/api/index.js
tfrcarvalho: Using the latter would be a problem if you ever need a more complex action name. So, I believe the former is more general and less likely to cause problems in the future.

app/assets/src/api/index.js
tfrcarvalho: I would still make it asynchronous (1) since it is easy to do to, (2) avoids any dependencies on segment API which we do not control, (3) we do not use the return value of the call.
Given your answer, it is not as critical... 

We should probably have a try catch block around handleEvent to still log the interaction. This to avoid logging before the handling...

app/assets/src/api/index.js
```diff
@@ -248,10 +248,78 @@ const createProject = params =>
...
+      };
+    }
+
+    // Evaluate lazy values, giving some time for React to set state
+    // See https://reactjs.org/docs/react-component.html#setstate
+    let shouldWait = false;
+    for (var k in eventData) {
+      if (typeof eventData[k] === "function") {
+        shouldWait = true;
+        setTimeout(() => (eventData[k] = eventData[k]()), 1);
```
gregdingle: @tfrcarvalho this is a solution to a problem I ran into. For a call such as 

```
withAnalytics(
                    this.handleEmailChange,
                    "CreateUser_email_changed",
                    {
                      email: this.state.email
                    }
                  )
```

The `email` value is before being updated by `handleEmailChange` which is not what an analyst would want. Wrapping in function to make the value lazy `email: () => this.state.email` is still pretty clean.

But there is a second problem. React setState is async, so in the common case the above wouldn't work. Adding a `setTimeout` fixes the problem, at the cost of some indeterminism. We can't be 100% sure when React will be done. 

What do you think? If it is all too complicated, I'm happy to throw out the lazy evaluation and avoid setting this type of event-specific dynamic data until we actually have a need for it. Logging the event names and static data is 95% of the value right now.

tfrcarvalho: Yeah, I would avoid using state whenever possible.
IIUC, this particular example can be changed to `email: (event) => event.target.value`. I would actually whenever possible avoid reading from state, for the reasons you mention before and most create logs should come from event data. We are logging the action, not the processing.

Do you have a compelling example that cannot be refactored?

gregdingle: No, I do not. I'm going to revert. Thanks!

app/assets/src/api/index.js
gregdingle: Ok I talked to Mark as well and he preferred the format `AccessionViz_more-reads-link_clicked` so I'll go with that.

app/assets/src/api/index.js
gregdingle: Maybe I'm not getting you completely, but I don't see how it's easy to make `withAnalytics` async. Wouldn't that require `await` before every call?

tfrcarvalho: No, you can call an async function and continue without waiting for it. 
You only need `await` if you are interested in the result.

gregdingle: I'm going to merge this PR now so I can make some progress and I trust we can discuss the async question later.

PR #2122, opened 2019-04-06
Revise narrow container width rules and apply to all view headers.

tfrcarvalho:
* Apply narrow containers to view headers to keep sync with the content.
* Fix a lack of sync bug between headers and contents in narrow windows.

# Type of change

Bug fix and clean up.

# Test and Reproduce

Check all pages that use NarrowContainer to check for components with different width behaviors.

Screen Shot 2019-04-05 at 5 49 17 PM

### DIFF
--- a/app/assets/src/components/layout/narrow_container.scss
+++ b/app/assets/src/components/layout/narrow_container.scss
@@ -1,4 +1,8 @@
 .narrowContainer {
+  // Per design, narrow container has a max-width of 1280px with a minimum margin of 14px.
+  // Since there is no min-margin property we set max-width to 2xmin_margin and
+  // ensure the min margin by removing the margin width from the actual width of
+  // the component (i.e. 100% - 2xmin_margin)
   max-width: 1308px;
   margin: 0 auto;
   width: calc(100% - 28px);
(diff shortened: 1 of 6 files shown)


app/assets/src/components/layout/narrow_container.scss
```diff
@@ -1,7 +1,7 @@
 .narrowContainer {
-  max-width: 1280px;
+  max-width: 1308px;
```
gregdingle: please add a comment to explain how this value was chosen

app/assets/src/components/layout/narrow_container.scss
```diff
@@ -1,7 +1,7 @@
...
   margin: 0 auto;
-  width: 100%;
+  width: calc(100% - 28px);
```
gregdingle: please add a comment to explain the `-28px`

app/assets/src/components/layout/narrow_container.scss
tfrcarvalho: done

app/assets/src/components/layout/narrow_container.scss
tfrcarvalho: done

PR #2127, opened 2019-04-10
Fix some issues in analytics

gregdingle:
# Description

This fixes some issues in analytics.

1. provide more context to GA
2. fix `request` arg in ruby
3. remove old events covered by auto events in ruby
4. add name scheme validation

# Test

1. Check new context in segment debugger
2. see no errors in ruby log
3. try valid and invalid names such as "AccessionViz_more-reads-link_clicked" and "AccessionViz_more-reads_link_clicked" and see appropriate errors


app/assets/src/api/index.js
```diff
@@ -262,6 +262,9 @@ const logAnalyticsEvent = (eventName, eventData = {}) => {
         czi_user: traits.biohub_user,
         demo_user: traits.demo_user,
         has_samples: traits.has_samples,
+        // For Google Analytics. See https://segment.com/docs/destinations/google-analytics/#track
+        label: JSON.stringify(eventData),
+        category: eventName.split("_")[0],
```
tfrcarvalho: Why category? This is the component name, right? 
Did we describe what category field should be for somewhere else?

gregdingle: `label` and `category` are special keys for GA. I'll update the comment.

PR #2129, opened 2019-04-10
[Data Discovery] Add placeholder column to align with the plus icon.

tfrcarvalho:
# Description

This PR adds an extra column to the table when selecting columns is enable. This allows the scrollbar to show at the right of the plus column. 
The plus icon is currently outside of the table because, if it is part of the table, will be re-rendered everytime we select an option and add a column to the table causing some flickering.

Apr-10-2019 09-43-07

# Test and Reproduce

Open a table that has the option to select columns enabled (e.g. Samples Table) and verify the scrollbar is on the rightmost side of the table.


app/assets/src/components/visualizations/table/BaseTable.jsx
```diff
@@ -212,6 +212,9 @@ class BaseTable extends React.Component {
                   />
                 );
               })}
+              {initialActiveColumns && (
+                <Column dataKey={"plusP  laceholder"} width={20} />
```
jshoe: typo `plusP  laceholder` ?

PR #2130, opened 2019-04-10
[MetadataFields] Fix display of custom fields on the heatmap with CSS escape

jshoe:
### Description
- Type of change: Bug fix
- Problem: Custom fields from CSV like "Temperature Exposure" or "Bashing (2x, 1min each)" would not display on the heatmap. (First one would not show values, second one would crash the page).
- Cause: There is one spot on the heatmap frontend that use the name value as a code-friendly key. (Metadata field search box list selector and the heatmap row displays).
- Solution: We can just escape the CSS selector.

### Test and Reproduce
- Tested by creating values for "Custom Field", "CustomField", "Bashing (2x, 1min each)", "Bashing |2x, 1min each|", and verifying they work on the heatmap with distinct correct values.

### Demo
Screen Shot 2019-04-10 at 12 09 00 PM
Screen Shot 2019-04-10 at 12 11 20 PM

Screen Shot 2019-04-10 at 9 49 00 AM
Screen Shot 2019-04-10 at 9 57 24 AM

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
@@ -242,6 +242,7 @@ class SamplesHeatmapVis extends React.Component {
   getSelectedMetadata() {
     const sortByLabel = (a, b) => (a.label > b.label ? 1 : -1);
 
+    console.log("this.metadataTypes: ", this.metadataTypes);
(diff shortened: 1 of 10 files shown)


jshoe:
@MarkAZhang @tfrcarvalho I actually did a 180 and did the local selector change instead. It worked well to use CSS.escape and allows us to not sanitize names. See updated description.

PR #2133, opened 2019-04-10
Performance optimization for `/samples/{sample_id}/report_info` endpoint.

davidrissato:
# Description

Optimizations in this code change:
1) ActiveRecord `.each` method creates an instance of TaxonLineage for each row, which is resource intensive and extremely slow. That would only be useful if we needed to invoke any additional method of TaxonLineage. Since we only need the raw values, we can avoid this transformation by using `.pluck`. Pluck returns just an array of values instead of an array of TaxonLineage instances.
2) Previous code was also iterating using `.each` and adding results to a Hash one by one. These steps were optimized by chaining `.map` and `.to_h`.
3) The `.as_json` invocation on each record is much slower than zipping (`.zip`) the column names to the values returned by `.pluck`. (this performance gain is directly related to item 1).
4) Small optimization on the query used to retrieve lineage_version. This version is resolving in one query (previous was using 2).
5) Added # frozen_string_literal: true to this module to save resources
6) Removed unused fields from lineage sub-structure: created_at, ended_at, started_at, updated_at, version_end, version_start

(additional details: https://docs.google.com/presentation/d/1bjyjv6Heqtf_dsb20H8auqt1b1ArNpUON2aNgp7nddU/edit?usp=sharing)

## Type of change

- [x] Performance tuning

# Test and Reproduce

1. On `/home`, click on a sample that has a status "Complete"
2. On `/samples` page, table view should be displayed as usual
2. On `/samples` page, taxonomic tree view should be displayed as usual

Which pages?

- [x] Report Page

# Testing Script:

###Sample Details Page
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

### DIFF
--- a/app/models/taxon_lineage.rb
+++ b/app/models/taxon_lineage.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # The TaxonLineage model gives the taxids forming the taxonomic lineage of any given species-level taxid.
 require 'elasticsearch/model'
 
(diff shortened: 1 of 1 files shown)


app/models/taxon_lineage.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
tfrcarvalho: Maybe add a comment for this line...
What is the impact of this?

davidrissato: This magic comment needs to be at the very first line of the source file, and it is used to implicitly mark all string literals as frozen.
It provides better resource usage, since Ruby knows that is safe to reuse string literals across multiple calls because it is guaranteed the string cannot be changed.
The original proposal can be found here: https://bugs.ruby-lang.org/issues/8976#note-30

davidrissato: (by the way, the original PR description mentions it:

> 5) Added # frozen_string_literal: true to this module to save resources

gregdingle:
Nice work! Nice to see the pluck still effective with the larger number of columns. 

there is one more thing you could look at for this change. We set up perf monitoring in airbrake. Once this is deployed in prod, you could check to see what decrease there is in the endpoint over all prod traffic.

PR #2134, opened 2019-04-10
Add analytics tracking to report page

gregdingle:
# Description

This adds event logging to `PipelineSampleReport` component and its immediate children. My basic strategy was to instrument inline with `withAnalytics` or `logAnalyticsEvent`. If there was an important `setState` that happened in the handler, I put the logging in the handler as a callback to `setState`. 

**Please** check that the event names make sense. You probably know better than me!

I did a small bit of refactoring along the way such as changing JS to camelCase and grouping imports. 

# Test and Reproduce
1. open a sample report
2. try clicking on some tracked events such as switching view
3. see log data in segment chrome extension
4. see no errors in console

### DIFF
--- a/app/assets/src/components/Login.jsx
+++ b/app/assets/src/components/Login.jsx
@@ -1,8 +1,11 @@
 import React from "react";
 import axios from "axios";
 import $ from "jquery";
-import LogoIcon from "./ui/icons/LogoIcon";
+
 import { openUrl } from "~utils/links";
+import { logAnalyticsEvent, withAnalytics } from "~/api/analytics";
+
+import LogoIcon from "./ui/icons/LogoIcon";
 
 class Login extends React.Component {
   constructor(props, context) {
--- a/app/assets/src/components/Login.jsx
+++ b/app/assets/src/components/Login.jsx
@@ -66,23 +66,37 @@ class Login extends React.Component {
   }
 
   isFormInValid() {
-    if (this.refs.email.value === "" && this.refs.password.value === "") {
-      this.setState({
-        showFailedLogin: true,
-        errorMessage: "Please enter email and password"
+    const logError = () =>
+      logAnalyticsEvent("Login_login-form-error_displayed", {
+        errorMessage: this.state.errorMessage
       });
+
+    if (this.refs.email.value === "" && this.refs.password.value === "") {
+      this.setState(
+        {
+          showFailedLogin: true,
+          errorMessage: "Please enter email and password"
+        },
+        logError
+      );
       return true;
     } else if (this.refs.email.value === "") {
-      this.setState({
-        showFailedLogin: true,
-        errorMessage: "Please enter email"
-      });
+      this.setState(
+        {
+          showFailedLogin: true,
+          errorMessage: "Please enter email"
+        },
+        logError
+      );
       return true;
     } else if (this.refs.password.value === "") {
-      this.setState({
-        showFailedLogin: true,
-        errorMessage: "Please enter password"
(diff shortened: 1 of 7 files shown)


app/assets/src/components/Login.jsx
```diff
@@ -66,23 +69,37 @@ class Login extends React.Component {
   }
 
   isFormInValid() {
-    if (this.refs.email.value === "" && this.refs.password.value === "") {
-      this.setState({
-        showFailedLogin: true,
-        errorMessage: "Please enter email and password"
+    const logError = () =>
```
jshoe: do you want to add a call a few lines above this one to `errorMessage: "Invalid Email and Password"`?

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -800,14 +839,36 @@ class PipelineSampleReport extends React.Component {
...
+        onNcbiActionClick={withAnalytics(
+          this.gotoNCBI,
+          "PipelineSampleReport_ncbi-link_clicked",
+          analyticsContext
+        )}
         fastaEnabled={fastaEnabled}
-        onFastaActionClick={this.downloadFastaUrl}
+        onFastaActionClick={withAnalytics(
+          this.downloadFastaUrl,
+          "PipelineSampleReport_fasta-link_clicked",
```
jshoe: maybe `taxon-fasta-link` because many links could be called a fasta link on this page

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -833,13 +894,25 @@ class PipelineSampleReport extends React.Component {
...
+          <span
+            className="taxon-sidebar-link"
+            onClick={withAnalytics(
+              onTaxonClickHandler,
+              "PipelineSampleReport_taxon-sidebar-link_clicked"
+            )}
+          >
             {taxonNameDisplay}
           </span>
         );
```
jshoe: you're probably saving this for later but I've always wondered if people are hovering over the PathogenPreview dots below

app/assets/src/components/Login.jsx
```diff
@@ -66,23 +69,37 @@ class Login extends React.Component {
...
+      logAnalyticsEvent("Login_login-form-error_displayed", {
+        errorMessage: this.state.errorMessage
       });
+
+    if (this.refs.email.value === "" && this.refs.password.value === "") {
```
tfrcarvalho: This could be more compact and without using state... something like
```
errors = compact([
  this.refs.email.value === "" ? "email": null,
  this.refs.password.value === "" ? “Password” : null
]);
if (errors.length > 0) {
  const errorMessage = `Please enter ${errors.join(“ and “)}`;
  this.setState({
    showFailedLogin: true,
    errorMessage
  });
  logError(errorMessage);
  return true;
}
```

app/assets/src/components/Login.jsx
gregdingle: good catch

app/assets/src/components/PipelineSampleReport.jsx
gregdingle: `PathogenPreview` and `ReportInsightIcon` are tricker to instrument because they don't have event hooks, so I think we should do them later.

app/assets/src/components/Login.jsx
gregdingle: True, but I'd prefer to keep with purely additive changes to avoid extra manual testing.

PR #2137, opened 2019-04-11
[Data Discovery] Remove dates from details columns and move them into their own.

tfrcarvalho:
## Description

The dates that we previously shown in the main column, have their own column now. On non-infinite tables (projects and visualizations) that column is sortable.

Apr-11-2019 09-32-49

Smaller changes also in this PR:
* Refactor: move static functions for rendering table fields to their own class `TableRenderers`
* Changed date format to `DD-MM-YYYY` both on the table and sidebar
* Fixed an alignment issue of cells in same table rows (all cells need to align to the top text of the main cell)

# Test and Reproduce

On projects, visualizations and samples view notice the presence of a date column whose header might vary (Created, Uploaded, Updated, ...). The columns must:
* Be the first column after the main column with details for the model
* Have content
* Have two rows: absolute date and approx. time elapsed since then.
The details column should not have date information (neither absolute nor elapsed).

### DIFF
--- a/app/assets/src/components/views/projects/ProjectsView.jsx
+++ b/app/assets/src/components/views/projects/ProjectsView.jsx
@@ -4,11 +4,11 @@ import PropTypes from "prop-types";
 import PrivateProjectIcon from "~ui/icons/PrivateProjectIcon";
 import PublicProjectIcon from "~ui/icons/PublicProjectIcon";
 import BaseDiscoveryView from "~/components/views/discovery/BaseDiscoveryView";
+import TableRenderers from "~/components/views/discovery/TableRenderers";
 // CSS file must be loaded after any elements you might want to override
 import cs from "./projects_view.scss";
 
 import { find, merge, pick } from "lodash/fp";
-import moment from "moment";
 
 class ProjectsView extends React.Component {
   constructor(props) {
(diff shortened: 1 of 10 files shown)


app/assets/src/components/views/projects/ProjectsView.jsx
```diff
@@ -33,22 +33,28 @@ class ProjectsView extends React.Component {
...
         width: 200,
         disableSort: true,
-        cellRenderer: BaseDiscoveryView.renderList
+        cellRenderer: TableRenderers.renderList
       },
       {
         dataKey: "number_of_samples",
         width: 140,
-        label: "No. of Samples"
+        label: "No. of Samples!!"
```
jshoe: "!!" typo?

jshoe:
Personally I'm not used to seeing DD-MM-YYYY (vs MM/DD/YYYY or YYYY-MM-DD).. was this from @happyimadesignr ?

tfrcarvalho:
> Personally I'm not used to seeing DD-MM-YYYY (vs MM/DD/YYYY or YYYY-MM-DD).. was this from @happyimadesignr ?

Yes, and from @eggerdesigns.

happyimadesignr:
I personally prefer MMDDYYYY =] DDMMYYYY is what most of the world uses.

Screen Shot 2019-04-04 at 3 32 14 PM

happyimadesignr:
@eggerdesigns, FYI, we went back to year first b/c that format is also very common and creates less confusion.

PR #2139, opened 2019-04-11
Add hover tooltips to coverage viz.

MarkAZhang:
Updated mock data to reflect the data produced from the pipeline step.

Mock data now contains additional metrics for contigs, reads, coverage bins, and the accession itself.
Mock data now aggregates contigs and reads into hit groups.
Mock data now treats coverage as a sparse array. (the data is a lot smaller now)

Coverage viz now has tooltips for coverage bins and hit groups.

Histogram and GenomeViz now support hover tooltips.

Tooltips will display within a buffer distance, since bars can be very small.

coverage

## Type of change

# Test and Reproduce

1. Verified that coverage viz displays without crashing.
2. Verified that non-admin users are not affected.


app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
```diff
@@ -234,15 +272,51 @@ export default class CoverageVizBottomSidebar extends React.Component {
...
+        <div className={cs.headerControls}>
+          <Dropdown
+            options={this.getAccessionOptions()}
+            value={get("id", currentAccessionSummary)}
+            label="Accession"
+            onChange={this.setCurrentAccession}
+            rounded
+          />
+          <div className={cs.vizLinkContainer}>
+            <a
```
tfrcarvalho: I have seen a few times... Should we create a link object to guarantee this is done correctly?

app/assets/src/components/visualizations/Histogram.js
```diff
@@ -271,6 +336,18 @@ export default class Histogram {
       }
     }
 
+    this.midpointsToDataIndices = midpointsToDataIndices;
+    this.sortedMidpoints = midpoints.sort((a, b) => a - b);
+
+    this.svg
```
tfrcarvalho: Are you doing that refactor soon? I see we perpetuating this appending without data binding issue...

app/assets/src/components/visualizations/Histogram.js
```diff
@@ -173,6 +176,57 @@ export default class Histogram {
     return this.options.barOpacity || 0.8;
   };
 
+  // Find the bar midpoint that is closest to hoverX, within hoverBuffer.
```
tfrcarvalho: Can you clarify what a "bar midpoint" is ?
This is not something that applies to all histograms, right?

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
```diff
@@ -137,21 +145,47 @@ export default class CoverageVizBottomSidebar extends React.Component {
...
-    }));
+  handleHistogramHover = (pageX, pageY, hoverData) => {
+    const { currentAccessionData } = this.state;
+
+    if (hoverData && hoverData[0] === 0) {
+      this.setState({
+        histogramTooltipParams: {
+          left: pageX,
+          top: pageY,
+          tooltipData: getHistogramTooltipData(
```
tfrcarvalho: I think there is a better way to do this... I would have to understand the context better.
But you typically set state fo the hovered element, and get the data on re-rendering if it changes. 
Definitely, left and top should not be in state.

app/assets/src/components/visualizations/Histogram.js
MarkAZhang: Changed to "bar x-center". It's the center of the bar on the x-dimension.

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
MarkAZhang: Refactored into separate "enter", "hover", "exit" events as discussed.

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
```diff
@@ -174,23 +230,30 @@ export default class CoverageVizBottomSidebar extends React.Component {
...
+      data.coverage_bin_size
     );
 
     this.contigReadViz = new GenomeViz(
       this.contigReadVizContainer,
       contigReadVizData,
       {
-        domain: [0, data.total_length]
+        domain: [0, data.total_length],
+        colors: ["#A9BDFC", "#3768FA"],
```
tfrcarvalho: Is it worth moving this colors to some kind of color scheme for later use?

PR #2140, opened 2019-04-12
Add No Data Message to Coverage Viz.

MarkAZhang:
Also tweak the title and read-level viz link.
Show the accession name instead of the accession id.

Screenshot of the "no data" message:

Screen Shot 2019-04-11 at 5 58 42 PM

The accession name with ellipsis:

Screen Shot 2019-04-11 at 5 59 09 PM


app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
```diff
@@ -92,11 +93,11 @@ export default class CoverageVizBottomSidebar extends React.Component {
...
 
   componentDidUpdate(prevProps, prevState) {
-    const { visible, params } = this.props;
+    const { params } = this.props;
     const { currentAccessionData } = this.state;
 
-    if (!prevProps.visible && visible) {
-      this.setCurrentAccession(params.accessionSummaries[2].id);
+    if (params.accessionSummaries !== prevProps.params.accessionSummaries) {
+      this.setCurrentAccession(get([2, "id"], params.accessionSummaries));
```
tfrcarvalho: This is not new... but a good pattern for this magic numbers to improve readability is to name it... I do not know what 2 means but you could define a constant like: `ACCESSION_THAT_MATTERS_INDEX=2`.
This would allows use to understand what we trying to load...

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
```diff
@@ -214,16 +220,18 @@ export default class CoverageVizBottomSidebar extends React.Component {
...
-          >
-            {currentAccessionSummary.id}
-          </a>
+          <div className={cs.ncbiLinkWrapper}>
+            <a
+              href={`https://www.ncbi.nlm.nih.gov/nuccore/${
+                currentAccessionSummary.id
+              }?report=genbank`}
+              target="_blank"
+              rel="noopener noreferrer"
```
tfrcarvalho: I referred to this in another PR. I believe we should create a Link component for this.

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
```diff
@@ -341,8 +350,43 @@ export default class CoverageVizBottomSidebar extends React.Component {
     );
   }
 
+  renderNoDataContents() {
+    const { params } = this.props;
+
+    return (
+      <NarrowContainer className={cs.contents}>
+        <div className={cs.header}>
```
tfrcarvalho: I have a NoResultsBanner in discovery view, that is similar to this one. I think we should try to consolidate the two of them.

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
MarkAZhang: This is temporary code for mock data. Will add a TODO.

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
MarkAZhang: Hey, talked to Jenn and we decided it's okay to special case this one for now. It's different from the other one (longer text, different icon), and not clear if it's re-usable like the "NoResultsBanner" is.

PR #2142, opened 2019-04-12
[Data Discovery] Update URLs with state we want to save.

tfrcarvalho:
## Description

This PR loads the state to save in Data Discovery to the URL. This enables shareable URLs and solves some navigations issues.
As per current UX decision, clicking "My Data" and "Public" resets the whole state. Later, we might consider keeping some state either in URL or session state.

Apr-11-2019 21-24-46

## Test and Reproduce

Navigate within the Data Discovery Views, and verify that the URL is updated with the correct values for `filters`, `currentTab`, `project`, `showFilters`, `showStats`. 
Verify that clicking `My Data` or `Public` takes you back to a reset view like in a first load.


jshoe:
@tfrcarvalho did you intend to request review?

app/assets/src/components/utils/UrlQueryParser.js
```diff
@@ -0,0 +1,63 @@
+import QueryString from "query-string";
+
+class UrlQueryParser {
+  // This util class enhances query-string to be enable parsing
```
jshoe: 'to enable'

app/assets/src/components/utils/UrlQueryParser.js
```diff
@@ -0,0 +1,63 @@
+import QueryString from "query-string";
+
+class UrlQueryParser {
```
gregdingle: This kind of class is crying out for a unit test.

PR #2148, opened 2019-04-15
Store number metadata as decimal instead of float.

MarkAZhang:
This avoids loss of accuracy in metadata.

Unlike float, decimal columns can cause mysql range errors.
We ensure that range errors are properly caught.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# Test and Reproduce

1. Verify that normal values can still be input on Sample Sidebar in both Report page and Heatmap page.
2. Verify that integer and decimal values for "Host Age" are not rounded weirdly when input.
3. Verified that migration can be done forward and backward without error, and that errors are automatically converted with no rounding errors for integers. For floats, we will need to do a check in prod after the migration happens, as 1.0f can become 1.0000000001 as a decimal. A little scary to automatically round in the migration, so we can compare to the raw value manually.
4. Verified that large values throw errors gracefully in the sidebar and metadata upload.

Screen Shot 2019-04-15 at 12 14 15 PM

Screen Shot 2019-04-15 at 12 07 25 PM

5. Verified that new sample upload with metadata still works.

(anywhere with metadata)
Report Page
Phylo Tree
Heatmap
Metadata Upload Modal

### DIFF
--- a/app/helpers/error_helper.rb
+++ b/app/helpers/error_helper.rb
@@ -7,6 +7,7 @@ module ErrorHelper
     INVALID_OPTION = "Invalid option for metadata field.".freeze
     INVALID_DATE = "Invalid date for metadata field.".freeze
     INVALID_NUMBER = "Invalid number for metadata field.".freeze
+    NUMBER_OUT_OF_RANGE = "Number is out of range.".freeze
   end
 
   module MetadataUploadErrors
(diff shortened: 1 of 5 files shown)


MarkAZhang:
Float is more performant and memory-efficient, but can lose precision compared to decimal. I would recommend switching over on a case-by-case basis, and being more conscious about which one we use for new columns. Probably fine to use float for most cases.

https://stackoverflow.com/questions/1056323/difference-between-numeric-float-and-decimal-in-sql-server

PR #2151, opened 2019-04-16
[Mapping] Start map_playground endpoint and MapPlayground components

jshoe:
### Description
- Starts a locations/map_playground endpoint that will return currently viewable values for MetadataField collection_location. Eventually there will be new Location objects to use instead.
- Endpoint tests in locations_controller_test.rb.
- Starts MapPlayground and BaseMap React components.
- Includes setup for react-map-gl and tiles from MapTiler.
- NPM changes were produced by running `npm install react-map-gl`

### Demo
- Will look like this (but more zoomed in)
Screen Shot 2019-04-15 at 5 58 07 PM

### Test and Reproduce
- Pull the branch, retrieve MAPTILER_API_KEY from Parameter Store, start locally with `MAPTILER_API_KEY=key_here docker-compose up web`
- Visit http://localhost:3000/locations/map_playground
- See map displayed and explore


test/controllers/locations_controller_test.rb
```diff
@@ -67,4 +67,23 @@ class LocationsControllerTest < ActionDispatch::IntegrationTest
     assert_response :error
     assert_equal JSON.dump(@api_key_error), @response.body
   end
+
```
cdebourcy: Could you write some "joe CANNOT..." type access control tests for the new `map_playground` endpoint?
"joe cannot see collection location for another user's private sample in map playground"

app/assets/src/components/views/discovery/BaseMap.jsx
```diff
@@ -0,0 +1,63 @@
...
+        zoom
+      }
+    };
+  }
+
+  render() {
+    const { mapTilerKey, renderMarker, markers } = this.props;
+    const { viewport } = this.state;
+
+    const styleID = "a63d8846-0896-4fd0-b062-c9d992a49b63";
```
tfrcarvalho: Is this key not sensitive?

app/controllers/locations_controller.rb
```diff
@@ -31,6 +31,24 @@ def external_search
     }, status: :internal_server_error
   end
 
+  def map_playground
+    # Show all viewable locations
+    field_id = MetadataField.find_by(name: "collection_location")
+    viewable = current_power.samples.pluck(:id)
```
tfrcarvalho: As discussed, we should try to stop using large lists of sample ids, in lieu of using the actual conditions to the database. In this case, something like:
```
samples_with_metadata = samples
                            .includes(metadata: :metadata_field)
                            .where(metadata: { metadata_field_id: metadata_field.metadata_field_id })
```

app/assets/src/components/views/discovery/BaseMap.jsx
tfrcarvalho: Never mind... seems like it is just a public id.

test/controllers/locations_controller_test.rb
jshoe: 👍

app/controllers/locations_controller.rb
jshoe: 👍

PR #2152, opened 2019-04-16
Add metadata to accession select in coverage viz.

MarkAZhang:
Update mock data to latest format.

Accession select now shows contig count and coverage.
Augmented BareDropdown and Dropdown slightly to handle this.
Accession select is now sorted by best score, descending.

Accession select now has popup showing the accession name.

Screen Shot 2019-04-15 at 6 23 34 PM

## Type of change

# Test and Reproduce

1. Open the coverage viz sidebar and select different accessions. Note that the best accession is at the top. 
2. Non-admin users are not affected.

### DIFF
--- a/app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
+++ b/app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
@@ -1,6 +1,15 @@
 // This component allows you to easily add a dropdown of options to any trigger element
 // Wraps around Semantic UI Dropdown, with react-popper as an alternative.
 
+// NOTE ABOUT CUSTOMIZING THE DROPDOWN:
+// If you don't want the user to be able to select a value, and you want to specify
+// custom onClick behavior yourself, use props.options. For example, if you want to provide the user
+// with a menu of options.
+// If you do want the user to be able to select a value, and you want the option to be rendered in a
+// custom way, use props.items and specify a "customNode" for each item.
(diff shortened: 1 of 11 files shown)


app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
```diff
@@ -1,6 +1,13 @@
 // This component allows you to easily add a dropdown of options to any trigger element
 // Wraps around Semantic UI Dropdown, with react-popper as an alternative.
 
+// NOTE ABOUT CUSTOMIZING THE DROPDOWN:
+// If you don't want the user to be able to select a value, and you want to specify
```
jshoe: it sounds a little confusing to say "you don't want the user to be able to select a value"? you just mean not select within this component right?

MarkAZhang: Yeah, for example in the User Menu, the user isn't selecting a value, they're picking from a list of navigation. Will think about how to word this. Thanks for feedback.

tfrcarvalho: A good way to structure the use cases is a bullet list with a key title. Something like: 
```
* options with default rendering that return a value: ...
* options with custom rendering that return a value: ...
* options that trigger an immediate action (e.g. navigations links): ...
```

PR #2153, opened 2019-04-16
Load any validation errors surfaced by the pipeline and display them on the sample page

cdebourcy:
When available, this can provide visibility into the reason for the failure and save on-call work.

image

Summoning @happyimadesignr to run the frontend addition by her (see screenshot).

### DIFF
--- a/app/assets/src/components/views/SampleView/SampleView.jsx
+++ b/app/assets/src/components/views/SampleView/SampleView.jsx
@@ -296,6 +296,9 @@ class SampleView extends React.Component {
         return (
           <div className={cs.failedContainer}>
             <h6 className={cs.failed}>Sample run failed</h6>
+            <h6 className={cs.failed}>
+              {this.props.pipelineRun.error_message}
+            </h6>
           </div>
         );
       }
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -397,9 +397,9 @@ class PipelineRun < ApplicationRecord
 
   def db_load_input_validations
     file = Tempfile.new
-    Syscall.s3_cp(s3_file_for("input_validations"), file.path)
-    dict = JSON.parse(File.read(file))
-    error_message = dict["Validation error"]
+    downloaded = PipelineRun.download_file_with_retries(s3_file_for("input_validations"),
+                                                        file.path, 3, false)
+    error_message = downloaded ? JSON.parse(File.read(file))["Validation error"] : nil
     update(error_message: error_message) if error_message
     file.unlink
   end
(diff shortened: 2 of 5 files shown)


app/models/pipeline_run.rb
```diff
@@ -393,6 +395,15 @@ def succeeded?
     job_status == STATUS_CHECKED
   end
 
+  def db_load_input_validations
+    file = Tempfile.new
+    downloaded = PipelineRun.download_file_with_retries(s3_file_for("input_validations"),
+                                                        file.path, 3, false)
+    error_message = downloaded ? JSON.parse(File.read(file))["Validation error"] : nil
+    update(error_message: error_message) if error_message
```
jshoe: may want to append to existing error_message if we use this field for other messages in the future

app/models/pipeline_run.rb
```diff
@@ -967,22 +980,25 @@ def local_amr_drug_summary_path
...
     end
   end
 
-  def self.download_file(s3_path, destination_dir)
-    command = "mkdir -p #{destination_dir};"
-    command += "aws s3 cp #{s3_path} #{destination_dir}/;"
-    _stdout, _stderr, status = Open3.capture3(command)
-    return nil unless status.exitstatus.zero?
-    "#{destination_dir}/#{File.basename(s3_path)}"
+  def self.download_file(s3_path, destination, dest_is_dir = true)
```
MarkAZhang: Maybe add a comment "If the destination path is a directory, create the directory if necessary and append the file name to the end of the destination path. If the destination path includes a file name, assume that the parent directories already exist."

MarkAZhang: Took a little while to understand this function.

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -296,6 +296,9 @@ class SampleView extends React.Component {
         return (
           <div className={cs.failedContainer}>
             <h6 className={cs.failed}>Sample run failed</h6>
+            <h6 className={cs.failed}>
+              {this.props.pipelineRun.error_message}
```
jshoe: Design-wise the error message could have a separate style to be more friendly (just normalcase, non-red)

app/models/pipeline_run.rb
cdebourcy: ok, added your comment

app/models/pipeline_run.rb
cdebourcy: good point

app/assets/src/components/views/SampleView/SampleView.jsx
cdebourcy: true

happyimadesignr:
Thanks for including me @cdebourcy. We have this style for errors that @MarkAZhang created.

At some point, it would be nice to take some time to audit all the places we have error states and make them consistent.

Screen Shot 2019-04-23 at 9 48 45 AM

PR #2154, opened 2019-04-16
[Data Discovery] Refactor search

tfrcarvalho:
* Split between client and server searches:
	- Client searches uses information from dimensions to feed the suggestions on the search box
	- Server searches, as before, triggers a server on server side (for models with a high count like samples and taxons)
* Refactor server searches to be able to restrain by domain (i.e. public, my_data, etc.)
* Created a new component LiveSearchBox (which is a refactor of SearchBox) that is more flexible on how search suggestions is loaded. Also solves some bugs on triggering concurrent searches and avoid using outdated results.
* Some UI styling adjustments to the search box results

Testing: 
* Search for each of the models that search box in the header supports (projects, samples, taxa, location, host and tissues) and verify if the correct filter or action is triggered.
* Switch tabs (public and my data) and verify that results are constrained to the corresponding domain.
* As a follow up of this change, I will work on a manual script to validate data discovery for more comprehensive testing.


app/helpers/elasticsearch_helper.rb
```diff
@@ -15,47 +15,83 @@ def prefix_match(model, field, prefix, condition)
...
 
-  def taxon_search(prefix, tax_levels = TaxonCount::NAME_2_LEVEL.keys, project_id = nil)
+  def taxon_search(prefix, tax_levels = TaxonCount::NAME_2_LEVEL.keys, filters = {})
     return {} if Rails.env == "test"
     prefix = sanitize(prefix)
-    matching_taxa = {}
 
+    matching_taxa = []
+    taxon_ids = []
     tax_levels.each do |level|
```
cdebourcy: now that we started to use aggregations... can we get rid of the iteration over tax levels and handle that using an aggregation as well?

tfrcarvalho: Maybe... What did you have in mind?

The aggregations is to help guarantee that the genus query returns distinct results. When we were querying by genus we could have n equal results all for the same genus.
The two separate queries avoid skewing the number of results towards the species side (since there are many more of them).

Also, because we do not have an explicit level number/name, we cannot filter by a particular set of levels. What we can do is something like an additional filter like `taxid=genus_id OR taxonid=species_id ...`. I might explore that in a future PR.

PR #2155, opened 2019-04-16
Show total number of accessions in coverage viz, even hidden ones.

MarkAZhang:
Also adjust names and ordering of metric.

Use en-dash for ranges of numbers.

Add Help Icon to explain hidden accessions.

Screen Shot 2019-04-16 at 2 35 29 PM

### DIFF
--- a/app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
+++ b/app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
@@ -3,7 +3,9 @@ import { find, get } from "lodash/fp";
 import cx from "classnames";
 import ReactDOM from "react-dom";
 
-import Sidebar from "~/components/ui/containers/Sidebar";
+import { formatPercent } from "~/components/utils/format";
+import Sidebar from "~ui/containers/Sidebar";
+import HelpIcon from "~ui/containers/HelpIcon";
 import PropTypes from "~/components/utils/propTypes";
 import Histogram from "~/components/visualizations/Histogram";
 import GenomeViz from "~/components/visualizations/GenomeViz";
@@ -50,7 +52,7 @@ const METRIC_COLUMNS = [
     {
       key: "referenceLength",
       name: "Reference Length",
-      tooltip: "Length in bp of the reference accession sequence."
+      tooltip: "Length in base pairs of the reference accession."
     }
   ],
   [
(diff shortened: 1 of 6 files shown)


app/assets/src/components/common/CoverageVizBottomSidebar/utils.js
```diff
@@ -15,7 +15,7 @@ export const getHistogramTooltipData = memoize(
         data: [
           [
             "Base Pair Range",
-            `${Math.round(coverageObj[0] * binSize)} - ${Math.round(
+            `${Math.round(coverageObj[0] * binSize)}\u2013${Math.round(
```
gregdingle: are we not allowing full utf-8 literals in our codebase?

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
```diff
@@ -59,19 +60,19 @@ const METRIC_COLUMNS = [
       name: "Aligned Contigs",
       tooltip: "Number of contigs for which this accession was the best match."
     },
+
```
tfrcarvalho: nit: remove empty line...

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
```diff
@@ -333,23 +332,43 @@ export default class CoverageVizBottomSidebar extends React.Component {
       referenceLength: currentAccessionData.total_length,
       alignedContigs: currentAccessionSummary.num_contigs,
       maxAlignedLength: currentAccessionData.max_aligned_length,
-      coverageDepth: currentAccessionData.coverage_depth,
-      coverageBreadth: currentAccessionData.coverage_breadth,
+      coverageDepth: `${currentAccessionData.coverage_depth}x`,
+      coverageBreadth: `${100 * currentAccessionData.coverage_breadth}%`,
```
tfrcarvalho: nit: Should percent be a function in case we need to enforce changes (e.g. precision) for all?

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
```diff
@@ -525,7 +544,7 @@ export default class CoverageVizBottomSidebar extends React.Component {
         onClose={onClose}
         direction="bottom"
       >
-        {params.accessionSummaries && params.accessionSummaries.length > 0
+        {params.accessionData && params.accessionData.best_accessions.length > 0
```
tfrcarvalho: Is there any possibility of best_accessions not being filled?

app/assets/src/components/common/CoverageVizBottomSidebar/utils.js
MarkAZhang: I decided to keep it in the character encoding because the en-dash looks a lot like a hyphen, and I don't want someone to accidentally change it to a hyphen in the future. Added a comment that "\u2013 is en-dash". We can change this approach if we start using a lot of these unicode characters.

PR #2158, opened 2019-04-17
Hook coverage viz up to back-end endpoints.

MarkAZhang:
Remove mock data from front-end.

Back-end uses mock data from S3.
This mock data is the coverage viz pipeline step run on our standard benchmark sample.
We will remove the mock data and change the back-end to point to the correct files when we add the coverage-viz pipeline step to idseq-web.

--
Testing steps:
Verified that non-admin users can still open old alignment viz without error.
If admin user goes to benchmark sample run, they can see coverage viz with data.
If admin user goes to non-benchmark sample, they can see coverage viz with data if the taxon they clicked on is also included in the benchmark sample (as expected). Otherwise, they see the 'no-data' message without error.

Screen Shot 2019-04-17 at 1 35 23 PM


app/assets/src/components/visualizations/GenomeViz.js
```diff
@@ -118,14 +118,6 @@ export default class GenomeViz {
       }
     }
 
-    if (this.options.onGenomeVizHover) {
```
jshoe: why did this go away?

app/controllers/samples_controller.rb
```diff
@@ -1111,6 +1111,34 @@ def upload_heartbeat
...
+    if coverage_viz_summary_s3_path
+      @coverage_viz_summary = get_s3_file(coverage_viz_summary_s3_path)
+      render json: @coverage_viz_summary
+    else
+      render json: {
+        error: "coverage viz summary file does not exist for this sample"
+      }
+    end
+  end
+
```
jshoe: you may still want some sort of rescue => json error for these two in case a fetching step fails

app/models/pipeline_run.rb
```diff
@@ -421,6 +421,16 @@ def db_load_ercc_counts
     update(total_ercc_reads: total_ercc_reads)
   end
 
+  def coverage_viz_summary_s3_path
+    # TODO(mark): Replace mock data with real data once we are running the sample.
+    return "s3://idseq-samples-development/markazhang/samples_12452/generate_coverage_viz_test/3.5/coverage_viz_summary.json"
+  end
+
```
jshoe: these files could've gone in assets.idseq.net

cdebourcy: Seconded. James led an effort a couple months ago to remove all explicit s3 paths from our open source code because it is too revealing

app/assets/src/components/visualizations/GenomeViz.js
MarkAZhang: It was replaced with `this.options.onGenomeVizBarHover`, but forgot to remove.

app/models/pipeline_run.rb
MarkAZhang: Okay, moved to s3://assets.idseq.net

PR #2162, opened 2019-04-17
[Mapping] Add CircleMarker and zoom controls to MapPlayground

jshoe:
### Description
- Type of change: Enhancement
- Add NavigationControl (zoom in/out) to the bottom-right of the map
- Add a CircleMarker with adaptable sizing based on zoom levels and a basic hover popup (next up is onClick popup)
- Improve coordinate regex matching
- Moved files into a mapping sub-folder, so the diff appears larger :3
- Note: For drawing markers we could also put the points onto a defined map layer if needed. For drawing popups we could use Popup from react-map-gl instead if needed.

### Demo
- These coordinates were randomly generated
Screen Shot 2019-04-17 at 4 09 43 PM
Screen Shot 2019-04-17 at 4 14 49 PM

### Test and Reproduce
- Pull the branch, retrieve MAPTILER_API_KEY from Parameter Store, start locally with `MAPTILER_API_KEY=key_here docker-compose up web`
- Visit http://localhost:3000/locations/map_playground
- Try out the navigation controls and hover popups


app/assets/src/components/views/discovery/mapping/BaseMap.jsx
```diff
@@ -20,6 +20,11 @@ class BaseMap extends React.Component {
     };
   }
 
+  updateViewport = viewport => {
+    this.setState({ viewport });
+    this.props.updateViewport && this.props.updateViewport(viewport);
```
tfrcarvalho: As a note for the future (need to update dev guidelines), we going towards a pattern where we load all props and state at the beginning of the function, like

```javascript
updateViewport = viewport => {
	const { updateViewPort } = this.props;
    this.setState({ viewport });
    updateViewport && updateViewport(viewport);
};
```

Becomes more relevant for more complex functions

app/assets/src/components/views/discovery/mapping/CircleMarker.jsx
```diff
@@ -0,0 +1,49 @@
...
+        onClick={onClick}
+        className={cx(cs.circle, hoverContent && cs.hoverable)}
+      />
+    );
+    return (
+      <svg
+        height={size}
+        viewBox={`0 0 ${size} ${size}`}
+        // Proper centering
+        style={{ transform: `translate(${-size / 2}px, ${-size / 2}px)` }}
```
tfrcarvalho: So, you get center at 50% but then you translate it back?

app/assets/src/components/views/discovery/mapping/CircleMarker.jsx
```diff
@@ -0,0 +1,49 @@
...
+class CircleMarker extends React.Component {
+  render() {
+    const { size, onClick, hoverContent } = this.props;
+
+    const circleBody = (
+      <circle
+        cx="50%"
+        cy="50%"
+        // Don't let edges get cut off in the viewBox
+        r={Math.max(size - 4, 1) / 2}
```
tfrcarvalho: Are edges size 4? Can you make it into a property?

app/assets/src/components/views/discovery/mapping/MapPlayground.jsx
```diff
@@ -0,0 +1,86 @@
...
+class MapPlayground extends React.Component {
+  state = {
+    locationsToSamples: {},
+    viewport: {}
+  };
+
+  componentDidMount() {
+    const { results } = this.props;
+
+    const locationsToSamples = Object.assign({}, this.state.locationsToSamples);
```
tfrcarvalho: This seems to be part of initializing state for the first time and is not DOM related. You should do this on the constructor (or better, from an initialization function called from the constructor)

app/assets/src/components/views/discovery/mapping/BaseMap.jsx
jshoe: 👍

app/assets/src/components/views/discovery/mapping/CircleMarker.jsx
jshoe: 👍 I'm not sure why it was around 4 but lower than that and I noticed some edge distortion 
Screen Shot 2019-04-18 at 11 07 47 AM

app/assets/src/components/views/discovery/mapping/MapPlayground.jsx
jshoe: thnx

app/assets/src/components/views/discovery/mapping/CircleMarker.jsx
jshoe: So this was going off an example but I think the svg transform places the viewbox over the center, and then the circle center places the circle in the center of the viewbox. is there another way to do this?

Without the viewbox transform:
Screen Shot 2019-04-18 at 11 02 58 AM

Without the circle center:
Screen Shot 2019-04-18 at 11 03 35 AM

Both:
Screen Shot 2019-04-18 at 11 04 08 AM

PR #2165, opened 2019-04-18
Move human filter for non-human hosts after subsampling step

cdebourcy:
For performance reasons.


cdebourcy:
> LGTM. What more is required for moving a pipeline step?

This should be all... just left "WIP" because I haven't tested it yet.

PR #2166, opened 2019-04-19
Add coverage_viz pipeline step to experimental dag template json

MarkAZhang:
Add coverage viz to idseq-web dag template.

Also a slight modification to the postprocessing step.

Currently gated with pipeline version >= 3.6.

As soon as the coverage viz step gets pushed to idseq_dag master, idseq_web will start
running that pipeline (due to this change).
We plan to test the coverage viz step in staging and prod using admin options before pushing
the coverage viz step to master.

The front-end is further gated on allowed_feature=coverage_viz_enabled and admin-only,
which we will remove in an upcoming PR.

---

Tested that:
* Existing samples can still be viewed without issue.
* A new sample that is run without admin-options will have pipeline-version v3.5 and succeed without issue.
* A new sample that is run with idseq-dag branch = markzhang/coverage_viz will have pipeline-version v3.6 and succeed without issue. The coverage viz properly loads on the sample.

### DIFF
--- a/app/assets/src/components/utils/sample.js
+++ b/app/assets/src/components/utils/sample.js
@@ -3,7 +3,17 @@ import { last } from "lodash/fp";
 export const pipelineVersionHasAssembly = pipelineVersion => {
   if (!pipelineVersion) return false;
   const versionNums = pipelineVersion.split(".");
-  return +versionNums[0] >= 3 && +versionNums[1] >= 1;
+  return (
+    +versionNums[0] >= 4 || (+versionNums[0] === 3 && +versionNums[1] >= 1)
+  );
+};
+
+export const pipelineVersionHasCoverageViz = pipelineVersion => {
+  if (!pipelineVersion) return false;
+  const versionNums = pipelineVersion.split(".");
+  return (
+    +versionNums[0] >= 4 || (+versionNums[0] === 3 && +versionNums[1] >= 6)
+  );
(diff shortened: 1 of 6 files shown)


app/assets/src/components/utils/sample.js
```diff
@@ -6,6 +6,12 @@ export const pipelineVersionHasAssembly = pipelineVersion => {
   return +versionNums[0] >= 3 && +versionNums[1] >= 1;
 };
 
+export const pipelineVersionHasCoverageViz = pipelineVersion => {
+  if (!pipelineVersion) return false;
+  const versionNums = pipelineVersion.split(".");
+  return +versionNums[0] >= 3 && +versionNums[1] >= 6;
```
jshoe: this will fail starting 4.0 ?

app/lib/dags/experimental.json.erb
```diff
@@ -7,6 +7,18 @@
       "gsnap.hitsummary.tab",
       "rapsearch2.hitsummary.tab"
     ],
+    <% if attribute_dict[:pipeline_version].to_f >= 3.6 %>
```
jshoe: this would fail if we push a "3.10"?

MarkAZhang: This check will get much more verbose when we get to 3.10...

PR #2168, opened 2019-04-19
Add frontend analytics to SampleView, data discovery views

gregdingle:
# Description

image

This adds analytics to the remaining controls of the sample report page and to the new data discovery views. See https://czi.quip.com/67RCAIiHN0Qc/IDseq-product-analytics-How-to-log for methodology.

image

# Test and Reproduce

Click on all new added analytics such as download button 
Observe events in chrome debugger
Observe no new errors


app/assets/src/components/views/SampleView/SampleViewControls.jsx
```diff
@@ -3,14 +3,15 @@ import SvgSaver from "svgsaver";
...
+import { logAnalyticsEvent } from "~/api/analytics";
 import DownloadButtonDropdown from "~/components/ui/controls/dropdowns/DownloadButtonDropdown";
 import PrimaryButton from "~/components/ui/controls/buttons/PrimaryButton";
 import {
   getDownloadDropdownOptions,
   getLinkInfoForDownloadOption
 } from "~/components/views/report/utils/download";
 
-class Controls extends React.Component {
+class SampleViewControls extends React.Component {
```
gregdingle: I renamed this globally

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -187,7 +202,17 @@ export default class DiscoverySidebar extends React.Component {
       return (
         <div className={cs.barChartRow} key={`${value}_row_${i}`}>
           <dt className={cs.barLabel} key={`${value}_label_${i}`}>
-            <a onClick={() => this.handleFilterClick(value)}>
+            <a
+              onClick={() => {
+                this.handleFilterClick(value);
```
gregdingle: this does not appear to be working yet

app/assets/src/components/views/Landing.jsx
```diff
@@ -56,6 +56,9 @@ class Landing extends React.Component {
         }
       })
       .then(() => {
+        // Although this event does not follow current naming conventions for
+        // frontend events, we keep it for continuity. See
```
jshoe: I just realized this event probably doesn't work anymore if we're not loading Segment on the landing page.. it could be moved to a backend event in HomeController#sign_up

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -408,28 +419,62 @@ class SampleView extends React.Component {
...
               <div className={cs.sampleDetailsLinkContainer}>
                 <span
                   className={cs.sampleDetailsLink}
-                  onClick={this.toggleSampleDetailsSidebar}
+                  onClick={withAnalytics(
+                    this.toggleSampleDetailsSidebar,
+                    "SampleView_sample-details-link_clicked",
+                    {
+                      sampleId: sample.id,
+                      sampleName: sample.name
```
jshoe: out of curiosity, do you always want to log the sample name with the sample ID?

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -163,7 +172,13 @@ export default class DiscoverySidebar extends React.Component {
         {extraRows.length > 0 && (
           <div
             className={cs.showHide}
-            onClick={() => this.toggleExpandedMetadataGroup(field)}
+            onClick={() => {
+              this.toggleExpandedMetadataGroup(field);
+              logAnalyticsEvent("DiscoverySidebar_show-more-toggle_clicked", {
```
jshoe: for the sake of analytics, would you recommend breaking "toggle" functions into a show function and a hide function?

app/assets/src/components/views/Landing.jsx
gregdingle: good catch. I confirmed none of those events have been logged lately. I will move it to the backend.

app/assets/src/components/views/SampleView/SampleView.jsx
gregdingle: In general, I advocate logging the name of an object with the ID as a convenience for analysis downstream. It avoids a JOIN with little cost.

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
gregdingle: We should just make sure the info of showing or hiding is available in the logged event, which it was not here, so good catch!

PR #2171, opened 2019-04-22
Log status changes in active record objects

gregdingle:
# Description

See comment in code. 

# Test and Reproduce
Upload a local sample
See log message 

`Event 'sample_created' logged with properties {:id=>12305, :name=>"RR004_water_2_S23A", :user_id=>94, :project_id=>165, "status"=>"created"`


jshoe:
@gregdingle tip: when you only have 1 commit on the PR, Github won't use your PR title in the commit history :|

gregdingle:
@jshoe Thanks for the tip, I didn't know that!

PR #2173, opened 2019-04-22
Add analytics to new upload flow and site header

gregdingle:
# Description

This adds analytics tracking to the last big parts of the frontend of the app. 

image

image

image

# Test and Reproduce

Click on all new added analytics such as project selector
Observe events in chrome debugger
Observe no new errors

### DIFF
--- a/app/assets/src/components/common/Header.jsx
+++ b/app/assets/src/components/common/Header.jsx
@@ -1,14 +1,17 @@
 import React from "react";
 import PropTypes from "prop-types";
+import { forbidExtraProps } from "airbnb-prop-types";
+import cx from "classnames";
+
 import { RequestContext } from "~/components/common/RequestContext";
 import ToastContainer from "~ui/containers/ToastContainer";
 import BareDropdown from "~ui/controls/dropdowns/BareDropdown";
 import LogoIcon from "~ui/icons/LogoIcon";
 import { openUrl } from "~utils/links";
 import { deleteAsync } from "~/api/core";
-import { forbidExtraProps } from "airbnb-prop-types";
+import { logAnalyticsEvent, withAnalytics } from "~/api/analytics";
+
 import cs from "./header.scss";
-import cx from "classnames";
 
 class Header extends React.Component {
   render() {
@@ -119,6 +122,9 @@ const UserMenuDropDown = ({
           <a
             className={cs.option}
             href={`mailto:${email}?Subject=Report%20Feedback`}
+            onClick={() =>
+              logAnalyticsEvent("Header_dropdown-feedback-option_clicked")
+            }
           >
             Report Feedback
           </a>
@@ -132,6 +138,9 @@ const UserMenuDropDown = ({
             target="_blank"
             rel="noopener noreferrer"
             href="https://assets.idseq.net/Terms.pdf"
+            onClick={() =>
+              logAnalyticsEvent("Header_dropdown-terms-option_clicked")
+            }
           >
             Terms of Use
           </a>
@@ -145,12 +154,22 @@ const UserMenuDropDown = ({
             target="_blank"
             rel="noopener noreferrer"
             href="https://assets.idseq.net/Privacy.pdf"
+            onClick={() =>
+              logAnalyticsEvent("Header_dropdown-privacy-policy-option_clicked")
+            }
           >
             Privacy Policy
           </a>
         }
       />,
-      <BareDropdown.Item key="7" text="Logout" onClick={signOut} />
+      <BareDropdown.Item
+        key="7"
+        text="Logout"
+        onClick={withAnalytics(
+          signOut,
+          "Header_dropdown-logout-option_clicked"
+        )}
+      />
     );
     return userDropdownItems;
   };
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -16,6 +16,7 @@ import _fp, {
   isEqual
 } from "lodash/fp";
 
+import { logAnalyticsEvent } from "~/api/analytics";
 import MultipleDropdown from "~ui/controls/dropdowns/MultipleDropdown";
 import DataTable from "~/components/visualizations/table/DataTable";
 import PropTypes from "~/components/utils/propTypes";
@@ -151,7 +152,7 @@ class MetadataManualInput extends React.Component {
   // Convert metadata headers and fields to a CSV-like format before passing to parent.
   onMetadataChange = (newHeaders, newFields) => {
     // Only send fields for the selected samples to the parent component.
-    // If a sample was added, and then later removed, that sample's metadata will not be sent up,
+    // If a sample was added, and then later removed, that sample's metadata thll not be sent up,
     // but will still persist in this component.
     const sampleNames = map("name", this.props.samples);
     const fieldsForSamples = pickBy(
@@ -186,6 +187,9 @@ class MetadataManualInput extends React.Component {
 
   handleColumnChange = selectedFieldNames => {
     this.setState({ selectedFieldNames });
+    logAnalyticsEvent("MetadataManualInput_column-selector_changed", {
+      selectedFieldNames: selectedFieldNames.length
+    });
   };
 
   getHostGenomeOptions = () =>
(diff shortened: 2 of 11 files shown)


app/assets/src/components/common/MetadataManualInput.jsx
```diff
@@ -257,7 +267,7 @@ class MetadataManualInput extends React.Component {
   isHostGenomeIdValidForField = (hostGenomeId, field) =>
     includes(
       hostGenomeId,
-      this.props.projectMetadataFields[field].host_genome_ids
+      get([field, "host_genome_ids"], this.props.projectMetadataFields) || false
```
gregdingle: @MarkAZhang I was getting an "undefined" error here when I tried "apply all" . This seems like an appropriate fix. Please confirm!

app/assets/src/components/views/SampleUploadFlow/RemoteSampleFileUpload.jsx
```diff
@@ -53,12 +62,26 @@ class RemoteSampleFileUpload extends React.Component {
...
+
+      logAnalyticsEvent("RemoteSampleFileUpload_connect_succeeded", {
+        projectId: this.props.project.id,
+        bulkPath: this.state.remoteS3Path,
+        newSamples: newSamples.length
+      });
     } catch (e) {
       if (e.status.startsWith("No samples imported")) {
         this.setState({
+          // TODO (gdingle): we should have an error state for
```
gregdingle: cc @MarkAZhang

app/assets/src/components/common/Header.jsx
```diff
@@ -145,12 +154,22 @@ const UserMenuDropDown = ({
             target="_blank"
             rel="noopener noreferrer"
             href="https://assets.idseq.net/Privacy.pdf"
+            onClick={() =>
+              logAnalyticsEvent("Header_dropdown-privacy-policy-option_clicked")
+            }
```
MarkAZhang: Do these onClick events consistently fire before the browser navigates to the new page? It's an edge case I'm not sure about.

app/assets/src/components/common/MetadataManualInput.jsx
```diff
@@ -151,7 +152,7 @@ class MetadataManualInput extends React.Component {
   // Convert metadata headers and fields to a CSV-like format before passing to parent.
   onMetadataChange = (newHeaders, newFields) => {
     // Only send fields for the selected samples to the parent component.
-    // If a sample was added, and then later removed, that sample's metadata will not be sent up,
+    // If a sample was added, and then later removed, that sample's metadata thll not be sent up,
```
MarkAZhang: Typo.

app/assets/src/components/common/MetadataManualInput.jsx
MarkAZhang: Hey, this is definitely safer and should be correct.

I don't think you need the `|| false`. `get(...)` will default to undefined, and `includes` should be able to accept `undefined` as a parameter and default gracefully to `false`. One of the great things about lodash.

I'm curious how you got this error! I wasn't able to reproduce it when I tried briefly.

app/assets/src/components/visualizations/table/BaseTable.jsx
```diff
@@ -73,6 +74,10 @@ class BaseTable extends React.Component {
   handleColumnChange = selectedColumns => {
     const { protectedColumns } = this.props;
     this.setState({ activeColumns: concat(protectedColumns, selectedColumns) });
+    logAnalyticsEvent("BaseTable_column-selector_changed", {
```
MarkAZhang: I worry that if we use BaseTable across the app, it can be difficult to distinguish the analytics event of a particular BaseTable. Not sure what we can do about that...Maybe have `BaseTable` take a `logId` prop that it logs? So I can set `logId="SamplesTable"` when I use it in the samples page (bad example) and the BaseTable will pass that id to snowflake.

MarkAZhang: Same concern goes for all generic components like this.

app/assets/src/components/views/SampleUploadFlow/RemoteSampleFileUpload.jsx
MarkAZhang: I can see the value of having a more specific error message for an invalid bucket path. We'll have to distinguish this case on the back-end. Good idea!

app/assets/src/components/views/SampleUploadFlow/UploadMetadataStep.jsx
```diff
@@ -102,13 +105,22 @@ class UploadMetadataStep extends React.Component {
           <div className={cs.controls}>
             <PrimaryButton
               text="Continue"
-              onClick={this.handleContinue}
+              onClick={withAnalytics(
```
MarkAZhang: Probably also good to put a log step at line 57, which is where manual metadata gets validated and any issues are displayed.

app/assets/src/components/common/MetadataUpload.jsx
```diff
@@ -57,6 +61,16 @@ class MetadataUpload extends React.Component {
       issues,
       validatingCSV
     });
+    if (validatingCSV) {
```
MarkAZhang: IMO I don't think you need this first one.

This code is a bit confusing, but whenever the user validates metadata, this function is called twice.

The first time is before the request is sent. It just clears the errors and sets validatingCSV to true so <MetadataUpload> can display a loader.
The second time is after the request is returned. This function is called with the actual issues and sets validating CSV to false.

So when validatingCSV is true, issues.errors.length is always 0. It's just to reset the state.

MarkAZhang:
Thanks so much for doing this!!

app/assets/src/components/common/MetadataManualInput.jsx
gregdingle: Thanks. I'll remove the `false`. I got the error using the test files someone shared with the team around month ago. 

image

app/assets/src/components/common/Header.jsx
gregdingle: That's a good question. It was for sure a problem a number of years ago (10y?!) when I was doing something similar. However, in this case it is not an issue because of `target="_blank"`. In general, it should not much of an issue for IDSeq analytics because 1) we track pageviews as well when a page loads and we have the referrer for ordinary internal links 2) we don't have any external links that open in the same browser tab.

app/assets/src/components/visualizations/table/BaseTable.jsx
gregdingle: I share the concern. That's why in general I tried to log in the outermost component, wrapping passed-in callbacks. In this case, the behavior is entirely internal to `BaseTable`. So, as a fallback, we need to look at the url path that is logged with every analytics event to establish context.

That said, the argument could be made in this case we really do care more about the column-select behavior across pages.

app/assets/src/components/views/SampleUploadFlow/UploadMetadataStep.jsx
gregdingle: Good point. I'll move it inside handleContinue.

app/assets/src/components/common/MetadataUpload.jsx
gregdingle: Thanks for explaining. I noticed the double logging and I didn't know what to do exactly. I will change it to only log when `validatingCSV` is false.

app/assets/src/components/views/SampleUploadFlow/UploadMetadataStep.jsx
```diff
@@ -67,6 +70,13 @@ class UploadMetadataStep extends React.Component {
         issues: this.state.issues
       });
     }
+    logAnalyticsEvent("UploadMetadataStep_continue-button_clicked", {
```
MarkAZhang: I would log "this.state.wasManual" as well, since there are two code paths in this function.

Also, this.setState is asynchronous, so this.state might not immediately update after you call it. `this.state.issues` here might not reflect the `setState({ issues })` call on line 58. You'll want to use `result.issues`.

app/assets/src/components/visualizations/table/BaseTable.jsx
MarkAZhang: Ah, the URL is captured as well. That should help.

app/assets/src/components/common/Header.jsx
MarkAZhang: Makes sense!

MarkAZhang:
Just one comment. Otherwise looks good.

app/assets/src/components/views/SampleUploadFlow/UploadMetadataStep.jsx
gregdingle: Good catch! I normally watch out for `setState` but I missed this one. 

I'll log `wasManual` as well.

PR #2174, opened 2019-04-23
[Mapping] Start initial migration for Locations model

jshoe:
### Description
- Open for discussion!
- Create a model for storing Locations with multiple levels of specificity as discussed previously. I will add the cleaned up db/schema.rb file once before merging.

- Why do we need to store different broken down levels of regions for each location?
The main reason for wanting to store explicit levels (country/state/subdivision/city) is for efficient filtering and also strict compliance before storing anything (e.g. never storing belong state level on sensitive samples).

- Why do we have these particular 4 levels + coordinates?
This was a simplification of all the types of potential administrative levels (https://wiki.openstreetmap.org/wiki/Key:place#Relationship_with_admin_level) and also the geosearch providers will return these levels in almost all cases, so we are offloading most of the burden of categorization onto their APIs. These will not be perfect categorizations for every location but they should work well enough for our uses.

- Why not have additional columns country_id/state_id/subdivision_id/etc. referencing other entries in the table?
If the user added a new entry for MyCountry|MyState|MySubdivision|MyCity, you'd need to do a query to fill in MyCountry with the existing country entry id if any, or disambiguate between two subdivisions named MySubdivision for the sake of assigning the right unique ID. I don't think these additional columns would yield much benefit over composite queries over country_name|state_name|etc, but I could be convinced otherwise..

- How will you query between ambiguous names, e.g. 31 counties in the U.S. named Washington County?
We will assume that 'taxonomically' there are no duplicates with identical upwards lineages. E.g. that there are not multiple Washington Counties in the same state and country. So queries will need to specify their ancestors.

- Why set the fields to default empty string and not null?
This simplifies some queries because "!=" queries would not return null values, even though you might expect them to logically.

- Is there not a unique place identifier like OpenStreetMaps (OSM) IDs that we can store?
Unfortunately OSM IDs are not stable so we should not rely on them ("the element node 1 represents has changed many times: tree in Germany (v13), buoy in the Atlantic Ocean (v15), memorial stone in London (v17), bike rental in Edinburgh (v19)"). Also depending on the geosearch API, we may get unique IDs from the API, but I would rather not couple ourselves to a particular geosearch provider in that way. https://help.openstreetmap.org/questions/66577/how-often-the-osm_id-was-unsolicitedly-changed-what-the-risk-before-2020 and https://gis.stackexchange.com/questions/279755/can-i-depend-on-the-country-city-ids-of-osm

- How will we associate these with the actual Metadata/MetadataFields/Samples?
Open to suggestions but I plan to special-case MetadataField base_type #3 and have the Metadatum entry have a "location_id" pointer value.

### Test and Reproduce
- Pull the branch, run `docker-compose run web "rake db:migrate"`, see the migration run. Reload and make sure the app works. Run `docker exec -ti idseq-web_web rails console development` or similar. Type Location.all in the console and see `#<ActiveRecord::Relation []>`.

### DIFF
--- a/db/migrate/20190422185437_add_location_table.rb
+++ b/db/migrate/20190422185437_add_location_table.rb
@@ -1,0 +1,31 @@
+class AddLocationTable < ActiveRecord::Migration[5.1]
+  def up
+    # Check conditions
+
+    create_table :locations do |t|
+      t.string name, default: "", null: false, comment: "Full display name, like a complete address"
+      t.string level_name, default: "", null: false, limit: 20, comment: "Level of specificity (country, state, subdivision, or city)"
+      t.string country_name, default: "", null: false, limit: 100, comment: "The country of this location if available"
+      t.string state_name, default: "", null: false, limit: 100, comment: "The state of this location if available"
+      t.string subdivision_name, default: "", null: false, limit: 100, comment: "Simplification for second-level administrative division (e.g. county/district/division/province/etc.) of this location if available"
+      t.string city_name, default: "", null: false, limit: 100, comment: "The city of this location (or closest administrative equivalent) if available"
+
+      # Recommended settings for lat/lon
+      t.decimal lat, precision: 10, scale: 6, comment: "The latitude of this location if available"
+      t.decimal lon, precision: 10, scale: 6, comment: "The longitude of this location if available"
+
+      t.timestamps
+    end
+
+    # Add indexes
+    add_index :locations, :name
+    add_index :locations, :level_name
+    add_index :locations, :country_name
+    add_index :locations, :state_name
+    add_index :locations, :subdivision_name
+    add_index :locations, :city_name
+  end
+
+  def down
+  end
+end
--- a/db/migrate/20190422185437_add_location_table.rb
+++ b/db/migrate/20190422185437_add_location_table.rb
@@ -11,7 +11,7 @@ class AddLocationTable < ActiveRecord::Migration[5.1]
       t.integer :osm_id, comment: "OpenStreetMap ID for traceability. May change at any time."
       t.integer :locationiq_id, comment: "Data provider API ID for traceability."
 
-      # Recommended settings for lat/lon
+      # Recommended settings for lat/lng
       t.decimal :lat, precision: 10, scale: 6, comment: "The latitude of this location if available"
       t.decimal :lng, precision: 10, scale: 6, comment: "The longitude of this location if available"
 
--- a/db/migrate/20190422185437_add_location_table.rb
+++ b/db/migrate/20190422185437_add_location_table.rb
@@ -2,15 +2,18 @@ class AddLocationTable < ActiveRecord::Migration[5.1]
   def up
     create_table :locations, force: true do |t|
       t.string :name, default: "", null: false, comment: "Full display name, such as a complete address"
-      t.string :level_name, default: "", null: false, limit: 20, comment: "Level of specificity (country, state, subdivision, or city)"
+      t.string :geo_level, default: "", null: false, limit: 20, comment: "Level of specificity (country, state, subdivision, or city)"
       t.string :country_name, default: "", null: false, limit: 100, comment: "Country (or equivalent) of this location if available"
+      t.string :country_code, default: "", null: false, limit: 5, comment: "ISO 3166 alpha-2 country codes. Can be used to resolve country_name if data sources ever change."
       t.string :state_name, default: "", null: false, limit: 100, comment: "State (or equivalent) of this location if available"
       t.string :subdivision_name, default: "", null: false, limit: 100, comment: "Second-level administrative division (e.g. county/district/division/province/etc.) of this location if available"
       t.string :city_name, default: "", null: false, limit: 100, comment: "City (or equivalent) of this location if available"
+      t.integer :osm_id, comment: "OpenStreetMap ID for traceability. May change at any time."
+      t.integer :locationiq_id, comment: "Data provider API ID for traceability."
 
       # Recommended settings for lat/lon
       t.decimal :lat, precision: 10, scale: 6, comment: "The latitude of this location if available"
-      t.decimal :lon, precision: 10, scale: 6, comment: "The longitude of this location if available"
+      t.decimal :lng, precision: 10, scale: 6, comment: "The longitude of this location if available"
 
       t.timestamps
     end
--- a/db/migrate/20190422185437_add_location_table.rb
+++ b/db/migrate/20190422185437_add_location_table.rb
@@ -20,7 +20,7 @@ class AddLocationTable < ActiveRecord::Migration[5.1]
 
     # Add indexes
     add_index :locations, :name, comment: "Index for lookup by location name"
-    add_index :locations, :level_name, comment: "Index for lookup by level of specificity"
+    add_index :locations, :geo_level, comment: "Index for lookup by level of specificity"
     add_index :locations, [:country_name, :state_name, :subdivision_name, :city_name], name: "index_locations_levels", comment: "Index for lookup within regions. Composite works for any left subset of columns."
   end
 
--- a/db/migrate/20190422185437_add_location_table.rb
+++ b/db/migrate/20190422185437_add_location_table.rb
@@ -1,31 +1,27 @@
 class AddLocationTable < ActiveRecord::Migration[5.1]
   def up
-    # Check conditions
-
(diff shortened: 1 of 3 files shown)


db/migrate/20190422185437_add_location_table.rb
```diff
@@ -0,0 +1,27 @@
...
+    create_table :locations, force: true do |t|
+      t.string :name, default: "", null: false, comment: "Full display name, such as a complete address"
+      t.string :level_name, default: "", null: false, limit: 20, comment: "Level of specificity (country, state, subdivision, or city)"
+      t.string :country_name, default: "", null: false, limit: 100, comment: "Country (or equivalent) of this location if available"
+      t.string :state_name, default: "", null: false, limit: 100, comment: "State (or equivalent) of this location if available"
+      t.string :subdivision_name, default: "", null: false, limit: 100, comment: "Second-level administrative division (e.g. county/district/division/province/etc.) of this location if available"
+      t.string :city_name, default: "", null: false, limit: 100, comment: "City (or equivalent) of this location if available"
+
+      # Recommended settings for lat/lon
+      t.decimal :lat, precision: 10, scale: 6, comment: "The latitude of this location if available"
```
MarkAZhang: I learned from https://gis.stackexchange.com/questions/8650/measuring-accuracy-of-latitude-and-longitude that 6 decimal places is accuracy within 11.1cm. So I guess this is sufficient! :)

db/migrate/20190422185437_add_location_table.rb
```diff
@@ -0,0 +1,27 @@
+class AddLocationTable < ActiveRecord::Migration[5.1]
+  def up
+    create_table :locations, force: true do |t|
+      t.string :name, default: "", null: false, comment: "Full display name, such as a complete address"
+      t.string :level_name, default: "", null: false, limit: 20, comment: "Level of specificity (country, state, subdivision, or city)"
```
MarkAZhang: Maybe call this `specificity`? Like the metadata field we never used. :(

db/migrate/20190422185437_add_location_table.rb
```diff
@@ -0,0 +1,27 @@
...
+      t.string :country_name, default: "", null: false, limit: 100, comment: "Country (or equivalent) of this location if available"
```
MarkAZhang: Do we need all the `_name` suffixes?

MarkAZhang:
I agree with storing country_name in each location, instead of having a country_id pointer. Less complicated queries.

Regarding "How will you query between ambiguous names", it is a little sad. There totally might be multiple subdivisions in Washington named Washington County. I think we can rely on the lat/long to be unique. Maybe assume the lat/long is unique but the taxonomic lineage might not be?

Also, would it make sense to store a pointer in location back to sample_id? Or maybe have a many-to-one relationship between samples and locations? I'm not sure if that will make the querying easier, but it might be nice to have a direct way of knowing which samples have a given location.

cdebourcy:
For what purpose will this table be queried and what are the queries going to look like? In which cases is a new row going to be inserted into this table?

db/migrate/20190422185437_add_location_table.rb
gregdingle: I think `level_name` seems good to me, or `geo_level`. Some other candidates are `granularity` and `scale`. See https://english.stackexchange.com/questions/94398/whats-the-best-term-for-geographical-level  .

db/migrate/20190422185437_add_location_table.rb
gregdingle: I think it's wise if we want to store `country_code` later... which begs the question of exactly which standardized format do we store in `country_name`?

db/migrate/20190422185437_add_location_table.rb
gregdingle: Do we have any desire to use mysql geo functions? https://dev.mysql.com/doc/refman/5.7/en/spatial-analysis-functions.html . Or use a gem like https://github.com/geokit/geokit-rails  ? If yes, we should optimize our schema for their designs.

db/migrate/20190422185437_add_location_table.rb
```diff
@@ -0,0 +1,27 @@
...
+      t.string :state_name, default: "", null: false, limit: 100, comment: "State (or equivalent) of this location if available"
+      t.string :subdivision_name, default: "", null: false, limit: 100, comment: "Second-level administrative division (e.g. county/district/division/province/etc.) of this location if available"
+      t.string :city_name, default: "", null: false, limit: 100, comment: "City (or equivalent) of this location if available"
+
+      # Recommended settings for lat/lon
+      t.decimal :lat, precision: 10, scale: 6, comment: "The latitude of this location if available"
+      t.decimal :lon, precision: 10, scale: 6, comment: "The longitude of this location if available"
```
gregdingle: I'm pretty sure the convention for longitude is `lng`  or `long` but not `lon`.

gregdingle:
> Also depending on the geosearch API, we may get unique IDs from the API, but I would rather not couple ourselves to a particular geosearch provider in that way. 

If we are populate this table directly from some API in a one-to-one way, it would be nice to the have the external ID for traceability. I would just name the column `apiname_id`.

db/migrate/20190422185437_add_location_table.rb
jshoe: @gregdingle I think we'd like to leave the door open to geo functions in MySQL or Elasticsearch so I'll take a look at those schemas too. Our coordinate operations would probably look like "is this point in the polygon defined by these points" (with polygons in Elasticsearch or MySQL JSON if we upgrade to 5.7) but I'm not planning coordinate arithmetic beyond that yet

db/migrate/20190422185437_add_location_table.rb
jshoe: 👍 ok I think I'll keep level_name here

db/migrate/20190422185437_add_location_table.rb
jshoe: This is a bit of a cop-out answer, but I think all the tools we use are going to be based on OpenStreetMap data potentially mixed with other sources, so the fields will align with what's in the name:en field and https://wiki.openstreetmap.org/wiki/Nominatim/Country_Codes Maybe we can store ISO 3166-1 country_code to help disambiguate?

If we need to, it could be viable to have a curated list of countries we store ourselves, but it would probably get unwieldy for below country level.

jshoe:
@MarkAZhang Actually I was going to hope that the lineage is unique, but there could be an additional constraint on the lat/lon (although that might be too specific if the data sources change). There are indeed no Washington Counties in the same state https://en.wikipedia.org/wiki/Washington_County

For querying I think yes we should have a many-to-one relationship for samples to locations so I think I will make a location_samples table (either here or later).

jshoe:
> For what purpose will this table be queried and what are the queries going to look like? In which cases is a new row going to be inserted into this table?

Discussed offline with @cdebourcy . This table is for storing all our locations and we plan to have one entry per unique location. So "Uganda" would look like country_name=Uganda and the fields below it blank. Queries could look like "show me samples where the state is California" or "show me all the Location entries for this group of samples". Insertions will happen whenever someone selects an entry from the geosearcher results that we don't already have stored (based on matching each level).

db/migrate/20190422185437_add_location_table.rb
jshoe: I found some examples for both (OSM and LocationIQ use `lon` https://wiki.openstreetmap.org/wiki/Nominatim https://locationiq.com/docs). But Google/Mapbox/Leaflet use `lng` so I will switch and change everything to `lng`.

db/migrate/20190422185437_add_location_table.rb
jshoe: Ok Mark convinced me to change this so that it doesn't imply `level`_name is parallel to `country/state/subdivision`_name. I will change to geo_level

jshoe:
^ going to add/rename a few columns based on these comments

jshoe:
> If we are populate this table directly from some API in a one-to-one way, it would be nice to the have the external ID for traceability. I would just name the column `apiname_id`.

That sounds good to me. I will add it

db/migrate/20190422185437_add_location_table.rb
gregdingle: Storing ISO country codes in addition to full names seems wise.

jshoe: added :)

PR #2176, opened 2019-04-23
More robust pipeline version checking.

MarkAZhang:
When we get to 3.10, we will run into issues when using the current comparison method.

Tested that features such as the list of report files and the coverage viz sidebar show up as expected. Also manually tested the pipeline version function in Ruby and JS.

### DIFF
--- a/app/assets/src/components/utils/sample.js
+++ b/app/assets/src/components/utils/sample.js
@@ -1,20 +1,37 @@
 import { last } from "lodash/fp";
 
-export const pipelineVersionHasAssembly = pipelineVersion => {
+export const pipelineVersionAtLeast = (pipelineVersion, testVersion) => {
   if (!pipelineVersion) return false;
-  const versionNums = pipelineVersion.split(".");
-  return (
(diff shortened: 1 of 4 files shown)


app/assets/src/components/utils/sample.js
```diff
@@ -1,21 +1,38 @@
...
+  const pipelineNums = pipelineVersion.split(".");
+  const testNums = testVersion.split(".");
+
+  if (toInt(pipelineNums[0]) > toInt(testNums[0])) {
+    return true;
+  } else if (toInt(pipelineNums[0]) === toInt(testNums[0])) {
+    if (toInt(pipelineNums[1]) > toInt(testNums[1])) {
+      return true;
+    } else if (toInt(pipelineNums[1]) === toInt(testNums[1])) {
+      if (toInt(pipelineNums[2]) >= toInt(testNums[2])) {
```
jshoe: for pipelineNums[2] I might've seen some version strings passed along without the third 'digit'.. something to look out for

PR #2177, opened 2019-04-23
Add ContigDownloader to Coverage Viz.

MarkAZhang:
Add functions to fetch byteranges of the contigs.fasta file.

Refactor the HitGroupViz into a separate component.
Add new Copy Icon.
Fix issue where tooltip was overflowing the window on the right,
and becoming narrower as it did so.

Make tooltips and contig downloader position:fixed instead of
position:absolute, so they'll stay in the correct position if
the user scrolls the page.

Verified:
* For single contigs and aggregated contigs, download file and copy to clipboard work as expected.
* For hit groups that are only reads, clicking on the hit group does nothing.
* Tooltips and contig downloader are correctly positioned after scrolling.

Screen Shot 2019-04-23 at 1 31 37 PM

Tooltip message changes when user clicks button. Changes back after user mouses out.
Screen Shot 2019-04-23 at 1 31 50 PM

Tooltip shifts to the right if the mouse is too close to right-side of the screen.
Screen Shot 2019-04-23 at 1 32 08 PM


app/assets/src/components/visualizations/GenomeViz.js
```diff
@@ -71,23 +71,18 @@ export default class GenomeViz {
...
 
-  // Find the bar that is closest to hoverX, within hoverBuffer.
-  onMouseMove = () => {
-    if (this.sortedEndpoints.length == 0 || !this.endpointToDataIndex) {
-      return;
-    }
-
-    const hoverX = mouse(this.svg.node())[0];
+  // Find the bar that is closest to mouseX, within hoverBuffer.
+  getDataIndexForMouseX = mouseX => {
```
MarkAZhang: A lot of the changes here are factoring out `getDataIndexForMouseX` into a separate function so it can be used in both `onMouseMove` and `onMouseClick`.

app/assets/src/components/common/CoverageVizBottomSidebar/HitGroupViz.jsx
```diff
@@ -0,0 +1,293 @@
...
+          contigDownloaderData &&
+          this.renderContigDownloader()}
+      </div>
+    );
+  }
+}
+
+HitGroupViz.propTypes = {
+  accessionData: PropTypes.shape({
+    avg_prop_mismatch: PropTypes.number,
```
jshoe: why were these ones snake_case again? because they come from the Python files?

app/assets/src/components/ui/icons/CopyIcon.jsx
```diff
@@ -0,0 +1,52 @@
...
+    <svg
+      className={props.className}
+      viewBox="0 0 15 17"
+      version="1.1"
+      xmlns="http://www.w3.org/2000/svg"
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+    >
+      <g id="Coverage-map" stroke="none" strokeWidth="1" fillRule="evenodd">
+        <g
+          id="2.6-Sample-Report-Copy-9"
```
jshoe: nit: could remove the 2.6-Sample-Report-Copy-9 and noun_copy_1801473_000000. looks a little odd

app/assets/src/components/visualizations/GenomeViz.js
jshoe: is there any difference between clientX and mouseX?

app/assets/src/components/common/CoverageVizBottomSidebar/HitGroupViz.jsx
MarkAZhang: They come directly from a JSON file where it's snake_case. :(

app/assets/src/components/visualizations/GenomeViz.js
MarkAZhang: mouseX is actually a point with respect to the svg container.
clientX is with respect to the page viewport.

I changed "mouseX" to "svgX" and added a comment so it's clearer.

PR #2181, opened 2019-04-24
Fix bug that crashes old Sample Upload page.

MarkAZhang:
If the user selects a file before they input a sample name in the old sample upload flow, the page would crash.

Verified that after this fix, the page no longer crashes.


gregdingle:
My bad. Thanks for the fix.

PR #2186, opened 2019-04-24
IDSEQ-913 Add error message when min contig size is too small.

MarkAZhang:
This reduces confusion for the user. Right now, when the user types an invalid value, the value resets to the old one, but there is no error message to explain what is going on.

Screen Shot 2019-04-24 at 3 15 16 PM


MarkAZhang:
idseq_web doesn't load contigs less than 4 from the pipeline. (See `db_load_contigs`). I believe this was a choice by YF because due to paired reads, a contig of just 2 reads could just be the two paired reads assembled together. So a contig with less than 4 reads is very unreliable.

PR #2187, opened 2019-04-24
Add chicken genome

tfrcarvalho:
Indices generated as per [instructions](https://github.com/chanzuckerberg/idseq-web/wiki/%5BDev%5D-How-to-update-host-genomes).


db/migrate/20190423182934_add_chicken_genome.rb
```diff
@@ -0,0 +1,20 @@
+class AddChickenGenome < ActiveRecord::Migration[5.1]
+  def up
```
cdebourcy: Recently added an additional line to the migration template! See https://github.com/chanzuckerberg/idseq-web/blob/0efb9d50d08cd6f85ea9d1024c8b420bcb799565/db/migrate/20190417183014_add_carp_host.rb#L3

cdebourcy: Mostly for the case where we might want to create the host genome on staging / prod manually before we actually deploy web... then we wouldn't want the deploy to cause a duplicate host genome

tfrcarvalho: I see, that edge case makes the migration non-symmetric though...

cdebourcy: true

PR #2188, opened 2019-04-25
Fix error in logging

gregdingle:
# Description

Fixes 
image (6)

To be clear, the continue was still working, only the logging was broken.

# Test
Upload a sample
Click continue
See no more error


gregdingle:
> Is this on prod? If so, we should push asap.

This is in prod but only the logging is broken so no big deal.

PR #2193, opened 2019-04-25
 Show Notification toast after Metadata Upload.

MarkAZhang:
2 commits.
The first commit contains code changes.
The second commit just rearranges files.
To review code, just look at the first commit.
    
Commit 1:
Add new ListNotification component for notifications that contain a list of items.
Refactor Accordion so it can be used in more places. This was mainly removing opinionated styling.
    
Add Success state to NotificationComponent.
    
Refactor PublicSampleNotification to use ListNotification.

Screen Shot 2019-04-25 at 2 42 08 PM

Screen Shot 2019-04-25 at 2 41 49 PM

Commit 2:
Rearrange file structure of notifications and toasts.
    
Notifications are now grouped together in /ui/notifications.
They are components that show info, warnings, errors, and success states.
    
Toasts describe anything that appears at the top-right and can be dismissed.
Often you will create toasts containing a notification.
Notifications can also be used outside of toasts, for example to display errors in a form.

Verified that:
* Existing notifications, including the DD CollectionModal and PublicSampleNotification, work as before.
* The metadata upload modal IssueGroup (which got moved) works as before.


app/assets/src/components/common/Header.jsx
```diff
@@ -4,7 +4,7 @@ import { forbidExtraProps } from "airbnb-prop-types";
 import cx from "classnames";
 
 import { RequestContext } from "~/components/common/RequestContext";
-import ToastContainer from "~ui/containers/ToastContainer";
+import ToastContainer from "~/components/common/ToastContainer";
```
tfrcarvalho: Why the move? It is a UI element...

MarkAZhang: `<ToastContainer>` can only be rendered once in the application. I felt like it belonged with `<Header>`, which is also in `~/components/common`. It felt confusing to put it in `/ui`, because it's not re-usable.

tfrcarvalho: I think of it more like what would happen if we wanted to package the IDseq UI. In that case, `ToastContainer` would need to be included... that it is only instantiated once, is just a consequence of the type of element.

PR #2197, opened 2019-04-26
[Mapping] Round LatLng geosearch responses

jshoe:
- Implements a recommendation from legal to round geosearch responses to 2 decimal places to enhance privacy and non-specificity. Also changes the responses to float.
- If you're curious, that could be like rounding to an area the size of SoMA. 1 decimal place would get very broad and be entire SF.

Screen Shot 2019-04-26 at 11 51 24 AM

### Test and Reproduce
- Load the code, go to http://localhost:3000/locations/external_search?query=UCSF or similar and see results


app/controllers/locations_controller.rb
```diff
@@ -17,8 +17,10 @@ def external_search
             state: c["address"]["state"] || "",
             county: c["address"]["county"] || "",
             city: c["address"]["city"] || "",
-            lat: c["lat"] || "",
-            lng: c["lng"] || ""
+            # Round coordinates to enhance privacy
+            lat: c["lat"] ? c["lat"].to_f.round(2) : nil,
+            # LocationIQ uses 'lon'
```
gregdingle: interesting... I'm not opposed to using `lon` if that is more standard

jshoe: I think you're right that `lng` is more standard actually. based on Google searching of LatLon vs. LatLng.

PR #2198, opened 2019-04-26
[Data Discovery] Make "My Data" the home page for users that have data discovery enabled.

tfrcarvalho:
* `My Data` is now the home page for users that have the `data_discovery` feature enabled (remaining users will see the current home page)
* Create `legacy` endpoint (accessible by typing in URL) that always serves the previous home page


app/controllers/home_controller.rb
```diff
@@ -29,13 +29,21 @@ def landing
   end
 
   def index
+    if current_user.allowed_feature_list.include?("data_discovery")
+      render 'my_data'
+    else
+      legacy
+      render 'legacy'
```
MarkAZhang: This feels a little weird, since the call to `legacy` is just setting up variables here and you have to call `render` afterward.
But if you call the `legacy` function directly using in an http request,  it will automatically render `legacy`.
So the function behaves differently based on where it is being called because of Rails magic.

Could you maybe move `render 'legacy`' into the `legacy` function? I think it's less confusing.

tfrcarvalho: Yeap, makes sense. Not sure why i did not do that initially.

PR #2203, opened 2019-04-30
 Address subtle edge case with metadata upload.

MarkAZhang:
Edge case:
Our official "Blood Fed" metadata field is only for Mosquitoes. If you try to upload a value for "Blood Fed" to a human sample, we currently create a new Blood Fed custom field that only accepts Human Samples. This is quite confusing, as we now have two metadata fields with the same name for the project (the official one for mosquitoes, and the custom one for humans). It also defeats the point of specifying a schema for each host genome.

With this fix, if a user tries to upload Blood Fed for humans, and Blood Fed is an official metadata field, an error is thrown. The user can either rename their column (which will create a custom field), or message us to let us know that our schema should be more permissive. It was probably an accident that the user uploaded a value for Blood Fed for a human sample...so this will help them catch errors.

Screen Shot 2019-04-29 at 6 15 08 PM

### DIFF
--- a/app/helpers/metadata_helper.rb
+++ b/app/helpers/metadata_helper.rb
@@ -4,7 +4,7 @@ module MetadataHelper
   include ErrorHelper
 
   def get_available_matching_field(sample, name)
-    available_fields = (sample.project.metadata_fields & sample.host_genome.metadata_fields)
+    available_fields = sample.project.metadata_fields
     return available_fields.find { |field| field.name == name || field.display_name == name }
   end
 
@@ -153,17 +153,12 @@ module MetadataHelper
 
     processed_samples = []
 
-    # Make a best effort to guess which custom fields will be created.
-    # This validation actually misses one non-critical edge case:
-    # If a user uploads "Blood Fed" for a Human Sample (our default Blood Fed field is only for Mosquito),
-    # our system will create a custom field for Blood Fed for that project (assigned to Human).
-    # That custom field won't be listed here, since we don't restrict the fields to host genomes here.
-    # TODO(mark): Detect this edge case and output a warning.
     if samples[0].project
       matching_fields =
         (samples[0].project.metadata_fields | MetadataField.where(is_core: 1))
         .select { |field| metadata["headers"].include?(field.name) || metadata["headers"].include?(field.display_name) }
 
+      # Add a warning for each custom field that will be created.
       metadata["headers"].each_with_index do |col, index|
         next if ["sample_name", "Sample Name", "host_genome", "Host Genome"].include?(col)
 
@@ -253,7 +248,7 @@ module MetadataHelper
           error_aggregator.add_error(:sample_not_found, [index + 1, sample.name])
         end
 
-        if val_errors[:invalid_key_for_host_genome].present?
+        if val_errors[:invalid_field_for_host_genome].present?
           error_aggregator.add_error(:invalid_key_for_host_genome, [index + 1, sample.name, sample.host_genome_name, field])
         end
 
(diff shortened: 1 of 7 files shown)


app/helpers/metadata_helper.rb
```diff
@@ -253,7 +248,7 @@ def validate_metadata_csv_for_samples(samples, metadata, new_samples = false)
           error_aggregator.add_error(:sample_not_found, [index + 1, sample.name])
         end
 
-        if val_errors[:invalid_key_for_host_genome].present?
+        if val_errors[:invalid_host_genome].present?
```
gregdingle: was this a bug being fixed? and it seems like `invalid_key_for_host_genome` is a better name, so maybe rename where `val_errors` is updated instead.

test/controllers/samples_bulk_upload_test.rb
```diff
@@ -334,6 +334,57 @@ class SamplesBulkUploadTest < ActionDispatch::IntegrationTest
     assert_equal 0, Sample.where(name: "RR004_water_2_S23A").length
   end
 
+  # If the metadata field isn't supported by the sample's host genome, throw an error.
+  test 'bulk upload invalid key for  host genome' do
+    post user_session_path, params: @user_params
+    post user_session_path, params: @user_params
```
gregdingle: why is this line twice? copy paste error?

test/controllers/samples_bulk_upload_test.rb
```diff
@@ -334,6 +334,57 @@ class SamplesBulkUploadTest < ActionDispatch::IntegrationTest
...
+
+    assert_response :success
+    assert_equal 1, @response.parsed_body["errors"].length
+    assert_equal MetadataUploadErrors.save_error('blood_fed', 'Yes'), @response.parsed_body["errors"][0]
+
+    # The sample should still upload, minus the invalid metadata.
+    assert_equal 1, Sample.where(name: "RR004_water_2_S23A").length
+    sample_id = Sample.where(name: "RR004_water_2_S23A").first.id
+    # The invalid metadata should not be uploaded.
+    assert_equal 5, Metadatum.where(sample_id: sample_id).length
```
gregdingle: helpful comments 😄

app/helpers/metadata_helper.rb
```diff
@@ -4,7 +4,7 @@ module MetadataHelper
   include ErrorHelper
 
   def get_available_matching_field(sample, name)
-    available_fields = (sample.project.metadata_fields & sample.host_genome.metadata_fields)
+    available_fields = sample.project.metadata_fields
```
gregdingle: so are the host genome fields already included in the project fields?

app/helpers/metadata_helper.rb
MarkAZhang: Good call.

test/controllers/samples_bulk_upload_test.rb
MarkAZhang: Good catch!

app/helpers/metadata_helper.rb
MarkAZhang: We are making the available fields less restrictive. Before, it was fields that match both the project and host_genome. Now, it's just fields that match the project.

This means that fields on the project but not compatible with the host genome will now be returned, which is part of the intention of this PR. We want the upload process to attempt to use metadata fields that are incompatible with the host genome and fail, instead of ignoring these metadata fields and creating a confusing custom field.

test/controllers/metadata_validate_new_samples_test.rb
```diff
@@ -272,6 +272,41 @@ class MetadataValudateNewSamplesTest < ActionDispatch::IntegrationTest
     assert_equal 0, @response.parsed_body['issues']['warnings'].length
   end
 
+  test 'invalid metadata fields for host genome' do
```
cdebourcy: Yay tests!

PR #2205, opened 2019-04-30
Fix user name regex

gregdingle:
# Description

See https://jira.czi.team/browse/IDSEQ-925 . I decided to increase the char class to include accent chars. 

# Test and Reproduce

`rails t test/models/user_test.rb`

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -24,7 +24,8 @@ class User < ApplicationRecord
   has_many :phylo_trees
 
   validates :email, presence: true
-  validates :name, presence: true, format: { with: /\A[a-zA-Z -]+\z/, message: "only allows letters" }
+  # TODO: (gdingle): write unit test
+  validates :name, presence: true, format: { with: /\A(\w|[ -])+\z/, message: "Name must be letters or spaces" }
   attr_accessor :email_arguments
   ROLE_ADMIN = 1
   DEMO_USER_EMAILS = ['idseq.guest@chanzuckerberg.com'].freeze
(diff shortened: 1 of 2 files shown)


test/models/user_test.rb
```diff
@@ -1,7 +1,45 @@
 require 'test_helper'
 
 class UserTest < ActiveSupport::TestCase
-  # test "the truth" do
-  #   assert true
-  # end
+  test "user name validates" do
```
cdebourcy: Yay tests!

app/models/user.rb
```diff
@@ -24,7 +24,10 @@ class User < ApplicationRecord
   has_many :phylo_trees
 
   validates :email, presence: true
-  validates :name, presence: true, format: { with: /\A[a-zA-Z -]+\z/, message: "only allows letters" }
+  validates :name, presence: true, format: {
+    # See https://www.ascii-code.com/
+    with: /\A[- 'a-zA-ZÀ-ÖØ-öø-ÿ]+\z/, message: "Name must contain only letters, apostrophes, dashes or spaces"
```
MarkAZhang: What was the rationale for picking these characters in particular? (are they the most popular?) I notice there are some additional Latin characters in www.ascii-code.com, such as 131 "Latin small letter f with hook", but I guess they aren't compatible with UTF-8. Figuring out a good regex for international names seems like a rabbit hole, but it would be good to add a comment explaining why you choose these characters.

gregdingle: Good question. I added this comment

```    # See https://www.ascii-code.com/. These were the ranges that captured the
    # common accented chars I knew from experience, leaving out pure symbols.
```

test/models/user_test.rb
gregdingle: 🕺

PR #2208, opened 2019-05-01
Show warning on low number of reads

gregdingle:
# Description

Following a high percent of failed runs (18/96) in Vida's project, I thought we should show a warning when the number of reads makes meaningful results very unlikely. 

In discussion with @cdebourcy , we decided on initial threshold of 1000 total reads, then I added a threshold 100 passed filter reads. Happy to change either. 

image

Note: I artificially increased the thresholds here to see more warnings. 

@happyimadesignr please review the visual look

# Test and Reproduce

1. Load a project with low read samples
2. See warning

### DIFF
--- a/app/assets/src/components/views/SampleView/SampleView.jsx
+++ b/app/assets/src/components/views/SampleView/SampleView.jsx
@@ -327,6 +327,9 @@ class SampleView extends React.Component {
             <h6 className={cs.failed}>
               {this.props.pipelineRun.error_message}
             </h6>
+            <h6 className={cs.failed}>
+              {this.props.pipelineRun.total_reads} Total Reads
+            </h6>
           </div>
         );
       }
--- a/app/assets/src/components/views/SampleView/SampleView.jsx
+++ b/app/assets/src/components/views/SampleView/SampleView.jsx
@@ -330,6 +330,12 @@ class SampleView extends React.Component {
             <h6 className={cs.failed}>
               {this.props.pipelineRun.total_reads} Total Reads
             </h6>
+            {this.props.pipelineRun.adjusted_remaining_reads && (
+              <h6 className={cs.failed}>
+                {this.props.pipelineRun.adjusted_remaining_reads} Reads Passed
+                Filters
+              </h6>
+            )}
           </div>
         );
       }
--- a/app/assets/src/components/views/discovery/discovery_api.js
+++ b/app/assets/src/components/views/discovery/discovery_api.js
@@ -76,7 +76,11 @@ const processRawSample = sample => {
       status: get(
         "run_info.result_status_description",
         sample.details
-      ).toLowerCase()
+      ).toLowerCase(),
+      lowReadsWarning: get(
+        "derived_sample_output.summary_stats.low_reads_warning",
+        sample.details
+      )
     },
     collectionLocation: get("metadata.collection_location", sample.details),
     createdAt: sample.created_at,
--- a/app/assets/src/components/views/samples/SamplesView.jsx
+++ b/app/assets/src/components/views/samples/SamplesView.jsx
@@ -224,6 +224,9 @@ class SamplesView extends React.Component {
               <div className={cx(cs.sampleStatus, cs[sample.status])}>
                 {sample.status}
               </div>
+              <div className={cx(cs.sampleStatus, cs[sample.status])}>
+                {sample.status}
+              </div>
             </div>
           ) : (
             <div className={cs.sampleNameAndStatus} />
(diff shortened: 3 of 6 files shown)


MarkAZhang:
In addition (doesn't have to be in this PR), it would be great to show these warnings in a box at the top the report page, even if it succeeds. Perhaps using the `<Notification>` components we've been developing. Would be great to get some design help on that.

MarkAZhang:
I think the error message should said a little more than "X total reads. X reads passed filters". There should be an explanation of what happened and what action the user can take. "passed filters" is also potentially confusing. (what does "filters" mean?)

Something like:
"The pipeline failed because there were too few high-quality non-host reads in your sample input files. 5 reads passed host and quality filtering. At least 10 reads are required." (and maybe a button to re-submit their input files? Not sure if that makes sense here.)

------

Also maybe we can change the styling of the error message to match our other errors?

Ex:
Screen Shot 2019-04-30 at 5 46 50 PM

You should be able to use the `<Notification>` component.

gregdingle:
@MarkAZhang 
> In addition (doesn't have to be in this PR), it would be great to show these warnings in a box at > the top the report page, even if it succeeds. Perhaps using the <Notification> components we've been developing. Would be great to get some design help on that.

That's a good idea. While we're at it, I think we should also redesign the failure message that appears on a sample page when the run failed. 

> I think the error message should said a little more than "X total reads. X reads passed filters". > There should be an explanation of what happened and what action the user can take. "passed filters" is also potentially confusing. (what does "filters" mean?)

I agree, this PR is the most minimal improvement. I like your suggesed message, but I think we should review it with more knowledgeable people. 

By the way, how many reads are required for the contig visualization to make sense?

happyimadesignr:
Hi, I chatted with Greg about this a bit. We are going to collaborate on this.

**There are some considerations:**
- Sample column is already running out of space even for the “Complete” Tag
- We have 2 columns called for reads that “Passed Filters” and “Passed QC” so if we were to indicate something like low reads, so if we have it, it would belong closer to those.
- For the no read left after filtering, I’m working on a design to give a message on the sample report, so we should figure out how to be consistent.
- I want to understand the problem better because I wasn’t sure what prompted the need for this since we do have the 2 columns and whether or not Vida was expressing she couldn’t gather that there were low reads from the columns we already had. I do see that the tag highlights it more visually though. Greg and I are going to chat with Vida a bit b/c it’s not clear at the moment.
- With all of these notifications, tags, error messages, it’s a good time to think about these components at a higher level and define how things work consistently. And I’m doing some of this work with the ticket I was working on for all the errors you might get on the Sample Report page.

**Errors** that I was chatting with Charles about yesterday, where we can also give users more indication of why things happened and tell them what the next step is.
* **No reads after filtering** (ex: at the quality filter step, check your quality.)
* **Failed** (ex: at alignment stage, email us to rerun)
    * They could look at intermediate files that have produced.
* **We detected their input was faulty**, if they didn’t have the right format.
    * There was a problem with your files because
        * Read length is too long
        * number of paired reads don’t match
        * Format is incorrect
        * input is truncated
        * And more
        * CTA: check the format and reupload.

gregdingle:
Good comments @happyimadesignr 

> Sample column is already running out of space even for the “Complete” Tag

Good to know. That definitely affects this PR. 

> We have 2 columns called for reads that “Passed Filters” and “Passed QC” so if we were to indicate something like low reads, so if we have it, it would belong closer to those.

What is the difference between these two columns? I actually don't know. 

Also, the low-reads warning in this PR shows up when the *total reads* are below 1000, and total reads is not shown as a column.

> Errors that I was chatting with Charles about yesterday, where we can also give users more indication of why things happened and tell them what the next step is.

I would love to know this myself for oncall duties!

gregdingle:
Here are some interesting stats for designing thresholds for low reads, based on RR025 Pool2 iseq
 project. 

```
17 failures out of 96 samples

8 failed <1000 total reads
4 succeeded <1000 total reads

9 failed <10 passed filter reads
47 succeeded <10 passed filter reads

9 failed <4 passed filter reads
14 succeeded <4 passed filter reads
```

happyimadesignr:
@gregdingle 

Here the description of “Passed Filters” and “Passed QC” in the first set of bullet points in the link. I wasn't sure either, @katrinakalantar provided this link! https://github.com/chanzuckerberg/idseq-dag/wiki/IDseq-Pipeline-Step-%231:-Host-Filtering-and-QC

Thanks for pointing out that this is referring to total reads. I thought it was referring to reads that passed filters. When we were designing the table, I learned that people care more about the passed filters and the passed QC so Total Reads wasn't chosen to be shown in the default view, but it exists, you can customize the table to include it. Users said they can get a sense of Total Reads from the percentage on Passed Filters. 

The errors I was referring to are those 3 main bullet points listed below. We'll need to expand on the details a bit more though.

happyimadesignr:
Was the intent to warn people during the pipeline process of the low reads or after it was completed?

PR #2209, opened 2019-05-01
Fix edge case with new custom metadata fields.

MarkAZhang:
Previously, when someone uploads a custom metadata field for a sample,
the new field is only linked to that sample's host genome (as opposed to
all host genomes). This causes issues when someone wants to upload the same
custom field for samples with different host genomes. (you get a bunch of
identical custom fields created, each linked to a different host genome)

We fix this issue by automatically linking custom fields
to all host genomes by default.

Verification steps:
* Added new tests for this edge case.
* Verified that all tests pass.
* Verified in the app that you can upload a custom field for multiple samples with different host genomes, and the app behaves as expected. Verified this both in "upload metadata to existing samples" and "bulk upload new samples." 
* Also verified that updating individual metadata fields in the sidebar works as expected, since we modified `metadata_add_or_update`.

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -573,8 +573,43 @@ class Sample < ApplicationRecord
     return metadata.find { |metadatum| metadatum.metadata_field.name == key || metadatum.metadata_field.display_name == key }
   end
 
+  # Ensure that an appropriate metadata field exists for the given key.
+  # Add core fields to the sample's project if needed.
+  # Create custom fields if needed.
+  # Return a string describe the status of the metadata field.
+  def ensure_metadata_field_for_key(key)
+    metadata_field = get_existing_metadatum(key.to_s) || get_available_matching_field(self, key.to_s)
+
+    if metadata_field
+      # Return ok if the field already exists and no additional actions were needed.
+      return "ok"
+    end
+
+    metadata_field = get_matching_core_field(self, key.to_s)
+
+    # This is a core field that isn't currently part of the project.
+    # Add it to the project.
+    if metadata_field
+      project.metadata_fields.append(metadata_field)
+      return "core"
+    end
+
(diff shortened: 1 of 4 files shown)


app/models/sample.rb
```diff
@@ -573,40 +573,54 @@ def get_existing_metadatum(key)
...
+      return "core"
+    end
+
+    metadata_field = get_new_custom_field(key.to_s)
+    metadata_field.save
+
+    # Add the new custom field to the project
+    # and all host genomes.
+    project.metadata_fields.append(metadata_field)
+    HostGenome.all.each do |genome|
```
gregdingle: Maybe I'm missing something but it seems like the data model is messed up if we need to add each custom metadata field to *all* host genomes in idseq unconditionally. Why not just skip the existence checks then for custom fields? Or persist in a single global table of custom metadata fields?

gregdingle: Or why not append to only the genomes in the current batch?

MarkAZhang: Hmm you're right. It does seem a little weird.

The issue is, if I upload "Custom Field X" to a bunch of human samples, then later I upload "Custom Field X" to a bunch of mosquito samples, I don't want two different custom fields named "Custom Field X" to be created.

A different approach is: when uploading metadata for a sample, look for a matching metadata field in the same project. If one exists but its host genome doesn't match the sample's, then check if it's custom and if so, add the host genome to metadata field. So instead of adding the custom field to all host genomes at the beginning, we do it on-demand.

How does that approach sound?

gregdingle: That does sound better but could you explain more in person so I can be sure I understand well?

gregdingle: Talked with @MarkAZhang offline. All good now.

PR #2214, opened 2019-05-02
[Data Discovery] Fix link to project page

tfrcarvalho:
## Description

* Redirects previous project links to new DD project page
* The project page will load under `my_data`, `public` tabs depending on if the user owns the project or the project is public
* The user will be redirect to default `my_data` if the user has not permissions to see the project.
* backwards compatible with previous links

Note: a future change should replace this by the normal RESTful route `/projects/<project_id>`. To do that, we need to implement the admin features we have for the current project page.

## Test

* Add controller tests user access control
* Load `/home?project_id=<project_id>`, and check if the project opens on the correct tab.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -32,14 +32,15 @@ class HomeController < ApplicationController
     if current_user.allowed_feature_list.include?("data_discovery")
       project_id = params[:project_id]
       project = Project.find(params[:project_id])
-      if project
-        if current_user.owns_project?(project.id)
-          redirect_to action: "my_data", project_id: project_id
-        elsif project.public_access != 0
-          redirect_to action: "public", project_id: project_id
-        else current_user.admin?
-          redirect_to action: "all_data", project_id: project_id
-        end
+
+      redirect_to my_data_path if project.blank?
+
+      if current_user.owns_project?(project.id)
+        redirect_to action: "my_data", project_id: project_id
+      elsif project.public_access != 0
+        redirect_to action: "public", project_id: project_id
+      elsif current_user.admin?
+        redirect_to action: "all_data", project_id: project_id
       else
         redirect_to my_data_path
       end
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -30,7 +30,19 @@ class HomeController < ApplicationController
 
   def index
     if current_user.allowed_feature_list.include?("data_discovery")
-      redirect_to my_data_path
+      project_id = params[:project_id]
+      project = Project.find(params[:project_id])
+      if project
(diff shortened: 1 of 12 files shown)


test/controllers/home_controller_test.rb
```diff
@@ -1,7 +1,38 @@
...
+    @joe_project = projects(:joe_project)
+    get "/home?project_id=#{@joe_project.id}"
+    res = @response.body
+
+    assert_redirect_to(action: my_data, )
+    assert_response :success
+    puts res
   end
+
+  test 'joe redirect to my_data when seeing joe_project with data_discovery' do
```
gregdingle: why is the same test printed three times over?

tfrcarvalho: Test was still in the works... thats why I had `[WIP] Adding some extra tests` in the top of the description. :)

test/controllers/home_controller_test.rb
```diff
@@ -1,7 +1,68 @@
...
 class HomeControllerTest < ActionDispatch::IntegrationTest
-  test 'the truth' do
-    assert true
+  test 'joe redirected to project page in my_data when linking to joe_project with data_discovery' do
+    sign_in(:joe_dd)
+    @joe_project = projects(:joe_project)
+    get "/home?project_id=#{@joe_project.id}"
+
+    assert_response :redirect
+    assert_redirected_to action: "my_data", controller: "home", project_id: @joe_project.id
```
gregdingle: I'm not seeing HomeController#my_data method... what am I missing?

app/views/home/all_data.html.erb
```diff
@@ -3,6 +3,7 @@
   <%= javascript_tag do %>
     react_component('DiscoveryView', {
       domain: "all_data",
+      projectId: <%= params[:project_id] || "null" %>,
```
gregdingle: why quote `"null"` ? deserves a comment

app/controllers/home_controller.rb
```diff
@@ -30,7 +30,20 @@ def landing
...
   def index
     if current_user.allowed_feature_list.include?("data_discovery")
-      redirect_to my_data_path
+      project_id = params[:project_id]
+      project = Project.find(params[:project_id])
+
+      redirect_to my_data_path if project.blank?
+
+      if current_user.owns_project?(project.id)
+        redirect_to action: "my_data", project_id: project_id
```
gregdingle: I don't see `my_data` action as a method in `HomeController`... what am I missing about rails routing?

tfrcarvalho: The action does nothing it goes directly to view where it adds the React component. In that case you do not need to define it.

PR #2215, opened 2019-05-03
Speed up travis build

gregdingle:
# Description

See https://jira.czi.team/browse/IDSEQ-954 . This speeds up the build by using existing images as a layer cache, `from_cache`.

This should speed up build time by ~5min because in most PRs there are no changes to external dependencies and therefore no need to download and install gems and node modules. 

The docker cache should invalidate itself when desired because of the gem and npm lock files changing. See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache#leverage-build-cache . 

The largest remaining time is in webpack build, around 2min. @MarkAZhang is there an easy way you can see to trim this? Would it be safe to have a custom travis webpack config that does less? I don't see how we actually need webpack in travis except to test that it will work in prod. 

# Test

The changes are visible in the travis log of this PR.

The total time in this run decreased by from ~15min to ~7min.

```
Building web
Step 1/24 : FROM ruby:2.5
2.5: Pulling from library/ruby
Digest: sha256:063f661153e37b60c8345c63d76c02ba914f54e0df9f6de48e7111d49d1b2828
Status: Downloaded newer image for ruby:2.5
 ---> e86557c9a8ab
Step 2/24 : RUN apt-get update && apt-get install -y build-essential nodejs mysql-client python-dev python-pip apt-transport-https
 ---> Using cache
 ---> 2bed7b309dff
Step 3/24 : RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
 ---> Using cache
 ---> 2e352c320037
Step 4/24 : RUN apt-get update && apt-get install -y nodejs
 ---> Using cache
 ---> 586355434e30
Step 5/24 : RUN pip install --upgrade pip
 ---> Using cache
 ---> 145d0aabb98c
Step 6/24 : RUN pip install --upgrade setuptools
 ---> Using cache
 ---> 20945fd1ebd8
Step 7/24 : RUN pip install --upgrade pyOpenSSL
 ---> Using cache
 ---> 770c0e7edb02
Step 8/24 : RUN pip install --upgrade aegea
 ---> Using cache
 ---> e15ed5354d71
Step 9/24 : ADD https://github.com/segmentio/chamber/releases/download/v2.2.0/chamber-v2.2.0-linux-amd64 /bin/chamber
 ---> Using cache
 ---> 8114e91af4d1
Step 10/24 : RUN chmod +x /bin/chamber
 ---> Using cache
 ---> 060945592020
Step 11/24 : RUN mkdir -p /app
 ---> Using cache
 ---> 26dc389c889a
Step 12/24 : WORKDIR /app
 ---> Using cache
 ---> 2f5ca9f5f8f1
Step 13/24 : COPY Gemfile Gemfile.lock ./
 ---> Using cache
 ---> 58598c35439e
Step 14/24 : RUN gem install bundler && bundle install --jobs 20 --retry 5
 ---> Using cache
 ---> 771679b1a9e7
Step 15/24 : RUN npm update -g
 ---> Using cache
 ---> 6cdb19f97dbe
Step 16/24 : COPY package.json package-lock.json ./
 ---> Using cache
 ---> 812639531096
Step 17/24 : RUN npm install
 ---> Using cache
 ---> a0a75648ed5f
Step 18/24 : COPY . ./
 ---> e7c00c874e6a
Step 19/24 : RUN mkdir -p app/assets/dist && npm run build-img && ls -l app/assets/dist/
 ---> Running in 9433ee84431c
...
```

### DIFF
--- a/bin/build-docker
+++ b/bin/build-docker
@@ -7,6 +7,7 @@ BRANCH="$1"
 COMMIT="$2"
 BUILD_NUMBER="$3"
 COMPOSE_TAG="$REPO:compose"
+LATEST_TAG="$REPO:latest"
 
 echo $BRANCH
 echo $COMMIT
@@ -14,10 +15,14 @@ echo $REPO
 echo $TAG
 echo '--'
 
+# need to pull images for cache_from first
+docker pull $LATEST_TAG
+docker pull $COMPOSE_TAG
+
 docker-compose build --pull --parallel --build-arg GIT_COMMIT=${COMMIT} web
 
 if [ "$BRANCH" == "master" ]; then
-  docker tag $COMPOSE_TAG $REPO:"latest"
+  docker tag $COMPOSE_TAG $LATEST_TAG
 fi
 docker tag $COMPOSE_TAG $REPO:"sha-$COMMIT"
 BRANCH_CLEANED=`echo $BRANCH | sed 's/\//-/g'`
--- a/bin/build-docker
+++ b/bin/build-docker
@@ -16,8 +16,8 @@ echo $TAG
 echo '--'
 
 # need to pull images for cache_from first
-docker pull $LATEST_TAG
-docker pull $COMPOSE_TAG
+docker pull $LATEST_TAG || true
+docker pull $COMPOSE_TAG || true
 
 docker-compose build --pull --parallel --build-arg GIT_COMMIT=${COMMIT} web
 
--- a/bin/build-docker
+++ b/bin/build-docker
@@ -6,7 +6,7 @@ export REPO=chanzuckerberg/idseq-web
 BRANCH="$1"
 COMMIT="$2"
 BUILD_NUMBER="$3"
-COMPOSE_TAG="chanzuckerberg/idseq-web:compose"
+COMPOSE_TAG="$REPO:compose"
 
 echo $BRANCH
 echo $COMMIT
@@ -14,7 +14,7 @@ echo $REPO
 echo $TAG
 echo '--'
 
-docker-compose build --build-arg GIT_COMMIT=${COMMIT} web
+docker-compose build --pull --parallel --build-arg GIT_COMMIT=${COMMIT} web
 
 if [ "$BRANCH" == "master" ]; then
   docker tag $COMPOSE_TAG $REPO:"latest"
--- a/bin/build-docker
+++ b/bin/build-docker
@@ -19,7 +19,7 @@ echo '--'
 docker pull $LATEST_TAG || true
 docker pull $COMPOSE_TAG || true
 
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -101,7 +101,13 @@ services:
...
     command: bash
   web:
-    build: .
+    build:
+      context: .
+      # See https://testdriven.io/blog/faster-ci-builds-with-docker-cache/
+      # and https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache#leverage-build-cache
+      cache_from:
+        - chanzuckerberg/idseq-web:latest
+        - chanzuckerberg/idseq-web:compose
```
gregdingle: `compose` should be the image of the last travis run branch, which will provide even more cache hits over `latest`

tfrcarvalho:
Sounds good! Is it worth to run a quick test with some new gem/npm package to guarantee the invalidation works as expected?

bin/build-docker
```diff
@@ -6,18 +6,23 @@ export REPO=chanzuckerberg/idseq-web
 BRANCH="$1"
 COMMIT="$2"
 BUILD_NUMBER="$3"
-COMPOSE_TAG="chanzuckerberg/idseq-web:compose"
+COMPOSE_TAG="$REPO:compose"
```
ryanking: This script later got refactored out to https://github.com/chanzuckerberg/infra-tools/blob/master/docker-helper, which a bunch of teams use. That script includes a version of this optimization, but without the compose label.

Might be worth getting on that script version to see if it helps.

Usage example [here](https://github.com/chanzuckerberg/oauth2_proxy/blob/ea0160d7a50320fde398e785694f08242da98ccb/.travis.yml#L4), though make sure you use the latest version of the script from infra-tools.

gregdingle: Ok that’s good to know, and independent validation of this change. 

How do you recommend we make use of that shared script? Copy it into our web repo? It looks like it takes care of pushing as well, so it could replace our dockerpush script.

gregdingle:
Tiago, good point. I’ll add a temporary commit that changes a lock file to see if the cache is invalidated.

gregdingle:
Test of updating deps worked: 

```
Step 14/24 : RUN gem install bundler && bundle install --jobs 20 --retry 5
 ---> Running in 08078d849cf8
Successfully installed bundler-2.0.1
1 gem installed
```

bin/build-docker
gregdingle: In discussion with Tiago, we decided to defer this work. See https://jira.czi.team/browse/IDSEQ-970.

PR #2216, opened 2019-05-04
[Data Discovery] Remove projects that do not have samples passing filters.

tfrcarvalho:
## Description

When using filters, we remove any projects that do not have any samples that pass the filters.
This can be tricky when considering empty projects. In that case we only filter them out if there are any filters defined.


tfrcarvalho:
Not sure what you mean about hidden projects in this context... maybe you can explain better in person.

PR #2219, opened 2019-05-07
Precache report info on success of pipeline run

gregdingle:
# Description

This fixes the bad perf of the sample report page by caching the http response as a file on disk by using the Rails standard action caching. The browser HTTP client-side caching behavior is preserved in prod, and enabled in dev.

My tests of a large sample show 
1) report_info time decreased from 8s to 1s by server side caching
2) report_info time decreased to less than 1s by client side caching

I refactored some of the `report_info` and child methods to make the param dependencies more clear.

I still have some questions about the correct defaults for params like `background_id` which will affect the cache hit rate. Note we can always precache with more param values. 

NOTE: the choice of cache keys means that the server cache of a report will be invalidated on 
* new pipeline run
* any update of sample object
* new code push

# Test and Reproduce

Test caching
1. Open a report such as http://localhost:3000/samples/12303
2. See SQL queries in rails log 
3. Reload the report 
4. See cache line in rails log such as 

```
ActiveSupport::Cache::FileStore : Cache read: /tmp/rails_cache_store/BD7/1E3/views%2Flocalhost%3A3000%2Fsamples%2F10399%2Freport_info.json%3Fbackground_id%3D29%26git_version%3D%26pipeline_version%3D3.3%26report_ts%3D1548725583%26sort_by%3Dnt_aggregatescore ({:expires_in=>365 days})
```

5. Verify that the decoded URL is correctly unique in param values
```/tmp/rails_cache_store/BD7/1E3/views/localhost:3000/samples/10399/report_info.json?background_id=29&git_version=&pipeline_version=3.3&report_ts=1548725583&sort_by=nt_aggregatescore```

6. Verify cache on further reloads of report page

7. Verify cache miss on changing background

Test precaching:

1.  In `rails console`, load a pipeline run
2. Precache  method and see the following:

```
irb(main):002:0> PipelineRun.find(18660).precache_report_info!
...
[2019-05-07T19:28:38] DEBUG Rails : Precaching URL http://localhost:3000/samples/12295/report_info?background_id=93&format=json&git_version=&pipeline_version=3.3&report_ts=1549325153&scoring_model=aggregate_score&sort_by=nt_aggregatescore
=> #<Tempfile:/tmp/open-uri20190507-144-u0xcg3>
```
3. Run it again and see a cache hit in the rails logs

Overall perf testing:

I'll review the stats before and after launch to prod in the airbrake dashboard.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -76,6 +76,10 @@ gem 'shortener'
 # Explicit load to avoid an "unable to load" warning
 gem 'http-2'
 
+# Caching for action responses. Added for slow pages such sample report.
+# See https://github.com/rails/actionpack-action_caching .
+gem 'actionpack-action_caching'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'bundler-audit'
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -214,7 +214,7 @@ class PipelineSampleReport extends React.Component {
     });
     let params = `?${window.location.search.replace("?", "")}&report_ts=${
       this.report_ts
-    }&version=${this.gitVersion}`;
+    }&git_version=${this.gitVersion}&format=json`;
 
     const [sampleReportInfo, summaryContigCounts] = await Promise.all([
       getSampleReportInfo(this.sampleId, params),
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -30,8 +30,10 @@ class SamplesController < ApplicationController
                    :dimensions, :all, :show_sample_names, :cli_user_instructions, :metadata_fields, :samples_going_public,
                    :search_suggestions, :stats, :upload, :validate_sample_files].freeze
 
-  before_action :authenticate_user!, except: [:create, :update, :bulk_upload, :bulk_upload_with_metadata]
-  before_action :authenticate_user_from_token!, only: [:create, :update, :bulk_upload, :bulk_upload_with_metadata]
+  # For API-like access
+  TOKEN_AUTH_ACTIONS = [:create, :update, :bulk_upload, :bulk_upload_with_metadata, :report_info].freeze
+  before_action :authenticate_user!, except: TOKEN_AUTH_ACTIONS
+  before_action :authenticate_user_from_token!, only: TOKEN_AUTH_ACTIONS
 
   before_action :admin_required, only: [:reupload_source, :resync_prod_data_to_staging, :kickoff_pipeline, :retry_pipeline, :pipeline_runs]
   before_action :no_demo_user, only: [:create, :bulk_new, :bulk_upload, :bulk_import, :new]
@@ -50,6 +52,14 @@ class SamplesController < ApplicationController
   DEFAULT_MAX_NUM_TAXONS = 30
   MAX_PAGE_SIZE_V2 = 100
   MAX_BINS = 34
+  REPORT_INFO_CACHE_EXPIRES = 365.days
+
+  # before_action filters placed after the caches_action directive will not run
+  # when serving from the cache.
+  caches_action :report_info, expires_in: REPORT_INFO_CACHE_EXPIRES, cache_path: lambda do
+    pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
+    params.permit(pipeline_run.report_info_params.keys)
+  end
 
   # GET /samples
   # GET /samples.json
@@ -570,7 +580,7 @@ class SamplesController < ApplicationController
   # GET /samples/1/metadata.json
   def metadata
     # Information needed to show the samples metadata sidebar.
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     summary_stats = nil
     pr_display = nil
     ercc_comparison = nil
@@ -645,7 +655,7 @@ class SamplesController < ApplicationController
   # GET /samples/1
   # GET /samples/1.json
   def show
-    @pipeline_run = select_pipeline_run(@sample, params)
+    @pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
     @amr_counts = nil
     can_see_amr = (current_user.admin? || current_user.allowed_feature_list.include?("AMR"))
     if can_see_amr && @pipeline_run
@@ -654,7 +664,7 @@ class SamplesController < ApplicationController
         @amr_counts = @pipeline_run.amr_counts
       end
     end
-    @pipeline_version = @pipeline_run.pipeline_version || PipelineRun::PIPELINE_VERSION_WHEN_NULL if @pipeline_run
+    @pipeline_version = @pipeline_run.report_info_params[:pipeline_version] if @pipeline_run
     @pipeline_versions = @sample.pipeline_versions
 
     @pipeline_run_display = curate_pipeline_run_display(@pipeline_run)
@@ -667,8 +677,7 @@ class SamplesController < ApplicationController
     @host_genome = @sample.host_genome ? @sample.host_genome : nil
     @background_models = current_power.backgrounds.where(ready: 1)
     @can_edit = current_power.updatable_sample?(@sample)
-    @git_version = ENV['GIT_VERSION'] || ""
-    @git_version = Time.current.to_i if @git_version.blank?
+    @git_version = ENV['GIT_VERSION'] || "" # used for cache busting
 
     @align_viz = false
     align_summary_file = @pipeline_run ? "#{@pipeline_run.alignment_viz_output_s3_path}.summary" : nil
@@ -681,7 +690,7 @@ class SamplesController < ApplicationController
     # Check if the report table should actually show
     if background_id && @pipeline_run && @pipeline_run.report_ready?
       @report_present = true
-      @report_ts = @pipeline_run.updated_at.to_i
+      @report_ts = @pipeline_run.report_info_params[:report_ts]
       @all_categories = all_categories
       @report_details = report_details(@pipeline_run, current_user.id)
       @ercc_comparison = @pipeline_run.compare_ercc_counts
@@ -722,8 +731,8 @@ class SamplesController < ApplicationController
   end
 
   def report_info
-    expires_in 30.days
-    @pipeline_run = select_pipeline_run(@sample, params)
+    expires_in REPORT_INFO_CACHE_EXPIRES
+    @pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
 
     ##################################################
     ## Duct tape for changing background id dynamically
@@ -734,7 +743,12 @@ class SamplesController < ApplicationController
       pipeline_run_id = @pipeline_run.id
     end
 
-    @report_info = external_report_info(pipeline_run_id, background_id, params)
+    @report_info = external_report_info(
+      pipeline_run_id,
+      background_id,
+      params[:scoring_model],
+      params[:sort_by]
+    )
 
     # Fill lineage details into report info.
     # @report_info[:taxonomy_details][2] is the array of taxon rows (which are hashes with keys like tax_id, name, NT, etc)
@@ -773,14 +787,14 @@ class SamplesController < ApplicationController
   end
 
   def contig_taxid_list
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     render json: pr.get_taxid_list_with_contigs
   end
 
   def taxid_contigs
     taxid = params[:taxid]
     return if HUMAN_TAX_IDS.include? taxid.to_i
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     contigs = pr.get_contigs_for_taxid(taxid.to_i)
     output_fasta = ''
     contigs.each { |contig| output_fasta += contig.to_fa }
@@ -788,7 +802,7 @@ class SamplesController < ApplicationController
   end
 
   def summary_contig_counts
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     min_contig_size = params[:min_contig_size] || PipelineRun::MIN_CONTIG_SIZE
     contig_counts = pr.get_summary_contig_counts(min_contig_size)
     render json: { min_contig_size: min_contig_size, contig_counts: contig_counts }
@@ -796,7 +810,7 @@ class SamplesController < ApplicationController
 
   def show_taxid_fasta
     return if HUMAN_TAX_IDS.include? params[:taxid].to_i
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     if params[:hit_type] == "NT_or_NR"
       nt_array = get_taxid_fasta_from_pipeline_run(pr, params[:taxid], params[:tax_level].to_i, 'NT').split(">")
       nr_array = get_taxid_fasta_from_pipeline_run(pr, params[:taxid], params[:tax_level].to_i, 'NR').split(">")
@@ -816,7 +830,7 @@ class SamplesController < ApplicationController
       return
     end
 
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
 
     @tax_level = @taxon_info.split("_")[1]
     @taxon_name = taxon_name(@taxid, @tax_level)
@@ -855,7 +869,7 @@ class SamplesController < ApplicationController
   end
 
   def contigs_fasta
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     contigs_fasta_s3_path = pr.contigs_fasta_s3_path
 
     if contigs_fasta_s3_path
@@ -869,7 +883,7 @@ class SamplesController < ApplicationController
   end
 
   def contigs_summary
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     local_file = pr.generate_contig_mapping_table
 
     @contigs_summary = File.read(local_file)
@@ -890,7 +904,7 @@ class SamplesController < ApplicationController
   end
 
   def contigs_fasta_by_byteranges
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     byteranges = params[:byteranges]
 
     contig_fasta = pr.contigs_fasta_s3_path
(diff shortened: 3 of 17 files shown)

### CODE REVIEW

File: config/environments/prod.rb [hunk 1]
```diff
@@ -13,6 +13,7 @@
   # Full error reports are disabled and caching is turned on.
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
+  config.cache_store = :file_store, "/tmp/rails_cache_store"
```
gregdingle: in prod it may be better to use redis or memcache to ensure a single global cache... I'm not sure how many prod web machines we are running right now, or will be soon.

app/controllers/samples_controller.rb
```diff
@@ -30,8 +30,10 @@ class SamplesController < ApplicationController
                    :dimensions, :all, :show_sample_names, :cli_user_instructions, :metadata_fields, :samples_going_public,
                    :search_suggestions, :stats, :upload, :validate_sample_files].freeze
 
-  before_action :authenticate_user!, except: [:create, :update, :bulk_upload, :bulk_upload_with_metadata]
-  before_action :authenticate_user_from_token!, only: [:create, :update, :bulk_upload, :bulk_upload_with_metadata]
+  # For API-like access
+  TOKEN_AUTH_ACTIONS = [:create, :update, :bulk_upload, :bulk_upload_with_metadata, :report_info].freeze
```
cdebourcy: nice

app/models/pipeline_run.rb
```diff
@@ -1390,4 +1391,35 @@ def outputs_by_step(can_see_stage1_results = false)
   def self.viewable(user)
     where(sample_id: Sample.viewable(user).pluck(:id))
   end
+
+  # Keys here are used as cache keys for report_info action in SamplesController.
+  # The values here are used as defaults for PipelineSampleReport.jsx.
+  def report_info_params
+    {
+      pipeline_version: pipeline_version || PipelineRun::PIPELINE_VERSION_WHEN_NULL,
+      # TODO: (gdingle): is this good enough for cache hits? see get_background_id
```
cdebourcy: Actually background_id selection is persisted in a cookie. So all reports will be opened with the background that the user last selected. `get_background_id` just checks if the user is allowed to access the selected background (if not, it uses default public background instead).

app/models/pipeline_run.rb
```diff
@@ -845,6 +845,7 @@ def update_job_status
       # all stages succeeded
       self.finalized = 1
       self.job_status = STATUS_CHECKED
+      precache_report_info!
```
cdebourcy: Could you do this asynchronously using resque? I'm weary of adding more processing steps to `update_job_status`. It needs to be fast -- the round-robin of pipeline checks should not be delayed by the precaching of the reports

cdebourcy: Also, logically the precache action should probably be associated with the "result monitor" (and the results_finalized attribute of pipeline_run) rather than the "job monitor" (and the finalized attribute of pipeline_run). See https://github.com/chanzuckerberg/idseq-web/blob/9c55396980692f860c4bc961372f355c75017b16/app/models/pipeline_run.rb#L763 for the result-monitor equivalent of the `update_job_status` function.

app/controllers/samples_controller.rb
```diff
@@ -667,8 +677,7 @@ def show
     @host_genome = @sample.host_genome ? @sample.host_genome : nil
     @background_models = current_power.backgrounds.where(ready: 1)
     @can_edit = current_power.updatable_sample?(@sample)
-    @git_version = ENV['GIT_VERSION'] || ""
-    @git_version = Time.current.to_i if @git_version.blank?
+    @git_version = ENV['GIT_VERSION'] || "" # used for cache busting
```
cdebourcy: Does this mean the cache will be invalidated every time we deploy a new commit? I think we would need more persistence than that?

Gemfile
```diff
@@ -76,6 +76,10 @@ gem 'shortener'
 # Explicit load to avoid an "unable to load" warning
 gem 'http-2'
 
+# Caching for action responses. Added for slow pages such sample report.
+# See https://github.com/rails/actionpack-action_caching .
+gem 'actionpack-action_caching'
```
cdebourcy: does it have eviction policies other than the expiration time you set? I'm assuming it has some kind of mechanism that prevents it from filling up too much space? Where are those parameters set?

config/environments/prod.rb
cdebourcy: I agree that would be best

app/controllers/samples_controller.rb
cdebourcy: Actually, no! I think `report_info` must not be in TOKEN_AUTH_ACTIONS. Will follow up offline

cdebourcy:
This is nice but I think the following problems would ruin the benefit unless they are addressed:

(1) Any new code push invalidates cache. 
I think we would need to put a version number on the report_info method. The overall git version will change far too frequently and cache will only be useful in a very small minority of cases.

(2) Any update of sample object invalidates cache.
Surely it must be possible to avoid that? User might go and edit the sample name and it would be sad if that caused the report to become slower.

(3) Only the "default" background is cached.
Probably won't cut it, because the initial background selection is just based on user's last session -- probably will be different from default_background_id. If we're using redis / memcached I'd vote for precaching every background id. But then the controller will need to ensure that access permission to the background is verified for the specific user before it serves the report from cache.

(4) Other access control issues to be discussed offline

config/environments/prod.rb
gregdingle: It looks like redis requires some configuration to be used as a cache. See https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore . Who is familiar with redis on our team? 

And if we are running a single web machine right now, I don't think it will result in any faster performance, FYI.

app/models/pipeline_run.rb
gregdingle: Ok so what would you recommend given the expected access pattern? I noticed this line
```
viewable_background_ids = current_power.backgrounds.pluck(:id)
```
We could simply cache a report for all `viewable_background_ids`.

app/models/pipeline_run.rb
gregdingle: Thanks for the guidance. I wanted to ask about that. I'll move `precache_report_info!` to inside `monitor_results`.

app/models/pipeline_run.rb
cdebourcy: Yes. Would you use the `Power` of the user that uploaded the sample? What about other users who have access by way of project membership? I think we could precache for *all* background_ids, but need to bear in mind we will need to enforce access control at request time

app/models/pipeline_run.rb
gregdingle: Do you still think this should be done non-blocking thru resque? My assumption is that the precaching happens only once per pipeline run success, which seems relatively rare.

app/controllers/samples_controller.rb
gregdingle: Yes it does. It is conservative in that way. I'd prefer that for now. Keep in mind also that the cache will be regenerated on any report load, not just precaching.

Gemfile
gregdingle: No,  I suppose it could run out of disk space but I doubt that... each report is 20MB max, so we could store 100k in 1TB of disk space. 

When we get to that point I would recommend doing something smarter with redis (LRU).

app/models/pipeline_run.rb
cdebourcy: Yes, I still think if we want to open ourselves up to the possibility of precaching with all the different backgrounds or other parameters, it would be best to make it non-blocking

gregdingle:
Good review! 

> (1) Any new code push invalidates cache.

I agree this is the most pessimistic invalidation. However, if we see the speedup we want in prod, why risk it? 

Anyway, instead of versioning the report info method, I would recommend adding a cache-busting retry in the javascript fetch, which would be more robust. 

> (2) Any update of sample object invalidates cache.

Again, this is the most pessimistic approach. In general, it's hard to know whether a change in a report object requires a refresh. Actually, the most pessimistic would be to make the report_ts include the update time of the related data as well, the taxon counts. 


> (3) Only the "default" background is cached.

I agree, I'll change this. 

> If we're using redis / memcached I'd vote for precaching every background id. 

Why not for the file system? 

> But then the controller will need to ensure that access permission to the background is verified for the specific user before it serves the report from cache.

Is it possible for a user to lose permission to a background? I actually don't even know where backgrounds come from. I didn't know they are user controlled.

> (4) Other access control issues to be discussed offline

To be clear, the request to the cached report_info goes through the normal access control--critically, `set_sample`.

app/models/pipeline_run.rb
```diff
@@ -1390,4 +1391,35 @@ def outputs_by_step(can_see_stage1_results = false)
...
+  # Keys here are used as cache keys for report_info action in SamplesController.
+  # The values here are used as defaults for PipelineSampleReport.jsx.
+  def report_info_params
+    {
+      pipeline_version: pipeline_version || PipelineRun::PIPELINE_VERSION_WHEN_NULL,
+      # TODO: (gdingle): is this good enough for cache hits? see get_background_id
+      background_id: sample.default_background_id,
+      # scoring_model is currrently static
+      scoring_model: TaxonScoringModel::DEFAULT_MODEL_NAME,
+      # TODO: (gdingle): why does PipelineSampleReport and SamplesController have different default sort_by?
```
gregdingle: @cdebourcy any thoughts about this? the backend default is `highest_nt_aggregatescore`

Gemfile
tfrcarvalho: Also, report info should not be 20MB. Probably not even close.

app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -214,7 +214,7 @@ class PipelineSampleReport extends React.Component {
     });
     let params = `?${window.location.search.replace("?", "")}&report_ts=${
       this.report_ts
-    }&version=${this.gitVersion}`;
+    }&git_version=${this.gitVersion}&format=json`;
```
tfrcarvalho: Why is format sent as param?
Should be in the endpoitn i.e. `report_info.json?...`

gregdingle: format is sent because the endpoint is not currently requested as `report_info.json`, and it was necessary to make the cache response header correct. I don't know why `report_info.json`  is not in use. I could fix that.

gregdingle:
I played around with calling an action internally for a while without success. I started with https://coderwall.com/p/gghtkq/rails-internal-requests and went deeper. I gave up for now and implemented the local IP check which I think provides good enough security for report info. If the reviewers here agree with this approach, I'll write controller tests.

Gemfile
gregdingle: I just double checked and the largest cached file is 19M of uncompressed JSON.

config/environments/prod.rb
davidrissato: I think it is required to use redis or memcache, since we currently have 6 instances.

app/models/pipeline_run.rb
davidrissato: I agree with Charles, some of these reports might be slow to generate, and it can affect the stability of update_status.

Gemfile
davidrissato: I don't think it is a good design caching on disk. We need to move this cache to Redis or even S3. We have 6 servers, caching on disk will have a hit cache miss of more than 83%. We will be able to use the cache only 16% of the time.
Not to mention that we will need to evict caches on 6 different machines.
Also, it can lead to inconsistent results, since caches on different machines can potentially present different data.

davidrissato: Also, since this data is big, it might worth compressing that before sending to the cache, specially if we use memory caches like redis.

davidrissato:
Another thing to take in consideration is that using Redis cache is only useful if you are regenerating cache on the fly when you don't have one.
I have the impression that we are generating these cache results during update_job_status, which means this cache needs to be persistent. So Redis is not an option either, since it is supposed to be used as a fast cache, not as a persistent one.
Persistent caching need a persistent technology, like Elastic Search, S3, or even mysql with a blob field.

cdebourcy:
> Is it possible for a user to lose permission to a background? I actually don't even know where backgrounds come from. I didn't know they are user controlled.
> 
> To be clear, the request to the cached report_info goes through the normal access control--critically, `set_sample`.

A background is generated from a collection of samples. Users have the ability to build backgrounds that contain samples of their choice. You have access to the background if you have access to every single one of these samples. `set_sample` checks your permission to view the sample. `get_background_id` checks your permission to view the background: https://github.com/chanzuckerberg/idseq-web/blob/master/app/controllers/samples_controller.rb#L733
Is this access control line still going to be executed if a report is served from cache?

On a sidenote, we should definitely separate that background access control logic out more in the code.

Gemfile
yunfangjuan: why don't we just do memcache?

yunfangjuan: agreed with @davidrissato's point on caching centrally and compression. There's a rails memcache gem that's pretty turn key I think. Why don't we just setup a memcache tier and use that one? https://redislabs.com/lp/rails-memcached/

app/controllers/samples_controller.rb
```diff
@@ -50,6 +53,15 @@ class SamplesController < ApplicationController
   DEFAULT_MAX_NUM_TAXONS = 30
   MAX_PAGE_SIZE_V2 = 100
   MAX_BINS = 34
+  REPORT_INFO_CACHE_EXPIRES = 365.days
```
yunfangjuan: Isn't this a bit long? I think 30 days should suffice.

app/controllers/samples_controller.rb
yunfangjuan: an alternative way to do this is with a constant in initializers called WEB_REPORT_CACHE_VERSION, every time there's a change in format, up the version.

app/models/pipeline_run.rb
yunfangjuan: this is only for ranking purposes where the front end can manipulate., so I don't think it really matters.

config/environments/prod.rb
yunfangjuan: please use memcache. we don't need a persitent cache on this. Doing a decent size memcache tier should suffice.

gregdingle: Does anybody on the team have experience setting up memcache or redis, ideally for rails cache? 

And do you think we should make that change in this PR?

app/models/pipeline_run.rb
gregdingle: Ok sounds good. I'll see about access control at request time.

app/models/pipeline_run.rb
gregdingle: Ok I'll do that in this PR.

app/controllers/samples_controller.rb
gregdingle: Yea, that's what Charles suggested as well. If people here feel it's worth the risk then I will implement manually versioning of the cache.

Gemfile
gregdingle: Yea, same comments as above. Does anybody have experience setting up memcache or redis ideally with rails? And do you want to see that change in this PR?

app/controllers/samples_controller.rb
gregdingle: I'm not sure ... I was thinking about samples such as medical detectives that may not be accessed for 30 days but we still want them fast.

gregdingle:
@yunfangjuan 

> Also, we should also cache the contigs and coverage viz.

Could you elaborate or point to the code in mind? 

> I will also suggest doing a database queries analysis and see what frequent/heavy queries that we can cache.

The interesting thing about report_info, and I also the heatmap I expect, is that most of the time is now spent **after** the SQL query in JSON serialization. Thus the importance of analyzing by endpoint. 

We now have good data on perf of endpoints here: https://chanzuckerberg.airbrake.io/projects/158403/performance?dur=90d%2F1d&last_n=30&environment=prod&sortColumn=total.p95&sortDesc=true&page=1 .

gregdingle:
@yunfangjuan 

I'm also wondering about this comment. What was the clean up you imagined? 

```
  def report_info
    expires_in REPORT_INFO_CACHE_EXPIRES
    @pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])

    ##################################################
    ## Duct tape for changing background id dynamically
    ## TODO(yf): clean the following up.
    ####################################################
    if @pipeline_run && (((@pipeline_run.adjusted_remaining_reads.to_i > 0 || @pipeline_run.finalized?) && !@pipeline_run.failed?) || @pipeline_run.report_ready?)
      background_id = get_background_id(@sample)
      pipeline_run_id = @pipeline_run.id
    end
```

gregdingle:
I just changed the way the client http caching works for the report_info. Previously it was set to 30 days, but it was not working in the case of precached reports, because, as I discovered, Rails strangely does not set the cache-control header on cached responses or have an option to do so. (See https://stackoverflow.com/questions/1251992/how-to-get-caches-action-to-set-expires-headers-when-there-is-a-hit-in-memcached)

On deeper thought, I decided it was best to use only the last-modified header, so the server has complete control over client cache behavior. The cost is an additional http request from the client. In this context, I feel that extra 150 to 500ms is a small price to pay. 

I verified that the server is returning 304 response appropriately, and that it is faster.

gregdingle:
I just changed the precache call to go through resque queue. It seems to be working. Not having used resque before, I'm not sure how it is supposed to behave. 

```
irb(main):002:0> Resque.enqueue(PrecacheReportInfo, 18660)
=> true
```

image

The job showed up in the queue though it didn't run. I'm pretty sure nothing will run in my local resque queue. I'm not sure if that's by design.

Gemfile
yunfangjuan: Amazon has the elasticcache tier and you can add a docker container for the local use.

cdebourcy:
> I just changed the precache call to go through resque queue. It seems to be working. Not having used resque before, I'm not sure how it is supposed to behave.
> 
> ```
> irb(main):002:0> Resque.enqueue(PrecacheReportInfo, 18660)
> => true
> ```
> 
> image
> 
> The job showed up in the queue though it didn't run. I'm pretty sure nothing will run in my local resque queue. I'm not sure if that's by design.

It should run locally. But you may have to rebuild your Docker images and start the container. It's possible it didn't pick up your code change.

gregdingle:
Now that I've added the unit tests and counters, I'd like to merge this as-is and deploy to staging so I can see that everything is working. Then I will switch to redis or memcache, and focus on cache hit efficiency.

test/controllers/samples_controller_test.rb
```diff
@@ -349,4 +349,118 @@ class SamplesControllerTest < ActionDispatch::IntegrationTest
...
+    second_runtime = @response.headers["X-Runtime"]
+    cache_header = @response.headers["X-IDseq-Cache"]
+    assert_equal "requested", cache_header
+
+    # Second request should always be at least 2x faster than the first.
+    # Typically it will be 10x faster.
+    assert first_runtime.to_f > (second_runtime.to_f * 2)
+  end
+
+  test 'report_info should override background when background is not viewable' do
```
cdebourcy: thanks

PR #2224, opened 2019-05-10
[Geosearch] Modify search box component to use as an input field + new GeoSearchInputBox

jshoe:
### Description
- Adapt LiveSearchBox with new props to be suitable for use as an input field.
- Create a new GeoSearchInputBox component and move handleSearchTriggered logic into there for better composition.
- In inputMode, let them use just the plain text of what they put in if they hit enter without selecting anything and keep the value in the box when they select something.
- handleSelectionChange copies the onSelectionChange in SearchBox.
- Also add a rectangular style to match the other metadata input fields.
- Fix props results object strategy for prop validation.
- Alternative approach would be to not have inputMode and instead add props to control the 'value' and 'results' from outside the LiveSearchBox (to set the value and close the results box when they press enter without selection).

### Demo
May-10-2019 10-57-01

### Test and Reproduce
- Pull the branch, start with `LOCATION_IQ_API_KEY=key MAPTILER_API_KEY=key docker-compose up web`, go to `http://localhost:3000/locations/map_playground`.
- Type in search queries like California, White House, Uganda, Bangladesh, etc. and test the search box. See behavior description above.


app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
```diff
@@ -0,0 +1,51 @@
...
+      results: [{ title: query }]
+    };
+    return categories;
+  };
+
+  render() {
+    const { onResultSelect } = this.props;
+    return (
+      <LiveSearchBox
+        onSearchTriggered={this.handleSearchTriggered}
```
gregdingle: might as well put in the analytics tracking now... see https://czi.quip.com/67RCAIiHN0Qc/IDseq-product-analytics-How-to-log

jshoe:
@happyimadesignr has some ideas for the other 2 searches but is still figuring it out!

Added 2 analytics events! :)

PR #2225, opened 2019-05-10
Fix precaching

gregdingle:
# Description

This is a quick fix to solve the auth error observed in staging. 

# Test and Reproduce

Will deploy to staging and test.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -87,15 +87,12 @@ class ApplicationController < ActionController::Base
 
   # This provides IP whitelisting on top of token authentication for added security.
   def authenticate_internal_user_from_token!
-    if request.remote_ip == "127.0.0.1" ||
-       # is the request coming from the server's own public IP address?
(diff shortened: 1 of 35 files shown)


app/controllers/application_controller.rb
```diff
@@ -87,7 +87,9 @@ def authenticate_user_from_token!
 
   # This provides IP whitelisting on top of token authentication for added security.
   def authenticate_internal_user_from_token!
```
cdebourcy: I think this should be a PR to the branch "gdingle/action-cache"? I.e. " gregdingle wants to merge 1 commit into gdingle/action-cache from gdingle/fix-sample-viewable"? `authenticate_internal_user_from_token!` no longer exists in master because I reverted your last commit

gregdingle: Yea, I wasn't planning on merging this at all, only deploy it to staging, but if I did merge, it woudl be on `gdingle/action-cache`.

cdebourcy:
This can be closed, correct?

PR #2229, opened 2019-05-10
Fix airbrake filter

gregdingle:
# Description

Airbrake's docs are buggy: https://github.com/airbrake/airbrake/blob/406ec71c04c0ca2d3559c3428b45a4adf5456c02/CHANGELOG.md#L12 . 

I confirmed with support.
image

# Test and Reproduce
Test in staging

### DIFF
--- a/config/initializers/airbrake.rb
+++ b/config/initializers/airbrake.rb
@@ -71,13 +71,10 @@ if ENV['AIRBRAKE_PROJECT_ID'] && ENV['AIRBRAKE_PROJECT_KEY']
     if resource.route.include?('health_check')
       resource.ignore!
     end
-  end
-
-  # Ignore CZI users from perf stats because they tend to be outliers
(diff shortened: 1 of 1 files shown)


config/initializers/airbrake.rb
```diff
@@ -71,13 +71,10 @@
...
     end
-  end
-
-  # Ignore CZI users from perf stats because they tend to be outliers
-  Airbrake.add_performance_hook do |resource|
+    # Ignore CZI users from perf stats because they tend to be outliers
     if resource.stash.key?(:user)
       email = resource.stash[:user][:email]
-      if email.present? && ["chanzuckerberg.com"].include?(email.split("@").last)
+      if !email.nil? && ["chanzuckerberg.com"].include?(email.split("@").last)
```
gregdingle: I'm still learning about ruby `nil` :/

PR #2237, opened 2019-05-15
Cache report info in redis

gregdingle:
# Description

This is a replacement for https://github.com/chanzuckerberg/idseq-web/pull/2219 using Redis instead of file cache and low level cache instead of action level cache. 

It fixes the bad perf of the sample report page by caching the http response. The browser HTTP client-side caching behavior is also optimized.

My tests of a large sample show 
1) report_info time decreased from 8s to 1s by server side caching
2) report_info time decreased to less than 1s by client side caching

I refactored some of the `report_info` and child methods to make the param dependencies more clear. I refactored the `ReportHelper` class to be mostly private class methods because that's what they were--they didn't need instance state or external access.

I chose to precache all `background_id` after all because Redis should act as a LRU cache when `expires_in` is enabled, so unused data will be evicted. 

We currently have 6 GB free in a cache.m3.large according to https://us-west-2.console.aws.amazon.com/elasticache/home?region=us-west-2. 

We I chose to share the Redis machines with Resque because Redis will act both as LRU and persistent in the default `volatile-lru` mode. 

I chose the default expires to be one month, on the assumption that anything worth seeing will be viewed within a month of precaching. 

I expanded the definition of `report_ts` to be the max `updated_at` of all child objects of a pipeline run, such as `taxon_counts`. This should capture all changes in underlying data, now and going forward. 

The git version is no longer in the cache key though I left it in the params for debugging. 

I removed the `sort_by` and `scoring_model` as action params, because they were always the defaults anyway. 

# Test and Reproduce

Automated tests
`rails t`

Test caching
1. Open a report such as http://localhost:3000/samples/12303
2. See SQL queries in rails log 
3. Reload the report 
4. See cache line in rails log such as 

```
ActiveSupport::Cache::FileStore : Cache read: /tmp/rails_cache_store/BD7/1E3/views%2Flocalhost%3A3000%2Fsamples%2F10399%2Freport_info.json%3Fbackground_id%3D29%26git_version%3D%26pipeline_version%3D3.3%26report_ts%3D1548725583%26sort_by%3Dnt_aggregatescore ({:expires_in=>365 days})
```

5. Verify that the decoded URL is correctly unique in param values
```/tmp/rails_cache_store/BD7/1E3/views/localhost:3000/samples/10399/report_info.json?background_id=29&git_version=&pipeline_version=3.3&report_ts=1548725583&sort_by=nt_aggregatescore```

6. Verify cache on further reloads of report page

7. Verify cache miss on changing background

Test precaching:

1.  In `rails console`, load a pipeline run
2. Precache  method and see the following:

```
irb(main):002:0> PipelineRun.find(18660).precache_report_info!
...
[2019-05-07T19:28:38] DEBUG Rails : Precaching URL http://localhost:3000/samples/12295/report_info?background_id=93&format=json&git_version=&pipeline_version=3.3&report_ts=1549325153&scoring_model=aggregate_score&sort_by=nt_aggregatescore
=> #<Tempfile:/tmp/open-uri20190507-144-u0xcg3>
```
3. Run it again and see a cache hit in the rails logs

Overall perf testing:

I'll review the stats before and after launch to prod in the airbrake dashboard.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -683,8 +683,8 @@ class SamplesController < ApplicationController
     if background_id && @pipeline_run && @pipeline_run.report_ready?
       @report_present = true
       @report_ts = @pipeline_run.report_info_params[:report_ts]
-      @all_categories = all_categories
-      @report_details = report_details(@pipeline_run, current_user.id)
+      @all_categories = ReportHelper::ALL_CATEGORIES
+      @report_details = ReportHelper.report_details(@pipeline_run, current_user.id)
       @ercc_comparison = @pipeline_run.compare_ercc_counts
     end
 
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -749,7 +749,7 @@ class SamplesController < ApplicationController
     response.headers["X-IDseq-Cache"] = 'requested'
 
     pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
-    json = Rails.cache.fetch(cache_key) do
+    json = Rails.cache.fetch(cache_key, expires_in: 30.days) do
       MetricUtil.put_metric_now("samples.cache.miss", 1)
       response.headers["X-IDseq-Cache"] = 'missed'
       ReportHelper.report_info_json(pipeline_run, params[:background_id])
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -30,8 +30,11 @@ class SamplesController < ApplicationController
                    :dimensions, :all, :show_sample_names, :cli_user_instructions, :metadata_fields, :samples_going_public,
                    :search_suggestions, :stats, :upload, :validate_sample_files].freeze
 
-  before_action :authenticate_user!, except: [:create, :update, :bulk_upload, :bulk_upload_with_metadata]
-  before_action :authenticate_user_from_token!, only: [:create, :update, :bulk_upload, :bulk_upload_with_metadata]
+  # For API-like access
+  TOKEN_AUTH_ACTIONS = [:create, :update, :bulk_upload, :bulk_upload_with_metadata].freeze
+  before_action :authenticate_user!, except: TOKEN_AUTH_ACTIONS + [:report_info]
+  before_action :authenticate_internal_user_from_token!, only: [:report_info]
+  before_action :authenticate_user_from_token!, only: TOKEN_AUTH_ACTIONS
 
   before_action :admin_required, only: [:reupload_source, :resync_prod_data_to_staging, :kickoff_pipeline, :retry_pipeline, :pipeline_runs]
   before_action :no_demo_user, only: [:create, :bulk_new, :bulk_upload, :bulk_import, :new]
@@ -50,6 +53,33 @@ class SamplesController < ApplicationController
   DEFAULT_MAX_NUM_TAXONS = 30
   MAX_PAGE_SIZE_V2 = 100
   MAX_BINS = 34
+
+  # TODO: (gdingle): refactor this to function
+  # before_action filters placed after the caches_action directive will not run
+  # when serving from the cache.
+  # rubocop:disable Style/Lambda
+  caches_action :report_info, cache_path: -> do
+    MetricUtil.put_metric_now("samples.cache.requested", 1)
+
+    pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
+    if pipeline_run.nil?
+      message = "Pipeline run not found for sample #{@sample.id}"
+      raise ActiveRecord::RecordNotFound, message
+    end
+    report_info_params = pipeline_run.report_info_params
+
+    # This allows 304 Not Modified to be returned so that the client can use its
+    # local cache and avoid the large download.
+    httpdate = Time.at(report_info_params[:report_ts]).utc.httpdate
+    response.headers["Last-Modified"] = httpdate
+    # This is a custom header for testing and debugg
+    response.headers["X-IDseq-Cache"] = 'requested'
+
+    cache_keys = params.permit(report_info_params.keys)
+    # Set background_id to a viewable background, same behavior as in report_info itself
+    cache_keys[:background_id] = get_background_id(@sample)
+    cache_keys
+  end
 
   # GET /samples
   # GET /samples.json
@@ -570,7 +600,7 @@ class SamplesController < ApplicationController
   # GET /samples/1/metadata.json
   def metadata
     # Information needed to show the samples metadata sidebar.
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     summary_stats = nil
     pr_display = nil
     ercc_comparison = nil
@@ -645,7 +675,7 @@ class SamplesController < ApplicationController
   # GET /samples/1
   # GET /samples/1.json
   def show
-    @pipeline_run = select_pipeline_run(@sample, params)
+    @pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
     @amr_counts = nil
     can_see_amr = (current_user.admin? || current_user.allowed_feature_list.include?("AMR"))
     if can_see_amr && @pipeline_run
@@ -654,7 +684,7 @@ class SamplesController < ApplicationController
         @amr_counts = @pipeline_run.amr_counts
       end
     end
-    @pipeline_version = @pipeline_run.pipeline_version || PipelineRun::PIPELINE_VERSION_WHEN_NULL if @pipeline_run
+    @pipeline_version = @pipeline_run.report_info_params[:pipeline_version] if @pipeline_run
     @pipeline_versions = @sample.pipeline_versions
 
     @pipeline_run_display = curate_pipeline_run_display(@pipeline_run)
@@ -667,8 +697,7 @@ class SamplesController < ApplicationController
     @host_genome = @sample.host_genome ? @sample.host_genome : nil
     @background_models = current_power.backgrounds.where(ready: 1)
     @can_edit = current_power.updatable_sample?(@sample)
-    @git_version = ENV['GIT_VERSION'] || ""
-    @git_version = Time.current.to_i if @git_version.blank?
+    @git_version = ENV['GIT_VERSION'] || "" # used for cache busting
 
     @align_viz = false
     align_summary_file = @pipeline_run ? "#{@pipeline_run.alignment_viz_output_s3_path}.summary" : nil
@@ -681,7 +710,7 @@ class SamplesController < ApplicationController
     # Check if the report table should actually show
     if background_id && @pipeline_run && @pipeline_run.report_ready?
       @report_present = true
-      @report_ts = @pipeline_run.updated_at.to_i
+      @report_ts = @pipeline_run.report_info_params[:report_ts]
       @all_categories = all_categories
       @report_details = report_details(@pipeline_run, current_user.id)
       @ercc_comparison = @pipeline_run.compare_ercc_counts
@@ -722,8 +751,10 @@ class SamplesController < ApplicationController
   end
 
   def report_info
-    expires_in 30.days
-    @pipeline_run = select_pipeline_run(@sample, params)
+    MetricUtil.put_metric_now("samples.cache.miss", 1)
+    response.headers["X-IDseq-Cache"] = 'missed'
+
+    @pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
 
     ##################################################
     ## Duct tape for changing background id dynamically
@@ -734,7 +765,12 @@ class SamplesController < ApplicationController
       pipeline_run_id = @pipeline_run.id
     end
 
-    @report_info = external_report_info(pipeline_run_id, background_id, params)
+    @report_info = external_report_info(
+      pipeline_run_id,
+      background_id,
+      params[:scoring_model],
+      params[:sort_by]
+    )
 
     # Fill lineage details into report info.
     # @report_info[:taxonomy_details][2] is the array of taxon rows (which are hashes with keys like tax_id, name, NT, etc)
@@ -773,14 +809,14 @@ class SamplesController < ApplicationController
   end
 
   def contig_taxid_list
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     render json: pr.get_taxid_list_with_contigs
   end
 
   def taxid_contigs
     taxid = params[:taxid]
     return if HUMAN_TAX_IDS.include? taxid.to_i
-    pr = select_pipeline_run(@sample, params)
+    pr = select_pipeline_run(@sample, params[:pipeline_version])
     contigs = pr.get_contigs_for_taxid(taxid.to_i)
     output_fasta = ''
     contigs.each { |contig| output_fasta += contig.to_fa }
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

File: app/jobs/result_monitor_loader.rb [hunk 1]
```diff
@@ -15,7 +15,7 @@ def self.perform(pipeline_run_id, output)
       # TODO: revisit this
       sleep(Time.now.to_i % 30)
       output_state.update(state: PipelineRun::STATUS_LOADING_ERROR)
-      message = "SampleFailedEvent: Pipeline Run #{pr.id} for Sample #{pr.sample.id} by #{pr.sample.user.email} failed loading #{output} with #{pr.adjusted_remaining_reads} reads remaining after #{pr.duration_hrs}. See: #{pr.status_url}"
+      message = "SampleFailedEvent: Pipeline Run #{pr.id} for Sample #{pr.sample.id} by #{pr.sample.user.email} failed loading #{output} with #{pr.adjusted_remaining_reads || 0} reads remaining after #{pr.duration_hrs} hours. See: #{pr.status_url}"
```
gregdingle: This was an opportunistic change of an unrelated error message

app/controllers/samples_controller.rb
```diff
@@ -721,30 +722,78 @@ def samples_going_public
...
+    end
+
+    render json: json
+  end
+
+  # Example cache key:
+  # /samples/12303/report_info?background_id=93&format=json&pipeline_version=3.3&report_ts=1549504990
+  def self.report_info_cache_key(path, kvs)
+    kvs = kvs.to_h.sort.to_h
+    # Increment this is you ever change the response structure of report_info
```
jshoe: s/is/if

app/controllers/samples_controller.rb
```diff
@@ -721,30 +722,78 @@ def samples_going_public
...
+    httpdate = Time.at(report_info_params[:report_ts]).utc.httpdate
+    response.headers["Last-Modified"] = httpdate
+    # This is a custom header for testing and debugging
+    response.headers["X-IDseq-Cache"] = 'requested'
+
+    pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
+    json = Rails.cache.fetch(cache_key) do
+      MetricUtil.put_metric_now("samples.cache.miss", 1)
+      response.headers["X-IDseq-Cache"] = 'missed'
+      SamplesController.report_info_json(pipeline_run, params[:background_id])
```
jshoe: it looks a little odd to call controller methods statically, so report_info_json could also move to ReportHelper

app/lib/report_helper.rb
```diff
@@ -970,7 +1040,7 @@ def taxonomy_details(pipeline_run_id, background_id, params)
     rows_passing_filters = rows.length
 
     # Compute sort key and sort.
-    sort_by = decode_sort_by(params[:sort_by]) || decode_sort_by(DEFAULT_SORT_PARAM)
+    sort_by = decode_sort_by(sort_by) || decode_sort_by(DEFAULT_SORT_PARAM)
```
jshoe: where is sort_by coming from without the `params[:sort_by]`?

config/environments/test.rb
```diff
@@ -20,14 +20,21 @@
 
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
-  config.action_controller.perform_caching = false
+  # config.action_controller.perform_caching = false
```
jshoe: nit can delete line 23

app/models/pipeline_run.rb
```diff
@@ -1405,4 +1408,43 @@ def status_url
...
+  # Gets last update time of all has_many relations and current pipeline run
+  def max_updated_at
+    assocs = PipelineRun.reflect_on_all_associations(:has_many)
+    assocs_max = assocs.map { |assoc| send(assoc.name).pluck(:updated_at).max }
+    [updated_at, assocs_max.compact.max].compact.max
+  end
+
+  # This precaches reports for *all* backgrounds. Each report took between 1 and
+  # 10s in testing.
+  def precache_report_info!
```
davidrissato: If it is pre-caching for all Backgrounds, shouldn't it be a class method of Background model?

```
class Background
    def self.precache_all_ready_reports
```

test/controllers/samples_controller_test.rb
```diff
@@ -349,4 +349,118 @@ class SamplesControllerTest < ActionDispatch::IntegrationTest
...
+    get url
+    assert_response :success
+
+    second_runtime = @response.headers["X-Runtime"]
+    cache_header = @response.headers["X-IDseq-Cache"]
+    assert_equal "requested", cache_header
+
+    # Second request should always be faster than the first.
+    # Typically it will be 10x faster.
+    assert first_runtime.to_f > second_runtime.to_f
```
davidrissato: I think asserting for time spent is not a good way to check if the report has been properly cached. That assertion is brittle and might lead to random failures between runs.

I suggest assertion would only check for the response header `["X-IDseq-Cache"]` AND expected response content. The way this assertion is currently done, if the second run returned an empty report, this test would give a false positive, hurting the confidence level on our tests.

app/controllers/samples_controller.rb
gregdingle: good point

app/lib/report_helper.rb
gregdingle: It is hard coded to the default here: 

```    
report_info = ReportHelper.external_report_info(
      pipeline_run_id,
      background_id,
      TaxonScoringModel::DEFAULT_MODEL_NAME,
      ReportHelper::DEFAULT_SORT_PARAM
    )
```

test/controllers/samples_controller_test.rb
```diff
@@ -349,4 +349,118 @@ class SamplesControllerTest < ActionDispatch::IntegrationTest
...
+  end
+
+  test 'report_info should return last-modified' do
+    post user_session_path, params: @user_params
+
+    report_ts = Time.now.utc
+    url = report_info_url(report_ts: report_ts.to_i)
+    get url
+    last_modified = Time.httpdate(@response.headers["Last-Modified"])
+    # Last-Modified of test data will be creation time, so we just match to the day
```
davidrissato: This assertion is unstable if the test is running at midnight, that might give us false negatives.

Consider stubbing Time.now using ActiveSupport [TimeHelper](https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-freeze_time):

```ruby
test 'report_info should return last-modified' do
   freeze_time do
      report_ts = Time.now.utc
      
      post user_session_path, params: @user_params

      url = report_info_url(report_ts: report_ts.to_i)
      get url
      last_modified = Time.httpdate(@response.headers["Last-Modified"])

      # now you can assert the entire date
   end

```

config/environments/development.rb
```diff
@@ -15,8 +15,11 @@
   # Enable/disable caching. By default caching is disabled.
   if Rails.root.join('tmp', 'caching-dev.txt').exist?
     config.action_controller.perform_caching = true
-
-    config.cache_store = :memory_store
+    config.cache_store = :redis_store, 'redis://redis:6379/0/cache',
+                         {
+                           # Needed for redis to evict keys in volatile-lru mode
+                           expires_in: 30.days
```
davidrissato: I'm still wondering if redis is the right call for this cache layer. Do we have a good reason to expire the entry after 30 days?

config/environments/prod.rb
```diff
@@ -13,6 +13,11 @@
   # Full error reports are disabled and caching is turned on.
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
+  config.cache_store = :redis_store, ENV['REDISCLOUD_URL'] + '/0/cache',
+                       {
+                         # Needed for redis to evict keys in volatile-lru mode
+                         expires_in: 30.days
```
davidrissato: I'm still wondering if redis is the right call for this cache layer. Do we have a good reason to expire the entry after 30 days? Why 30 days? Why not 1 or 365?

test/controllers/samples_controller_test.rb
gregdingle: Fair point about the runtime. I added that initially for my interest, not for reliability. 

I will add the assertion about the content. That should have been the first thing!

test/controllers/samples_controller_test.rb
gregdingle: Nice, I didn't know about `freeze_time` 😄

config/environments/prod.rb
gregdingle: I chose the default expires to be one month, on the assumption that anything worth seeing will be viewed within a month of precaching.

I realize you advocated for a persistent long term cache.

In any case, I think the best way to decide will be to look at the cache hit rate provided by the metrics in this PR.

davidrissato: Question: what are the conditions where this cache needs to be invalidated?

davidrissato: I feel like we have no reason to set an arbitrary number. The cache eviction should be either based on memory (ex: Redis should keep up to 1 GB of pre-cached files to avoid memory and performance issues) or would be persistent and rely on well known cache invalidation conditions.

app/models/pipeline_run.rb
davidrissato: However, I might be wrong, but it feels it is not correct to pre-cache ALL ready Backgrounds. Shouldn't we be caching only data that is related to this specific pipeline_run?

test/models/pipeline_run_test.rb
```diff
@@ -10,6 +10,10 @@ class PipelineRunTest < ActiveSupport::TestCase
   end
 
   test "duration is correct" do
-    assert_equal 0.0, pipeline_runs(:four).duration_hrs
+    assert pipeline_runs(:four).duration_hrs >= 0.0
+  end
+
+  test "precaching executes" do
+    pipeline_runs(:four).precache_report_info!
```
davidrissato: What is the assertion for this test?
Also, based on the method implementation, this method doesn't use any info related to the pipeline_run model, since it pre-caches all reports. Looks like it should be a class method instead or be changed to cache a specific pipeline_run.

cdebourcy:
@davidrissato : the background data is completely orthogonal to the pipeline run data. To make a report, the user could choose any pipeline run with any background model of their preference. There isn't necessarily one background that fits better than another

MarkAZhang:
You can change the `sort_by` param that's sent to the report_info endpoint by clicking on these arrows here (the contig_r, L, %id column headers). Not sure you can just use a default value for `sort_by` when passing to `external_report_info`. 

Screen Shot 2019-05-16 at 11 47 26 AM

davidrissato:
@gregdingle if the background data is completely orthogonal to the pipeline run data, that seems is a good reason to move that method out of the pipeline run model. My suggestion is moving it into a class method of `Background` model class, rename it to a more descriptive name, and remove the bang (`!`) from its name

app/controllers/samples_controller.rb
```diff
@@ -721,30 +722,40 @@ def samples_going_public
...
+      params.permit(report_info_params.keys)
+    )
 
-    # Label top-scoring hits for the executive summary
-    @report_info[:topScoringTaxa] = label_top_scoring_taxa!(@report_info[:taxonomy_details][2])
-    @report_info[:contig_taxid_list] = @pipeline_run.get_taxid_list_with_contigs
+    # This allows 304 Not Modified to be returned so that the client can use its
+    # local cache and avoid the large download.
+    httpdate = Time.at(report_info_params[:report_ts]).utc.httpdate
+    response.headers["Last-Modified"] = httpdate
```
MarkAZhang: Based on my understanding, you need to do a little more to get 304 Not Modified to work.
https://blog.appsignal.com/2018/05/01/client-side-caching-in-rails-conditional-get-requests.html

Did you get 304 Not Modified working? If not, maybe defer to a follow-up commit? (and remove the relevant from this one)

app/models/pipeline_run.rb
gregdingle: I chose to precache all background_id after all because Redis should act as a LRU cache when expires_in is enabled, so unused data will be evicted.

It's still not clear to me how the default background is selected and what cache hit rate would result. When @cdebourcy and I discussed this on the original PR, he recommended caching all backgrounds.

config/environments/prod.rb
gregdingle: By my reading, Redis will act as a LRU cache up to `maxmemory`, according to https://redis.io/topics/lru-cache , so long as we set any `expires_in`.

gregdingle: See https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/ParameterGroups.Redis.html#ParameterGroups.Redis.NodeSpecific also. The `maxmemory` currently is 6501171200 or 6.2GB.

gregdingle: > what are the conditions where this cache needs to be invalidated?

The conditions are all the cache key param values plus the sample ID plus the expires.

test/models/pipeline_run_test.rb
gregdingle: > this method doesn't use any info related to the pipeline_run model

It uses the `pipeline_version` and the `sample` and others. See `report_info_params` method.

gregdingle:
> You can change the sort_by param that's sent to the report_info endpoint by clicking on these column headers (the contig r column header is highlighted below). Not sure you can just use a default value for sort_by when passing to external_report_info.

Interesting. In any case, you can see in the `decode_sort_by` method that only three underscore strings are considered, which is not what appears in the browser URL on click. My working assumption is that rails only ever used the default sort. 

```
  def self.decode_sort_by(sort_by)
    return nil unless sort_by
    parts = sort_by.split "_"
    return nil unless parts.length == 3
```

app/controllers/samples_controller.rb
gregdingle: I did get the 304 working in my browser though I'm not sure if it is the most robust way.

app/models/pipeline_run.rb
davidrissato: In this logic, how do we know if Redis has already evicted the entry or not? I mean, I'm assuming this `ready: 1` is some sort of flag that represents if it is `ready` to be cached and it is turned to 0 when caching has been completed, is that correct? 
When redis evicts an entry, does it revert the flag back to `1`? Or are we reprocessing ALL backgrounds, every time when we invoke? This list seems to grow indefinitely, so there will be a day (maybe currently), where this query will return hundreds of thousands of lines, and we will be caching literally everything, defeating the purpose of an LRU cache.
Shouldn't we have a method to cache each Background individually? Shouldn't we invoke this method only when that data has been changed somehow? Shouldn't we do that only for things that are currently in use?

davidrissato: And I also believe we should move this method to `Background` model. It seems to be unrelated to `PipelineRun`.

config/environments/prod.rb
davidrissato: Nice to know that we have a `maxmemory`. So why would we need the `expires_in` clause? I see no value on it.

test/models/pipeline_run_test.rb
davidrissato: I miss an `assert` line in this test. If someone puts a `return true` in that method this test will pass. If we don't have an assertion, this test can generate false positives, and should be removed from the test suite.

config/environments/prod.rb
gregdingle: Discussed offline. I  will add an `expires_in` in the fetch.

test/models/pipeline_run_test.rb
gregdingle: Discussed offline. Will mock out.

app/models/pipeline_run.rb
davidrissato: Discussed offline. Greg clarified the relationship between a background and a pipeline_run.

PR #2239, opened 2019-05-17
[Locations] Update geosearch fields + add functions for writing new Location Metadata values

jshoe:
### Description
- This PR got a bit unwieldy but the crux of it is: User selects the location they want -> We pass down the locationiq_id/osm_id/osm_type -> At write time, we make another API call for just that 1 entry for extra security -> We make a Location entry to associate with the Metadatum.

- Align geosearch responses with the fields we decided on for the location schema for consistency.
- Add osm_type field since apparently osm_ids are numbered by type (Node, Way, or Relation).
- Refactor some of the previous location functions, add LOCATION_TYPE to metadatum.rb, and add some indexes.
- Since we handle metadata by generating values and then writing 'validated values' to db once the user submits/saves, we'll make the Location entries in the location validator as well.
- For the sake of not trusting user input, we want to make another call (just for the 1 location they selected) to get those details directly from the API service. We could also do some sort of caching to avoid this (but it'd have to cache all the options presented in the typeahead before they make their selection).
- This starts in the check_and_set_location_type function. We check if we already have the location details, or else fetch them and make a new Location entry. The Metadatum number_validated_value becomes the foreign key to the Location id.
- When we do batch writing with Metadatum.import, the check_and_set_location_type function will still be called. To speed up batch writing later we can reduce check calls to the db by memoizing location IDs and fields.
- Add a bunch of Location tests. I still used Minitest but will probably try to get us into Rspec.
- Up next: Migration for the new_collection_location field, hooking up the React components.

### Test and Reproduce
- See tests and generate similar cases manually.
- You could add new_collection_location locally with `MetadataField.new( name: "new_collection_location", display_name: "New collection Location", description: "Location sample was originally collected", base_type: 3, is_core: 1, is_default: 1, is_required: 1, group: "Sample", host_genomes: HostGenome.all)`


app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
```diff
@@ -17,10 +17,15 @@ class GeoSearchInputBox extends React.Component {
...
         name: locationsCategory,
-        // LiveSearchBox/Search tries to use 'title' as 'key'. Use title + i instead.
-        results: serverSideSuggestions.map((r, i) =>
-          Object.assign({}, r, { key: `${r.title}-${i}` })
-        )
+        // Format title/description for the text box.
+        results: serverSideSuggestions.map((r, i) => {
+          const nameParts = r.name.split(", ");
+          r.title = nameParts[0];
+          // LiveSearchBox/Search tries to use 'title' as 'key'. Use title + i instead.
```
tfrcarvalho: What problem are you trying to solve? 
Whats wrong with title in this case? Is it duplicate?

app/controllers/locations_controller.rb
```diff
@@ -9,19 +11,7 @@ def external_search
       success, resp = Location.geosearch(query)
       if success
         resp.each do |c|
```
tfrcarvalho: Maybe use `.map(&:method_name)`?

app/models/location.rb
```diff
@@ -1,16 +1,57 @@
...
-    raise "No API key for geosearch" unless ENV['LOCATION_IQ_API_KEY']
+  include LocationHelper
 
-    base_url = "https://us1.locationiq.com/v1/search.php?key=#{ENV['LOCATION_IQ_API_KEY']}&format=json&addressdetails=1&normalizecity=1"
-    query_url = "#{base_url}&q=#{query}"
+  # Base request to LocationIQ API
+  def self.location_api_request(endpoint_query)
+    raise "No location API key" unless ENV["LOCATION_IQ_API_KEY"]
+
+    query_url = "https://us1.locationiq.com/v1/#{endpoint_query}&key=#{ENV['LOCATION_IQ_API_KEY']}&format=json"
```
tfrcarvalho: Can you move this to a constant or at least the URL?

tfrcarvalho: Not sure if it is worth it...

app/models/location.rb
```diff
@@ -1,16 +1,57 @@
...
+    raise ArgumentError, "No query for geosearch" if query.blank?
+    endpoint_query = "search.php?addressdetails=1&normalizecity=1&q=#{query}"
+    location_api_request(endpoint_query)
+  end
+
+  # Create a Location from parameters
+  def self.create_from_params(location_params)
+    # Ignore fields that don't match columns
+    location_params = location_params.select { |x| Location.attribute_names.index(x.to_s) }
+    # Name sanitization is imperfect but we can have a sanity check
```
tfrcarvalho: What does this mean?

app/models/metadatum.rb
```diff
@@ -101,6 +102,21 @@ def check_and_set_date_type
...
+    # Based on our metadata structure, the location details selected by the user will end up in
+    # raw_value.
+    loc = JSON.parse(raw_value, symbolize_names: true)
+
+    # Set to existing Location or create a new one based on the external IDs. For the sake of not
+    # trusting user input, we'll potentially re-fetch location details based on the API and OSM IDs.
+    result = Location.find_or_create_by_api_ids(loc[:locationiq_id], loc[:osm_id], loc[:osm_type])
+
+    # Re-use number_validated_value column to be the foreign key to the Location table
+    self.number_validated_value = result.id
```
MarkAZhang: Could you create a new column `location_id` instead of re-using an old column?

Even though it creates an extra column, it keeps things clean. Otherwise, in the future, every time we do anything with number_validated_value, we have to remember the special case where location also re-uses that column.

app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
jshoe: Yeah, in the "short name" case there were a lot of duplicates (so changing the list would only remove one) but I'll change this to `loc-{locationiq_id}` instead

app/controllers/locations_controller.rb
jshoe: sure, will change

app/models/location.rb
jshoe: sure will change

app/models/location.rb
jshoe: @tfrcarvalho I felt compelled to have some kind of sanitization but maybe it's not necessary..

jshoe:
@tfrcarvalho re session cache: Now that Greg merged his Redis PR, I think I will familiarize myself with Redis and add it in a follow-up relatively soon

app/models/metadatum.rb
jshoe: sure will change

PR #2240, opened 2019-05-17
Rake task to copy samples to a new project

tfrcarvalho:
# Description

Copy samples between projects, based on CSV list.


lib/tasks/copy_samples.rake
```diff
@@ -0,0 +1,161 @@
+require "aws-sdk-s3"
+
+DEFAULT_S3_REGION = "us-west-2".freeze
+BUCKET_NAME = "idseq-samples-#{Rails.env}".freeze
+
+# Original CSV headers:
+# Source Project Name,	Source Sample Name,	Sample Name	Source,	Day, Time, Sample Type, Nucleotyde Type, Collection Date, Collection Location
```
jshoe: s/Nucleotyde/Nucleotide / the spacing around "Sample Name" is weird

jshoe:
looks pretty good / I don't see anything missing

app/models/pipeline_run.rb
```diff
@@ -295,21 +295,24 @@ def failed?
...
-    output_state_entries = []
-    target_outputs.each do |output|
-      output_state_entries << OutputState.new(
-        output: output,
-        state: STATUS_UNKNOWN
-      )
-    end
-    self.output_states = output_state_entries
+    # If we create a sample with results, we should not trigger a new run
+    unless results_finalized?
```
cdebourcy: I feel like this is the wrong place to make a change. It makes the code really confusing ("how could `results_finalized` be true if the pipeline_run is just getting created?"). Can you not move the logic for skipping the callbacks to your rake task instead? https://stackoverflow.com/questions/632742/how-can-i-avoid-running-activerecord-callbacks

lib/tasks/copy_samples.rake
```diff
@@ -0,0 +1,167 @@
+require "aws-sdk-s3"
+
+DEFAULT_S3_REGION = "us-west-2".freeze
+BUCKET_NAME = "idseq-samples-#{Rails.env}".freeze
```
cdebourcy: use `SAMPLES_BUCKET_NAME` which we already have?

app/models/pipeline_run.rb
tfrcarvalho: Thats a good suggestion for this particular case. I also think we should avoid callbacks

cdebourcy:
Fun fact: copying the S3 files is not enough to make results_folder work, because https://github.com/chanzuckerberg/idseq-web/blob/a5456d1b63389783aa48d050e343b0bbe514af73/app/models/pipeline_run.rb#L1353 uses the dag json as it was originally specified (i.e. with the old project id)

lib/tasks/copy_samples.rake
tfrcarvalho: Done

app/models/pipeline_run.rb
tfrcarvalho: > Fun fact: copying the S3 files is not enough to make results_folder work, because
> 
> https://github.com/chanzuckerberg/idseq-web/blob/a5456d1b63389783aa48d050e343b0bbe514af73/app/models/pipeline_run.rb#L1353
> 
> uses the dag json as it was originally specified (i.e. with the old project id)

Should this be dynamic from https://github.com/chanzuckerberg/idseq-web/blob/master/app/models/sample.rb#L75 ?

tfrcarvalho:
> Fun fact: copying the S3 files is not enough to make results_folder work, because
> 
> https://github.com/chanzuckerberg/idseq-web/blob/a5456d1b63389783aa48d050e343b0bbe514af73/app/models/pipeline_run.rb#L1353
> 
> uses the dag json as it was originally specified (i.e. with the old project id)

Should this be dynamic from https://github.com/chanzuckerberg/idseq-web/blob/master/app/models/sample.rb#L75 ?

cdebourcy:
> > Fun fact: copying the S3 files is not enough to make results_folder work, because
> > https://github.com/chanzuckerberg/idseq-web/blob/a5456d1b63389783aa48d050e343b0bbe514af73/app/models/pipeline_run.rb#L1353
> > 
> > uses the dag json as it was originally specified (i.e. with the old project id)
> 
> Should this be dynamic from https://github.com/chanzuckerberg/idseq-web/blob/master/app/models/sample.rb#L75 ?

Right now, results_folder looks for the S3 paths specified in pipeline_run_stage.dag_json which was generated when the pipeline_run was kicked off. Replacing these hardcoded S3 paths with dynamic ones based on the line you referenced would make it work correctly for the duplicated samples, yes.

app/models/pipeline_run.rb
tfrcarvalho: @cdebourcy Suggestion does not work since it is an old method.
There are suggestions that involve monkey patching `ActiveRecord::skip_callbacks` which seems bad practice.

cdebourcy:
@davidrissato We want to deep copy a bunch of records without triggering any of the relevant rails callbacks, do you know an easy way to do it?

PR #2247, opened 2019-05-21
[Locations] Finish backend/frontend adaptation for new location values

jshoe:
### Description
- Update Metadata components to display the location objects (generalized for if we have any other JSON metadata types in the future)
- Add GeoSearchInputBox to MetadataInput
- Modify GeoSearchInputBox to work with showing a value from parent props
- Finish check_and_set_location_type validation to associate with Locations table
- Add "data migration" for adding a "new_collection_location" MetadataField. Adds the field to the project "CI" for admins (can expand once we start releasing to users).
- Later: Updating component on the /samples/upload page to pop out of the inner modal t_t

### Demo
May-21-2019 11-09-44
May-21-2019 11-10-23
May-21-2019 11-14-35
May-21-2019 11-18-44
May-21-2019 11-19-31
May-21-2019 11-58-52

### Test and Reproduce
- Pull the branch, run migration, go to project CI, test out the flows above. You can verify the values in the console like Sample.find(10).metadata.last.location_id.

### DIFF
--- a/app/assets/src/components/common/DetailsSidebar/SampleDetailsMode/MetadataTab.jsx
+++ b/app/assets/src/components/common/DetailsSidebar/SampleDetailsMode/MetadataTab.jsx
@@ -120,8 +120,8 @@ class MetadataTab extends React.Component {
       <div className={cs.emptyValue}>--</div>
     ) : (
       <div className={cs.metadataValue}>
-        {/* If asked to display an object, try the name field (e.g. location object) */}
-        {isObject(val) ? val.name : val}
+        {/* If we want to display an object (e.g. location object), provide a 'name' field */}
+        {isObject(val) && val.name ? val.name : val}
       </div>
     );
   };
--- a/app/assets/src/components/common/DetailsSidebar/SampleDetailsMode/MetadataTab.jsx
+++ b/app/assets/src/components/common/DetailsSidebar/SampleDetailsMode/MetadataTab.jsx
@@ -120,8 +120,8 @@ class MetadataTab extends React.Component {
       <div className={cs.emptyValue}>--</div>
     ) : (
       <div className={cs.metadataValue}>
-        {/* If asked to display an arbitrary object, try the name field */}
-        {isObject(val) && val.name ? val.name : val}
+        {/* If asked to display an object, try the name field */}
+        {isObject(val) ? val.name : val}
       </div>
     );
   };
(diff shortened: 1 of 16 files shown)


app/assets/src/components/common/DetailsSidebar/SampleDetailsMode/MetadataTab.jsx
```diff
@@ -119,7 +119,10 @@ class MetadataTab extends React.Component {
     return val === undefined || val === null || val === "" ? (
       <div className={cs.emptyValue}>--</div>
     ) : (
-      <div className={cs.metadataValue}>{val}</div>
+      <div className={cs.metadataValue}>
+        {/* If asked to display an arbitrary object, try the name field */}
+        {isObject(val) && val.name ? val.name : val}
```
tfrcarvalho: Does it still make sense to display the object? In what circumstances would that happen?

jshoe: probably better not to try to display the object (you'll get `Objects are not valid as a React child`). I'll change to `isObject(val) ? val.name : val`

db/migrate/20190521175543_add_new_collection_location.rb
```diff
@@ -0,0 +1,19 @@
+class AddNewCollectionLocation < ActiveRecord::Migration[5.1]
+  # Data migration to add new intelligent collection location MetadataField
+
+  def up
+    existing = MetadataField.find_by(name: "new_collection_location")
+    unless existing
+      field = MetadataField.create(name: "new_collection_location", display_name: "New Collection Location", description: "Location sample was originally collected", base_type: 3, group: "Sample", host_genomes: HostGenome.all)
```
tfrcarvalho: What is the migraiton plan for the transition from collection_location to new_collection_location?

app/helpers/error_helper.rb
```diff
@@ -31,11 +31,16 @@ def self.missing_required_metadata(sample, missing_metadata_fields)
     end
   end
 
+  LOCATION_INVALID_ERROR = "Please input a location.".freeze
   DATE_INVALID_ERROR_HUMAN = "Please input a date in the format YYYY-MM or MM/YYYY. (for human samples)".freeze
   DATE_INVALID_ERROR = "Please input a date in the format YYYY-MM-DD, YYYY-MM, MM/DD/YY, or MM/YYYY.".freeze
   NUMBER_INVALID_ERROR = "Please input a number.".freeze
 
   def get_field_error(field, is_human = false)
+    if field.base_type == 3
```
tfrcarvalho: There are not constants to use instead of this magic numbers?

jshoe: fixed 👍

db/migrate/20190521175543_add_new_collection_location.rb
jshoe: added "Data Migration Plan" here https://czi.quip.com/gzWdAafIX8Yb/LocationMapping-Sprint-Plan

PR #2252, opened 2019-05-22
[Locations] Add new Location rendering adaptations to metadata display points

jshoe:
### Description
- Now new_collection_location will work in all the places we currently show metadata. Metadata sidebar, projects "Add Metadata" page, heatmap, phylo tree.
- Follows the pattern of using object.name for rendering location objects
- In PhyloTreeVis I noticed that metadata fields selection was broken because id was changed to sample_id :|. Fixed.
Screen Shot 2019-05-21 at 6 01 05 PM
- Special case location validated values on the metadata endpoints.
- Other small fixes
- Later: Plugging new_collection_location into data discovery filtering and search.

### Demo
Screen Shot 2019-05-21 at 6 15 27 PM
Screen Shot 2019-05-21 at 6 16 10 PM
Screen Shot 2019-05-21 at 5 45 00 PM
Screen Shot 2019-05-21 at 6 12 30 PM

### Test and Reproduce
- Pull the branch, visit the UI components described above, try inputting values/saving/refreshing to make sure the values persist.

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -624,8 +624,6 @@ class Sample < ApplicationRecord
       raise RecordNotFound("No matching field for #{key}") unless m.metadata_field
       m.key = m.metadata_field.name
     end
-    puts "foobar 3:49pm. val: ", val, m.raw_value
-    puts val.class.name
     if val.present? && m.raw_value != val
       m.raw_value = val.is_a?(ActionController::Parameters) ? val.to_json : val
       return {
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -733,7 +733,7 @@ class Sample < ApplicationRecord
       resp = m.attributes
       base_type = m.metadata_field.base_type
 
-      # Special-case for locations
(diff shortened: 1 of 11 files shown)


app/models/sample.rb
```diff
@@ -730,9 +730,15 @@ def private_until
...
   def metadata_with_base_type
     metadata.map do |m|
-      m.attributes.merge(
-        "base_type" => Metadatum.convert_type_to_string(m.metadata_field.base_type)
-      )
+      resp = m.attributes
+      base_type = m.metadata_field.base_type
+
+      # Special-case for locations (avoid extra call to #metadata_field.base_type in #validated_value)
+      if base_type == Metadatum::LOCATION_TYPE
```
tfrcarvalho: Could this mean a bunch of extra calls to location in a bulk operation?

jshoe: it looks like you'd get 1 more call to the Locations table per sample (assuming every sample has 1 Location). This could be batched at the place before metadata_with_base_type is called. Looks like only one call in report_helper#samples_taxons_details is in a batch context, and that one is still making a separate metadata call per sample

PR #2253, opened 2019-05-22
[MapPlayground] Switchover to new Locations to display

jshoe:
### Description
- Updates the MapPlayground to use the new Locations model. Now it will only display 'resolved' locations since parsing lat/lon name strings was a temporary demo.
- Now you can test out a whole flow by going to a sample in project "CI", adding a New Collection Location in the sidebar, picking a location, and then seeing it in /locations/map_playground !

### Demo
Screen Shot 2019-05-22 at 10 40 17 AM

### Test and Reproduce
- Start with map keys loaded, go to /locations/map_playground, see new_collection_locations displayed. Add locations as described above.

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -40,12 +40,13 @@ class LocationsController < ApplicationController
     end
 
     # Show all viewable locations in a demo format
-    field_id = MetadataField.find_by(name: "collection_location").id
+    field_id = MetadataField.find_by(name: "new_collection_location").id
     sample_info = current_power.samples
                                .includes(metadata: :metadata_field)
                                .where(metadata: { metadata_field_id: field_id })
-                               .pluck(:id, :name, :string_validated_value)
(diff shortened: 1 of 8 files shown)


app/controllers/locations_controller.rb
```diff
@@ -40,12 +40,13 @@ def map_playground
     end
 
     # Show all viewable locations in a demo format
-    field_id = MetadataField.find_by(name: "collection_location").id
+    field_id = MetadataField.find_by(name: "new_collection_location").id
```
gregdingle: Is `new_collection_location` a name we keeping long-term? It's a bad practice to name anything "new" because it isn't "new" for long and the old "collection_location" still sounds more authoritative.

jshoe: I made a note to rename to 'collection_location' after the old one is gone but I might as well change to collection_location_v2 now since people keep mentioning it ;) https://czi.quip.com/gzWdAafIX8Yb/LocationMapping-Sprint-Plan

jshoe: added 👍

PR #2262, opened 2019-05-24
Jira 761: Respect sample selection in DD bulk downloads

cdebourcy:
Works only for the new UI. The legacy UI (which includes Host Gene Counts download) is unchanged.

### DIFF
--- a/app/assets/src/components/views/samples/ReportsDownloader.js
+++ b/app/assets/src/components/views/samples/ReportsDownloader.js
@@ -49,11 +49,7 @@ export default class ReportsDownloader {
     }, 2000);
   }
 
-  generateReport = ({
-    makeAction,
-    statusAction,
-    retrieveAction
-  }) => {
+  generateReport = ({ makeAction, statusAction, retrieveAction }) => {
     // TODO(tiago): stop using cookies
     const backgroundId = Cookies.get("background_id");
     let queryString = [];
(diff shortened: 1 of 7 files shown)


app/assets/src/components/views/samples/ReportsDownloader.js
```diff
@@ -4,8 +4,14 @@ import { openUrl } from "~utils/links";
...
 export default class ReportsDownloader {
-  constructor({ projectId = "all", onDownloadFail = null, downloadOption }) {
+  constructor({
+    projectId = "all",
+    onDownloadFail = null,
+    downloadOption,
+    selectedSampleIds
+  }) {
     this.projectId = projectId;
+    this.selectedSampleIds = Array.from(selectedSampleIds);
```
tfrcarvalho: Should we impose a limit on the number of samples? 
The url may become too large.

cdebourcy: Good question. Do we currently have a limit on the sample selection for the heatmap? If not, I'd suggest punting on the limit for now

tfrcarvalho: My concern is that it can crash. But apparently, the limits are not as strict as they used to be it, for chrome in particular...
So lets leave as is for now.

cdebourcy: Ok sounds good. Let's see how it goes

PR #2264, opened 2019-05-28
[Maps] Add location_v2 loading to data discovery map display

jshoe:
### Description
- This completes the flow and add location_v2 loading to the map display.
- Matches the patterns for stats/dimensions and the left filtering sidebar still works as expected.
- Location endpoint returns the location info keyed by location_id for frontend lookups along with the list of sample_ids at that point. The main interaction (coming next) will be clicking a bubble -> seeing just those samples in the right sidebar.
- Most of the DiscoveryMap code was copied from MapPlayground.

### Demo
May-28-2019 10-06-44

### Test and Reproduce
- Load the branch with map keys, go to Data Discovery samples tab, click the map toggle, play around with the map and filters.
- See endpoint tests.

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -63,6 +63,7 @@ class LocationsController < ApplicationController
 
   # GET /locations/sample_locations.json
   # Get location data for a set of samples with filters
+  # TODO(jsheu): Consider consolidating if similar location data is loaded w/ data discovery tables.
   def sample_locations
     unless feature_access?
       render(json: {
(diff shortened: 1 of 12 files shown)


app/controllers/locations_controller.rb
```diff
@@ -60,6 +61,57 @@ def map_playground
...
+    # Get the samples
+    domain = params[:domain]
+    samples = samples_by_domain(domain) # access controlled
+    samples = filter_samples(samples, params)
+
+    # Get the relevant location_ids and sample_ids
+    field_id = MetadataField.find_by(name: "collection_location_v2").id
+    sample_info = samples
+                  .includes(metadata: :metadata_field)
+                  .where(metadata: { metadata_field_id: field_id })
```
tfrcarvalho: You can just add the previous query here, no?

PR #2268, opened 2019-05-29
Fix heatmap perf part 2

gregdingle:
# Description

This PR reduces the time to render the heatmap by about 40% on top of previous gains. In my test of a 49 sample project, time for the `samples_taxons` endpoint went down from 14s to 8s consistently. 

The improvement was achieved mostly by ranking and filtering in SQL instead of ruby. See `top_n_query`. 

Next, about 2 seconds was saved by converting two N+1 queries into a subquery to get the most recent pipeline run. 

Thirdly, the bottleneck SQL query was filtered more with `MINIMUM_ZSCORE_THRESHOLD` and `MINIMUM_READ_THRESHOLD`.

In the process, I refactored `HeatmapHelper` and `ReportHelper` a bunch. I moved all methods used only by `HeatmapHelper` into that class. There is still a lot of duplication of code in the two classes. I will leave that for another day. 

Other changes:
* I added a suite of unit tests to cover the public methods of `ReportHelper`. The tests cover different permutations of inputs but not all edge cases. 
* I removed the `aggregatescore` from the heatmap UI and code on recommendation by @cdebourcy .

# Notes

* **failed pipeline runs**: The original logic which I reproduced in SQL is to use the most recent run regardless of status. This doesn't seem right to me. 
* **missing pipeline runs**: I literally spent 3 hours debugging an issue which turned out to be caused by taxon counts referencing pipeline runs that did not exist. There appears to be 87 of such counts in prod. See https://modeanalytics.com/editor/chanzuckerberg/reports/46d9ac9db658 . It would be nice to enforce data integrity with either database constraints or garbage collecting scripts to avoid such head-scratching situations. 
* I **duplicated definitions** of user facing metrics such as zscore for performance reasons. This could easily lead to inconsistencies if not now then later which may be spotted by users. That said, there were **already** several duplicate definitions in the reports codebase, which is more troubling. 
* There are a bunch of **magic constants and hidden filtering** in the reports codebase without explanation. For example, phages are excluded by default without any indication in the UI.
* It seems like the **aggregate score** is no longer loved. If so we should kill it in the report page as well. 

# Test

`rails t` to cover basic param functionality

To cover metric definitions:

print the values of zscore and rpm, computed in ruby and in SQL
see the values are the same

To cover correct top N behavior: 

Load a large heatmap in dev branch
Load a large heatmap in master branch
See that the values on hover are the same
See that the grid of colors is very similar, differing only in equally ranked samples or taxons
Try all different metrics, different taxons per sample

To cover metric definitions:

Load a large heatmap in dev branch
print the values of zscore and rpm, computed in ruby and in SQL
see the values are the same

### DIFF
--- a/app/controllers/visualizations_controller.rb
+++ b/app/controllers/visualizations_controller.rb
@@ -136,7 +136,6 @@ class VisualizationsController < ApplicationController
         Viruses: ["Phage"]
       },
       metrics: [
-        { text: "Aggregate Score", value: "NT.aggregatescore" },
         { text: "NT Z Score", value: "NT.zscore" },
         { text: "NT rPM", value: "NT.rpm" },
         { text: "NT r (total reads)", value: "NT.r" },
@@ -149,7 +148,6 @@ class VisualizationsController < ApplicationController
       end,
       thresholdFilters: {
         targets: [
-          { text: "Aggregate Score", value: "NT_aggregatescore" },
           { text: "NT Z Score", value: "NT_zscore" },
           { text: "NT rPM", value: "NT_rpm" },
           { text: "NT r (total reads)", value: "NT_r" },
--- a/app/controllers/visualizations_controller.rb
+++ b/app/controllers/visualizations_controller.rb
@@ -136,12 +136,12 @@ class VisualizationsController < ApplicationController
         Viruses: ["Phage"]
       },
       metrics: [
-        { text: "NT Z Score", value: "NT.zscore" },
         { text: "NT rPM", value: "NT.rpm" },
+        { text: "NT Z Score", value: "NT.zscore" },
         { text: "NT r (total reads)", value: "NT.r" },
+        { text: "NR rPM", value: "NR.rpm" },
         { text: "NR Z Score", value: "NR.zscore" },
-        { text: "NR r (total reads)", value: "NR.r" },
-        { text: "NR rPM", value: "NR.rpm" }
+        { text: "NR r (total reads)", value: "NR.r" }
       ],
       backgrounds: current_power.backgrounds.map do |background|
         { name: background.name, value: background.id }
--- a/app/controllers/visualizations_controller.rb
+++ b/app/controllers/visualizations_controller.rb
@@ -126,9 +126,69 @@ class VisualizationsController < ApplicationController
     }, status: :internal_server_error
   end
 
+  # START OF HEATMAP METHODS
+
+  def heatmap
+    {
+      taxonLevels: %w[Genus Species],
+      categories: ReportHelper::ALL_CATEGORIES.pluck('name'),
+      subcategories: {
+        Viruses: ["Phage"]
+      },
+      metrics: [
+        { text: "Aggregate Score", value: "NT.aggregatescore" },
+        { text: "NT Z Score", value: "NT.zscore" },
+        { text: "NT rPM", value: "NT.rpm" },
+        { text: "NT r (total reads)", value: "NT.r" },
+        { text: "NR Z Score", value: "NR.zscore" },
+        { text: "NR r (total reads)", value: "NR.r" },
+        { text: "NR rPM", value: "NR.rpm" }
+      ],
+      backgrounds: current_power.backgrounds.map do |background|
+        { name: background.name, value: background.id }
+      end,
+      thresholdFilters: {
+        targets: [
+          { text: "Aggregate Score", value: "NT_aggregatescore" },
+          { text: "NT Z Score", value: "NT_zscore" },
+          { text: "NT rPM", value: "NT_rpm" },
+          { text: "NT r (total reads)", value: "NT_r" },
+          { text: "NT %id", value: "NT_percentidentity" },
+          { text: "NT L (alignment length in bp)", value: "NT_alignmentlength" },
+          { text: "NT log(1/e)", value: "NT_neglogevalue" },
+          { text: "NR Z Score", value: "NR_zscore" },
+          { text: "NR r (total reads)", value: "NR_r" },
+          { text: "NR rPM", value: "NR_rpm" },
+          { text: "NR %id", value: "NR_percentidentity" },
+          { text: "NR L (alignment length in bp)", value: "NR_alignmentlength" },
+          { text: "R log(1/e)", value: "NR_neglogevalue" }
+        ],
+        operators: [">=", "<="]
+      },
+      allowedFeatures: current_user.allowed_feature_list
+    }
+  end
+
+  def download_heatmap
+    @sample_taxons_dict = HeatmapHelper.sample_taxons_dict(params, samples_for_heatmap)
+    output_csv = generate_heatmap_csv(@sample_taxons_dict)
+    send_data output_csv, filename: 'heatmap.csv'
+  end
+
+  def samples_taxons
+    @sample_taxons_dict = HeatmapHelper.sample_taxons_dict(params, samples_for_heatmap)
+    render json: @sample_taxons_dict
+  end
+
   private
 
   def visualization_params
     params.permit(:domain, :type, :id, :url, :search, data: {})
+  end
+
+  def samples_for_heatmap
+    current_power.samples
+                 .where(id: params[:sampleIds])
+                 .includes([:host_genome, :pipeline_runs, metadata: [:metadata_field]])
   end
 end
--- a/app/helpers/heatmap_helper.rb
+++ b/app/helpers/heatmap_helper.rb
@@ -107,17 +107,19 @@ module HeatmapHelper
       sample_id = pr.sample_id
 
       tax_2d = ReportHelper.taxon_counts_cleanup(taxon_counts)
-      ReportHelper.only_species_or_genus_counts!(tax_2d, species_selected)
+      HeatmapHelper.only_species_or_genus_counts!(tax_2d, species_selected)
 
       rows = []
       tax_2d.each do |_tax_id, tax_info|
         rows << tax_info
       end
 
-      ReportHelper.compute_aggregate_scores_v2!(rows)
+      HeatmapHelper.compute_aggregate_scores_v2!(rows)
       rows = rows.select do |row|
         row["NT"]["maxzscore"] >= MINIMUM_ZSCORE_THRESHOLD &&
-          ReportHelper.check_custom_filters(row, threshold_filters)
+          # Note: these are applied *after* SQL filters, so results may not be
+          # 100% as expected .
+          HeatmapHelper.check_custom_filters(row, threshold_filters)
       end
 
       # Get the top N for each sample. This re-sorts on the same metric as in
@@ -151,8 +153,6 @@ module HeatmapHelper
     min_reads = MINIMUM_READ_THRESHOLD,
     sort_by = DEFAULT_TAXON_SORT_PARAM
   )
-    pipeline_run_ids = samples.map { |s| s.first_pipeline_run ? s.first_pipeline_run.id : nil }.compact
-
     categories_map = ReportHelper::CATEGORIES_TAXID_BY_NAME
     categories_clause = ""
     if categories.present?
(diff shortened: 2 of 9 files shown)


app/controllers/visualizations_controller.rb
```diff
@@ -126,9 +126,67 @@ def shorten_url
     }, status: :internal_server_error
   end
 
+  # START OF HEATMAP METHODS
+
+  def heatmap
```
gregdingle: I moved all action methods from the `HeatmapHelper` to the controller itself so that the helper could be a pure static class.

app/controllers/visualizations_controller.rb
```diff
@@ -126,9 +126,67 @@ def shorten_url
...
+  end
+
   private
 
   def visualization_params
     params.permit(:domain, :type, :id, :url, :search, data: {})
   end
+
+  def samples_for_heatmap
+    current_power.samples
```
tfrcarvalho: Since we are moving, this looks like a model kind of function.

app/helpers/heatmap_helper.rb
```diff
@@ -1,61 +1,25 @@
+# This is a class of static helper methods for generating data for the heatmap
+# visualization. See HeatmapHelperTest.
+# See selectedOptions in SamplesHeatmapView for client-side defaults, and
+# heatmap action in VisualizationsController.
 module HeatmapHelper
   DEFAULT_MAX_NUM_TAXONS = 30
+  # Zscore is best for heatmaps because it weighs the frequency against the background
+  DEFAULT_TAXON_SORT_PARAM = 'highest_nt_zscore'.freeze
```
tfrcarvalho: Where did you get this information?
Z-score does not differentiate much by itself.

tfrcarvalho: Is this new?

app/helpers/heatmap_helper.rb
```diff
@@ -71,25 +35,434 @@ def sample_taxons_dict(params)
...
+        taxon_counts.percent_concordant  AS  percentconcordant
+      FROM taxon_counts
+      LEFT OUTER JOIN taxon_summaries ON
+        #{background_id.to_i}   = taxon_summaries.background_id   AND
+        taxon_counts.count_type = taxon_summaries.count_type      AND
+        taxon_counts.tax_level  = taxon_summaries.tax_level       AND
+        taxon_counts.tax_id     = taxon_summaries.tax_id
+      WHERE
+        pipeline_run_id IN (
+          SELECT MAX(id)
```
tfrcarvalho: Max id is not necessarily the last run... This is something we need to fix in the entire app...eventually.

app/helpers/heatmap_helper.rb
```diff
@@ -71,25 +35,434 @@ def sample_taxons_dict(params)
...
+          a.*
+        FROM (
+          #{query}
+        ) a
+        ORDER BY
+          pipeline_run_id,
+          count_type = '#{sort[:count_type]}' DESC,
+          #{sort[:metric]} #{sort[:direction] == 'highest' ? 'DESC' : 'ASC'}
+      ) b
+      -- Overfetch by a factor of 4 to allow for post-SQL threshold filters
```
cdebourcy: Would appreciate more explanation here. Which filters are they? Why are they applied post-SQL? Is there a way to apply them in SQL? Is there a risk that the results are going to be "inexact"? Is there a way to make them exact?

app/controllers/visualizations_controller.rb
gregdingle: isn't `current_power` required to be in a controller because it uses session state?

app/helpers/heatmap_helper.rb
gregdingle: yes, this was my reasoning and it could be wrong -- thanks for asking. @cdebourcy told me to remove the aggregate score but he didn't say what should replace it as a default.

app/helpers/heatmap_helper.rb
```diff
@@ -1,61 +1,25 @@
...
+  # Zscore is best for heatmaps because it weighs the frequency against the background
+  DEFAULT_TAXON_SORT_PARAM = 'highest_nt_zscore'.freeze
+  READ_SPECIFICITY = true
+  MINIMUM_READ_THRESHOLD = 5
+  # Note: this is activated from the heatmap page by selecting "Viruses -
+  # Phages". The default categories are all BUT phages, though the UI does not
+  # indicate this.
+  INCLUDE_PHAGE = false
 
-  def heatmap
```
gregdingle: moved to controller

app/helpers/heatmap_helper.rb
```diff
@@ -1,61 +1,25 @@
...
-      allowedFeatures: current_user.allowed_feature_list
-    }
-  end
-
-  def download_heatmap
-    @sample_taxons_dict = sample_taxons_dict(params)
-    output_csv = generate_heatmap_csv(@sample_taxons_dict)
-    send_data output_csv, filename: 'heatmap.csv'
-  end
+  MINIMUM_ZSCORE_THRESHOLD = 1.7
```
gregdingle: moved from report helper

app/helpers/heatmap_helper.rb
```diff
@@ -1,61 +1,25 @@
...
-    @sample_taxons_dict = sample_taxons_dict(params)
-    render json: @sample_taxons_dict
-  end
+  def self.sample_taxons_dict(params, samples)
+    return {} if samples.empty?
 
-  def sample_taxons_dict(params)
-    sample_ids = (params[:sampleIds] || []).map(&:to_i)
     num_results = params[:taxonsPerSample] ? params[:taxonsPerSample].to_i : DEFAULT_MAX_NUM_TAXONS
+    min_reads = params[:minReads] ? params[:minReads].to_i : MINIMUM_READ_THRESHOLD
```
gregdingle: this is now dynamic just so I could have fewer fixtures in the unit test

app/helpers/heatmap_helper.rb
```diff
@@ -71,25 +35,434 @@ def sample_taxons_dict(params)
                           JSON.parse(params[:thresholdFilters] || "[]")
                         end
     subcategories = if params[:subcategories] && params[:subcategories].respond_to?(:to_h)
-                      params[:subcategories].permit!.to_h
+                      params[:subcategories].to_h
```
gregdingle: to make it a static method... not sure if it's important to always `permit`

app/helpers/heatmap_helper.rb
```diff
@@ -71,25 +35,434 @@ def sample_taxons_dict(params)
...
+    HeatmapHelper.samples_taxons_details(
+      samples,
+      taxon_ids,
+      background_id,
+      species_selected,
+      threshold_filters
+    )
+  end
+
+  def self.top_taxons_details(
```
gregdingle: moved from heatmap helper and functionally untouched internally

app/helpers/heatmap_helper.rb
```diff
@@ -71,25 +35,434 @@ def sample_taxons_dict(params)
...
+        taxon = if candidate_taxons[row["tax_id"]]
+                  candidate_taxons[row["tax_id"]]
+                else
+                  { "tax_id" => row["tax_id"], "samples" => {} }
+                end
+        taxon["max_aggregate_score"] = row[sort[:count_type]][sort[:metric]] if
+          taxon["max_aggregate_score"].to_f < row[sort[:count_type]][sort[:metric]].to_f
+        taxon["samples"][sample_id] = [count, row["tax_level"], row["NT"]["zscore"], row["NR"]["zscore"]]
+        candidate_taxons[row["tax_id"]] = taxon
+        break if count >= num_results
```
gregdingle: the original sorting and ranking behavior is preserved and applied on top of the SQL sort and rank so that results are guaranteed to be the same as before

app/helpers/heatmap_helper.rb
```diff
@@ -71,25 +35,434 @@ def sample_taxons_dict(params)
...
+        count += 1
+      end
+    end
+
+    candidate_taxons.values.sort_by { |taxon| -1.0 * taxon["max_aggregate_score"].to_f }
+  end
+
+  def self.fetch_top_taxons(
```
gregdingle: moved from report helper... original SQL query is slightly modified and wrapped in outer SQL query that sorts and ranks emulating logic that is in ruby... this way we avoid processing 100k+ rows in ruby

app/helpers/heatmap_helper.rb
gregdingle: will discuss in person

app/helpers/heatmap_helper.rb
gregdingle: good point ... should disucss in person

app/helpers/heatmap_helper.rb
gregdingle: After discussion, I decided to move the logic of the threshold filters into the SQL.

app/helpers/heatmap_helper.rb
```diff
@@ -71,25 +35,434 @@ def sample_taxons_dict(params)
...
+        row["zscore"] = z_max if row["zscore"] > z_max &&
+                                 row["zscore"] != ReportHelper::ZSCORE_WHEN_ABSENT_FROM_BACKGROUND
+        row["zscore"] = z_min if row["zscore"] < z_min
+        result_hash[pipeline_run_id]["taxon_counts"] << row
+      end
+    end
+    result_hash
+  end
+
+  def self.samples_taxons_details(
```
gregdingle: this method and the rest below were moved with no change in logic from report helper

app/controllers/visualizations_controller.rb
tfrcarvalho: Yeah, nevermind, I guess the key here for samples is the includes statement which is heatmap specific. 
I would just see the rest as something like `current_power.visualization.authorized_samples` which you would define through power controller. 

In any case, disregard this comment on the context of this PR.

app/helpers/heatmap_helper.rb
tfrcarvalho: I belive R and/or RPMs are the best since the absolute value of the difference is also relevant. The problem with Z is that it relies on an (often) invalid distribution and loses precision quickly (lots of entries close max value 100)

app/helpers/heatmap_helper.rb
tfrcarvalho: One of the reasons we are moving to factories... need to get my PR out soon.

app/helpers/heatmap_helper.rb
```diff
@@ -71,25 +35,434 @@ def sample_taxons_dict(params)
...
+    count_type = sort[:count_type]
+    metric = sort[:metric]
+    candidate_taxons = {}
+    results_by_pr.each do |_pr_id, res|
+      pr = res["pr"]
+      taxon_counts = res["taxon_counts"]
+      sample_id = pr.sample_id
+
+      tax_2d = ReportHelper.taxon_counts_cleanup(taxon_counts)
+      HeatmapHelper.only_species_or_genus_counts!(tax_2d, species_selected)
```
tfrcarvalho: Cant this be a filter on SQL too?

gregdingle:
@cdebourcy @tfrcarvalho 

I spent another 1.5 hours today trying to convert the custom threshold filters and all the remaining ruby filtering to SQL. However, I gave up because it was too hard to get exactly the same results in all cases. The key difference is that the ruby filters are applied at the level of taxons but the inner SQL query is applied on counts. 

The good news is that custom filtering is a small percent of queries and we could easily route to the old code path, which I implemented in the lastest commit, so we get nearly all the performance benefit at the cost of code complexity. 

Before working any further on heatmaps, I believe we need to refactor the code a lot more and write tests for each part, assuming we want to maintain the current behavior exactly.

tfrcarvalho:
I am ok with moving on with this change. I understand some refactors are too long for the current scope.

PR #2269, opened 2019-05-30
[Locations] Automatically restrict overly-specific locations to the state-level + map attribution compaction

jshoe:
### Description
- First change was an opportunistic tweak to display map attribution tags in a compact mode by default to make the map look more clean (still shows attribution on hover).
- Add restrictions for locations for Human. If they enter a location below the state-level, automatically do a search for just the (State, Country) and save that instead.

- TODO: Adding nice UI warnings when this occurs + auto-refreshing the values. Alternatively I am leaning towards having a special-case Human location input component that only accepts and geosearches for "State, Country".

### Test and Reproduce
- Start with geosearch keys, go to a sample, go to metadata, type in a location like "UCSF", save, refresh and see the value was changed to "California, USA".
- Can also call the flow manually in Rails console.
- See tests.

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -23,6 +23,11 @@ class Location < ApplicationRecord
     location_api_request(endpoint_query)
   end
 
+  def self.geosearch_by_country_and_state(country_name, state_name)
+    endpoint_query = "search.php?addressdetails=1&normalizecity=1&country=#{country_name}&state=#{state_name}"
+    location_api_request(endpoint_query)
+  end
+
   # Create a Location from parameters
   def self.create_from_params(location_params)
     # Ignore fields that don't match columns
@@ -54,6 +59,19 @@ class Location < ApplicationRecord
 
       resp = LocationHelper.adapt_location_iq_response(resp)
       create_from_params(resp)
+    end
+  end
+
+  def self.check_and_restrict_specificity(location, host_genome_name)
+    if host_genome_name == "Human"
+      if location.subdivision_name.present? || location.city_name.present?
+        # Redo the search for just the state/country and pick the first result (should be unique)
+        success, results = self.geosearch("#{location.state_name}, #{location.country_name}")
+        if success
+          results = resp.map { |r| LocationHelper.adapt_location_iq_response(r) }
+        end
+        puts "foobar 11:34am", results
+      end
     end
   end
 end
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -30,15 +30,15 @@ class Location < ApplicationRecord
     location_api_request(endpoint_query)
   end
 
-  # Create a Location from parameters
-  def self.create_from_params(location_params)
+  # Instantiate a new Location from parameters WITHOUT saving
+  def self.new_from_params(location_params)
     # Ignore fields that don't match columns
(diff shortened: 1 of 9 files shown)


app/models/location.rb
```diff
@@ -56,4 +63,27 @@ def self.find_or_create_by_api_ids(locationiq_id, osm_id, osm_type)
...
+    # We don't want Human locations with subdivision or city
+    if host_genome_name == "Human" && (location.subdivision_name.present? || location.city_name.present?)
+      # Redo the search for just the country/state
+      success, resp = geosearch_by_country_and_state(location.country_name, location.state_name)
+      unless success
+        raise "Couldn't find #{location.state_name}, #{location.country_name} (state, country)"
+      end
+
+      result = LocationHelper.adapt_location_iq_response(resp[0])
+      existing = Location.find_by(locationiq_id: result[:locationiq_id])
```
tfrcarvalho: there is an `exists?` function if thats all you want...

app/models/location.rb
```diff
@@ -56,4 +63,27 @@ def self.find_or_create_by_api_ids(locationiq_id, osm_id, osm_type)
...
+  def self.check_and_restrict_specificity(location, host_genome_name)
+    # We don't want Human locations with subdivision or city
+    if host_genome_name == "Human" && (location.subdivision_name.present? || location.city_name.present?)
+      # Redo the search for just the country/state
+      success, resp = geosearch_by_country_and_state(location.country_name, location.state_name)
+      unless success
+        raise "Couldn't find #{location.state_name}, #{location.country_name} (state, country)"
+      end
+
+      result = LocationHelper.adapt_location_iq_response(resp[0])
```
tfrcarvalho: Cant an empty `[]` response be a success?

app/models/location.rb
```diff
@@ -56,4 +63,27 @@ def self.find_or_create_by_api_ids(locationiq_id, osm_id, osm_type)
...
+      existing = Location.find_by(locationiq_id: result[:locationiq_id])
+      new_location = existing ? existing : create_from_params(result)
+
+      # Delete the overly specific location if not used anymore
+      location.delete if Metadatum.where(location: location).empty?
```
tfrcarvalho: Why is this the case? We are temporarily saving it before?

cdebourcy: I agree it would be better to never save it to the database in the first place. (BTW, deleted rows are still retained in fivetran)

tfrcarvalho: @cdebourcy how does that work? Only if they deleted after the fivetran sync, right?

cdebourcy: Yes -- the point is, it's possible. As you allude to, it also just seems bad form to create a record only to delete it later (there is actually another metadata endpoint that does something similar, which really bugs me as well...)

tfrcarvalho: Thanks, just wanted to understand we are talking about the same case.
Which endpoint is that?

app/models/location.rb
jshoe: good catch 👍

app/models/location.rb
jshoe: refactored 'create_from_params' to just 'new_from_params' to avoid this 👍

PR #2270, opened 2019-05-30
Fix heatmap perf part 3: caching

gregdingle:
# Description

This adds caching to the slow ajax endpoint of the heatmap page, reducing the render time of the heatmap from ~10s to ~2s for 50 samples. Note, it only works after the first load. 

The cache is invalidated on the change of any filters, which change the URL. I added a special `heatmapTs` param similar to the `reportTs` param to bust the cache when a new pipeline run occurs or is updated. 

# Test

open a heatmap 
see the speedup
see the `heatmapTs` in the URL


gregdingle:
Yes and yes.  There’s not much benefit to caching a sunset because we now push most Params down to sql.

PR #2271, opened 2019-05-31
Handle insufficient inputs gracefully

cdebourcy:
Detects invalid_step_input.json produced by idseq-dag. If a known user error is detected, no pipeline alert is sent and the sample is displayed as "complete with issues" rather than "failed". The logic is all wired -- but the frontend for this is bare bones: I'm relying on @MarkAZhang to implement the actual design when he comes back.

### DIFF
--- a/app/assets/src/components/views/SampleView/SampleView.jsx
+++ b/app/assets/src/components/views/SampleView/SampleView.jsx
@@ -325,17 +325,22 @@ class SampleView extends React.Component {
           <div className={cs.failedContainer}>
             {this.props.pipelineRun.known_user_error === "FAULTY_INPUT" ? (
               <div>
-                Sorry, something was wrong with your file.{" "}
-                {this.props.pipelineRun.error_message}
+                Sorry, something was wrong with your file ({
+                  this.props.pipelineRun.error_message
+                }). Please check the format and reupload your file.
               </div>
             ) : this.props.pipelineRun.known_user_error ===
             "INSUFFICIENT_READS" ? (
               <div>
                 Oh no! No matches were identified because there weren't any
-                reads left after filtering.
+                reads left after filtering. Please check the quality of your
+                samples.
               </div>
             ) : (
-              <h6 className={cs.failed}>Oh no! The sample run failed.</h6>
+              <h6 className={cs.failed}>
+                Oh no! The sample run failed. Please contact us for help
+                rerunning your sample.
+              </h6>
             )}
           </div>
         );
--- a/app/assets/src/components/views/SampleView/SampleView.jsx
+++ b/app/assets/src/components/views/SampleView/SampleView.jsx
@@ -335,7 +335,7 @@ class SampleView extends React.Component {
                 reads left after filtering.
               </div>
             ) : (
-              <h6 className={cs.failed}>Sample run failed</h6>
+              <h6 className={cs.failed}>Oh no! The sample run failed.</h6>
             )}
           </div>
         );
--- a/app/assets/src/components/views/SampleView/SampleView.jsx
+++ b/app/assets/src/components/views/SampleView/SampleView.jsx
@@ -323,10 +323,20 @@ class SampleView extends React.Component {
       } else {
         return (
           <div className={cs.failedContainer}>
-            <h6 className={cs.failed}>Sample run failed</h6>
-            <h6 className={cs.failed}>
-              {this.props.pipelineRun.error_message}
-            </h6>
+            {this.props.pipelineRun.known_user_error === "FAULTY_INPUT" ? (
+              <div>
+                Sorry, something was wrong with your file.{" "}
+                {this.props.pipelineRun.error_message}
+              </div>
+            ) : this.props.pipelineRun.known_user_error ===
+            "INSUFFICIENT_READS" ? (
+              <div>
+                Oh no! No matches were identified because there weren't any
+                reads left after filtering.
+              </div>
+            ) : (
+              <h6 className={cs.failed}>Sample run failed</h6>
+            )}
           </div>
         );
       }
--- a/app/helpers/pipeline_runs_helper.rb
+++ b/app/helpers/pipeline_runs_helper.rb
@@ -181,4 +181,25 @@ module PipelineRunsHelper
   def pipeline_version_at_least_2(pipeline_version)
     pipeline_version_at_least(pipeline_version, PIPELINE_VERSION_2)
   end
+
+  def check_for_user_error(failed_stage)
+    result = [nil, nil]
+    if failed_stage.step_number == 1
+      user_input_validation_file = "#{failed_stage.pipeline_run.host_filter_output_s3_path}/#{PipelineRun::INPUT_VALIDATION_NAME}"
+      invalid_step_input_file = "#{failed_stage.pipeline_run.host_filter_output_s3_path}/#{PipelineRun::INVALID_STEP_NAME}"
+      if file_generated_since_run(failed_stage, user_input_validation_file)
+        file = Tempfile.new
+        downloaded = PipelineRun.download_file_with_retries(user_input_validation_file, file.path, 3, false)
+        validation_error = downloaded ? JSON.parse(File.read(file))["Validation error"] : nil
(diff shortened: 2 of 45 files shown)


db/schema.rb
```diff
@@ -71,7 +71,7 @@
     t.index ["sample_id"], name: "index_backgrounds_samples_on_sample_id"
   end
 
-  create_table "contigs", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
+  create_table "contigs", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=latin1" do |t|
```
cdebourcy: not sure why this happened again, I will fix before merging

app/helpers/pipeline_runs_helper.rb
```diff
@@ -181,4 +181,25 @@ def pipeline_version_has_coverage_viz(pipeline_version)
...
+        file = Tempfile.new
+        downloaded = PipelineRun.download_file_with_retries(user_input_validation_file, file.path, 3, false)
+        validation_error = downloaded ? JSON.parse(File.read(file))["Validation error"] : nil
+        file.unlink
+        return ["FAULTY_INPUT", validation_error] if validation_error
+      end
+      if file_generated_since_run(failed_stage, invalid_step_input_file)
+        # Currently 'invalid_step_input_file' only gets produced in a single failure mode:
+        # INSUFFICIENT_READS. If we extended the file to other error types in the future,
+        # we would download 'invalid_step_input_file' and parse the failure mode here.
```
gregdingle: why not future proof it now by downloading the file and parse it as above with `user_input_validation_file`?

and-or it would be good to double check that the output read count is actually zero for the previous pipeline step

app/models/pipeline_run.rb
```diff
@@ -853,8 +843,11 @@ def update_job_status
       if prs.failed?
         self.job_status = STATUS_FAILED
         self.finalized = 1
-        message = "SampleFailedEvent: Sample #{sample.id} by #{sample.user.email} failed #{prs.name} with #{adjusted_remaining_reads || 0} reads remaining after #{duration_hrs} hours. See: #{status_url}"
-        LogUtil.log_err_and_airbrake(message)
+        self.known_user_error, self.error_message = check_for_user_error(prs)
+        log_message = "SampleFailedEvent: Sample #{sample.id} by #{sample.user.admin? ? 'admin user' : 'non-admin user'} " \
+          "failed #{prs.name} with #{adjusted_remaining_reads || 0} reads remaining after #{duration_hrs} hours. " \
+          "See: #{status_url}"
+        LogUtil.log_err_and_airbrake(log_message) unless known_user_error
```
gregdingle: we should count the `known_user_error` somehow even if not showing as a pipeline failure

cdebourcy:
> It's really nice to see this change!
> 
> I suggested some improvements but the code looks like it should work.
> 
> How are you planning to test it ?

I tested the different failure/success modes locally by uploading samples with appropriate inputs (one that runs out of reads in the middle of host filtering, and one that just succeeds).

app/models/pipeline_run.rb
cdebourcy: Good idea, how/where would you like to be alerted to this? My thinking was that we're trying to denoise alerts. Anyone who is interested could still count known_user_errors in Mode or see them in the sample table as "complete - issue"

app/helpers/pipeline_runs_helper.rb
cdebourcy: The idseq-dag change does the read counting: https://github.com/chanzuckerberg/idseq-dag/commit/861e4d9f1382315ae16971c6985b31f08feca501
No need to revalidate here in my opinion.
Yes, I will parse `user_input_validation_file` then :)

cdebourcy: ok done

app/helpers/pipeline_runs_helper.rb
```diff
@@ -181,4 +181,25 @@ def pipeline_version_has_coverage_viz(pipeline_version)
...
   end
+
+  def check_for_user_error(failed_stage)
+    if failed_stage.step_number == 1
+      user_input_validation_file = "#{failed_stage.pipeline_run.host_filter_output_s3_path}/#{PipelineRun::INPUT_VALIDATION_NAME}"
+      invalid_step_input_file = "#{failed_stage.pipeline_run.host_filter_output_s3_path}/#{PipelineRun::INVALID_STEP_NAME}"
+      if file_generated_since_run(failed_stage, user_input_validation_file)
+        file = Tempfile.new
+        downloaded = PipelineRun.download_file_with_retries(user_input_validation_file, file.path, 3, false)
+        validation_error = downloaded ? JSON.parse(File.read(file))["Validation error"] : nil
```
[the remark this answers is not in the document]
cdebourcy: done, links added

app/helpers/pipeline_runs_helper.rb
```diff
@@ -181,4 +181,25 @@ def pipeline_version_has_coverage_viz(pipeline_version)
   def pipeline_version_at_least_2(pipeline_version)
     pipeline_version_at_least(pipeline_version, PIPELINE_VERSION_2)
   end
+
+  def check_for_user_error(failed_stage)
+    if failed_stage.step_number == 1
```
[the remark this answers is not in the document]
cdebourcy: ok done. Although I was thinking in the future it would probably be more like
```
if failed_stage.step_number == 1
  ...
elsif failed_stage.step_number == 2
  ...
```
etc

app/models/pipeline_run.rb
davidrissato: I’m coding the restart logic of our watchdog and I’m about to touch this piece of code. I’m planning to move this message to a different place, which aligns to Charles’ idea of “denoising” alerts (I can carry Greg’s logic with my change).
I’m planning to just log an error message here without sending it to air brake, and enqueue in resque an asynchronous pipeline restart in case this is the first error for this sample. I’ll only trigger air brake if this is not the first failure.

app/helpers/pipeline_runs_helper.rb
```diff
@@ -181,4 +181,25 @@ def pipeline_version_has_coverage_viz(pipeline_version)
...
+      user_input_validation_file = "#{failed_stage.pipeline_run.host_filter_output_s3_path}/#{PipelineRun::INPUT_VALIDATION_NAME}"
+      invalid_step_input_file = "#{failed_stage.pipeline_run.host_filter_output_s3_path}/#{PipelineRun::INVALID_STEP_NAME}"
+      if file_generated_since_run(failed_stage, user_input_validation_file)
+        file = Tempfile.new
+        downloaded = PipelineRun.download_file_with_retries(user_input_validation_file, file.path, 3, false)
+        validation_error = downloaded ? JSON.parse(File.read(file))["Validation error"] : nil
+        file.unlink
+        return ["FAULTY_INPUT", validation_error] if validation_error
```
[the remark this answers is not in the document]
cdebourcy: hmmmm. It seems a bit long and non-specific. 
Right now the 2 codes are FAULTY_INPUT and INSUFFICIENT_READS. I'm worried INPUT_VALIDATION_ERROR might sound like it would also encompass INSUFFICIENT_READS. I think I'll leave it is as for now but we can always change in the future

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -323,10 +323,25 @@ class SampleView extends React.Component {
...
+            ) : this.props.pipelineRun.known_user_error ===
+            "INSUFFICIENT_READS" ? (
+              <div>
+                Oh no! No matches were identified because there weren't any
+                reads left after filtering. Please check the quality of your
+                samples.
+              </div>
+            ) : (
+              <h6 className={cs.failed}>
+                Oh no! The sample run failed. Please contact us for help
```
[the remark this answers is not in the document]
cdebourcy: I agree. But Jenn's mocks have an actual link there I think. I'll leave it to the person that's implementing the mocks

PR #2273, opened 2019-05-31
Fix heatmap perf 4: genus query

gregdingle:
# Description

I noticed that a second large query is only needed for genus counts, but it was also executed for the default of species counts. This PR makes it only execute when needed.

# Note

I refactored some of the methods to make the call chain more flat and easier to follow. As a consequence I had to change a bunch of tests. 

I noticed that family counts are also being fetched in `ReportHelper.fetch_parent_ids` from the DB then discarded later in `fetch_samples_taxons_counts`. I'm not sure why family counts would ever be fetched, and it doesn't have much perf impact, so I left it alone until someone can confirm the correct behavior. 

# Test

`rails t`

Run a set of samples in species and genus modes
Check that heatmaps are the same as before

### DIFF
--- a/app/helpers/heatmap_helper.rb
+++ b/app/helpers/heatmap_helper.rb
@@ -48,43 +48,6 @@ module HeatmapHelper
     first_sample = samples.first
     background_id = params[:background] ? params[:background].to_i : get_background_id(first_sample)
 
-    taxon_ids = HeatmapHelper.top_taxons_details(
-      samples,
-      background_id,
-      num_results,
-      sort_by,
-      species_selected,
-      categories,
-      threshold_filters,
-      read_specificity,
-      include_phage,
-      min_reads
-    ).pluck('tax_id')
-
-    taxon_ids -= removed_taxon_ids
-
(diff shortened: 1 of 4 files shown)


app/helpers/heatmap_helper.rb
```diff
@@ -48,22 +48,35 @@ def self.sample_taxons_dict(params, samples)
...
+      sort_by,
+      species_selected,
+      threshold_filters
+    )
+
+    taxon_ids = details.pluck('tax_id')
     taxon_ids -= removed_taxon_ids
 
+    # Refetch at genus level using species level
+    parent_ids = species_selected ? [] : ReportHelper.fetch_parent_ids(taxon_ids, samples)
```
gregdingle: this is the key optimization

jshoe:
I think you're correct that family counts are not used (legacy plans to add 3rd level to report table)

cdebourcy:
I think family counts are not necessary but having the "parent" might be necessary for species-level organisms, to show in the taxon sidebar: 
image

PR #2274, opened 2019-05-31
Precache with less strict update time

gregdingle:
# Description

By debugging, I found that the update time in the cache key set by precahing was 15 min different from the update time on sample report view. This change truncates to the nearest day so there is little chance of mismatch. 

Any new pipeline runs are assumed to be run with a new pipeline version. The `report_ts` covers exceptional situations. 

# Note

I don't know why a pipeline run would be updated after this: 

```
    # Check if run is complete:
    if all_output_states_terminal?

      if all_output_states_loaded? && !compiling_stats_failed
        update(results_finalized: FINALIZED_SUCCESS)

        # Precache reports for all backgrounds
        Resque.enqueue(PrecacheReportInfo, id)
```

Any ideas? It is important to know if any such update would affect the report counts. 

# Test

load report page
see no errors

check datadog in prod for cache hits after preaching

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,8 +5,8 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
-gem 'airbrake', '~> 9.2.2'
-gem 'airbrake-ruby', '~> 4.4'
+gem 'airbrake', '~> 8'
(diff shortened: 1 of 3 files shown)


jshoe:
I don't see a reason the pipeline should update after all outputs were loaded and it was marked finalized

PR #2276, opened 2019-06-03
Add tests to projects_controller#index using RSpec and FactoryBot

tfrcarvalho:
## Description

* Adds multiple integration tests to `projects_controller#index`. 
* This is the first step in a refactor of that root to move most logic to the model. Thus, the integration tests might be a more thorough then necessary if we move more logic to the model.
* Added multiple factories as necessary to develop the tests.

## Testing

Code changes were driven by the integration tests added.
Integration tests should be self-explanatory. Please add a note for any test that is not clear.


spec/controllers/projects_controller_spec.rb
```diff
@@ -1,14 +1,460 @@
...
+      it "sees all projects" do
+        expected_projects = [
+          create(:project, users: [@admin]),
+          create(:project, users: [@joe]),
+          create(:project, users: [@joe, @admin]),
+          create(:public_project),
+          create(:project, samples_data: [{ created_at: 1.year.ago }]),
+          create(:project, users: [@admin], samples_data: [{ created_at: 1.year.ago }])
+        ]
+        expected_projects.reverse!
```
jshoe: what are the `reverse!` for?

tfrcarvalho: By default, projects controller sorts projects by id in decreasing order. Probably I should change the expectation not to test order...
Do you have any other suggestions?

app/controllers/application_controller.rb
```diff
@@ -109,4 +109,16 @@ def set_current_context_for_logging!
   rescue => e
     Rails.logger.error(e)
   end
+
+  def sanitize_order_by(model, order_by, default = nil)
```
gregdingle: how many of our controllers would ever use this? I'm not sure if this is the best way to share code for a minority of subclasses

app/controllers/projects_controller.rb
```diff
@@ -109,7 +115,7 @@ def index
           (tissues_by_project_id[s.project_id] ||= Set.new) << metadata[s.id][:sample_type] if (metadata[s.id] || {})[:sample_type]
           (locations_by_project_id[s.project_id] ||= Set.new) << metadata[s.id][:collection_location] if (metadata[s.id] || {})[:collection_location]
           # Assumes project owner is the uploader of the project's first sample
-          if !min_sample_by_project_id[s.project_id] || min_sample_by_project_id[s.project_id] < s.id
+          if !min_sample_by_project_id[s.project_id] || min_sample_by_project_id[s.project_id] > s.id
```
gregdingle: this seems like a pretty big fix!

app/helpers/samples_helper.rb
```diff
@@ -545,7 +545,7 @@ def samples_by_metadata_field(sample_ids, field_name)
   private
 
   def filter_by_time(samples, start_date, end_date)
-    samples.where("samples.created_at >= ? AND samples.created_at <= ?", start_date, end_date)
+    samples.where("samples.created_at >= ? AND samples.created_at <= ?", start_date.beginning_of_day, end_date.end_of_day)
```
gregdingle: by the way, how we do treat time in idseq? do we normalize it to PST or UTC? do we try to localize it for the user?

tfrcarvalho:
> I'm only seeing the factories, not the tests. What am I missing?

Probably you need to expand the file. Collapsed by default for having too many changes.

davidrissato:
I'm eager to see this PR merged. That's a really nice addition!

app/controllers/projects_controller.rb
tfrcarvalho: Indeed. The impact is not that visible because many projects have the same uploader.

spec/controllers/projects_controller_spec.rb
tfrcarvalho: Removed the expectations since they not core for the tests.

app/controllers/application_controller.rb
tfrcarvalho: Moved it to a concern.

app/helpers/samples_helper.rb
tfrcarvalho: Not sure, but I believe it gets localized on view.

spec/factories/projects.rb
```diff
@@ -0,0 +1,36 @@
...
+      host_genome_name { nil }
+      # Array of samples entries to create.
+      # The hash elements will be passed on to sample factory as keyword arguments.
+      samples_data { [] }
+    end
+
+    sequence(:name) { |n| "Project #{n}" }
+    public_access { 0 }
+
+    trait :with_samples do
```
davidrissato: I have a suggestion here: shouldn't we call this traits `with_sample` instead of `with_samples`, since it returns only one sample? We can apply same logic to `with_public_samples`

PR #2281, opened 2019-06-03
Skip test

gregdingle:
# Description

See comment in code

# Test

`bin/rails test test/models/pipeline_run_test.rb:16`


test/models/pipeline_run_test.rb
```diff
@@ -14,6 +14,8 @@ class PipelineRunTest < ActiveSupport::TestCase
   end
 
   test "precaching executes report_info_json" do
```
davidrissato: What if we remove the test instead of skipping it? The test isn't that long, and can be reworked later on.

gregdingle: done

PR #2284, opened 2019-06-04
Add guard configuration

davidrissato:
Guard is a command line tool to easily handle events on file system modifications. It is useful when using the TDD approach. (https://guardgem.org/).

Essentially this is a tool that watches your folders and automatically triggers related test every time a file is changed. Example: whenever you save `app/models/pipeline_run.rb`, it will run rspec with file `spec/models/pipeline_run_spec.rb`.

Usage is as easy as running the following command in a separate window:
`docker-compose run --rm web guard`.


gregdingle:
I've never used an auto-rerun test tool, and I'm not sure if I'd use it, but if you like it, I don't see much downside.

PR #2285, opened 2019-06-04
Add endpoint for pipeline run stage results

emilyzhong:
# Add endpoint for pipeline run stage results

For the Pipeline Monitoring feature
* Add `stage_results` route to `samples_controller`, which takes the most recent pipeline run for the sample and returns a JSON combining the `dag_json` field of each pipeline run stage. Should include all four stages for admin users and only three stages for non-admin users (no experimental stage)

## Type of change

# Test and Reproduce
Visit `samples/[sample_id]/stage_results` to fetch json

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -973,7 +973,7 @@ class SamplesController < ApplicationController
       if stage.name == "Experimental" && !current_user.admin?
         next
       end
-      @results[stage.name] = stage.dag_json
+      @results[stage.name] = JSON.parse stage.dag_json
     end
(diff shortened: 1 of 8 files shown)


jshoe:
Screen Shot 2019-06-04 at 2 19 50 PM

When I go to the endpoint it looks like it's returning the dag JSON all as a string, but you want to return the parsed JSON, right?

jshoe:
@emilyzhong @cdebourcy points out that you should add a test that users can't access the endpoint if they don't have access to the sample

emilyzhong:
Working on changes in https://github.com/chanzuckerberg/idseq-web/pull/2287

PR #2286, opened 2019-06-05
Fix filtering projects by id [IDseq-984] 

tfrcarvalho:
# Description

Filtering by id was broken for all data, where multiple projects were returned if there were empty projects. 
Jira ticket: https://jira.czi.team/browse/IDSEQ-984

# Testing

Added integration tests.
To verify the bug was fix: 
1. Go to All Data page
2. Click any project with samples except for the first one and memorize the project name
3. Click a sample to go to report page
4. Click the project link above the sample title
5. Verify that the project name is correct (matches memorized project)

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -85,7 +85,7 @@ class ProjectsController < ApplicationController
                      current_power.projects
                    end
 
-        projects = projects.where(id: params[:projectId]) if project_id
+        projects = projects.where(id: project_id) if project_id
         projects = projects.db_search(search) if search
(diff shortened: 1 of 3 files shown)


app/controllers/projects_controller.rb
```diff
@@ -84,6 +85,7 @@ def index
                      current_power.projects
                    end
 
+        projects = projects.where(id: params[:projectId]) if project_id
```
jshoe: nit you probably meant to use the project_id var inside the clause as well

tfrcarvalho: done

PR #2287, opened 2019-06-05
Add feature flag for pipeline visualization

emilyzhong:
# Add feature flag for pipeline visualization

* Feature flag for pipeline visualization
* Access-related tests for `stage_results`


app/controllers/samples_controller.rb
```diff
@@ -36,7 +36,8 @@ class SamplesController < ApplicationController
   before_action :authenticate_user!, except: TOKEN_AUTH_ACTIONS
   before_action :authenticate_user_from_token!, only: TOKEN_AUTH_ACTIONS
 
-  before_action :admin_required, only: [:reupload_source, :resync_prod_data_to_staging, :kickoff_pipeline, :retry_pipeline, :pipeline_runs]
+  before_action :admin_required, only: [:reupload_source, :resync_prod_data_to_staging, :kickoff_pipeline, :retry_pipeline,
```
tfrcarvalho: As discussed offline, lets replace this check by a feature check... 
You can declare a `before_action` that checks for you feature and configure it for your methods.
You can check using: `current_user.allowed_feature_list.include?("pipeline-viz")` and you can add the feature using `user.add_allowed_feature("pipeline-viz")`.
This way you can keep the rest of the code.

test/controllers/samples_controller_test.rb
```diff
@@ -207,31 +208,41 @@ class SamplesControllerTest < ActionDispatch::IntegrationTest
...
 
     results = @response.parsed_body["pipeline_stage_results"]
     assert_equal 4, results.length
-    @pipeline_run.pipeline_run_stages.each do |stage|
+    @public_pipeline_run.pipeline_run_stages.each do |stage|
       assert_equal JSON.parse(stage.dag_json), results[stage.name]
     end
   end
 
-  test 'joe cannot see pipeline experimental stage results' do
```
tfrcarvalho: Can you add an admin test to verify it can see the experiment stage? 
Note: As discussed, we want to move tests to rspec/factories... You can do this now, or later on a follow up PR.

emilyzhong: Added experimental stage check for admin in the test above (+ renamed for better clarity)

I'm thinking about moving tests to rspec/factories in a separate PR just so this PR is cleaner/ doesn't have a huge amount of code for new factories everywhere

test/controllers/samples_controller_test.rb
```diff
@@ -205,26 +206,27 @@ class SamplesControllerTest < ActionDispatch::IntegrationTest
     assert_redirected_to samples_url
   end
 
-  test 'should return pipeline stage results' do
+  test 'admin should return all pipeline stage results' do
```
tfrcarvalho: nit: "admin *sees all..."

app/controllers/samples_controller.rb
```diff
@@ -967,15 +967,18 @@ def results_folder
...
   def stage_results
-    @results = {}
-    @sample.first_pipeline_run.pipeline_run_stages.each do |stage|
-      if stage.name == "Experimental" && !current_user.admin?
-        next
+    if current_user.allowed_feature_list.include?("pipeline_viz")
+      @results = {}
+      @sample.first_pipeline_run.pipeline_run_stages.each do |stage|
+        if stage.name == "Experimental" && !current_user.admin?
+          next
```
tfrcarvalho: You can avoid `next` (not great) by changing the condition a bit:
`if stage.name != "Experimental" || current_user.admin?`

test/controllers/samples_controller_test.rb
```diff
@@ -233,6 +235,13 @@ class SamplesControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
+  test 'joe cannot fetch pipeline stage results for another user\'s private samples' do
```
tfrcarvalho: Can you add a test for the feature flag, i.e. does not see if feature flag not enabled?

PR #2289, opened 2019-06-05
Fix default background for heatmap

gregdingle:
# Description

This was a consequence of my refactoring to a static class. It looks like the original code was useless anyway because it used a different param name for background. 

# Test

Load endpoint http://localhost:3000/visualizations/samples_taxons.json?sampleIds%5B%5D=17822
See no more error

Try loading valid background, all good
Try loading invalid background, see default loaded

### DIFF
--- a/app/helpers/heatmap_helper.rb
+++ b/app/helpers/heatmap_helper.rb
@@ -23,7 +23,8 @@ module HeatmapHelper
     { text: "NR r (total reads)", value: "NR.r" }
   ].freeze
 
-  def self.sample_taxons_dict(params, samples)
+  # Samples and background are assumed here to be vieweable.
+  def self.sample_taxons_dict(params, samples, background_id)
     return {} if samples.empty?
 
     num_results = params[:taxonsPerSample] ? params[:taxonsPerSample].to_i : DEFAULT_MAX_NUM_TAXONS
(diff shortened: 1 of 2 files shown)


app/helpers/heatmap_helper.rb
```diff
@@ -55,8 +55,7 @@ def self.sample_taxons_dict(params, samples)
               HeatmapHelper::DEFAULT_TAXON_SORT_PARAM
     species_selected = params[:species] ? params[:species].to_i == 1 : false # Otherwise genus selected
 
-    first_sample = samples.first
-    background_id = params[:background] ? params[:background].to_i : get_background_id(first_sample)
+    background_id = params[:background] ? params[:background].to_i : samples.first.default_background_id
```
MarkAZhang: I notice that in `get_background_id` it checks whether the background is viewable by the user. It looks like it's currently possible for a user to access a background that they don't own? (I know that was before your change)

Nit: I think the `get_background_id` function could be designed better as a `is_background_id_viewable` function that returns true/false. `get_background_id` is super confusing because it looks like you're getting the background_id from the sample, but it's actually looking at the http request params and using the sample as a default. Would that be hard to refactor in this PR?

gregdingle: > It looks like it's currently possible for a user to access a background that they don't own? 

Is `host_genome.default_background.id` guaranteed to be viewable by all? In general, it seems like we should never have host genomes or samples that are accessible by a viewer but their  default backgrounds are not. 

If the sample on line 58 is guaranteed to be viewable, then should we need to check the sample background as well?

MarkAZhang: Totally agree. My issue wasn't with `samples.first.default_background_id`, but with `params[:background]`. A malicious user could supply any background_id. It doesn't look like we check whether the background_id is viewable by them before running the query.

gregdingle: Ok got it. I'll add that validation.

PR #2291, opened 2019-06-05
fix read count in alert

cdebourcy:
If the read count has not yet been calculated, the alert should omit it rather than claiming it is 0

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -844,9 +844,12 @@ class PipelineRun < ApplicationRecord
         self.job_status = STATUS_FAILED
         self.finalized = 1
         self.known_user_error, self.error_message = check_for_user_error(prs)
-        log_message = "SampleFailedEvent: Sample #{sample.id} by #{sample.user.admin? ? 'admin user' : 'non-admin user'} " \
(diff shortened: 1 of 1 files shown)


app/models/pipeline_run.rb
```diff
@@ -844,9 +844,12 @@ def update_job_status
         self.job_status = STATUS_FAILED
         self.finalized = 1
         self.known_user_error, self.error_message = check_for_user_error(prs)
-        log_message = "SampleFailedEvent: Sample #{sample.id} by #{sample.user.admin? ? 'admin user' : 'non-admin user'} " \
-          "failed #{prs.name} with #{adjusted_remaining_reads || 0} reads remaining after #{duration_hrs} hours. " \
-          "See: #{status_url}"
+        log_message = "SampleFailedEvent: Sample #{sample.id} " \
+          "by #{sample.user.admin? ? 'admin user' : 'non-admin user'} " \
```
tfrcarvalho: nit: can you add the processing of this interpolated variables before log_message like ```
user_role = sample.user.admin? ? 'admin user' : 'non-admin user'
log_message = "[...] by #{user_role}"
```
Just makes it more readable...

cdebourcy:
superseded by another PR

PR #2292, opened 2019-06-05
Fix taxon tree png when no tree

gregdingle:
# Description

This is a fix to https://jira.czi.team/browse/IDSEQ-915 by showing relevant options only during tree view. 

image

image

# Test

In report page
Open download dropdown
See no PNG

Switch to taxon tree view
See PNG 

Repeat for SVG

Check that analytics is still sent in logs

### DIFF
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -747,12 +747,6 @@ class PipelineSampleReport extends React.Component {
     );
   };
 
-  handleViewClicked = (_, data) => {
-    this.setState({ view: data.name });
-    const friendlyName = data.name.replace(/_/g, "-");
-    logAnalyticsEvent(`PipelineSampleReport_${friendlyName}-view-menu_clicked`);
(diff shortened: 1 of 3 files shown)


gregdingle:
@happyimadesignr I'd love your input here as well on how to handle the download option when the tree is not visible. I can think of three ways:

1) popup an alert
2) switch to the taxon tree then download
3) make the options only visible when the tree is visible 
3b) make the options grayed out when the tree is not visible

happyimadesignr:
Hi @gregdingle! I think your option 3 would be best, for now at least. That way we don't give options to the user they can't really interact with. This is good to keep in mind when we redesign the download experience. 

Idk if it's within this scope, but can you move the Results Folder link to the bottom of the dropdown? That one is meant to be a "more" type option after you don't see your d/l options, go to the Results Folder page.

gregdingle:
I implemented option 3) by control download options and tree view with same state. See updated PR description. 

@happyimadesignr the "results folder" is now at the bottom.

PR #2293, opened 2019-06-05
Generalize coverage viz to handle genus taxons.

MarkAZhang:
Also ensure coverage viz takes nameType into account (common vs scientific name)

When a genus-level taxon is opened, we show the species that the current accession belongs to. 
Screen Shot 2019-06-05 at 2 25 45 PM

A tooltip helps clarify this further. Notice that the species name is also listed in the dropdown options, and all names now respect the nameType (common vs scientific) 
Screen Shot 2019-06-05 at 2 26 23 PM

Verified that:
* Species-level taxons still work as before.
* Genus-level taxons work as expected, for genuses with multiple species taxons with coverage.
* The nameType option still works properly for the report table and the taxon tree (since the refactor touched those)
* The nameType is still saved properly in the cookie.

### DIFF
--- a/app/assets/src/components/PipelineSampleReport.jsx
+++ b/app/assets/src/components/PipelineSampleReport.jsx
@@ -3,8 +3,9 @@ import React from "react";
 import Cookies from "js-cookie";
 import $ from "jquery";
 import { Label, Menu, Icon, Popup } from "semantic-ui-react";
-import { omit } from "lodash/fp";
+import { omit, partition, keyBy, groupBy, map, mapValues } from "lodash/fp";
 import Nanobar from "nanobar";
+import PropTypes from "prop-types";
 
 import { getSampleReportInfo, getSummaryContigCounts } from "~/api";
 import { parseUrlParams } from "~/helpers/url";
@@ -78,7 +79,6 @@ class PipelineSampleReport extends React.Component {
     const cachedIncludedCategories = Cookies.get("includedCategories");
     const cachedIncludedSubcategories = Cookies.get("includedSubcategories");
 
-    const cached_name_type = Cookies.get("name_type");
     const cachedReadSpecificity = Cookies.get("readSpecificity");
     const cachedTreeMetric = Cookies.get("treeMetric");
     const cachedMinContigSize = parseInt(Cookies.get("minContigSize"), 10);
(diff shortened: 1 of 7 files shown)


app/assets/src/components/PipelineSampleReport.jsx
```diff
@@ -78,7 +79,6 @@ class PipelineSampleReport extends React.Component {
     const cachedIncludedCategories = Cookies.get("includedCategories");
     const cachedIncludedSubcategories = Cookies.get("includedSubcategories");
 
-    const cached_name_type = Cookies.get("name_type");
```
MarkAZhang: We had to move the "name_type" state into the parent component of PipelineSampleReport (SampleView) because the CoverageVizSidebar needed it.

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -2,7 +2,8 @@
 // Also handles the sidebar and tabs logic.
 import React from "react";
 import cx from "classnames";
-import { get } from "lodash/fp";
+import Cookies from "js-cookie";
```
tfrcarvalho: Why cookies? No cookies please.

tfrcarvalho: I realize now it was there before... but good opportunity to change.

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -404,11 +413,54 @@ class SampleView extends React.Component {
       return {};
     }
 
+    let accessionData = null;
+
+    // For genus-level taxons, we aggregate all the available species-level taxons for that genus.
+    if (coverageVizParams.taxLevel === "genus") {
```
gregdingle: For code clarity, put this whole block in a method `getCoverageVizParamsForGenus` or similar

app/assets/src/components/views/SampleView/SampleView.jsx
gregdingle: I'd prefer to see us refactor all the report page cookie state at once.

MarkAZhang: I'll put it on my todo list.

PR #2294, opened 2019-06-05
Make taxon tree scroll horizontally on overflow

gregdingle:
# Description

This fixes issue https://jira.czi.team/browse/IDSEQ-1005?filter=10801 by making the a horizontal scrollbar appear when the viewport width is less than 1200px.

# Notes

* I checked and there are no callers that set any of the affected viz options . 

* I tried a variety of approaches before settling on this very simple one, including 
  * shrinking to fit: made trees too small for many samples
  * determining size from container: didn't play nice with dynamic height resizing that depends on number of leaves

# Test

Make tree scroll:
image

Make tree not scroll:
image

### DIFF
--- a/app/assets/src/components/views/TaxonTreeVis.jsx
+++ b/app/assets/src/components/views/TaxonTreeVis.jsx
@@ -352,6 +352,7 @@ class TaxonTreeVis extends React.Component {
           ref={container => {
             this.treeContainer = container;
           }}
+          style={{ overflowX: "scroll" }}
         >
           <div className="pathogen-labels">{this.renderPathogenLabels()}</div>
         </div>
--- a/app/assets/src/components/views/TaxonTreeVis.jsx
+++ b/app/assets/src/components/views/TaxonTreeVis.jsx
@@ -352,7 +352,6 @@ class TaxonTreeVis extends React.Component {
           ref={container => {
             this.treeContainer = container;
           }}
-          style={{ overflowX: "scroll" }}
         >
           <div className="pathogen-labels">{this.renderPathogenLabels()}</div>
         </div>
--- a/app/assets/src/components/visualizations/TidyTree.js
+++ b/app/assets/src/components/visualizations/TidyTree.js
@@ -14,7 +14,7 @@ export default class TidyTree {
         addOverlays: true,
         attribute: "aggregatescore",
         leafNodeHeight: 35,
-        minWidth: 1000,
+        minWidth: 960,
         minHeight: 300,
         transitionDuration: 500,
         onNodeHover: null,
@@ -27,7 +27,7 @@ export default class TidyTree {
         largerFont: 12,
         onCollapsedStateChange: () => {},
         collapsed: new Set(),
-        svgBackgroundColor: "white"
+        svgBackgroundColor: "white",
       },
       options || {}
     );
@@ -36,7 +36,7 @@ export default class TidyTree {
       top: 20,
       right: 200,
       left: 40,
-      bottom: 20
+      bottom: 20,
     };
 
     this.range = [0, 1];
@@ -123,7 +123,7 @@ export default class TidyTree {
       Math.min(
         ...this.root.leaves().map(l => l.data.values[this.options.attribute])
       ),
-      this.root.data.values[this.options.attribute]
+      this.root.data.values[this.options.attribute],
     ];
 
     let collapsedScale = scaleLinear()
@@ -174,8 +174,8 @@ export default class TidyTree {
                 ),
                 scientificName: `(${d.collapsedChildren.length})`,
                 lineageRank: d.collapsedChildren[0].data.lineageRank,
-                values: d.collapsedChildren[0].data.values
-              }
+                values: d.collapsedChildren[0].data.values,
+              },
             });
           }
         }
@@ -218,7 +218,7 @@ export default class TidyTree {
     this.update(
       Object.assign(node, {
         x0: node.x,
-        y0: node.y
+        y0: node.y,
       })
     );
     this.options.onCollapsedStateChange &&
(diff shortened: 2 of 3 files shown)


gregdingle:
@eggerdesigns welcome to review as well

tfrcarvalho:
Do you know why these trees are short (only 3 levels). Would be good to see examples with all the levels.

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -324,10 +326,15 @@ export default class TidyTree {
     }
 
     // adjust dimensions
-    let width = this.options.minWidth;
+    let width = Math.min(
+      Math.max(this.options.minWidth, this.options.width),
+      document.documentElement.clientWidth -
```
tfrcarvalho: We should adjust to the width of the parent container/div and not view port.
Specially because this is a generic tree. In other pages with elements left and right this might not work.

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -14,8 +14,10 @@ export default class TidyTree {
         addOverlays: true,
         attribute: "aggregatescore",
         leafNodeHeight: 35,
-        minWidth: 1000,
-        minHeight: 300,
+        width: 1000,
```
tfrcarvalho: Do not set width and height directly. Just adjust to parent div, so users can set the size of the container as desired (css or js)

tfrcarvalho: Or at list make the defaults match the div below...

tfrcarvalho:
Forgot to mention... to fully fix the issue you should add scrolling on the parent container (I believe heatmap does this).

gregdingle:
> Do you know why these trees are short (only 3 levels)? Would be good to see examples with all the levels.

I don't know. My understanding is that the number of levels shown is determined dynamically. Are you suggesting to expand the tree? 

And I thought the mosquito project was a good example project. What example would recommend? 


> Do not set width and height directly. Just adjust to parent div, so users can set the size of the container as desired (css or js)

The width and height were already being set directly. I just changed the way the min and max were computed. Do you know why the size would not have been determined by the container originally? I guess it was something to do with SVG. 

> Dont think view port is the right reference to use. See comments.

To be clear, I added the viewport logic to cover a case outside of the motivating issue. It simply prevents the size from being set to a value that would always make the tree too big to be seen 100%. It's not clear that fitting the tree in some container is better than fitting the contain elements to the tree. I'd be happy to remove that logic if it is too confusing.

Related, I'm not side scrolling is better than fitting the tree to the content area, which is how it seems like it was designed to behave. I'd like to discuss that more.

MarkAZhang:
Hey, I think there's a more elegant approach using `viewBox`, an avg attribute.

TLDR:
Instead of `<svg width={width} height={height}>`, do `<svg width="100%" viewBox="0 0 width height">`. (you'll need to do string interpolation, hard to do in a Github comment)

This will allow the visualization to scale with the width of the parent container, much like an image would scale. This works even if you resize the browser after the viz has loaded.
---

Basically, the `viewBox` specifies the coordinate system of all the elements inside the svg. If you do `viewBox="0 0 1 1"`, and then draw a point at x=0.5, y=0.5, then the point will be drawn in the center of the svg. (point is not a real svg element, this is just an example).

(if you don't specify a viewBox explicitly, the viewbox will default to `0 0 width height`

Once you have specified the viewBox, you can then specify a width and height for the svg element. The browser will then scale the svg to the width and height, while preserving the aspect ratio of the viewBox. So if I have `<svg viewBox="0 0 1 1" width="50" height="100">` then the browser will scale the svg to 50px by 50px (since the viewbox was 1 by 1), and center it.

app/assets/src/components/views/TaxonTreeVis.jsx
```diff
@@ -352,6 +352,7 @@ class TaxonTreeVis extends React.Component {
           ref={container => {
             this.treeContainer = container;
           }}
+          style={{ overflowX: "scroll" }}
```
gregdingle: I added this inline because there was no associated CSS file that I could find, and it seems likely this is will be a single instance

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -14,7 +14,7 @@ export default class TidyTree {
         addOverlays: true,
         attribute: "aggregatescore",
         leafNodeHeight: 35,
-        minWidth: 1000,
+        minWidth: 960,
```
gregdingle: This was necessary to prevent default scrolling on first load because the container maxes out at 1200px

app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -27,7 +27,7 @@ export default class TidyTree {
         largerFont: 12,
         onCollapsedStateChange: () => {},
         collapsed: new Set(),
-        svgBackgroundColor: "white"
+        svgBackgroundColor: "white",
```
gregdingle: ignore this and below... prettier

gregdingle:
@MarkAZhang @tfrcarvalho I spent another hour trying to make this work more elegantly but I didn't manage to do so. I think the fix here is good enough to ship. I talked with @eggerdesigns and she is okay with it as well.

app/assets/src/components/views/TaxonTreeVis.jsx
tfrcarvalho: Search for `taxon_tree_vis.scss`. It is old code, so still loaded in `loader.scss`.

tfrcarvalho:
> @MarkAZhang @tfrcarvalho I spent another hour trying to make this work more elegantly but I didn't manage to do so. See notes in above at top.
> 
> I think the fix here is good enough to ship. I talked with @eggerdesigns and she is okay with it as well.

@gregdingle Can you change the title to match the change?

app/assets/src/components/views/TaxonTreeVis.jsx
gregdingle: thanks!

PR #2295, opened 2019-06-05
Make metadata add icon always visible.

MarkAZhang:
Per designer's mocks.
Also made the dropdown right-aligned (instead of center-aligned).

Screen Shot 2019-06-05 at 4 12 03 PM

Screen Shot 2019-06-05 at 4 12 42 PM


MarkAZhang:
Yes! Thanks for catching that. Fixed.

PR #2296, opened 2019-06-05
IDSEQ-992 Make metadata sort when sort icon is clicked on.

MarkAZhang:


app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -842,11 +848,7 @@ export default class Heatmap {
       )
       .style("dominant-baseline", "central")
       .style("text-anchor", "end")
-      .on("click", d => {
-        this.options.onColumnMetadataLabelClick
-          ? this.options.onColumnMetadataLabelClick(d.value, d3.event)
```
tfrcarvalho: What is rationale to remove the onColumnMetadataLabelClick?
This is just part of a generic pattern for the heatmap...

PR #2297, opened 2019-06-05
Put precache resque job in own queue and prioritize

gregdingle:
# Description

The precache jobs were starving the higher pri jobs. 

# Notes

* until this PR, all the resque jobs were in the same queue, `@queue = :q03_pipeline_run`... that seems like poor design

* five resque workers are created... I'm wondering if that is optimized `QUEUE=* COUNT=5 rake resque:workers`

* In the interest of preserving existing behavior, I set the queue priority to `q03_pipeline_run,precache_report_info,*`. However, we could easily come up with a more rational system such as `high,*,low`.

* to be sure, I checked the ruby source of queue priorities: https://github.com/resque/resque/blob/master/lib/resque/worker.rb#L200

# Test

test in staging
wait for benchmark run to finish
see precache success in datadog logs
see precache in separate redis queue


gregdingle:
I checked in staging and everything looks good.

PR #2298, opened 2019-06-06
Automatic pipeline restart

davidrissato:
# Description

This change intends to reduce the on-call noise by automatically restarting the pipeline whenever a pipeline run fails for some random reason. The rspec results below describe this behavior in detail.

```
PipelineRun
  #update_job_status
    when all stages complete successfully
      changes status to finalized
    when a stage completes with an error
      and it is a known user error
        changes status to failed
        does not send error to airbrake
        does not enqueue new pipeline run
      and it is not a known user error
        and an automatic restart is allowed
          changes status to failed
          sends error to airbrake
          enqueues new pipeline run
        and an automatic restart is not allowed
          changes status to failed
          sends error to airbrake
          does not enqueue new pipeline run
  #automatic_restart_allowed?
    when user is an admin
      should be falsy
    when user is not an admin
      and sample has no previous pipeline runs with the same pipeline version
        should be truthy
      and sample has previous pipeline runs with the same pipeline version
        and they all succeeded
          should be truthy
        and at least one of them failed
          should be falsy
  #report_failed_pipeline_run_stage
    sends metric to datadog
    when send_to_airbrake is true
      sends error to airbrake and log error
    when send_to_airbrake is false
      do not send error to airbrake and log warn

Finished in 0.81819 seconds (files took 7.13 seconds to load)
17 examples, 0 failures
```

## Type of change

# Test and Reproduce

1. Upload a sample
2. Terminate the job in AWS console to trigger a failure
3. Wait a few minutes and a new pipeline run should kick off. No error messages so far.
4. Terminate the job in AWS console for this second pipeline run.
5. Pipeline shouldn't restart again. We should see an error being sent to Airbrake.

Specific components 

* [X] pipeline monitor

# Checklist:


davidrissato:
I noticed some pipeline runs in production where `pipeline_version` field is NULL.

I need to know if there is any special reason for that behavior or if it is a bug, because I added some logic to automatically restart the pipeline if I don't find a previous failure with the same pipeline version. 

It seems the method `sample#kickoff_pipeline` is not setting this attribute, and pipeline run will restart twice for this reason.

We should either fix the condition, or in case it is correct, I'll change the logic to not restart the pipeline if any previous run failed once, no matter which pipeline version.

cdebourcy:
> I noticed some pipeline runs in production where `pipeline_version` field is NULL.
> 
> I need to know if there is any special reason for that behavior or if it is a bug, because I added some logic to automatically restart the pipeline if I don't find a previous failure with the same pipeline version.
> 
> It seems the method `sample#kickoff_pipeline` is not setting this attribute, and pipeline run will restart twice for this reason.
> 
> We should either fix the condition, or in case it is correct, I'll change the logic to not restart the pipeline if any previous run failed once, no matter which pipeline version.

There is a lag before pipeline_version gets populated: job gets submitted --> batch container clones latest idseq-dag master and publishes the version to S3 --> pipeline monitor reads the version from S3

app/models/pipeline_run.rb
```diff
@@ -860,7 +859,31 @@ def update_job_status
       self.job_status = "#{prs.step_number}.#{prs.name}-#{prs.job_status}"
       self.job_status += "|#{STATUS_READY}" if report_ready?
     end
-    save
+    save!
+    enqueue_new_pipeline_run if automatic_restart
+  end
+
+  protected def report_failed_pipeline_run_stage(prs)
+    message = "SampleFailedEvent: Sample #{sample.id} by #{sample.user.admin? ? 'admin user' : 'non-admin user'} " \
```
cdebourcy: Do you want to incorporate this change here? :) 
https://github.com/chanzuckerberg/idseq-web/pull/2291/files
Then I'll close my PR

davidrissato: Sure, it is a good idea. I can do that.

davidrissato:
Ok, makes sense. Thank you for this explanation. So my logic should be fine. The worst it can happen is restarting twice, once for a record with an empty version field and once for the right version field.

> cdebourcy:
> There is a lag before pipeline_version gets populated: job gets submitted --> batch container clones latest idseq-dag master and publishes the version to S3 --> pipeline monitor reads the version from S3

app/models/pipeline_run.rb
davidrissato: Done

app/jobs/restart_pipeline_for_sample.rb
```diff
@@ -0,0 +1,20 @@
...
+  @queue = :q03_pipeline_run
+
+  def self.perform(sample_id)
+    s = Sample.find(sample_id)
+    if s.nil?
+      raise Exception("Sample #{sample_id} not found")
+    end
+    Rails.logger.info("RestartPipelineForSample #{sample_id} is being triggered")
+    unless s.kickoff_pipeline
+      raise Exception("not restarted")
```
tfrcarvalho: nit: I would just use the same lingo as the call, something like "Kickoff pipeline not successful" ? It is just easy to relate to the code...

davidrissato: done

app/models/sample.rb
```diff
@@ -459,6 +459,10 @@ def self.my_data(user)
     where("project_id in (?)", project_ids)
   end
 
+  def role_name
```
jshoe: I think you meant to put this on user.rb right? since you call sample.user.role_name above

davidrissato: Yes, that was my intention. Thank you for catching this one.

app/jobs/restart_pipeline_for_sample.rb
```diff
@@ -0,0 +1,18 @@
+class RestartPipelineForSample
+  @queue = :q03_pipeline_run
```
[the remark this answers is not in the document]
davidrissato: I wanted to use this one on purpose, since this restart task is actually part of the pipeline monitor.

davidrissato:
> Love the use of rspec here!
> 
> A couple general comments:
> 
> 1. it would be nice to send metrics to datadog for nice graphs there, using `MetricUtil.put_metric_now`
Sure. Let's discuss about that. I think I can do a complementary PR later on to add these metrics.

> 2. it would be nice to save the fact of restarting in the failed pipeline run for later understanding... for example `FAILED|RESTARTING`. See the precedent `      self.job_status += "|#{STATUS_READY}" if report_ready?`
Actually I'm not restarting the same pipeline run. I'm kicking off a new pipeline run from top, so FAILED is the correct state for the previous run.

app/models/sample.rb
davidrissato: I'm adding a test around that method to prevent regressions

davidrissato: done

PR #2300, opened 2019-06-06
[Maps] Add location_v2 to the dimensions endpoints, left pane filters, and main sample table

jshoe:
### Description
- Add locationV2 dropdown filter to the left sidebar (just based on the full name for now).
- Add locationV2 to the main data table.
- Modify sample and project dimensions endpoints to return locationV2 names. Moved code into LocationHelper sample_dimensions/project_dimensions.
- I decided to continue adding everything as locationV2 (instead of not rendering locationV1) because it makes it easier to verify changes and keep track of what has been updated.
- The test I added to location_helper_test.rb can be improved with some new RSpec factories and tests for SamplesHelper.

### Demo
Jun-05-2019 20-11-56
Jun-05-2019 20-13-39

### Test and Reproduce
- Go to data discovery samples, add "Location v2" column, see locations.
- Use the left pane filters on samples and projects and see Location v2 filter as expected.

### DIFF
--- a/app/assets/src/components/visualizations/table/InfiniteTable.jsx
+++ b/app/assets/src/components/visualizations/table/InfiniteTable.jsx
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { defaultTableRowRenderer, InfiniteLoader } from "react-virtualized";
+import { isObject } from "lodash/fp";
 import BaseTable from "./BaseTable";
 import cs from "./infinite_table.scss";
 import cx from "classnames";
@@ -72,7 +73,12 @@ class InfiniteTable extends React.Component {
     // Guarantees that we have at least one child div in the cell
     return (
       <div className={cs.cellContent}>
-        {cellData == null ? "" : String(cellData)}
+        {/* Use .name if val is an object (e.g. location object) */}
+        {cellData == null
+          ? ""
+          : isObject(cellData)
+            ? cellData.name
+            : String(cellData)}
       </div>
     );
   };
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -206,6 +206,15 @@ class SamplesController < ApplicationController
       locations << { value: "not_set", text: "Unknown", count: not_set_count }
     end
 
+    locations_v2 = samples_by_metadata_field(sample_ids, "collection_location_v2").count
+    locations_v2 = locations_v2.map do |location, count|
+      { value: location.name, text: location.name, count: count }
+    end
+    not_set_count = samples_count - locations_v2.sum { |l| l[:count] }
+    if not_set_count > 0
+      locations_v2 << { value: "not_set", text: "Unknown", count: not_set_count }
+    end
+
     tissues = samples_by_metadata_field(sample_ids, "sample_type").count
     tissues = tissues.map do |tissue, count|
       { value: tissue, text: tissue, count: count }
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -198,8 +198,8 @@ class SamplesController < ApplicationController
     sample_ids = samples.pluck(:id)
     samples_count = samples.count
 
-    locations = LocationHelper.dimensions(sample_ids, "collection_location", samples_count)
-    locations_v2 = LocationHelper.dimensions(sample_ids, "collection_location_v2", samples_count)
+    locations = LocationHelper.sample_dimensions(sample_ids, "collection_location", samples_count)
+    locations_v2 = LocationHelper.sample_dimensions(sample_ids, "collection_location_v2", samples_count)
 
     tissues = samples_by_metadata_field(sample_ids, "sample_type").count
     tissues = tissues.map do |tissue, count|
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -1,11 +1,12 @@
 class SamplesController < ApplicationController
   include ApplicationHelper
+  include ElasticsearchHelper
+  include ErrorHelper
+  include HeatmapHelper
+  include LocationHelper
+  include PipelineOutputsHelper
   include ReportHelper
   include SamplesHelper
-  include PipelineOutputsHelper
-  include ElasticsearchHelper
-  include HeatmapHelper
-  include ErrorHelper
 
   ########################################
   # Note to developers:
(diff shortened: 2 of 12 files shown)


app/assets/src/components/visualizations/table/InfiniteTable.jsx
```diff
@@ -72,7 +73,12 @@ class InfiniteTable extends React.Component {
     // Guarantees that we have at least one child div in the cell
     return (
       <div className={cs.cellContent}>
-        {cellData == null ? "" : String(cellData)}
+        {/* Use .name if val is an object (e.g. location object) */}
```
tfrcarvalho: Why do you need this?

app/controllers/samples_controller.rb
```diff
@@ -1,11 +1,12 @@
...
-  include SamplesHelper
-  include PipelineOutputsHelper
   include ElasticsearchHelper
-  include HeatmapHelper
   include ErrorHelper
+  include HeatmapHelper
+  include LocationHelper
+  include PipelineOutputsHelper
+  include ReportHelper
+  include SamplesHelper
```
tfrcarvalho: 💯 I like sorting!

spec/helpers/location_helper_spec.rb
```diff
@@ -0,0 +1,67 @@
+# See deprecated test/helpers/location_helper_test.rb
+
+require "rails_helper"
```
tfrcarvalho: 👍

spec/helpers/location_helper_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      result = LocationHelper.sanitize_name(name)
+      expect(result).to eq("California, USA DROP TABLE USERS divdiv")
+    end
+  end
+
+  describe "#truncate_name" do
+    it "truncates a name" do
+      name = "Chania, Chania Municipality, Chania Regional Unit, Region of Crete, Crete, 73135, Greece"
+      result = LocationHelper.truncate_name(name)
+      expect(result).to eq("Chania, Chania Municipality, Chania Regional ...")
```
tfrcarvalho: Should we be smarter about how we truncate this names? We should remove the middle part, no? 
E.g. Split by comma, get first and last part, and then just get from front or end until the limit...

spec/helpers/location_helper_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      expect(result).to eq("Alaska, USA")
+    end
+  end
+
+  # TODO: These tests would be improved with Factories and SamplesHelper tests
+  describe "#sample_dimensions" do
+    it "gets and formats sample location dimensions" do
+      sample_ids = [1, 2, 3]
+      field_name = "collection_location_v2"
+      mock_filtered = { [nil, "Alaska, USA"] => 1, [nil, "Chania, Chania Municipality, Chania Regional Unit, Region of Crete, Crete, 73135, Greece"] => 1, [nil, "Hanoi, Vietnam"] => 1, [nil, "Redwood City, San Mateo County, California, USA"] => 112, [nil, "Zimbabwe"] => 1 }
```
tfrcarvalho: I would put each option on its own line for readability.

spec/helpers/location_helper_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+                  { value: "Chania, Chania Municipality, Chania Regional Unit, Region of Crete, Crete, 73135, Greece", text: "Chania, Chania Municipality, Chania Regional ...", count: 1 },
+                  { value: "Hanoi, Vietnam", text: "Hanoi, Vietnam", count: 1 },
+                  { value: "Redwood City, San Mateo County, California, USA", text: "Redwood City, San Mateo County, California, U...", count: 1 },
+                  { value: "Zimbabwe", text: "Zimbabwe", count: 1 }]
+      result = LocationHelper.project_dimensions([1, 2, 3], field_name)
+      expect(result).to eq(expected)
+    end
+  end
+
+  describe "#filter_by_name" do
```
tfrcarvalho: If you really must have this placeholder, add a `pending "<todo description here>"`.
This will generate a warning when running tests.

app/helpers/samples_helper.rb
```diff
@@ -611,4 +627,7 @@ def filter_by_sample_ids(samples, requested_sample_ids)
     requested_sample_ids.class.name
     samples.where(id: requested_sample_ids.is_a?(Array) ? requested_sample_ids : JSON.parse(requested_sample_ids))
   end
+
+  # Callable without include/extend
+  module_function :samples_by_metadata_field
```
tfrcarvalho: Why?

app/assets/src/components/visualizations/table/InfiniteTable.jsx
jshoe: this one was because the location object is getting passed down altogether (like in other places) so it would show up as [object Object]. I could make a locationRenderer instead here?

spec/helpers/location_helper_spec.rb
jshoe: do you think I should just leave it out? or is there something I can still test without pulling in fixtures

```
    samples = samples_with_metadata.includes(metadata: :location)
    # Plain text locations in string_validated_value
    samples.where(metadata: { string_validated_value: query })
           .or(samples.where(metadata: { locations: { name: query } }))
```

app/helpers/samples_helper.rb
jshoe: this just let me call it as a class method `SamplesHelper.samples_by_metadata_field` without having to pull in all the other methods since it's defined as an instance method?

spec/helpers/location_helper_spec.rb
jshoe: great idea 👍 changed to get the first two or last two

spec/helpers/location_helper_spec.rb
jshoe: 👍

app/helpers/samples_helper.rb
tfrcarvalho: How hard would it be to just make them static?

spec/helpers/location_helper_spec.rb
jshoe: added the pending 👍

app/helpers/samples_helper.rb
jshoe: changed to static 👍

PR #2301, opened 2019-06-06
Add new warnings and constraints to phylo tree workflow

tfrcarvalho:
### Description

This PR:
* Replaces error message on number of samples and changes the lower limit.
* Add new warning message on number of reads.

Jun-05-2019 22-26-59

### Type of change
* New feature (non-breaking change which adds functionality)

### Test

* Verified that creation is blocked by too few (<4) or too many (>100) samples
* Verified that a warning is displayed for samples with too few reads.


app/assets/src/components/ui/notifications/Notification.jsx
```diff
@@ -20,19 +20,21 @@ class Notification extends React.Component {
...
+    const { children, className, displayStyle, onClose, type } = this.props;
     return (
       <div
         className={cx(
-          this.props.className,
+          className,
           cs.notification,
-          cs[this.props.type]
+          cs[type],
+          displayStyle && cs.flat
```
MarkAZhang: It looks like this just always applies `cs.flat`. Is that intentional?

app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
```diff
@@ -353,9 +355,30 @@ class PhyloTreeCreation extends React.Component {
...
+
+    const getReadsArray = (samples, indices, readType) =>
+      [...indices].map(i => samples[i].rawReads[readType]);
+
+    const projectSamplesWithFewReads = PhyloTreeChecks.countSamplesWithFewReads(
+      getReadsArray(projectSamples, selectedProjectSamples, "nt"),
+      getReadsArray(projectSamples, selectedProjectSamples, "nr")
+    );
+
+    const ptherSamplesWithFewReads = PhyloTreeChecks.countSamplesWithFewReads(
```
MarkAZhang: 'other', not 'pther'

app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
```diff
@@ -353,9 +355,30 @@ class PhyloTreeCreation extends React.Component {
     let nSamples =
       this.state.selectedProjectSamples.size +
       this.state.selectedOtherSamples.size;
-    return (
-      nSamples >= this.minNumberOfSamples && nSamples <= this.maxNumberOfSamples
+    return PhyloTreeChecks.isNumberOfSamplesValid(nSamples);
+  }
+
+  getNumberOfSamplesWithFewReads() {
```
MarkAZhang: I would add a comment explaining that this is the number of samples with few reads in EITHER nt or nr (an important detail).

app/assets/src/components/ui/notifications/Notification.jsx
tfrcarvalho: Not at all! Fixed

app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
tfrcarvalho: Added!

app/assets/src/components/views/phylo_tree/PhyloTreeCreation.jsx
tfrcarvalho: Fixed.

tfrcarvalho:
> Maybe a dumb question but if the message says
> 
> "Would you like additional samples..."
> 
> Shouldn't there be no minimum? Maybe the message needs to change as well.

Makes sense... I would probably change to something more neutral like "Add additional samples..."

PR #2302, opened 2019-06-06
Markzhang/long sample names

MarkAZhang:
Before:

Screen Shot 2019-06-06 at 1 52 34 PM

After:

Screen Shot 2019-06-06 at 1 46 48 PM


MarkAZhang:
Made a mistake with this branch.

PR #2304, opened 2019-06-06
Factories, rspec tests for pipeline visualization stage_results endpoint

emilyzhong:
# Factories, rspec tests for pipeline visualization stage_results endpoint
* `pipeline_run_stages` factory
* rspec tests in `samples_controller_spec.rb` specific to pipeline visualization stage_results endpoint (see https://github.com/chanzuckerberg/idseq-web/pull/2287, https://github.com/chanzuckerberg/idseq-web/pull/2285)

To run tests: `docker-compose run --rm web rspec -e "SamplesController"`

### DIFF
--- a/spec/controllers/samples_controller_spec.rb
+++ b/spec/controllers/samples_controller_spec.rb
@@ -1,0 +1,98 @@
+require 'rails_helper'
+
+RSpec.describe SamplesController, type: :controller do
+  create_users
+
+  pipeline_stages = [{
+    name: "Host Filtering",
+    dag_json: "{\"key1\": \"value1\"}"
+  }, {
+    name: "GSNAPL/RAPSEARCH alignment",
+    dag_json: "{\"key2\": \"value2\"}"
+  }, {
+    name: "Post Processing",
+    dag_json: "{\"key3\": \"value3\"}"
+  }, {
+    name: "Experimental",
+    dag_json: "{\"key4\": \"value4\"}"
+  }]
+
+  # Admin specific behavior
+  context "Admin user" do
+    before do
+      sign_in @admin
+      @admin.add_allowed_feature("pipeline_viz")
+      @admin.save!
+    end
+
+    describe "GET pipeline stage results" do
+      it "sees all pipeline stage results" do
+        project = create(:public_project)
+        sample = create(:sample, project: project,
+                                 pipeline_runs_data: [{ pipeline_run_stages_data: pipeline_stages }])
+
+        get :stage_results, params: { id: sample.id }
+
+        json_response = JSON.parse(response.body)["pipeline_stage_results"]
+        expect(json_response.length).to eq(4)
+        expect(json_response["Experimental"]).to be_truthy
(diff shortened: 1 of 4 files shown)


spec/controllers/samples_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
+require 'rails_helper'
+
+RSpec.describe SamplesController, type: :controller do
+  create_users
+
+  pipeline_stages = [{
```
jshoe: you could also call this one pipeline_run_stages_data to clarify that they're not pipeline stage objects

spec/controllers/samples_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  # Non-admin, aka Joe, specific behavior
+  context "Joe" do
+    before do
+      sign_in @joe
+      @joe.add_allowed_feature("pipeline_viz")
+      @joe.save!
+    end
+
+    describe "GET pipeline stage results" do
+      it "cannot see pipeline experimental stage results" do
```
jshoe: I think you could update this to say `it "can see pipeline stage results without the experimental stage results"` to clarify that this is testing the happy path

jshoe: and maybe also a test for "non-admin user w/o the feature flag can't see the results" although there's overlap

jshoe:
nit: usually we keep commit and PR messages in imperative tense

spec/controllers/samples_controller_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+  }, {
+    name: "Experimental",
+    dag_json: "{\"key4\": \"value4\"}"
+  }]
+
+  # Admin specific behavior
+  context "Admin user" do
+    before do
+      sign_in @admin
+      @admin.add_allowed_feature("pipeline_viz")
```
tfrcarvalho: This should be done through factories... something like:
```
create(:joe)
@admin = create(:admin, allowed_features: ["pipeline_viz"])
sign_in_admin
```
(instead of using the create_users macro)
Rule of thumb is we should not call inner methods to change state here but inside of factories.

spec/controllers/samples_controller_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+  context "Joe" do
+    before do
+      sign_in @joe
+      @joe.add_allowed_feature("pipeline_viz")
+      @joe.save!
+    end
+
+    describe "GET pipeline stage results" do
+      it "can see pipeline stage results without the experimental stage results" do
+        project = create(:public_project)
```
tfrcarvalho: You do not have a test for a project of the user, right? I.e.
`create(:project, users:[@joe])`

emilyzhong: Added! https://github.com/chanzuckerberg/idseq-web/pull/2304/files#diff-40e83e0600ff9945e686badfda7541a1R88

PR #2306, opened 2019-06-06
Fix server SQL error for heatmap if all taxons are filtered out.

MarkAZhang:
If the user enters a threshold filter that filters out everything, the server will throw a 500 error because the generated SQL is invalid. Add a guard against that.

Verified that:
* filtering everything out now shows "No data to display" as expected.
* the heatmap otherwise behaves normally.

### DIFF
--- a/app/helpers/heatmap_helper.rb
+++ b/app/helpers/heatmap_helper.rb
@@ -81,9 +81,11 @@ module HeatmapHelper
     taxon_ids = details.pluck('tax_id')
     taxon_ids -= removed_taxon_ids
 
-    # Refetch at genus level using species level
-    parent_ids = species_selected ? [] : HeatmapHelper.fetch_parent_ids(taxon_ids, samples)
-    results_by_pr = HeatmapHelper.fetch_samples_taxons_counts(samples, taxon_ids, parent_ids, background_id)
+    unless taxon_ids.empty?
+      # Refetch at genus level using species level
+      parent_ids = species_selected ? [] : HeatmapHelper.fetch_parent_ids(taxon_ids, samples)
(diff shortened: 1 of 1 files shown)


app/helpers/heatmap_helper.rb
```diff
@@ -81,9 +81,11 @@ def self.sample_taxons_dict(params, samples, background_id)
...
-    # Refetch at genus level using species level
-    parent_ids = species_selected ? [] : HeatmapHelper.fetch_parent_ids(taxon_ids, samples)
-    results_by_pr = HeatmapHelper.fetch_samples_taxons_counts(samples, taxon_ids, parent_ids, background_id)
+    unless taxon_ids.empty?
+      # Refetch at genus level using species level
+      parent_ids = species_selected ? [] : HeatmapHelper.fetch_parent_ids(taxon_ids, samples)
+      results_by_pr = HeatmapHelper.fetch_samples_taxons_counts(samples, taxon_ids, parent_ids, background_id)
+    end
 
     HeatmapHelper.samples_taxons_details(
```
tfrcarvalho: Shouldn't we skip this too?

MarkAZhang: For weird reasons, you still need to return the samples in the response, even if they don't have any results. I think SamplesHeatmapView clobbers this.state.sampleIds every time the response comes back. If you return {} in this endpoint, the heatmap page will forget about all sampleIds and you won't be able to get them back even if you remove the threshold filter...

Looks like a nontrivial refactor.

gregdingle:
Thanks for fixing this. I missed it in testing.

PR #2309, opened 2019-06-07
change description from unbiased to hypothesis-free

cdebourcy:
Following discussion with various partners, "hypothesis-free" better conveys what we're doing than "unbiased".


cdebourcy:
image

PR #2311, opened 2019-06-07
Add new amr_counts endpoint for AMR Heatmap

lvreynoso:
# Description

Implement a new endpoint: `GET /amr_counts.json`
- only accessible by admin users
- accepts a sample `id` parameter, required to work properly
- returns JSON including the `sample_name`, `sample_id`, and an array of `amr_counts`

Add a new Rails controller: `app/controllers/amr_heatmap_controller.rb`
- handles the `/amr_counts.json` route in its `index` method

Add a new Rspec test: `spec/controllers/amr_heatmap_controller_spec.rb`
- tests the AmrHeatmapController to ensure the data returned is correct, and to ensure that the endpoint can only be accessed by admin users

Add new Rspec factories: `spec/factories/amr_counts.rb` & `spec/factories/output_states.rb`
- required to properly create sample data to test the AmrHeatmapController

The purpose for these changes is to work towards creating an AMR Heatmap visualization. This is the first step, extracting amr_counts data from a given sample id.

## Type of change

- [✅ ] New feature (non-breaking change which adds functionality)

Specific components 

* 

# Testing Script:
`docker-compose run --rm web "RAILS_ENV=test rake db:drop db:create db:migrate db:seed"`
`docker-compose run --rm web "rspec"`
`docker-compose run --rm web "rake test"`


lvreynoso:
Made a one line fix in the Rspec test, should be ready for review

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,45 @@
+class AmrHeatmapController < ApplicationController
+  include ReportHelper
+
+  before_action :admin_required
+
+  current_power do # Put this here for CLI
```
MarkAZhang: There are special CLI-related situations in certain controllers where we'll need this block. We don't need this here since it isn't CLI-related. (CLI is our IDseq command line tool that users can use to upload data to IDseq)

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    Power.new(current_user)
+  end
+
+  power :viewable_samples, as: :samples_scope
```
MarkAZhang: You should be able to use `current_power.viewable_samples` directly without re-assigning to `samples_scope`.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  # first we grab the Sample from the id parameter passed to us
+  before_action :set_sample
+
+  # GET /amr_heatmap.json
+  # Return JSON information required to create a visualization,
+  # from submitted parameters
+  def index
```
MarkAZhang: It's not clear that this function should be the index. I would have expected the `index` to be the controller that renders the view for the amr heatmap. Maybe rename this function to `get_amr_counts`.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    @pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
```
MarkAZhang: I think it would make more sense to have this function accept multiple sample_ids, as opposed to a single one. If it only accepts a single id, we'd have to call the endpoint multiple times from the front-end in order to get the amr_counts for all of our samples.

Also, I don't think we need to allow for a `pipeline_version` param. We can just use the most recent pipeline_run for each sample, which you can get by calling `sample.first_pipeline_run` (it's poorly named).

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    @amr_counts = nil
+    can_see_amr = (current_user.admin? || current_user.allowed_feature_list.include?("AMR"))
```
MarkAZhang: You probably don't need this right now, since you already have `before_action : admin_required` above. You can always add the feature flag when it's time to remove `admin_required`.

As a side note, it's better to call your feature flag "amr_heatmap", instead of re-using "AMR" from the old amr table, since we'll want to control access to the two features separately.

app/views/visualizations/amr_heatmap.html.erb
```diff
@@ -0,0 +1,12 @@
+<div id="AMRHeatmapView_page">
```
MarkAZhang: I would remove this file, since it's not currently being used. Don't include files intended for future PRs.

spec/controllers/amr_heatmap_controller_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    describe "GET index" do
+      it "sees correct AMR data" do
+        # Create a test sample (in a project, as required) that contains Amr data.
+        # Note that pipeline_runs_data needs to have two attributes in order for the sample
+        # to be available for the AmrHeatmapController: a job_status equal to the constant
+        # STATUS_CHECKED as defined in the PipelineRun class, and an output_states array
+        # containing an OutputState for amr_counts with state equal to the constant
+        # STATUS_LOADED as defined in the PipelineRun class.
```
MarkAZhang: Great comments.

spec/controllers/amr_heatmap_controller_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+                     state: PipelineRun::STATUS_LOADED
+                   }]
+                 }]
+               }])
+
+        # Now that the sample has been created, grab it from the test db
+        power = Power.new(@joe) # for safe access
+        @viewable_samples = power.viewable_samples
+        @sample = @viewable_samples[0] # Because we only have one test sample
+        @amr_counts = @sample.first_pipeline_run.amr_counts[0] # Because we only have one AmrCount in amr_counts_data
```
MarkAZhang: Nit: Don't need this line.

spec/controllers/amr_heatmap_controller_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+                   job_status: PipelineRun::STATUS_CHECKED,
+                   output_states_data: [{
+                     output: "amr_counts",
+                     state: PipelineRun::STATUS_LOADED
+                   }]
+                 }]
+               }])
+
+        # Now that the sample has been created, grab it from the test db
+        power = Power.new(@admin) # for safe access
```
MarkAZhang: I'm not sure if using Power is the best way to get the newly created sample/amr_counts (though it does work). I believe `create()` returns the object it created. Will defer to @tfrcarvalho

app/controllers/amr_heatmap_controller.rb
tfrcarvalho: My 2 cents: I prefer feature flags to admin (which was our old default way to guard features). Feature flag is more granular and will allow us to test locally with both test and admin accounts, and to enable it for key users for external review and usability research.

spec/controllers/amr_heatmap_controller_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+                 }]
+               }])
+
+        # Now that the sample has been created, grab it from the test db
+        power = Power.new(@joe) # for safe access
+        @viewable_samples = power.viewable_samples
+        @sample = @viewable_samples[0] # Because we only have one test sample
+        @amr_counts = @sample.first_pipeline_run.amr_counts[0] # Because we only have one AmrCount in amr_counts_data
+
+        # Hit the controller, expect a redirect
```
tfrcarvalho: Another tip: make sure the comment is not redundant to the code... this test and expect languages are very clear by themselves.

spec/factories/pipeline_runs.rb
```diff
@@ -4,18 +4,31 @@
...
+      amr_counts_data { [] }
+      output_states_data { [] }
       # Array of pipeline_run_stage entries to create.
       # The hash elements will be passed on to pipeline_run_stage factory as keyword arguments.
       pipeline_run_stages_data { [] }
     end
 
     alignment_config { create(:alignment_config) }
 
+    trait :with_amr_count do
```
tfrcarvalho: This can be removed since you are only using the `amr_counts_data` variable below.

app/controllers/amr_heatmap_controller.rb
tfrcarvalho: @MarkAZhang, would it make sense for this to be the `show` endpoint?
we should avoid the `get_` since that is normally implicit on the GET action.

jshoe:
nit: we usually keep PR titles (and the squashed commit message) in imperative tense https://chris.beams.io/posts/git-commit/#imperative

app/controllers/amr_heatmap_controller.rb
lvreynoso: Removed!

app/controllers/amr_heatmap_controller.rb
lvreynoso: Done!

app/controllers/amr_heatmap_controller.rb
lvreynoso: I changed the endpoint to `amr_counts`, does that work?

app/controllers/amr_heatmap_controller.rb
lvreynoso: Done!

app/controllers/amr_heatmap_controller.rb
lvreynoso: I removed the feature flag since Mark is right, we already have `before_action :admin_required` restricting access and the feature flag can always be added back in later.

app/views/visualizations/amr_heatmap.html.erb
lvreynoso: Removed! Sorry about that, my oversight.

spec/controllers/amr_heatmap_controller_spec.rb
lvreynoso: Thank you!

spec/controllers/amr_heatmap_controller_spec.rb
lvreynoso: Removed!

spec/controllers/amr_heatmap_controller_spec.rb
lvreynoso: Done! Removed usage of Power from tests and created sample separately.

app/controllers/amr_heatmap_controller.rb
MarkAZhang: SGTM. Don't think `show` makes sense here since that's normally reserved for a view (e.g. `show.html.erb`).

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,39 @@
+class AmrHeatmapController < ApplicationController
+  include ReportHelper
```
MarkAZhang: I don't think you need `ReportHelper` any more.

spec/controllers/amr_heatmap_controller_spec.rb
lvreynoso: Removed, thank you!

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  before_action :admin_required
+
+  power :viewable_samples
```
MarkAZhang: I don't think this line is necessary either, as it's already done in `power.rb`. It looks like we only use `power` statements in controllers if we want to set different scopes for different controllers, e.g. 
`power :projects, map: { EDIT_ACTIONS => :updatable_projects }, as: :projects_scope`

spec/factories/pipeline_runs.rb
```diff
@@ -4,18 +4,31 @@
       # Array of taxon_counts entries to create.
       # The hash elements will be passed on to taxon_count factory as keyword arguments.
       taxon_counts_data { [] }
+      amr_counts_data { [] }
+      output_states_data { [] }
       # Array of pipeline_run_stage entries to create.
```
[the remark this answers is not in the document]
lvreynoso: Done!

spec/factories/pipeline_runs.rb
lvreynoso: Removed!

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  # GET /amr_heatmap.json
+  # Return JSON information required to create a visualization,
+  # from submitted parameters
```
MarkAZhang: I think it's better if this comment doesn't mention the visualization (since the amr_counts could conceivably be used for other purposes), and instead elaborates a little on what `amr_counts` is and the fact that this endpoint receives multiple samples.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  def amr_counts
+    samples = []
+    amr_data = []
+
+    params[:id].each do |param_id|
```
MarkAZhang: Rename `:id` to `:sampleIds`.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+      sample = current_power.viewable_samples.find(param_id)
+      assert_access
```
MarkAZhang: I'm honestly not 100% sure what the point of `assert_access` is (it's in `application_controller`), but you shouldn't need to do it here.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+      samples << sample
+    end
```
MarkAZhang: You should be able to do `current_power.viewable_samples.where(id: params[:sampleIds])` to fetch all the samples in one query.

config/routes.rb
```diff
@@ -106,6 +106,8 @@
   post 'visualizations/shorten_url', to: 'visualizations#shorten_url'
   get 'visualizations.json', to: 'visualizations#index'
 
+  get 'amr_heatmap.json', to: 'amr_heatmap#amr_counts'
```
MarkAZhang: Change `amr_heatmap.json` to `amr_heatmap/amr_counts.json`.

spec/controllers/amr_heatmap_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+                                                   gene: @amr_counts["gene"],
+                                                   allele: @amr_counts["allele"],
+                                                   coverage: @amr_counts["coverage"],
+                                                   depth: @amr_counts["depth"],
+                                                   pipeline_run_id: @amr_counts["pipeline_run_id"],
+                                                   drug_family: @amr_counts["drug_family"]
+                                                 }])
+      end
+    end
+  end
```
MarkAZhang: I would expand this test case to test 2 samples, since your endpoint accepts multiple samples.

I would also test the case where one of the samples doesn't exist.
And the case where one of the samples doesn't have amr_counts.

There's also the case where the user tries to access a sample that they don't have access to. Unfortunately, this isn't easy to test right now because admins have access to all samples. I would add a TODO to add this test later.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+        end
+      end
+      amr_data << {
+        sample_name: sample.name,
+        sample_id: sample.id,
+        amr_counts: amr_counts
+      }
+    end
+
+    render json: amr_data
```
MarkAZhang: I debated the need to add a `missingSampleIds` field to the response, which explicitly lists the sample ids that we couldn't find. I don't think this is necessary right now (since we can infer on the front-end which sample ids weren't returned). We might want to do something like this if there are multiple reasons a sample can fail to fetch, and we want to surface the failure reason to the user.

jshoe:
another nit: I think we should keep AMR capitalized (like AMRHeatmapController and AMRCounts) since you wouldn't find the acronym capitalized as Amr otherwise. Like we have AMRView.jsx and ERCCScatterPlot.jsx

lvreynoso:
> another nit: I think we should keep AMR capitalized (like AMRHeatmapController and AMRCounts) since you wouldn't find the acronym capitalized as Amr otherwise (so either all caps or all lowercase). Like we have AMRView.jsx and ERCCScatterPlot.jsx

Rails expects snake_case controller models and files to map to UpperCamelCase classes, unfortunately. `app/models/amr_count.rb` defines the AmrCount class, for example, and that's what's used in other parts of the Rails code. React seems less picky about this so I'll definitely use all caps AMR for future React components.

app/controllers/amr_heatmap_controller.rb
lvreynoso: Removed!

app/controllers/amr_heatmap_controller.rb
lvreynoso: Removed!

app/controllers/amr_heatmap_controller.rb
lvreynoso: Done! Updated with better comment.

app/controllers/amr_heatmap_controller.rb
lvreynoso: Done!

app/controllers/amr_heatmap_controller.rb
lvreynoso: Removed!

app/controllers/amr_heatmap_controller.rb
lvreynoso: Done!

config/routes.rb
lvreynoso: Done!

spec/controllers/amr_heatmap_controller_spec.rb
lvreynoso: Added!

app/controllers/amr_heatmap_controller.rb
lvreynoso: Sample ids that are invalid (for whatever reason) now just return the same object as valid ids, but with `sample_name` set to an empty string and `error` set to `"sample not found"`. I could undo it, but it works for now and I prefer it over failing silently.

MarkAZhang: Sounds good.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -0,0 +1,56 @@
+class AmrHeatmapController < ApplicationController
+  before_action :admin_required
+
+  # GET /amr_heatmap.json
```
MarkAZhang: Should be `/amr_heatmap/amr_counts.json`.

PR #2313, opened 2019-06-10
[Phylo Trees] Validate name on same step it is added

tfrcarvalho:
### Description

Validate the name of the phylo tree before moving on to the next step of phylo tree creation.

Jun-10-2019 09-52-18

### Testing

1. Check name of existing phylo trees
1. Click new phylo tree button
1. Choose any project and taxon
1. Input an existing phylo tree name
1. Click continue and verify invalid name error
1. Change phylo tree name and verify you can continue phylo tree creation.


app/controllers/phylo_trees_controller.rb
```diff
@@ -162,6 +163,17 @@ def new
     end
   end
 
+  def validate_name
+    # since current flow uses a sanitized name, lets use the same
```
jshoe: nit: s/lets/let's

spec/controllers/phylo_trees_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    context "with unique name" do
+      it "is valid" do
+        name = "unique_name"
+
+        get :validate_name, params: { format: "json", name: name }
+
+        json_response = JSON.parse(response.body)
+        puts json_response
```
jshoe: I'm assuming you meant to remove the `puts`

tfrcarvalho: Thanks!

app/controllers/phylo_trees_controller.rb
tfrcarvalho: Done

PR #2314, opened 2019-06-10
[Maps] Adjust map framing + other small fixes

jshoe:
### Description
- Small clean-up PR
- Fix some prop errors
- Fix a bug where you wouldn't see the project stats sidebar after switching over from the map
- Fix div on Safari/iOS (so that the map shows up at all)
- Adjust the default map framing to show most of the world

### Test and Reproduce
- Go to data discovery maps, click to map view, view the map framing, click over to the projects tab, see your stats right panel still open.

### DIFF
--- a/app/assets/src/components/views/discovery/mapping/BaseMap.jsx
+++ b/app/assets/src/components/views/discovery/mapping/BaseMap.jsx
@@ -68,6 +68,7 @@ class BaseMap extends React.Component {
           onViewportChange={this.updateViewport}
           mapStyle={styleURL}
           onLoad={this.setCompactAttribution}
+          style={{ position: "absolute" }} // Fixes div in Safari
         >
           {tooltip}
           {markers}
--- a/app/assets/src/components/views/discovery/mapping/BaseMap.jsx
+++ b/app/assets/src/components/views/discovery/mapping/BaseMap.jsx
@@ -102,18 +102,18 @@ BaseMap.propTypes = {
 BaseMap.defaultProps = {
   width: "100%",
   height: "100%",
-  // United States framed
-  latitude: 40,
-  longitude: -98,
-  zoom: 3,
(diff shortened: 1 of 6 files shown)


app/assets/src/components/views/discovery/mapping/BaseMap.jsx
```diff
@@ -68,6 +68,7 @@ class BaseMap extends React.Component {
           onViewportChange={this.updateViewport}
           mapStyle={styleURL}
           onLoad={this.setCompactAttribution}
+          style={{ position: "absolute" }} // Fixes div in Safari
```
tfrcarvalho: can you use css instead?

app/assets/src/components/views/discovery/mapping/BaseMap.jsx
```diff
@@ -101,18 +102,17 @@ BaseMap.propTypes = {
...
-  longitude: -98,
-  zoom: 3,
-  // These bounds prevent panning too far north or south, although you will still see those regions at the widest zoom levels. minZoom level frames most of the world. maxZoom level keeps you at an area about the size of a metropolitan area.
+  // Frame most of the world by default
+  latitude: 27,
+  longitude: 0,
+  zoom: 1.4,
   viewBounds: {
-    minLatitude: -60,
+    minLatitude: -60, // Limit panning too far north or south
```
tfrcarvalho: nit: can you put the comments on the line before?

app/assets/src/components/views/discovery/mapping/BaseMap.jsx
jshoe: in this case I think it was better to just pass it through (otherwise a lot of CSS overriding to do for the markers as well) 
Screen Shot 2019-06-10 at 12 54 29 PM

PR #2317, opened 2019-06-11
prettier config trailing commas

gregdingle:
# Description

As title. See comment in code. 

# Test

See js file changed


MarkAZhang:
Maybe use a config file instead? Easier to manage if we choose to add more rules.
https://prettier.io/docs/en/configuration.html

jshoe:
^ Greg mentioned he couldn't get the config file to get picked up for some reason

gregdingle:
Ya, I spent about 10min debugging then I gave up. I didn't think it was worth the time at this point.

PR #2323, opened 2019-06-11
Remove data discovery and new sample upload feature flags.

MarkAZhang:
Verify that for a non-admin user with no feature flags:
* Going to "/" redirects to DD.
* Clicking on "Upload" redirects to new sample upload.
* The "Save" button is visible on heatmap.


MarkAZhang:
Looks like `prettier` added a lot of trailing commas.

PR #2326, opened 2019-06-11
Fix styling bug with filter tags.

MarkAZhang:
Before:

Screen Shot 2019-06-11 at 3 58 49 PM

After:

Screen Shot 2019-06-11 at 3 58 58 PM


MarkAZhang:
`padding: 0 8px` was unnecessary and was overwriting a different padding rule.

PR #2327, opened 2019-06-12
Fix heatmap when no pipeline run

gregdingle:
# Description

Response to airbrake error. 

# Test

Force `first_pipeline_run` to return nil
See no more error

### DIFF
--- a/app/controllers/visualizations_controller.rb
+++ b/app/controllers/visualizations_controller.rb
@@ -214,7 +214,10 @@ class VisualizationsController < ApplicationController
 
   # The most recent update time of any samples pipeline run.
   def heatmap_ts
-    updated_ats = samples_for_heatmap.map { |sample| sample.first_pipeline_run && sample.first_pipeline_run.updated_at }
+    updated_ats = samples_for_heatmap
+                  .map(&:first_pipeline_run)
+                  .compact
+                  .map(&:updated_at)
     updated_ats.max.to_i
   end
 end
--- a/app/controllers/visualizations_controller.rb
+++ b/app/controllers/visualizations_controller.rb
@@ -214,7 +214,7 @@ class VisualizationsController < ApplicationController
 
   # The most recent update time of any samples pipeline run.
   def heatmap_ts
-    updated_ats = samples_for_heatmap.map { |sample| sample.first_pipeline_run.updated_at }
+    updated_ats = samples_for_heatmap.map { |sample| sample.first_pipeline_run && sample.first_pipeline_run.updated_at }
     updated_ats.max.to_i
   end
 end


app/controllers/visualizations_controller.rb
```diff
@@ -214,7 +214,7 @@ def background_for_heatmap
 
   # The most recent update time of any samples pipeline run.
   def heatmap_ts
-    updated_ats = samples_for_heatmap.map { |sample| sample.first_pipeline_run.updated_at }
+    updated_ats = samples_for_heatmap.map { |sample| sample.first_pipeline_run && sample.first_pipeline_run.updated_at }
```
tfrcarvalho: You are running the query twice here (inside first_pipeline_run)
Load the pipeline run and then use it...

gregdingle: I was wondering about that... does activerecord always re-execute a query?

davidrissato: There is an easier syntax for that:

`sample&.first_pipeline_run&.updated_at`

However I have a question: is it fine to return null for these cases? Shouldn't it be something like:

```
 updated_ats = samples_for_heatmap
                          .select { |sample| sample.first_pipeline_run } 
                          .map { |sample| sample.first_pipeline_run.updated_at }
```

gregdingle: It's okay to return null because of the `max` below but it's probably better to filter with `select`.

gregdingle: From the logs and the docs, it looks like the second call **is** cached: https://edgeguides.rubyonrails.org/caching_with_rails.html#sql-caching . Are we trying to avoid relying on the cache?

davidrissato: Actually, after reading Tiago's comments, I'm wondering if we should fix this query instead:

```
  def samples_for_heatmap
    id = visualization_params[:id]
    sample_ids = id ? Visualization.find(id).sample_ids : params[:sampleIds]
    current_power.samples
                 .where(id: sample_ids)
                 .includes([:host_genome, :pipeline_runs, metadata: [:metadata_field]])
  end
```

Maybe it shouldn't return samples that have no pipeline_runs.

davidrissato: In this case it might execute only once, since the `pipeline_runs` are being included in `samples_for_heatmap` query (see my comment below)

gregdingle: Samples that have no pipeline_runs are filtered out deeper inside the heatmap code.

davidrissato: And is there any use for samples that have no pipeline_run in the heatmap? I'm asking that because it seems that fixing the query would simplify the overall solution

gregdingle: As it happens, when I was working on heatmap perf, I found there are a small number of samples all dating to 2018 that have taxon counts but no pipeline runs. I'm not sure why that would be so I didn't want to change the logic.

davidrissato: I was discussing with Tiago and actually we are puzzled how could a Sample has no pipeline runs. Ok, let's just fix the issue now. We can review this flow more in depth later on.

gregdingle: I'd also love to know why!

PR #2329, opened 2019-06-12
Add duration to pipeline_runs page

gregdingle:
# Description

image

# Test

Load a sample that has run success and failures
See appropriate durations or blanks


app/views/samples/pipeline_runs.html.erb
```diff
@@ -34,8 +34,9 @@
           <th> Job Id </th>
           <th> Job Log </th>
           <th>Remaining reads</th>
-          <th>Pipeline Version</th>
```
gregdingle: removed to save col space

PR #2332, opened 2019-06-12
[Maps] Add a manual rake task for reference in migrating to locationV2 + bug fix for OSM ID integer limit

jshoe:
### Description
- Bug fix for location saving because OSM IDs currently go up to 5 billion, so we needed to change to bigint to save those.
- Adds a manual rake task (mostly for reference) for populating location_v2 values. This can help us do a first pass but will probably still need a lot of correction. Also while our provider contract is not finalized, we have an API rate limit of 2 requests/second.

### Test and Reproduce
- Open up local console and create a Location with an osm_id of a billion.
- Try out the rake task by running `docker-compose run web "rake migrate_to_location_v2"`


jshoe:
Closing this one to save it for later

PR #2333, opened 2019-06-12
[Maps] Add a bug fix for OSM ID integer limit

jshoe:
### Description
- Bug fix for location saving because OSM IDs currently go up to 5 billion, so we needed to change to bigint to save those.

### Test and Reproduce
- Open up local console and create a Location with an osm_id of a billion.


jshoe:
^ A couple locations in Bangladesh were above the previous max apparently and threw errors

PR #2335, opened 2019-06-14
Add missing analytics to heatmap

gregdingle:
# Description

Just some events I noticed were not recorded when doing usage analysis. 

# Tests

open heatmap
click on "collection location" 
see event in http log


app/assets/src/components/common/DetailsSidebar/SampleDetailsMode/MetadataTab.jsx
```diff
@@ -183,7 +183,10 @@ class MetadataTab extends React.Component {
           )}
         {validKeys.map(key => (
           <RequestContext.Consumer>
-            {({ allowedFeatures }) => {
+            {props => {
```
jshoe: thanks for catching and fixing this!

PR #2336, opened 2019-06-17
Add ability to create an AMR Heatmap from the samples page 

lvreynoso:
# Description

- Implement basic AMR Heatmap that displays allele coverage  …
- takes in multiple sample ids as parameters
- displays long sample names
- not able to switch view level or metric yet, but scaffolding is there
- new route at /amr_heatmap
- Can now select samples on the sample browser, click on the heatmap icon, and select "AMR Heatmap" from the dropdown menu.
- Will only be visible to admin users.


gregdingle:
Nice PR @lvreynoso ! 🥇  Is this is a work in progress or is it ready to merge?

lvreynoso:
> Nice PR @lvreynoso ! 🥇 Is this is a work in progress or is it ready to merge?

It should be ready to merge once I merge the latest changes from master!

app/views/amr_heatmap/visualization.html.erb
```diff
@@ -0,0 +1,10 @@
+<div id="AmrHeatmapVis_page">
+</div>
+
+<%= javascript_tag do %>
+  react_component(
+    'AMRHeatmapVis',
+    JSON.parse('<%= raw escape_json(@samples) %>'),
+    'AmrHeatmapVis_page'
```
MarkAZhang: Nit: Use `page_content` which already exists instead of creating another div `AmrHeatmapVis_page`.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -54,6 +54,7 @@ def amr_counts
     render json: amr_data
   end
 
-  def amr_heatmap
+  def visualization
```
MarkAZhang: I would move this controller function to the top and call it `index`.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -54,6 +54,7 @@ def amr_counts
...
+    @samples = { sampleIds: params[:sampleIds] }
```
MarkAZhang: Do `@sampleIds = params[:sampleIds]` here, and `{ sampleIds: JSON.parse('<%= raw escape_json(@sampleIds) %>') }` in the `.erb` file. This makes it more explicit in the `.erb` file exactly which values we're passing as props into `AMRHeatmapVis`.

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -191,21 +192,31 @@ class SamplesView extends React.Component {
         ? mapSidebarSelectedSampleIds
         : selectedSampleIds;
 
-    const log = () =>
-      logAnalyticsEvent("SamplesView_heatmap-icon_clicked", {
-        selectedSampleIds: targetSampleIds.length,
+    const heatmapOptions = [
```
MarkAZhang: If `props.admin` is false, keep the `<a>` link as before, instead of showing a dropdown with just one option.

MarkAZhang: We don't want to change the UI for non-admin users until we launch the feature.

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -191,21 +192,31 @@ class SamplesView extends React.Component {
...
     ) : (
-      <a
-        onClick={log}
-        href={`/visualizations/heatmap?sampleIds=${Array.from(
-          targetSampleIds
-        )}`}
-      >
-        <HeatmapIcon className={cx(cs.icon, cs.heatmap)} />
-      </a>
+      <HeatmapIconDropdown
```
MarkAZhang: It doesn't feel necessary to pull this out into a separate component. (I'm not sure why `DownloadIconDropdown` is a separate component.) We would want to create a separate component is if there is a substantial amount of logic in the component and we expect to use the component across multiple pages. Neither seems to apply here.

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -191,21 +192,31 @@ class SamplesView extends React.Component {
...
+        className={cx(cs.icon, cs.heatmapDropdown)}
+        iconClassName={cx(cs.icon, cs.heatmap)}
+        options={heatmapOptions}
+        onClick={heatmapOption => {
+          logAnalyticsEvent("SamplesView_heatmap-icon_clicked", {
+            selectedSampleIds: targetSampleIds.length,
+            heatmapOption,
+          });
+          let params = Array.from(targetSampleIds).join("&sampleIds[]=");
+          window.location.assign(`${heatmapOption}?sampleIds[]=${params}`);
```
MarkAZhang: Instead of calling `window.location.assign`, I would use `<a>` tags for each option. See how `UserDropdown` in `Header.jsx` does it. It's better to use `<a>` tags whenever possible, because they have better native browser support. (for example, when the user hovers over the link, the dest address will be displayed at the bottom)

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -191,21 +192,31 @@ class SamplesView extends React.Component {
...
+      <HeatmapIconDropdown
+        className={cx(cs.icon, cs.heatmapDropdown)}
+        iconClassName={cx(cs.icon, cs.heatmap)}
+        options={heatmapOptions}
+        onClick={heatmapOption => {
+          logAnalyticsEvent("SamplesView_heatmap-icon_clicked", {
+            selectedSampleIds: targetSampleIds.length,
+            heatmapOption,
+          });
+          let params = Array.from(targetSampleIds).join("&sampleIds[]=");
```
MarkAZhang: You can use `getURLParamString` to combine the sampleId params here.

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -191,21 +192,31 @@ class SamplesView extends React.Component {
...
-        )}`}
-      >
-        <HeatmapIcon className={cx(cs.icon, cs.heatmap)} />
-      </a>
+      <HeatmapIconDropdown
+        className={cx(cs.icon, cs.heatmapDropdown)}
+        iconClassName={cx(cs.icon, cs.heatmap)}
+        options={heatmapOptions}
+        onClick={heatmapOption => {
+          logAnalyticsEvent("SamplesView_heatmap-icon_clicked", {
```
MarkAZhang: Nit: I would use `heatmap-option` instead of `heatmap-icon` since you're picking an option.

app/assets/src/components/views/amr_heatmap/amr_heatmap_vis.scss
```diff
@@ -0,0 +1,19 @@
+// @import "~styles/themes/colors";
```
MarkAZhang: Remove unused and commented lines in this file.

app/assets/src/components/views/amr_heatmap/amr_heatmap_vis.scss
```diff
@@ -0,0 +1,19 @@
...
+  svg {
+    display: block;
+    margin: 0 auto;
+  }
+}
+
+.AMRHeatmapVis {
+}
+
+.tabs {
```
MarkAZhang: Just to make sure: Jenn's mocks had tabs (Taxon, AMR), but you can just do a separate page for now. Having two tabs complicates some of the logic around saving/sharing visualizations.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,262 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Heatmap from "~/components/visualizations/heatmap/Heatmap";
+import { getAMRCounts } from "~/api/amr";
```
MarkAZhang: We have a convention where we group our imports into three groups with a space between them, for readability. 
1) 3rd party libraries, 2) imports starting with `~/` 3) imports starting with `../` or `./`

app/assets/src/api/amr.js
```diff
@@ -0,0 +1,8 @@
+import { get } from "./core";
+
+export const getAMRCounts = sampleIds =>
+  get("amr_heatmap/amr_counts.json", {
+    params: {
+      sampleIds: sampleIds
```
MarkAZhang: Nit: You can just do
```
{
  sampleIds,
}
```

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,262 @@
...
+    let rawResponse = await getAMRCounts(sampleIds);
+    let [geneData, samples] = this.extractData(rawResponse);
+    let alleleToGeneMap = this.mapAllelesToGenes(geneData);
+    this.setState((state, props) => {
+      return {
+        rawData: rawResponse,
+        geneData: geneData,
+        samples: samples,
+        alleleToGeneMap: alleleToGeneMap,
+      };
```
MarkAZhang: In this case, you can just do
```
this.setState({
  rawData: rawResponse,
  geneData,
  samples,
})
```
without need for the function.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,262 @@
...
+    });
+  }
+
+  extractData(rawData) {
```
MarkAZhang: Instead of calling it `rawData`, I would call it something like `rawSamplesWithAMRCounts`. Gives a little more context to what that variable is when reading the code. Similarly, I would rename `extractData`. In general, try to be a bit more specific with naming.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,262 @@
+import React from "react";
```
MarkAZhang: Great job organizing the functions in this file.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,262 @@
...
+        samples: samples,
+        alleleToGeneMap: alleleToGeneMap,
+      };
+    });
+  }
+
+  extractData(rawData) {
+    let genes = {};
+    let sampleLabels = [];
+    rawData.filter(rawSample => rawSample.error === "").forEach(sample => {
```
MarkAZhang: The `genes` data structure seems overly complex. It's possible to implement `computeHeatmapValues` and `computeGeneLabels` with less overall code. You might not even need an intermediate data structure. 

I think the `lodash/fp` utility library will help you a lot. We have a small section on `lodash/fp` [here](https://github.com/chanzuckerberg/idseq-web/blob/master/DEV_GUIDELINES.md). The docs are [here](https://gist.github.com/jfmengels/6b973b69c491375117dc)

Here is some example code, for inspiration.

To get the unique gene labels:
```
// For each sample, map over each entry of sample.amr_counts, and get the "gene" field. 
// We use a lodash/fp shorthand of just specifying the field name "gene".
const getGeneLabelsForSample = sample => map("gene", sample.amr_counts)

// Map the above function over each sample.
const sampleGeneLabels = map(geneGeneLabelsForSample, rawData)
// Example: [["APH-Stph_AGly", "Dha1_Phe", "Tet-38_Tet"], ["APH-Stph_AGly", "Dha1_Phe", "Tet-38_Tet"]]

// Flatten the array and remove duplicates.
const uniqueGeneLabels = uniq(flatten(sampleGeneLabels))
// Example: ["APH-Stph_AGly", "Dha1_Phe", "Tet-38_Tet"]

// You can also do the above with one line. This is more compact but less readable.
// You should pick the option you think is most readable for the situation.
const uniqueGeneLabels = uniq(flatten(
  map(sample => map("gene", sample.amr_counts), rawData)
))
```

To determine, for a given gene label and sample name, whether there is a corresponding depth/coverage value (so that you can generate your heatmap values):

```
// Given sampleName, geneLabel (you'll need to map over all possible values)

// Get the matching sample object.
const sample = find(["sample_name", sampleName], rawData)

// Look for the matching amr count entry in the sample.
const amrCountData = find(["gene", geneLabel], sample.amr_counts)

// If the amr count entry doesn't exist, find() will return null. 
// Use get() to access the fields, since it can handle null values without throwing an error.
const depth = get("depth", amrCountData) || 0
const coverage = get("coverage", amrCountData) || 0
```

You can also use the `keyBy` function to key turn the array into an object, so that you no longer need to use the `find` function, which is O(n).

```
const sampleNameToAmrCounts = keyBy('sample_name', rawData)
```

I would recommend spending some time familiarizing yourself with `lodash/fp` and using it to simplify your logic here.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,262 @@
...
+  componentDidMount() {
+    this.requestAMRCountsData(this.props.sampleIds);
+  }
+
+  componentDidUpdate() {
+    let rows = this.createHeatmapLabels();
+    let columns = this.state.samples;
+    let values = this.computeHeatmapValues(rows);
+    let options = {};
+    this.renderHeatmap(rows, columns, values, options);
```
MarkAZhang: I don't think you need to define `options` here and then pass it in. If you do end up having options, you can just put them inside the `renderHeatmap` function.

MarkAZhang:
Good job on the PR overall! As mentioned in one of the comments, I would look at the `lodash/fp` library to simplify your logic in `<AMRHeatmapVis>`. It's a great library for writing more compact, elegant code.

Also, I'd recommend going ahead and adding the Allele/Gene dropdown to this PR, since the logic is already all there. Generally you want to avoid including "hanging logic" which is never used.

app/controllers/amr_heatmap_controller.rb
jshoe: maybe this is more of a 'show' action? since an 'index' would be for seeing many AMR Heatmaps?

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,262 @@
...
+    });
+    return alleleLabels;
+  }
+
+  createHeatmapLabels() {
+    let viewLevel = this.state.viewLevel;
+    switch (viewLevel) {
+      case METRIC_ALLELES: {
+        return this.createAlleleLabels();
+        break;
```
emilyzhong: nit: the return statement should exit out of the function already, so no need for break

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: Rewrote much of the logic to be more functional and removed the convoluted, unnecessary data structure. Thank you for your input!

lvreynoso:
> Good job on the PR overall! As mentioned in one of the comments, I would look at the `lodash/fp` library to simplify your logic in `<AMRHeatmapVis>`. It's a great library for writing more compact, elegant code.
> 
> Also, I'd recommend going ahead and adding the Allele/Gene dropdown to this PR, since the logic is already all there. Generally you want to avoid including "hanging logic" which is never used.

Adding a dropdown would mean including much more code in the PR, so I'll put in the next one (which is going to include dropdown code anyway).

app/views/amr_heatmap/index.html.erb
```diff
@@ -0,0 +1,10 @@
+<div id="page_content">
```
MarkAZhang: `page_content` is already included on every page (due to `application.html.erb`) so no need to put it here.

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -192,22 +193,63 @@ class SamplesView extends React.Component {
         ? mapSidebarSelectedSampleIds
         : selectedSampleIds;
 
-    const log = () =>
-      logAnalyticsEvent("SamplesView_heatmap-icon_clicked", {
-        selectedSampleIds: targetSampleIds.length,
+    const heatmapOptions = [
```
MarkAZhang: Nit: You can move heatmapOptions inside `if (this.props.admin)`. No need to `push` on a separate line.

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -192,22 +193,63 @@ class SamplesView extends React.Component {
...
+
+      return targetSampleIds.size < 2 ? (
+        <HeatmapIcon className={cx(cs.icon, cs.disabled, cs.heatmap)} />
+      ) : (
+        <BareDropdown
+          hideArrow
+          className={cx(cs.icon, cs.heatmapDropdown)}
+          closeOnClick={true}
+          items={heatmapOptions.map(option => {
+            const params = Array.from(targetSampleIds).join("&sampleIds[]=");
```
MarkAZhang: Nit: Use `getURLParamString` from `helpers/url`. Good practice to use existing code when possible, though I realize it's very little code in this case.

app/assets/src/components/views/amr_heatmap/amr_heatmap_vis.scss
```diff
@@ -0,0 +1,15 @@
...
+  overflow-x: auto;
+  margin-bottom: 300px;
+
+  svg {
+    display: block;
+    margin: 0 auto;
+  }
+}
+
+.AMRHeatmapVis {
```
MarkAZhang: Remove empty block.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,193 @@
...
+        return this.state.alleles;
+      }
+      case METRIC_GENES: {
+        return this.state.genes;
+      }
+    }
+  }
+
+  assembleAlleleValues(alleleLabels) {
+    const sampleData = this.state.samplesWithAMRCounts;
```
MarkAZhang: The code in this function is very similar to `assembleGeneValues`. I think you could consolidate the two functions if you pass in the metric ("gene" or "allele").

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,193 @@
...
+    const alleleValues = {
```
MarkAZhang: Also, it seems a little weird to assemble both depth and coverage values, and then throw one of them away in `computeHeatmapValues`. Could you pass whether you want "depth" or "coverage" into the function, and then just gather that metric?

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,193 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import Heatmap from "~/components/visualizations/heatmap/Heatmap";
+import { getAMRCounts } from "~/api/amr";
+
+import cs from "./amr_heatmap_vis.scss";
+
+const METRIC_ALLELES = "alleles";
+const METRIC_GENES = "genes";
```
MarkAZhang: It's awesome that you have these constants here. (It might make sense to do something similar with "coverage" vs "depth".

I think this might be better named as `VIEW_LEVEL_ALLELE` and `VIEW_LEVEL_GENE`. Since coverage and depth are the "metrics".

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,193 @@
...
+      return { label: allele };
+    });
+
+    return [sampleLabels, geneLabels, alleleLabels];
+  }
+
+  //*** Following functions must be called after the component has updated ***
+  //*** (i.e. after the component has requested AMR data and updated state) ***
+
+  createHeatmapLabels() {
```
MarkAZhang: I would call this `getHeatmapLabels` instead of create.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,193 @@
...
+    const [samples, genes, alleles] = this.createLabels(samplesWithAMRCounts);
+    this.setState({
+      samplesWithAMRCounts,
+      samples,
+      genes,
+      alleles,
+    });
+  }
+
+  createLabels(sampleData) {
```
MarkAZhang: Maybe `extractLabels`.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -0,0 +1,193 @@
...
+
+  async requestAMRCountsData(sampleIds) {
+    const rawSampleData = await getAMRCounts(sampleIds);
+    const samplesWithAMRCounts = rawSampleData.filter(
+      sampleData => sampleData.error === ""
+    );
+    const [samples, genes, alleles] = this.createLabels(samplesWithAMRCounts);
+    this.setState({
+      samplesWithAMRCounts,
+      samples,
```
MarkAZhang: I would call these `sampleLabels`, `geneLabels` instead. Calling it `genes` implies that they're gene objects of some kind, instead of just the names.

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -192,22 +193,63 @@ class SamplesView extends React.Component {
...
-      </a>
-    );
+
+      return targetSampleIds.size < 2 ? (
+        <HeatmapIcon className={cx(cs.icon, cs.disabled, cs.heatmap)} />
+      ) : (
+        <BareDropdown
+          hideArrow
+          className={cx(cs.icon, cs.heatmapDropdown)}
+          closeOnClick={true}
```
MarkAZhang: Nit: closeOnClick is already `true` by default (see BareDropdown's `defaultProps`) so no need for this. Also, when setting a boolean prop to true, you can just do `closeOnClick` instead of `closeOnClick={true}`.

MarkAZhang:
MUCH cleaner. Appreciate your attention to detail. A few more comments.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: Yeah, I noticed that too. Rewriting it.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: Thanks for catching that!

lvreynoso:
Incorporated a lot of suggested changes and cleaned it up. Should be good to go now.

MarkAZhang:
Looks great! Great job with the PR.

PR #2340, opened 2019-06-18
Initial Pipeline Viz React Component

emilyzhong:
# Initial Pipeline Viz React Component
Initial component for pipeline visualization

# Notes
Functionality/features **very likely** to change once final designs are in

# Tests
Go to `samples/[sample id]/stage_results`


app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+const START_NODE_ID = -1;
+const END_NODE_ID = -2;
+const STAGE_BG_COLOR = "#f8f8f8";
+const NODE_COLOR = "#eaeaea";
+const EDGE_COLOR = "#999999";
+
+class PipelineViz extends React.Component {
+  constructor(props) {
+    super(props);
+    this.stageResults = this.props.stageResults.stages;
```
jshoe: for convention, could you pull out all the this.props and this.state in this PR to the tops of the functions like

`const { stageResults } = this.props`

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+
+    if (this.props.admin) {
+      this.stageNames = [
+        "Host Filtering",
+        "GSNAPL/RAPSEARCH alignment",
+        "Post Processing",
+        "Experimental",
+      ];
+    } else {
+      this.stageNames = [
```
jshoe: nit: it'd be more DRY to pull out the first 3 and have admin add Experimental

app/assets/src/components/views/PipelineViz.jsx
jshoe: nit: it's a little confusing to have a prop stageResults and then have an internal var stageResults referring to stages in stageResults instead

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+const STAGE_BG_COLOR = "#f8f8f8";
+const NODE_COLOR = "#eaeaea";
+const EDGE_COLOR = "#999999";
+
+class PipelineViz extends React.Component {
+  constructor(props) {
+    super(props);
+    this.stageResults = this.props.stageResults.stages;
+    this.pipelineVersion = this.props.stageResults.pipeline_version;
+    this.stageGraphsInfo = [];
```
jshoe: since you actually push to stageGraphsInfo later, you might want to store it in the state

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+      zoom: 1,
+    };
+  }
+
+  componentDidMount() {
+    this.renderGraphs();
+  }
+
+  onMouseWheelZoom(e) {
+    const zoomChange = (e.deltaY < 0 ? 1 : -1) * 0.01;
```
jshoe: for the sake of not having magic numbers, could you add a comment explaining why -0.01 and +0.01

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+    this.setState({ zoom: this.state.zoom + zoomChange });
+  }
+
+  toggleStage(index) {
+    const updatedStagesOpened = [...this.state.stagesOpened];
```
jshoe: this spread ... is unnecessary

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+    updatedStagesOpened[index] = !updatedStagesOpened[index];
+    this.setState({ stagesOpened: updatedStagesOpened });
+  }
+
+  modifyStepNames() {
+    // Strips 'PipelineStep[Run/Generate]' from front of each step name.
+    // TODO(ezhong): Consider adding 'name' field to dag_json later.
+    forEach(this.stageResults, (stageData, _) => {
```
jshoe: I think it'd be better to move stageResults to the state as well since you're manipulating it directly here

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+                this[`container${i}`] = ref;
+              }}
+            />
+          </div>
+        </div>
+      );
+    });
+
+    return (
+      <div onWheel={e => this.onMouseWheelZoom(e)}>
```
jshoe: you can just do `this.onMouseWheelZoom` since it's just 1 param

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+  renderInterStageEdges() {
+    this.stageGraphsInfo.forEach((currStageInfo, i) => {
+      if (i == this.stageGraphsInfo.length - 1) {
+        // Create hidden edges to final node for vertical centering of nodes.
+        currStageInfo.data.nodes.forEach(nodeData => {
+          currStageInfo.data.edges.add({
+            from: nodeData.id,
+            to: END_NODE_ID,
+            color: {
+              color: STAGE_BG_COLOR,
```
jshoe: you may want to consider some sort of className based coloring instead of directly setting styles here

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+            currFileNameToOutputtingNode[fileName] = nodeId;
+          });
+        });
+
+        const nextStageData = this.stageResults[this.stageNames[i + 1]];
+        nextStageData.steps.forEach((step, nextNodeId) => {
+          step.in.forEach(inTarget => {
+            nextStageData.targets[inTarget].forEach(fileName => {
+              if (inTarget in nextStageData.given_targets) {
+                fileName = `${
```
jshoe: add const fileName

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+                }
+              }
+            });
+          });
+        });
+      }
+    });
+  }
+
+  renderStageGraph(stageData, container, index) {
```
jshoe: since this function isn't rendering anything you could call it prepareStageGraph or similar

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+    const edgeData = [];
+
+    this.populateNodeAndEdgeData(stageData.steps, nodeData, edgeData);
+
+    const data = {
+      nodes: new DataSet(nodeData),
+      edges: new DataSet(edgeData),
+    };
+
+    const options = {
```
jshoe: since renderStageGraph is going to be called many times and `options` isn't going to change, you might want to move options to a static variable

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,366 @@
...
+          <div
+            className={isOpened ? cs.hidden : cs.stageButton}
+            onClick={() => this.toggleStage(i)}
+          >
+            {stageName}
+          </div>
+
+          <div className={isOpened ? cs.openedStage : cs.hidden}>
+            <div
+              style={{
```
jshoe: you can move this style into a class as well

app/assets/src/components/views/pipeline_viz.scss
```diff
@@ -0,0 +1,51 @@
+@import "~styles/themes/colors";
+@import "~styles/themes/typography";
+
+.pipelineViz {
+  width: 100vw;
```
jshoe: I think you should take out the vw and vh units so that the component isn't bound to always taking up the full screen

tfrcarvalho:
Looks great. A lot of work here... I think we can make an effort to make it a bit more modular. We can discuss that offline.

jshoe:
This is not a blocker but for the future I would lean towards more popular libraries if possible (re: ajainarayanan/react-pan-zoom). Package hijacking is always possible https://blog.npmjs.org/post/180565383195/details-about-the-event-stream-incident

app/assets/src/components/views/PipelineViz.jsx
emilyzhong: Yep, changes to the graph doesn't trigger re-rendering.

jshoe:
also there was an end-of-life notice for the (old?) visjs repo https://github.com/almende/vis/issues/4259

emilyzhong:
> also there was an end-of-life notice for the (old?) visjs repo [almende/vis#4259](https://github.com/almende/vis/issues/4259)

I'll switch to [visjs-network](https://github.com/visjs-community/visjs-network), which seems to be the current version (?) (https://github.com/almende/vis/issues/4259#issuecomment-469318846)

app/assets/src/components/views/PipelineViz.jsx
emilyzhong: I'll be keeping it within the class due to https://github.com/chanzuckerberg/idseq-web/pull/2340/files/183678e556942d731bf8c831f32d9dd1e95d81f4#r296294515 — the `color` fields will be based on the class's proptypes

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,337 @@
...
+          step.class = step.class.substring(3);
+        }
+        if (step.class.substring(0, 8) == "Generate") {
+          step.class = step.class.substring(8);
+        }
+      });
+    });
+  }
+
+  populateNodeAndEdgeData(index, nodeData, edgeData) {
```
jshoe: not a blocker but populateNodeAndEdgeData got quite long and could be split off into more functions

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,337 @@
...
+    });
+    this.adjustGraphNodePositions();
+    this.closeNonactiveSteps();
+  }
+
+  closeNonactiveSteps() {
+    this.stageNames.forEach((stageName, i) => {
+      const graph = this.graphs[i];
+      const stageData = this.props.stageResults.stages[stageName];
+      if (!(stageData.job_status == "STARTED")) {
```
jshoe: nit could just be stageData.job_status !== "STARTED"

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,337 @@
...
+          // Connect beginning steps to input node
+          edgeData.push({ from: START_NODE_ID, to: i });
+        }
+      });
+    });
+    nodeData[stepData.length + 1].level = maxLevel + 1;
+
+    if (index < this.stageNames.length - 1) {
+      // Create edges to output node if it's output files appear in next stage's inputs.
+      const nextStageData = this.props.stageResults.stages[
```
jshoe: not a blocker but we still want to pull out all the props at the top of the function like `const { stageResults, backgroundColor } = this.props`

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,337 @@
...
+              {stageName}
+              <RemoveIcon
+                className={cs.closeButton}
+                onClick={() => this.toggleStage(i)}
+              />
+            </div>
+            <div
+              className={cs.graph}
+              ref={ref => {
+                this[`container${i}`] = ref;
```
jshoe: maybe add a dash after `container`?

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,337 @@
...
+    };
+
+    this.onMouseWheelZoom = this.onMouseWheelZoom.bind(this);
+  }
+
+  componentDidMount() {
+    this.drawGraphs();
+  }
+
+  onMouseWheelZoom(e) {
```
jshoe: our dev guidelines would have this as handleMouseWheelZoom since this is the handler function (and the hook is still called onMouseWheelZoom)

app/controllers/samples_controller.rb
```diff
@@ -981,12 +982,18 @@ def stage_results
...
-        pipeline_stage_results: {
-          pipeline_version: pipeline_run.pipeline_version,
-          stages: stage_info
-        }
+      @results = {
+        pipeline_version: pipeline_run.pipeline_version,
+        stages: stage_info
       }
+      respond_to do |format|
+        format.html do
```
jshoe: you could also just inline these like `format.html { render template: "samples/stage_results" }`

jshoe:
looks pretty good 👍 see last comments

jshoe:
ok and see Travis for an out of memory error T_T

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  render() {
+    const stageContainers = [];
+
+    this.stageNames.forEach((stageName, i) => {
+      const isOpened = this.state.stagesOpened[i];
+
+      stageContainers.push(
+        <div key={i} className={cs.stage}>
```
lvreynoso: Not sure if the order of items in this array will ever change, but if so React recommends against using the index in a key: https://reactjs.org/docs/lists-and-keys.html#keys

app/assets/src/components/views/PipelineViz.jsx
emilyzhong: Split into `populateIntraNodeAndEdgeData` and `populateInterEdgeData`

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,379 @@
+import React from "react";
+import { forEach } from "lodash";
```
tfrcarvalho: Please use `lodash/fp`. It is our team convention to avoid using the same library with two different paradigms. 
It follows a functional programming paradigm, were the subject is typically the last argument.

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,379 @@
...
+    updatedStagesOpened[index] = !updatedStagesOpened[index];
+    this.setState({ stagesOpened: updatedStagesOpened });
+  }
+
+  modifyStepNames() {
+    // Strips 'PipelineStep[Run/Generate]' from front of each step name.
+    // TODO(ezhong): Consider adding 'name' field to dag_json later.
+    const { stageResults } = this.props;
+    forEach(stageResults.stages, (stageData, _) => {
+      stageData.steps.forEach(step => {
```
dcwither: while I'm here, I figured I could help in the review - You should avoid mutating props in react. If you need to change those, you can create a new array and pass it up through a callback. More likely you just need the modified version from it (don't need to pass it back up).

dcwither: if steps is an array, you can use `stageData.steps.map` otherwise you can import `mapValues` from lodash. Btw, I think it's great you're importing single functions from lodash rather than the whole shebang, much easier to replace if necessary!

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,379 @@
...
+    if (this.props.admin) {
+      this.stageNames.push("Experimental");
+    }
+
+    this.state = {
+      stagesOpened: [true, true, true, true],
+      zoom: 1,
+    };
+
+    this.handleMouseWheelZoom = this.handleMouseWheelZoom.bind(this);
```
dcwither: if you're using babel with the right plugins, you should be able to declare the `handleMouseWheelZoom` function as `handleMouseWheelZoom = () =>` in place outside of the constructor

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,379 @@
...
+    this.pipelineVersion = this.props.stageResults.pipeline_version;
+    this.stageGraphsInfo = [];
+    this.graphs = [];
+
+    this.stageNames = [
+      "Host Filtering",
+      "GSNAPL/RAPSEARCH alignment",
+      "Post Processing",
+    ];
+    if (this.props.admin) {
```
dcwither: a nice syntactic sugar for this is the following 
```js
this.stageNames = [
  "Host Filtering",
  "GSNAPL/RAPSEARCH alignment",
  "Post Processing",
  ...(props.admin ? ['Experimental'] : [])
];
```

Probably one to discuss with your team if you like it since it can be divisive.

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,379 @@
...
+import ReactPanZoom from "@ajainarayanan/react-pan-zoom";
+
+import RemoveIcon from "~/components/ui/icons/RemoveIcon";
+import NetworkGraph from "~/components/visualizations/NetworkGraph.js";
+import cs from "./pipeline_viz.scss";
+
+const START_NODE_ID = -1;
+const END_NODE_ID = -2;
+
+class PipelineViz extends React.Component {
```
dcwither: Have you considered a function component with hooks? It looks like it could handle all of this, and be a little cleaner, though I don't want to assume.

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,379 @@
...
+
+  generateInterEdgeData(index) {
+    const { stageResults, backgroundColor } = this.props;
+    const stageData = stageResults.stages[this.stageNames[index]];
+    const stepData = stageData.steps;
+
+    const interEdgeData = [];
+    if (index == this.stageNames.length - 1) {
+      // For final stage, create hidden edges to final node for vertical centering of nodes.
+      stepData.forEach((_, i) => {
```
dcwither: it looks like this is mutating props as well - can have unintended consequences.

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,379 @@
...
+    );
+    currStageGraph.minimizeWidthGivenScale(1.0);
+
+    this.graphs.push(currStageGraph);
+  }
+
+  render() {
+    const stageContainers = [];
+
+    this.stageNames.forEach((stageName, i) => {
```
dcwither: a map would be appropriate here, and avoids creating an array and pushing onto it.

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -0,0 +1,379 @@
...
+              {stageName}
+              <RemoveIcon
+                className={cs.closeButton}
+                onClick={() => this.toggleStage(i)}
+              />
+            </div>
+            <div
+              className={cs.graph}
+              ref={ref => {
+                this[`container-${i}`] = ref;
```
dcwither: you could make an array `this.graphContainers[i] = ref` and initialize `this.graphContainers = []` in the constructor.

app/assets/src/components/views/PipelineViz.jsx
tfrcarvalho: It is possible...

app/assets/src/components/views/PipelineViz.jsx
tfrcarvalho: Just curious @dcwither , what observations triggered you to make that suggestion?

app/assets/src/components/views/PipelineViz.jsx
dcwither: https://reactjs.org/docs/components-and-props.html#props-are-read-only

app/assets/src/components/views/PipelineViz.jsx
dcwither: - use of a `componentDidMount` when a `componentDidUpdate` is probably also necessary (solved by dependencies in the `useEffect` hook). 
- multiple pieces of state that are modified independently, also some that could benefit from useReducer
- I think it would also help identify things that are being mutated that shouldn't be

app/assets/src/components/views/PipelineViz.jsx
emilyzhong: Should be resolved now — most (if not all) references to props have been changed!

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -14,64 +14,65 @@ class PipelineViz extends React.Component {
...
     const { stageResults } = this.props;
-    forEach(stageResults.stages, (stageData, _) => {
-      stageData.steps.forEach(step => {
-        if (step.class.substring(0, 12) == "PipelineStep") {
-          step.class = step.class.substring(12);
+    const stagesWithModifiedNames = mapValues(stageData => {
+      const modifiedStageData = Object.assign({}, stageData);
+      modifiedStageData.steps = modifiedStageData.steps.map(step => {
+        const modifiedStep = Object.assign({}, step);
+        if (modifiedStep.class.substring(0, 12) == "PipelineStep") {
```
dcwither: I think a regex would be simpler here
```js
return {
  ...step,
  class: step.replace(/(PipelineStep|Run|Generate)/, '')
}
```
I'm also replacing the `Object.assign` and the `return` here as well.
NOTE: this assumes the intended behavior was for the `if`s to be mutually exclusive.

It would require some modification if you wanted to do a transform like `PipelineStepRunGeneratehello` -> `hello`

app/assets/src/components/views/PipelineViz.jsx
tfrcarvalho: Good points! 
However this is just the first pass on this component with basic functionality, so I am not sure we want to go with a functional component. But we will definitely keep that in mind.

PR #2343, opened 2019-06-18
JIRA 1007: Fix sample delete button

cdebourcy:
# Description
TODO: test that indexing on create/update still works as expected.

(1) Work around a well-known issue with rails elasticsearch callbacks that caused an error when user deletes a sample (https://github.com/elastic/elasticsearch-rails/issues/567)
(2) Speed the sample destroy method up by deleting dependent records in bulk
(3) Add a nanobar so user has a sense of the delete action being in progress

# Tests
Bug fix, a priori no new tests required (test "joe cannot delete public_sample" existed). I added an additional one just to be safe: "joe cannot delete expired_sample".


app/assets/src/components/views/SampleView/SampleViewControls.jsx
```diff
@@ -10,6 +10,7 @@ import {
   getDownloadDropdownOptions,
   getLinkInfoForDownloadOption,
 } from "~/components/views/report/utils/download";
+import Nanobar from "nanobar";
```
jshoe: nit: can move to the top group for 3rd-party imports

app/assets/src/components/views/SampleView/SampleViewControls.jsx
```diff
@@ -34,7 +35,13 @@ class SampleViewControls extends React.Component {
 
   deleteSample = async () => {
     const { sample, project } = this.props;
+    let nanobar = new Nanobar({
+      id: "prog-bar",
+      class: "prog-bar",
+    });
+    nanobar.go(30);
     await deleteSample(sample.id);
```
jshoe: a further improvement would be to console.log any errors or display them but this is fine

cdebourcy: thx

app/assets/src/components/views/SampleView/SampleViewControls.jsx
cdebourcy: ok thx

PR #2344, opened 2019-06-19
[IDSEQ-1075] [Maps] Fix visual styles according to Designs

jshoe:
### Description
- See full list here: https://jira.czi.team/browse/IDSEQ-1075
- Assorted design fixes according to mock here: https://chanzuckerberg.invisionapp.com/d/main#/console/17624454/365364533/preview
- Tighten up map sidebar styles for the sample/project rows
- Adjust styles of the tables, map controls, tooltips, and font sizes.
- Last change is adding RequestContext to PhyloTreeListView (so that it can be properly loaded for its metadata sidebar). Last one!
- Also change it so that in the MapPreviewSidebar, when you select a project, it'll exit the map view and exit preview mode to take you to the whole samples list.

### Demo
Screen Shot 2019-06-18 at 6 28 13 PM
Screen Shot 2019-06-18 at 6 28 20 PM
Screen Shot 2019-06-18 at 6 30 19 PM

### Tests
- Go to the data discovery views, go to the map, select locations, observe the updated visual styles above.

### DIFF
--- a/app/assets/src/components/views/discovery/BaseDiscoveryView.jsx
+++ b/app/assets/src/components/views/discovery/BaseDiscoveryView.jsx
@@ -15,6 +15,7 @@ class BaseDiscoveryView extends React.Component {
       columns,
       data,
       handleRowClick,
+      headerClassName,
       initialActiveColumns,
       protectedColumns,
       rowClassName,
@@ -25,7 +26,8 @@ class BaseDiscoveryView extends React.Component {
       <Table
         columns={columns}
         data={data}
-        defaultRowHeight={rowHeight || 68}
+        defaultRowHeight={rowHeight}
+        headerClassName={headerClassName}
         initialActiveColumns={initialActiveColumns}
         onRowClick={handleRowClick}
         protectedColumns={protectedColumns}
@@ -39,12 +41,14 @@ class BaseDiscoveryView extends React.Component {
 BaseDiscoveryView.defaultProps = {
   columns: [],
   data: [],
+  rowHeight: 68,
 };
 
 BaseDiscoveryView.propTypes = {
   columns: PropTypes.array,
   data: PropTypes.array,
   handleRowClick: PropTypes.func,
+  headerClassName: PropTypes.string,
   initialActiveColumns: PropTypes.arrayOf(PropTypes.string),
   protectedColumns: PropTypes.arrayOf(PropTypes.string),
   rowClassName: PropTypes.string,
--- a/app/assets/src/components/views/discovery/BaseDiscoveryView.jsx
+++ b/app/assets/src/components/views/discovery/BaseDiscoveryView.jsx
@@ -1,5 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
+import cx from "classnames";
 
 import { Table } from "~/components/visualizations/table";
 // CSS file must be loaded after any elements you might want to override
@@ -16,17 +17,19 @@ class BaseDiscoveryView extends React.Component {
       handleRowClick,
       initialActiveColumns,
       protectedColumns,
+      rowClassName,
+      rowHeight,
     } = this.props;
 
     return (
       <Table
         columns={columns}
         data={data}
-        defaultRowHeight={68}
+        defaultRowHeight={rowHeight || 68}
         initialActiveColumns={initialActiveColumns}
         onRowClick={handleRowClick}
         protectedColumns={protectedColumns}
-        rowClassName={cs.tableDataRow}
+        rowClassName={cx(rowClassName, cs.tableDataRow)}
         sortable
       />
     );
@@ -44,6 +47,8 @@ BaseDiscoveryView.propTypes = {
   handleRowClick: PropTypes.func,
   initialActiveColumns: PropTypes.arrayOf(PropTypes.string),
   protectedColumns: PropTypes.arrayOf(PropTypes.string),
+  rowClassName: PropTypes.string,
+  rowHeight: PropTypes.number,
 };
 
 export default BaseDiscoveryView;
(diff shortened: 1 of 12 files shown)


app/assets/src/components/views/discovery/BaseDiscoveryView.jsx
```diff
@@ -16,17 +17,19 @@ class BaseDiscoveryView extends React.Component {
...
+      rowClassName,
+      rowHeight,
     } = this.props;
 
     return (
       <Table
         columns={columns}
         data={data}
-        defaultRowHeight={68}
+        defaultRowHeight={rowHeight || 68}
```
tfrcarvalho: Do you know the origin of the 68?
Can you switch it to a prop with default value ?

app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -608,13 +608,15 @@ class DiscoveryView extends React.Component {
     const { mapSidebarTab } = this.state;
     this.setState(
       {
+        currentDisplay: "table",
```
tfrcarvalho: Can we select projects from the map?

app/assets/src/components/views/discovery/mapping/base_map.scss
```diff
@@ -10,8 +10,12 @@
     position: absolute;
     bottom: 50px;
     right: 15px;
+    box-shadow: 0 0 0 1px rgba(0, 0, 0, 0.1);
```
tfrcarvalho: If we can't reuse a box-shadow already defined in `_elements` (always double check with Jenn to push for consistency), move this definition there and use a variable here.

app/assets/src/components/views/discovery/mapping/base_map.scss
```diff
@@ -10,8 +10,12 @@
...
 
     button {
```
tfrcarvalho: Should this use a class instead? Or is part of 3rd party lib?

app/assets/src/components/views/discovery/mapping/map_preview_sidebar.scss
```diff
@@ -48,7 +48,24 @@
 
 .tableDataRow {
   align-items: flex-start;
-  padding: 11px 0;
+  padding: 10px 0;
+  margin-left: 0;
+
+  // Override defaults
+  &:global(.ReactVirtualized),
```
tfrcarvalho: There are props to pass class names to customize this I believe. Can you search within `BaseTable` and `ReactVirtualize` libraries?

app/views/phylo_trees/index.html.erb
```diff
@@ -5,6 +5,7 @@
       taxon: JSON.parse('<%= raw escape_json(@taxon) %>'),
       phyloTrees: JSON.parse('<%= raw escape_json(@phylo_trees) %>'),
       allowedFeatures: JSON.parse('<%= raw escape_json(@current_user.allowed_feature_list) %>'),
-    }, 'PhyloTreeListView_page');
+    }, 'PhyloTreeListView_page',
+    JSON.parse('<%= raw escape_json(request_context)%>'));
```
tfrcarvalho: Why did you need to add this on this PR?

app/assets/src/components/views/discovery/BaseDiscoveryView.jsx
jshoe: no idea where the 68 was from actually. Moved to default props 👍

app/assets/src/components/views/discovery/DiscoveryView.jsx
jshoe: yes this is referring to clicking on the project in the right pane table while in the map

app/assets/src/components/views/discovery/mapping/base_map.scss
jshoe: changed 👍

app/assets/src/components/views/discovery/mapping/base_map.scss
jshoe: yes and yes (assuming you meant the button) 👍

app/assets/src/components/views/discovery/mapping/map_preview_sidebar.scss
jshoe: yes it looks like we get headerClassName and rowClassName https://github.com/bvaughn/react-virtualized/blob/master/docs/customizingStyles.md

I moved out the header but wasn't able to override rowColumn/headerColumn specifically

Screen Shot 2019-06-19 at 2 46 48 PM

app/views/phylo_trees/index.html.erb
jshoe: wasn't necessary 👍 probably could've bundled with a different PR

app/assets/src/components/views/discovery/DiscoveryView.jsx
tfrcarvalho: Shouldnt we continue on the map?

app/views/phylo_trees/index.html.erb
tfrcarvalho: OK, fine to change if you want to bundle it...

app/assets/src/components/views/discovery/DiscoveryView.jsx
jshoe: Jenn determined that it should set to the table view instead https://jira.czi.team/browse/IDSEQ-1087

jshoe: cc @happyimadesignr

tfrcarvalho: 👌

PR #2345, opened 2019-06-19
[Maps] improve perf metadata by sample

tfrcarvalho:
# Description

Improve performance of the `metadata_multiget` method by eager loading (v2) locations. 

# Tests

* Make a request for samples (the domain or specified samples should have regular metadata and location v2. e.g. `http://localhost:3000/samples/index_v2.json?domain=my_data&sampleIds=<sampleId>`
* Verify that metadata field loads correctly, include extra object for location v2.


app/models/metadatum.rb
```diff
@@ -277,7 +277,7 @@ def self.new_without_save(sample, key, value)
   def validated_value
     # Special case for Location objects
     if metadata_field.base_type == Metadatum::LOCATION_TYPE
-      location_id ? Location.find(location_id).attributes : string_validated_value
+      Hash[Location::DEFAULT_LOCATION_FIELDS.map { |k| [k, location[k]] }]
```
jshoe: I think you still want the conditional around this for `location ? : string_validated_value`. If location_id is nil then there's no Location object and it's just string_validated_value.

PR #2347, opened 2019-06-19
[Maps] Add more analytics tracking

jshoe:
# Description
- This adds more analytics tracking to the main map interactions. If you see more actions you think are missing, let me know.

# Tests
- Open up the map, click around to trigger these event handlers. See events in the Segment debugger (e.g. using staging key).

### DIFF
--- a/app/assets/src/components/views/discovery/mapping/DiscoveryMap.jsx
+++ b/app/assets/src/components/views/discovery/mapping/DiscoveryMap.jsx
@@ -2,7 +2,7 @@ import React from "react";
 import { Marker } from "react-map-gl";
 import { get, isEmpty, upperFirst } from "lodash/fp";
 
-import { withAnalytics } from "~/api/analytics";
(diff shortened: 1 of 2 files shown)


app/assets/src/components/views/discovery/mapping/DiscoveryMap.jsx
```diff
@@ -51,6 +53,10 @@ class DiscoveryMap extends React.Component {
       />
     );
     this.setState({ tooltip, tooltipShouldClose: false });
+
+    logAnalyticsEvent("DiscoveryMap_marker_hovered", {
+      locationId: hoverInfo.id,
```
tfrcarvalho: nit: `hoverInfo` could be `locationInfo` to be more explicit...

jshoe: 👍

PR #2348, opened 2019-06-20
Additional information for pipeline viz endpoint

emilyzhong:
# Additional information for pipeline viz endpoint

Adds the following information to the `stage_results` endpoint for pipeline visualization:
* Pipeline version
* Job status for each stage

# Notes

* Pipeline version needed for file matching for edges in front end ([see TODO here](https://github.com/chanzuckerberg/idseq-web/pull/2340/files#diff-749f700b60f6b84cf7f1ce76d874bff2R150)) + informing user of variations of pipeline due to version
* Job status for each stage also for front end: knowing which stage to expand initially ([see TODO here](https://github.com/chanzuckerberg/idseq-web/pull/2340/files#diff-749f700b60f6b84cf7f1ce76d874bff2R295))

# Tests

Updated both rails tests and rspec tests that are related to modified endpoint

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -969,14 +969,24 @@ class SamplesController < ApplicationController
   end
 
   # GET /samples/:id/stage_results
+  # GET /samples/:id/stage_results.json
   def stage_results
     if current_user.allowed_feature_list.include?("pipeline_viz")
-      @results = {}
-      @sample.first_pipeline_run.pipeline_run_stages.each do |stage|
+      pipeline_run = @sample.first_pipeline_run
+      pipeline_version = pipeline_run.report_info_params[:pipeline_version]
+
+      stage_info = {}
+      pipeline_run.pipeline_run_stages.each do |stage|
         if stage.name != "Experimental" || current_user.admin?
-          @results[stage.name] = JSON.parse stage.dag_json
+          stage_info[stage.name] = stage.dag_json.nil? ? {} : JSON.parse(stage.dag_json)
+          stage_info[stage.name][:job_status] = stage.job_status
         end
       end
+
+      @results = {
+        pipeline_version: pipeline_version,
+        stages: stage_info
+      }
       render json: { pipeline_stage_results: @results }
     else
       render(json: {
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -971,10 +971,9 @@ class SamplesController < ApplicationController
   # GET /samples/:id/stage_results
   # GET /samples/:id/stage_results.json
   def stage_results
-    if current_user.allowed_feature_list.include?("pipeline_viz")
-      pipeline_run = @sample.first_pipeline_run
-      pipeline_version = pipeline_run.report_info_params[:pipeline_version]
-
+    pipeline_run = @sample.first_pipeline_run
+    feature_allowed = current_user.allowed_feature_list.include?("pipeline_viz")
+    if feature_allowed && pipeline_run
       stage_info = {}
       pipeline_run.pipeline_run_stages.each do |stage|
         if stage.name != "Experimental" || current_user.admin?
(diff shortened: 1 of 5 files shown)


jshoe:
@emilyzhong did you mean to add reviewers?

emilyzhong:
I was planning on talking to Tiago first but now that I look at it hopefully my PR description is detailed enough to explain itself!

app/controllers/samples_controller.rb
```diff
@@ -969,14 +969,24 @@ def results_folder
   end
 
   # GET /samples/:id/stage_results
+  # GET /samples/:id/stage_results.json
   def stage_results
     if current_user.allowed_feature_list.include?("pipeline_viz")
-      @results = {}
-      @sample.first_pipeline_run.pipeline_run_stages.each do |stage|
+      pipeline_run = @sample.first_pipeline_run
+      pipeline_version = pipeline_run.report_info_params[:pipeline_version]
```
tfrcarvalho: Why not just use `pipeline_run.pipeline_version` ?

app/controllers/samples_controller.rb
```diff
@@ -969,14 +969,24 @@ def results_folder
...
+      stage_info = {}
+      pipeline_run.pipeline_run_stages.each do |stage|
         if stage.name != "Experimental" || current_user.admin?
-          @results[stage.name] = JSON.parse stage.dag_json
+          stage_info[stage.name] = stage.dag_json.nil? ? {} : JSON.parse(stage.dag_json)
+          stage_info[stage.name][:job_status] = stage.job_status
         end
       end
+
+      @results = {
```
tfrcarvalho: No need to assign to a variable. Just use it below directly

test/controllers/samples_controller_test.rb
```diff
@@ -214,10 +214,13 @@ class SamplesControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
 
     results = @response.parsed_body["pipeline_stage_results"]
```
tfrcarvalho: Arent these tests covered in spec? If so, you do not need to continue to update them. Should actually remove them.

app/controllers/samples_controller.rb
tfrcarvalho: Add a check on pipeline_run: Lets make sure we do not crash if there is no pipeline run yet... It is not supposed to happen when you access this endpoint but better to protect the code.

emilyzhong: Updated — didn't realize pipeline_version could be accessed straight from pipeline_run!

app/controllers/samples_controller.rb
```diff
@@ -969,20 +969,31 @@ def results_folder
...
-      @results = {}
-      @sample.first_pipeline_run.pipeline_run_stages.each do |stage|
+    pipeline_run = @sample.first_pipeline_run
+    feature_allowed = current_user.allowed_feature_list.include?("pipeline_viz")
+    if feature_allowed && pipeline_run
+      stage_info = {}
+      pipeline_run.pipeline_run_stages.each do |stage|
         if stage.name != "Experimental" || current_user.admin?
-          @results[stage.name] = JSON.parse stage.dag_json
+          stage_info[stage.name] = stage.dag_json.nil? ? {} : JSON.parse(stage.dag_json)
```
tfrcarvalho: nit can be more compact using `stage_info[stage.name] = JSON.PARSE(stage.dag_json || "{}")`

app/controllers/samples_controller.rb
```diff
@@ -969,20 +969,31 @@ def results_folder
   end
 
   # GET /samples/:id/stage_results
+  # GET /samples/:id/stage_results.json
   def stage_results
```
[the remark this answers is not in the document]
emilyzhong: Oops I added the html response in my React WIP pr, but not in this one — will move this comment over!

PR #2349, opened 2019-06-20
[AMR Heatmap] Add "Metric" filter and "View Level" options to AMR Heatmap

lvreynoso:
# Description

Adds a header and dropdown menu to the AMR Heatmap, with options to change between two metrics (depth and coverage) and select between two view levels (gene and allele). The heatmap will update instantly when a new option is selected.

# Notes

* Some state is lifted from AMRHeatmapVis to AMRHeatmapView, and so is the request for sample data

# Tests

* Added Rspec tests for `index` method (`amr_heatmap` endpoint)
* Go to All Data page, pick a bunch of samples with AMR Counts (like tb-flash), click on Heatmap Icon and select "AMR Heatmap" from the dropdown, when the heatmap loads click on the dropdowns and select different options to see if the heatmap is responsive.


app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+  updateOptions = options => {
+    let newOptions = Object.assign({}, this.state.selectedOptions, options);
+    this.setState({
+      selectedOptions: newOptions,
+    });
+  };
+
+  renderVisualization() {
+    if (this.state.loading === true) {
```
emilyzhong: Nit: doesn't seem that `this.state.loading` will be set as anything besides a boolean (unless I'm missing something), so can just put `if (this.state.loading) { ... }` (+ similar below)

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -104,10 +103,28 @@ export default class AMRHeatmapVis extends React.Component {
...
 
-  renderHeatmap() {
+  assembleHeatmapData() {
     const rows = this.getHeatmapLabels();
     const columns = this.state.sampleLabels;
     const values = this.computeHeatmapValues(rows);
+    return [rows, columns, values];
+  }
+
+  createHeatmap() {
```
emilyzhong: `createHeatmap` and `updateHeatmap` seem very similar (they both call `assembleHeatmapData`, then call functions that take in the same arguments) — I'd maybe combine the logic of whether to update vs render + corresponding update/rendering call in `renderHeatMap` and have all that in one method, so when reading the code there's less code-tracing going around between similar methods (could also just be a preferential thing though)

lvreynoso:
> 🌶 🗺 !!!
> A lot of these comments are based off my own coding preferences so feel free to take them with a grain of salt! 🧂

I appreciate them all the same :)

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
lvreynoso: It is supposed to be just a boolean, I just like being explicit about it since JS is dynamically typed

lvreynoso: Ah never mind, AirBnB style guide says to just do `if (this.state.loading) {...}` so I'll fix it

lvreynoso:
Looks like it's at a good state now, time for review!

jshoe:
^ For manually testing, you could write a little test plan like "Go to this URL, click the dropdown, change the filter to coverage, see coverage data" for the interactions.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -2,7 +2,7 @@ class AmrHeatmapController < ApplicationController
   before_action :admin_required
 
   def index
-    @sample_ids = params[:sampleIds]
+    @sample_ids = params[:sampleIds].map(&:to_i)
   end
 
   # GET /amr_heatmap/amr_heatmap.json
```
jshoe: this route should be amr_heatmap/amr_counts.json

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,65 @@
...
+import cs from "./amr_heatmap_view.scss";
+
+export default class AMRHeatmapControls extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  generateChangeFunction(filter) {
+    const changeFunction = option => {
+      if (option === this.props.selectedOptions[filter]) {
```
jshoe: we usually have a convention of pulling out all the props and state at the top of the function like

`const { selectedOptions, onSelectedOptionsChange } = this.props;`

jshoe: you could probably just replace this function with something like

`() => option !== selectedOptions[filter] && onSelectedOptionsChange({ [filter]: option })`

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,65 @@
...
+        return;
+      }
+      this.props.onSelectedOptionsChange({ [filter]: option });
+    };
+    return changeFunction;
+  }
+
+  renderFilterDropdowns() {
+    const filtersList = [];
+    this.props.filters.forEach((_, filter) => {
```
jshoe: this would be more succinct as `filtersList = filters.map` right into an array

jshoe: it might be easier to just have `filters` as an object instead of a Map

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,65 @@
...
+      filtersList.push(
+        <div className="col s3">
+          <Dropdown
+            fluid
+            rounded
+            options={this.props.filters.get(filter).options}
```
jshoe: since you're already iterating over `filter`, you can just do `filter.options`, right?

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,65 @@
...
+      );
+    });
+    return filtersList;
+  }
+
+  render() {
+    return (
+      <div className={cs.menu}>
+        <Divider />
+        <div className={`${cs.filterRow} row`}>
```
jshoe: you should actually use `classnames` instead of interpolation for this one (esp if you ever need conditional classnames)

`import cx from "classnames";`
and
`cx(cs.filterRow, row)`

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+AMRHeatmapControls.propTypes = {
+  filters: PropTypes.instanceOf(Map).isRequired,
+  selectedOptions: PropTypes.shape({
+    metric: PropTypes.string,
+    viewLevel: PropTypes.string,
+  }),
+  onSelectedOptionsChange: PropTypes.func.isRequired,
+  data: PropTypes.bool,
```
jshoe: since you have `data` as a bool I would recommend calling it `hasData`

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
lvreynoso: The difference is minimal and using a `Map` makes it easy to order the dropdowns, so I'll keep it a Map.

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
lvreynoso: `filter` is a key here, not the whole entry, but yeah I could just use the entry (which is the `_` parameter I didn't use). Updated

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
lvreynoso: I pulled out the props at the top of the function, but for the sake of readability I kept the control flow clearer (but shortened it). Because of the way the Dropdown component handles the `onChange` callback function, there's no way to access `filter` in the callback, which is why I decided to generate it beforehand.

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
MarkAZhang: I didn't know this about `Maps`! Very cool.

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,66 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import { Divider } from "~/components/layout";
+import { Dropdown } from "~ui/controls/dropdowns";
+
+import cs from "./amr_heatmap_view.scss";
+
+export default class AMRHeatmapControls extends React.Component {
+  constructor(props) {
```
MarkAZhang: If it's just a super() call, you can omit the constructor.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -0,0 +1,131 @@
...
+            </ViewHeader.Content>
+          </ViewHeader>
+        </NarrowContainer>
+        <StickyContainer>
+          <Sticky>
+            {({ style }) => (
+              <div style={style}>
+                <NarrowContainer>
+                  <AMRHeatmapControls
+                    filters={this.assembleControlOptions()}
```
MarkAZhang: Nit: Maybe we can pick one of "controls" or "filters" since it looks like we're using both.

To me, "controls" is the more generic term. It can mean filters or display options or other things. Might be better here.

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+AMRHeatmapControls.propTypes = {
+  filters: PropTypes.instanceOf(Map).isRequired,
+  selectedOptions: PropTypes.shape({
+    metric: PropTypes.string,
+    viewLevel: PropTypes.string,
+  }),
+  onSelectedOptionsChange: PropTypes.func.isRequired,
+  hasData: PropTypes.bool,
```
MarkAZhang: Nit: I think `isDataReady` is a bit more descriptive than `hasData`.

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,66 @@
...
+      );
+    });
+    return filtersList;
+  }
+
+  render() {
+    return (
+      <div className={cs.menu}>
+        <Divider />
+        <div className={`${cs.filterRow} row`}>
```
MarkAZhang: We have a convention to use `cx` for CSS classes in this case. See examples in other files.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -0,0 +1,131 @@
...
+      samplesWithAMRCounts,
+      sampleIds,
+      loading: false,
+    });
+  }
+
+  assembleControlOptions() {
+    // A Map is used here to preserve the order that the filters will be
+    // presented in the controls element
+    return new Map([
```
MarkAZhang: I was excited in a different comment to learn that Maps have ordered keys.

Still, I think it might be better to use an Array of objects here. You can just store the name of the option as a field in the object. No reason to use a map:
```
[
  {
    key: "viewLevel",
    options: VIEW_LEVELS,
    label: "View Level"
  },
  ...
]
```

A Map also works, but I don't think we use it anywhere else in our code base. I prefer not to introduce new ways of doing things unless we need them.

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,66 @@
...
+        onSelectedOptionsChange({ [filter]: option });
+      }
+    };
+    return changeFunction;
+  }
+
+  renderFilterDropdowns() {
+    const { filters, selectedOptions, hasData } = this.props;
+    const filtersList = [];
+    filters.forEach((filterOptionData, filter) => {
```
MarkAZhang: If you use an Array of objects for filters (see comment below), you can just do 

```
const filtersList = filters.map(...)
```

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -0,0 +1,131 @@
...
+    this.setState({
+      selectedOptions: newOptions,
+    });
+  };
+
+  renderVisualization() {
+    if (this.state.loading) {
+      return (
+        <p className={cs.loadingIndicator}>
+          <i className="fa fa-spinner fa-pulse fa-fw" />
```
jshoe: you should use <LoadingIcon> for this

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,66 @@
...
+    const { selectedOptions, onSelectedOptionsChange } = this.props;
+    const changeFunction = option => {
+      if (option !== selectedOptions[filter]) {
+        onSelectedOptionsChange({ [filter]: option });
+      }
+    };
+    return changeFunction;
+  }
+
+  renderFilterDropdowns() {
```
MarkAZhang: I really appreciate the effort to generalize functions and not duplicate code!

This is fine for now, but you'll probably need to generalize this further if we ever introduce a non-Dropdown control element.

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -0,0 +1,66 @@
...
+    const { filters, selectedOptions, hasData } = this.props;
+    const filtersList = [];
+    filters.forEach((filterOptionData, filter) => {
+      filtersList.push(
+        <div className="col s3" key={filter}>
+          <Dropdown
+            fluid
+            rounded
+            options={filterOptionData.options}
+            onChange={this.generateChangeFunction(filter)}
```
MarkAZhang: What you have works. You can pass `filter` into the function like this, which is a bit more concise:

```

onChange={option => this.handleFilterChange(option, filter)}

// Showing the handleFilterChange function for clarity
const handleFilterChange = (option, filter) => {
  if (option !== selectedOptions[filter]) {
    onSelectedOptionsChange({ [filter]: option });
  }
};
```

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -14,26 +13,14 @@ export default class AMRHeatmapVis extends React.Component {
     super(props);
 
     this.state = {
-      viewLevel: VIEW_LEVEL_GENES,
-      metric: "coverage",
+      loading: true,
```
MarkAZhang: Don't think we still need the loading code here anymore, since this component isn't displayed until the data is ready.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -104,10 +99,22 @@ export default class AMRHeatmapVis extends React.Component {
     return heatmapValues;
   }
 
-  renderHeatmap() {
+  prepareHeatmapData() {
```
MarkAZhang: I'd recommend calling this function `updateHeatmap` and renaming `renderHeatmap` to `initializeHeatmap`, to better describe what the functions are doing.

`render` typically suggests React rendering to me.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -37,18 +37,18 @@ def amr_counts
       end
       amr_data << {
         sample_name: sample.name,
-        sample_id: sample.id,
+        sample_id: sample.id.to_i,
```
jshoe: sample.id should be an int already

app/controllers/amr_heatmap_controller.rb
```diff
@@ -2,10 +2,10 @@ class AmrHeatmapController < ApplicationController
   before_action :admin_required
 
   def index
-    @sample_ids = params[:sampleIds]
+    @sample_ids = params[:sampleIds].map(&:to_i)
```
jshoe: if params[:sampleIds] could be strings, then you would want to also map to_i at the top of #amr_counts, right?

MarkAZhang:
Made some comments. Looks great overall!

jshoe:
> Made some comments. Looks great overall!

awesome heatmap! 🎉

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
lvreynoso: This is a more elegant solution, thank you!

PR #2351, opened 2019-06-21
[READY TO MERGE] JIRA 1014: add wikipedia license to taxon descriptions

cdebourcy:
image

### DIFF
--- a/app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/taxon_details_mode.scss
+++ b/app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/taxon_details_mode.scss
@@ -75,7 +75,11 @@
   line-height: $line-height-small;
 
   .wikiLink {
-    color: $primary-light !important;
+    color: $black !important;
     cursor: pointer;
+
+    &:hover {
+      text-decoration: underline;
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/taxon_details_mode.scss
```diff
@@ -67,3 +67,19 @@
     cursor: pointer;
   }
 }
+
+.wikiLicense {
+  padding-top: 10px;
+  color: $medium-grey;
+  font-size: $font-size-small;
```
tfrcarvalho: Can we define a font mixin?
@happyimadesignr Do we have a whole font style for the small attribution text?

happyimadesignr: @tfrcarvalho I'd use this for any attribution or caption type text. but i dont know yet where else this will be

@cdebourcy @MarkAZhang I thought our naming for this font size was $font-body-xxxs?

tfrcarvalho: This file still references the old `styles/_typography` and not the new one `styles/themes/_typography`. Need to make more explicit that the previous one is deprecated, but since this file was already using the previous on, I guess this is fine.

PR #2352, opened 2019-06-21
tweak rapsearch concurrency and chunk size for r5d.metal

boris-dimitrov:
Deploy this on prod ONLY AFTER deploying the matching idseq-infra change.

```
  SERVER      CHUNK     CHUNK           PARALLEL      COST           COST REDUCTION
  TYPE         SIZE     ELAPSED TIME    CHUNKS        (READS         FACTOR (UNITS OF
             (READS)    (SECONDS)       PER SERVER    PER DOLLAR)    1/i3.metal)
  -----------------------------------------------------------------------------

  r5d.metal   80,000       555             8          600,774        4.82

  r5d.metal   40,000       454             8          368,000        2.95

  i3.metal    20,000       463             4          124,655        1.00
```


boris-dimitrov:
The idseq infra change has been deployed.  Merging.

PR #2353, opened 2019-06-24
Save Country+State geographical levels on all locations and load them

jshoe:
# Description
- Add extra geosearching and saving for all Country and State levels when you save a Location. We agreed to skip Subdivison/City for now to simplify implementation and stay under provider per-second rate limits.
- Add extra_parents to LocationsController#sample_locations endpoint response. Frontend clustering is to-be-implemented so this may change.
- Better implementation of determining geo_levels based on a variety of similar OpenStreetMap names for similar levels.
- Add country_id/state_id/subdivision_id/city_id columns to Locations. This is for more efficient read comparisons so you can go straight to a parent or child level.
- Update tests.

# Notes
- Example of a merged sample_locations response with USA + California:
Screen Shot 2019-06-25 at 11 16 26 AM

# Tests
- See tests

### DIFF
--- a/app/helpers/location_helper.rb
+++ b/app/helpers/location_helper.rb
@@ -26,7 +26,6 @@ module LocationHelper
       country_name: address[country_key] || "",
       state_name: address[state_key] || "",
       subdivision_name: address[subdivision_key] || "",
-      # Normalize extra provider fields to city. normalizecity param doesn't work all the time.
       city_name: address[city_key] || "",
       # Round coordinates to enhance privacy
       lat: body["lat"] ? body["lat"].to_f.round(2) : nil,
--- a/app/helpers/location_helper.rb
+++ b/app/helpers/location_helper.rb
@@ -14,6 +14,8 @@ module LocationHelper
                   Location::STATE_LEVEL
                 elsif country_key
                   Location::COUNTRY_LEVEL
+                else
+                  ""
                 end
 
     loc = {
--- a/app/helpers/location_helper.rb
+++ b/app/helpers/location_helper.rb
@@ -92,4 +92,9 @@ module LocationHelper
     samples.where(metadata: { string_validated_value: query })
            .or(samples.where(metadata: { locations: { name: query } }))
   end
+
+  def self.set_parent_ids(location, parent_level_ids)
+    parent_level_ids.each { |level, id| location["#{level}_id"] = id }
+    location
+  end
 end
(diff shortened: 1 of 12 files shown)


app/models/location.rb
```diff
@@ -11,16 +11,32 @@ class Location < ApplicationRecord
...
+  SUBDIVISION_NAMES = %w[county state_district district].freeze
+  CITY_NAMES = %w[city city_distrct locality town borough municipality village hamlet quarter neighbourhood suburb].freeze
+
   # Base request to LocationIQ API
   def self.location_api_request(endpoint_query)
     raise "No location API key" unless ENV["LOCATION_IQ_API_KEY"]
 
     query_url = "#{LOCATION_IQ_BASE_URL}/#{endpoint_query}&key=#{ENV['LOCATION_IQ_API_KEY']}&format=json"
-    uri = URI.parse(query_url)
+    uri = URI.parse(URI.escape(query_url))
```
jshoe: this fixes a bug with special characters

app/helpers/location_helper.rb
```diff
@@ -2,18 +2,32 @@ module LocationHelper
...
     loc = {
       name: body["display_name"],
       geo_level: geo_level,
-      country_name: address["country"] || "",
-      state_name: address["state"] || "",
-      subdivision_name: address["county"] || "",
+      country_name: address[country_key] || "",
+      state_name: address[state_key] || "",
+      subdivision_name: address[subdivision_key] || "",
       # Normalize extra provider fields to city. normalizecity param doesn't work all the time.
```
tfrcarvalho: Does this comment still make sense?

tfrcarvalho: Maybe in the model?

test/models/location_test.rb
```diff
@@ -147,4 +149,101 @@ class LocationTest < ActiveSupport::TestCase
       assert_equal original, new_location
     end
   end
+
+  test "should fetch a missing Country parent level for a location" do
```
tfrcarvalho: Can we add a task to migrate this rspec?

app/helpers/location_helper.rb
jshoe: moved 👍

test/models/location_test.rb
jshoe: added 👍

PR #2354, opened 2019-06-24
[AMR Heatmap] Add dropdown option for coloring scale (log/lin) and a color legend

lvreynoso:
# Description

Add a dropdown option to select either a logarithmic or linear coloring scale for the values on the AMR Heatmap. Also adds a color legend with the minimum and maximum value displayed.

# Notes

* Controls and header code that was formerly in the AMRHeatmapView `render()` function have been put into separate functions for readability. 

# Tests

* No server side code modified
* For client side code, go to All Data page, pick a bunch of samples with AMR Counts (like tb-flash), click on Heatmap Icon and select "AMR Heatmap" from the dropdown, when the heatmap loads click on the dropdowns and select different options to see if the heatmap is responsive.

### DIFF
--- a/app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
+++ b/app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
@@ -9,6 +9,17 @@ import SequentialLegendVis from "~/components/visualizations/legends/SequentialL
 import cs from "./amr_heatmap_view.scss";
 
 export default class AMRHeatmapControls extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      legend: {
+        max: 0,
+        scale: "symlog",
+      },
+    };
+  }
+
   handleOptionChange(control, option) {
     const { selectedOptions, onSelectedOptionsChange } = this.props;
     if (option !== selectedOptions[control]) {
@@ -37,15 +48,33 @@ export default class AMRHeatmapControls extends React.Component {
   }
 
   renderLegend() {
-    if (!this.props.hasData) {
+    // Don't render a color legend if the heatmap is still loading
+    if (!this.props.isDataReady) {
+      return;
+    }
+    // This code is here so that when we need the legend to change
+    // because the scale has changed, React will unload the current
+    // legend from the DOM, and render the new, proper one when state updates.
+    if (
+      this.props.maxValueForLegend !== this.state.legend.max ||
+      this.props.selectedOptions.scale !== this.state.legend.scale
+    ) {
+      this.setState({
+        legend: {
+          max: this.props.maxValueForLegend,
+          scale: this.props.selectedOptions.scale,
+        },
+      });
       return;
     }
     return (
-      <SequentialLegendVis
-        min={0}
-        max={this.props.maxValueForLegend}
-        scale={this.props.selectedOptions.scale}
-      />
+      <div className="col s3" key="SequentialLegendVis">
+        <SequentialLegendVis
+          min={0}
+          max={this.state.legend.max}
+          scale={this.state.legend.scale}
+        />
+      </div>
     );
   }
 
--- a/app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
+++ b/app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
@@ -20,6 +20,20 @@ export default class AMRHeatmapControls extends React.Component {
     };
   }
 
+  componentDidUpdate() {
+    if (
+      this.props.maxValueForLegend !== this.state.legend.max ||
+      this.props.selectedOptions.scale !== this.state.legend.scale
+    ) {
+      this.setState({
+        legend: {
+          max: this.props.maxValueForLegend,
+          scale: this.props.selectedOptions.scale,
+        },
+      });
+    }
+  }
+
   handleOptionChange(control, option) {
     const { selectedOptions, onSelectedOptionsChange } = this.props;
     if (option !== selectedOptions[control]) {
@@ -59,12 +73,6 @@ export default class AMRHeatmapControls extends React.Component {
       this.props.maxValueForLegend !== this.state.legend.max ||
       this.props.selectedOptions.scale !== this.state.legend.scale
     ) {
-      this.setState({
-        legend: {
-          max: this.props.maxValueForLegend,
-          scale: this.props.selectedOptions.scale,
-        },
-      });
       return;
     }
     return (
--- a/app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
+++ b/app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
@@ -4,6 +4,7 @@ import cx from "classnames";
 
 import { Divider } from "~/components/layout";
 import { Dropdown } from "~ui/controls/dropdowns";
+import SequentialLegendVis from "~/components/visualizations/legends/SequentialLegendVis.jsx";
 
 import cs from "./amr_heatmap_view.scss";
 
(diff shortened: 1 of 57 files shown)


app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -35,12 +47,44 @@ export default class AMRHeatmapControls extends React.Component {
...
+      return;
+    }
+    // This code is here so that when we need the legend to change
+    // because the scale has changed, React will unload the current
+    // legend from the DOM, and render the new, proper one when state updates.
+    if (
+      this.props.maxValueForLegend !== this.state.legend.max ||
+      this.props.selectedOptions.scale !== this.state.legend.scale
+    ) {
+      this.setState({
```
MarkAZhang: Normally, you would put this kind of code (which updates the state based on the props) inside `componentDidUpdate` or `getDerivedStateFromProps`. For a short discussion on the difference, see: https://twitter.com/dan_abramov/status/977181473424932864.

It's risky to put `setState` calls inside the `render()` function, because `setState` will trigger a re-render, which if you're not careful will call setState again, causing an infinite loop. Changing the state inside the render function is generally considered bad practice in React, where render functions are supposed to be "pure" and not cause side effects.

However, in this case, it looks like you can just directly use `this.props` inside `<SequentialLegendVis>`, without storing it in the state at all.

lvreynoso: I moved this code to `componentDidUpdate()`. I did want to trigger a rerender to dismiss the `SequentialLegendVis`, then another render to put the new one in. (I couldn't get `SequentialLegendVis` to update the legend based on just props, I was getting some ugly visual artifacts). But yeah, I totally get wanting to keep render functions pure.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -82,30 +83,38 @@ export default class AMRHeatmapVis extends React.Component {
...
-    return heatmapValues;
+    return [heatmapValues, maxValue];
   }
 
   updateHeatmap() {
     const rows = this.getHeatmapLabels();
     const columns = this.state.sampleLabels;
-    const values = this.computeHeatmapValues(rows);
+    const [values, maxValue] = this.computeHeatmapValues(rows);
+    this.props.maxValueUpdater(maxValue);
```
MarkAZhang: This feels like the wrong place to calculate the max value. I would calculate it either in `<AMRHeatmapView>` or `<AMRHeatmapControls>`. The reason this feels like the wrong place is that this component `AMRHeatmapVis` doesn't need to know anything about the max value. It doesn't use it, and the data isn't fetched or stored here either. And it's a little awkward calculating it here, then passing it up to the parent via the `maxValueUpdater` function. It's better to put the logic closer to where it ends up getting used.

I realize it might require a little more duplicated logic to put it in AMRHeatmapView or AMRHeatmapControls (and you'll need to make sure you update the max value whenever samplesData or metric changes), but I think it's worth it to put the logic in a more natural place (for the reasons above). When you have many different variables and logic in a large application, following best practices of putting logic in the place you expect makes the code much easier to navigate.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -139,6 +152,7 @@ export default class AMRHeatmapVis extends React.Component {
 }
 
 AMRHeatmapVis.propTypes = {
+  maxValueUpdater: PropTypes.func,
   samplesWithAMRCounts: PropTypes.arrayOf(
```
MarkAZhang: Also, I didn't catch this last PR, but I don't think samplesWithAMRCounts needs to be on the state in this component. You can just directly use the props.

A rule of thumb in React is to only add to `this.state` when you need to, and to use `this.props` directly when you can. It makes it easier to track where a given value "lives"...whether the source of truth is that component or if the value is passed in from the parent.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -66,6 +73,43 @@ export default class AMRHeatmapView extends React.Component {
...
+  }
+
+  renderControls() {
+    return (
+      <AMRHeatmapControls
+        controls={this.assembleControlOptions()}
+        selectedOptions={this.state.selectedOptions}
+        onSelectedOptionsChange={this.updateOptions}
+        isDataReady={!this.state.loading}
+        maxValueForLegend={this.state.maxValue ? this.state.maxValue : 0}
```
MarkAZhang: Nit: You can do `this.state.maxValue || 0`

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -66,6 +73,43 @@ export default class AMRHeatmapView extends React.Component {
...
+        </ViewHeader.Content>
+      </ViewHeader>
+    );
+  }
+
+  renderControls() {
+    return (
+      <AMRHeatmapControls
+        controls={this.assembleControlOptions()}
+        selectedOptions={this.state.selectedOptions}
```
MarkAZhang: We have a convention of extracting `this.props` and `this.state` at the top of the function in the new code we write.

```
  const { selectedOptions } = this.state
```

MarkAZhang:
Looks great!

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
MarkAZhang: If it's not too much work, can you modify `SequentialLegendVis` to update on new props?

It looks like `SequentialLegendVis` only implements `componentDidMount` but not `componentDidUpdate`. If you add a `updateOptions` method to `SequentialLegend` that lets you update `this.options` with new options, and then in SequentialLegendVis do:

```
componentDidUpdate() {
  this.legend.updateOptions({
    scale: this.props.scale,
    ...
  })
  this.legend.update()
}
```

that might be enough to get it to work.

lvreynoso: I'll see if I can get that to work!

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -79,8 +135,9 @@ export default class AMRHeatmapView extends React.Component {
       <div className="row visualization-content">
         <ErrorBoundary>
           <AMRHeatmapVis
-            samplesWithAMRCounts={this.state.samplesWithAMRCounts}
-            selectedOptions={this.state.selectedOptions}
+            // maxValueUpdater={this.updateMaxValue}
```
emilyzhong: Nit: is this comment supposed to be here?

lvreynoso: removed!

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
lvreynoso: Got it!

app/assets/src/components/views/amr_heatmap/AMRHeatmapControls.jsx
```diff
@@ -4,10 +4,22 @@ import cx from "classnames";
 
 import { Divider } from "~/components/layout";
 import { Dropdown } from "~ui/controls/dropdowns";
+import SequentialLegendVis from "~/components/visualizations/legends/SequentialLegendVis.jsx";
 
 import cs from "./amr_heatmap_view.scss";
 
 export default class AMRHeatmapControls extends React.Component {
+  constructor(props) {
```
MarkAZhang: Doesn't seem like you need this.

PR #2359, opened 2019-06-25
Add front-end for Basespace integration.

MarkAZhang:
Guarded with admin-only.
Add "Upload with Basespace" option to Upload Sample options.
Add pop-up window for Basespace OAuth integration.
Add alternate components to render Basespace sample metadata in upload flow.
Add projects and samples_for_project endpoints that proxy to Basespace API.
Add metrics instrumentation.

# Screenshots

Screen Shot 2019-06-24 at 5 18 35 PM
Screen Shot 2019-06-24 at 5 18 39 PM
Screen Shot 2019-06-24 at 5 20 04 PM
Screen Shot 2019-06-24 at 5 20 11 PM
Screen Shot 2019-06-24 at 5 20 24 PM
Screen Shot 2019-06-24 at 5 20 44 PM

# Tests

Include tests for basespace_controller and http_helper.

Verified that Basespace upload flow succeeds and passes the expected object.
Verified that changing projects after the samples are selected re-validates the sample names.
Verified that picking samples from multiple projects works as expected (and the auth pop-up requests permissions from all projects). 
Verify that non-admin users don't see the Basespace tab and can't access the endpoints.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -107,6 +107,7 @@ end
 
 group :test do
   gem 'rspec-json_expectations'
+  gem "webmock", "~> 3.6"
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
--- a/app/assets/src/api/upload.js
+++ b/app/assets/src/api/upload.js
@@ -4,9 +4,17 @@ import {
   markSampleUploaded,
   uploadFileToUrlWithRetries,
 } from "~/api";
+import { map } from "lodash/fp";
 
 import { bulkUploadWithMetadata } from "~/api/metadata";
 import { putWithCSRF } from "./core";
+
+// TODO(mark): Implement back-end for basespace sample uploading.
+export const bulkUploadBasespace = async ({ samples, metadata }) => {
+  return {
+    sample_ids: map("id", samples),
+  };
+};
 
 export const bulkUploadRemote = ({ samples, metadata }) =>
   metadata
--- a/app/assets/src/components/utils/format.js
+++ b/app/assets/src/components/utils/format.js
@@ -2,3 +2,19 @@ export const formatPercent = number => `${(100 * number).toFixed(1)}%`;
 
 export const limitToRange = (number, min, max) =>
   Math.min(max, Math.max(min, number));
+
+// Converts file sizes to readable format.
+// Example: 1234 to 1.2 kb
+export const formatFileSize = bytes =>
+  formatWithUnits(bytes, 1024, ["B", "kB", "MB", "GB", "TB", "PB"]);
+
+export const formatWithUnits = (number, unitFactor, units) => {
+  let unitIndex = 0;
+
+  while (Math.abs(number) >= unitFactor && unitIndex < units.length - 1) {
+    number /= unitFactor;
+    unitIndex++;
+  }
+
+  return `${unitIndex > 0 ? number.toFixed(1) : number} ${units[unitIndex]}`;
+};
--- a/app/assets/src/components/views/SampleUploadFlow/BasespaceSampleImport.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/BasespaceSampleImport.jsx
@@ -1,0 +1,221 @@
+import React from "react";
+import { set, isEmpty, map, get, head, find } from "lodash/fp";
+
+import PropTypes from "~/components/utils/propTypes";
+import PrimaryButton from "~/components/ui/controls/buttons/PrimaryButton";
+import Dropdown from "~ui/controls/dropdowns/Dropdown";
+import LoadingIcon from "~ui/icons/LoadingIcon";
+import Notification from "~ui/notifications/Notification";
+import {
+  getBasespaceProjects,
+  getSamplesForBasespaceProject,
+} from "~/api/basespace";
+import { withAnalytics } from "~/api/analytics";
+
+import { openBasespaceOAuthPopup } from "./utils";
+import cs from "./basespace_sample_import.scss";
+
+export default class BasespaceSampleImport extends React.Component {
+  state = {
+    basespaceProjects: null,
+    selectedProjectId: null,
+    loadingSamples: false,
+    error: "",
+    errorType: "",
+  };
+
+  componentDidMount() {
+    const { accessToken } = this.props;
+    window.addEventListener("message", this.handleMessageEvent);
+    if (accessToken) {
+      this.fetchBasespaceProjects(accessToken);
+    }
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener("message", this.handleMessageEvent);
+  }
+
+  handleMessageEvent = async event => {
+    const { onAccessTokenChange } = this.props;
+    if (
+      event.source === this._window &&
+      event.origin === window.location.origin &&
+      event.data.basespaceAccessToken
+    ) {
+      const accessToken = event.data.basespaceAccessToken;
+      this.fetchBasespaceProjects(accessToken);
+      onAccessTokenChange(accessToken);
+    }
+  };
+
+  handleProjectSelect = projectId => {
+    this.setState({
+      selectedProjectId: projectId,
+    });
+  };
+
+  requestBasespaceBrowseGlobalPermissions = () => {
+    const { basespaceClientId, basespaceOauthRedirectUri } = this.props;
+
+    this._window = openBasespaceOAuthPopup({
+      client_id: basespaceClientId,
+      redirect_uri: basespaceOauthRedirectUri,
+      scope: "browse+global",
+    });
+  };
+
+  getProjectOptions = () => {
+    const { basespaceProjects } = this.state;
+
+    return map(
+      project => ({
+        value: project.id,
+        text: project.name,
+      }),
+      basespaceProjects
+    );
+  };
+
+  fetchBasespaceProjects = async accessToken => {
+    const projects = await getBasespaceProjects(accessToken);
+    this.setState({
+      basespaceProjects: projects,
+      selectedProjectId: get("id", head(projects)),
+    });
+  };
+
+  fetchSamplesForBasespaceProject = async () => {
+    const { selectedProjectId, basespaceProjects } = this.state;
+    const { onChange, accessToken, project } = this.props;
+
+    if (!this.props.project) {
+      this.setState({
+        error: "Please select a project.",
+        errorType: "error",
+      });
+      return;
+    } else {
+      this.setState({
+        error: "",
+      });
+    }
+
+    this.setState({ loadingSamples: true });
+
+    let samples = await getSamplesForBasespaceProject(
+      accessToken,
+      selectedProjectId
+    );
+
+    this.setState({ loadingSamples: false });
+
+    const currentProjectName = get(
+      "name",
+      find(["id", selectedProjectId], basespaceProjects)
+    );
+
+    if (samples.error) {
+      this.setState({
+        error: `There was an error accessing project ${currentProjectName}. Please contact us for help.`,
+        errorType: "error",
+      });
+      return;
+    }
+
+    // Add the current project id to each sample.
+    samples = map(set("project_id", project.id), samples);
+
+    if (isEmpty(samples)) {
+      this.setState({
+        error: `No valid samples could be found in project ${currentProjectName}`,
+        errorType: "warn",
+      });
+    }
+
+    onChange(samples);
+  };
+
+  renderConnectButton = () => {
+    return (
+      <React.Fragment>
+        <div className={cs.helpText}>
+          Please connect to Basespace and authorize IDseq to view your projects
+          and samples.
+        </div>
+        <PrimaryButton
+          text="Connect to Basespace"
+          rounded={false}
+          onClick={withAnalytics(
+            this.requestBasespaceBrowseGlobalPermissions,
+            "BasespaceSampleImport_connect-btn_clicked",
+            {}
+          )}
+        />
+      </React.Fragment>
+    );
+  };
+
+  renderProjectSelect = () => {
+    const { selectedProjectId, loadingSamples, error, errorType } = this.state;
+    return (
+      <React.Fragment>
+        <div className={cs.label}>Select Basespace Project</div>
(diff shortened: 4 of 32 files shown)


happyimadesignr:
@MarkAZhang, the flow is so smooth and clear!

A few suggestions/questions:
- From the success message, can we remove the IDseq top nav?
- Should the success message add in something about “You can return to IDseq upload flow now.”
- After you connect to a project, why is the button dark blue? Is it inactive? 
- Is the basespace project column necessary?

app/controllers/basespace_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+        error: "basespace access token required"
+      }
+      return
+    end
+
+    # 1024 is the maximum limit allowed by Basespace.
+    # If users reach this limit, we will need to implement multiple requests to fetch all the projects.
+    response = HttpHelper.get(
+      "https://api.basespace.illumina.com/v1pre3/users/current/projects",
+      { limit: 1024 },
```
tfrcarvalho: Can we move the number (and maybe the url) to a constant?

app/controllers/basespace_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    if access_token.nil?
+      render json: {
+        error: "basespace access token required"
+      }
+      return
+    end
+
+    # 1024 is the maximum limit allowed by Basespace.
+    # If users reach this limit, we will need to implement multiple requests to fetch all the projects.
```
tfrcarvalho: Multiple questions here:
Do we have to set the limit? Does basespace alerts if there are more projects? Will we pass this information to the user at some point?
(In the future, I think we should probably be a bit more strict and load less projects at a time...)

app/controllers/basespace_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+        error: "basespace project id required"
+      }
+      return
+    end
+
+    # 1024 is the maximum limit allowed by Basespace.
+    # If users reach this limit, we will need to implement multiple requests to fetch all the samples.
+    response = HttpHelper.get(
+      "https://api.basespace.illumina.com/v2/projects/#{basespace_project_id}/datasets?limit=1024",
+      { limit: 1024 },
```
tfrcarvalho: Same comments as for the projects...
Also, why do we use 2 different versions of the api (`v1pre3` above does inspire much confidence :) )?

app/controllers/basespace_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+      "Authorization" => "Bearer #{access_token}"
+    )
+
+    # Just return selected fields.
+    formatted_samples = response["Items"].map do |dataset|
```
tfrcarvalho: Shouldnt we check if response is good before accessing attributes on it? 
I see from your test that in case of errors response might be nil...

app/helpers/http_helper.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    response = http.request(request)
+
+    unless response.is_a?(Net::HTTPSuccess)
+      Rails.logger.warn("POST request to #{url} failed: #{response.message}")
+      return nil
+    end
+
+    # Parse the body as JSON and return it.
+    JSON.parse(response.body)
```
tfrcarvalho: Can we `begin...rescue` here to avoid trusting responses are perfect JSONs?

app/helpers/http_helper.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    response = http.request(request)
+
+    unless response.is_a?(Net::HTTPSuccess)
+      Rails.logger.warn("GET request to #{url} failed: #{response.message}")
+      return nil
+    end
+
+    # Parse the body as JSON and return it.
+    JSON.parse(response.body)
```
tfrcarvalho: Same as above.

spec/controllers/basespace_controller_spec.rb
```diff
@@ -0,0 +1,239 @@
...
+  create_users
+
+  context "Admin user" do
+    before do
+      sign_in @admin
+    end
+
+    describe "GET oauth" do
+      before do
+        allow(HttpHelper).to receive(:post).and_return("access_token" => "12345")
```
tfrcarvalho: Note: a general good practice is to assign these strings to constants to make sure they are in sync with checks below...

spec/controllers/basespace_controller_spec.rb
```diff
@@ -0,0 +1,239 @@
...
+
+          json_response = JSON.parse(response.body)
+          expect(json_response).to include_json(error: "basespace access token required")
+        end
+      end
+    end
+
+    describe "GET samples_for_project" do
+      before do
+        allow(HttpHelper).to receive(:get).and_return("Items" => [
```
tfrcarvalho: nit: Is there a way to reformat this to prevent rubocop from breaking this apart so much?

app/helpers/http_helper.rb
```diff
@@ -0,0 +1,47 @@
+require "net/http"
+
+module HttpHelper
+  def self.post(url, body)
```
tfrcarvalho: If we assume this always returns a JSON... should we encode it the name somewhere? 
Alternatively is not commit to return a JSON below.

tfrcarvalho:
> * From the success message, can we remove the IDseq top nav

@happyimadesignr I think we should at least keep the logo and the black bar. Should definitely remove the links and user dropdown.

app/assets/src/components/utils/format.js
```diff
@@ -2,3 +2,21 @@ export const formatPercent = number => `${(100 * number).toFixed(1)}%`;
 
 export const limitToRange = (number, min, max) =>
   Math.min(max, Math.max(min, number));
+
+// Converts file sizes to readable format.
+// Example: 1234 to 1.2 kb
+export const formatFileSize = bytes => {
+  const unitFactor = 1000;
```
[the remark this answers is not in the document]
MarkAZhang: Good catch. Sometimes people use 1000 and sometimes 1024, but it looks like Illumina uses 1024.

app/assets/src/components/utils/format.js
```diff
@@ -2,3 +2,21 @@ export const formatPercent = number => `${(100 * number).toFixed(1)}%`;
 
 export const limitToRange = (number, min, max) =>
   Math.min(max, Math.max(min, number));
+
+// Converts file sizes to readable format.
+// Example: 1234 to 1.2 kb
+export const formatFileSize = bytes => {
```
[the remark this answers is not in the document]
MarkAZhang: I'm a little hesitant. It feels a bit premature, and I could see other use cases wanting a different set of units. For example, "mm, cm". Would prefer to generalize when other use cases come up.

app/assets/src/components/views/SampleUploadFlow/BasespaceSampleImport.jsx
```diff
@@ -0,0 +1,217 @@
...
+    const params = getURLParamString({
+      client_id: basespaceClientId,
+      redirect_uri: basespaceOauthRedirectUri,
+      response_type: "code",
+      scope: "browse+global",
+    });
+    this._window = openUrlInPopupWindow(
+      `https://basespace.illumina.com/oauth/authorize?${params}`,
+      "BASEPACE_OAUTH_WINDOW",
+      1000,
```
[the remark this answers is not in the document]
MarkAZhang: Moved to new util method.

app/assets/src/components/views/SampleUploadFlow/BasespaceSampleUploadTable.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+    return (
+      <div className={cs.basespaceSampleUploadTable}>
+        <div className={cs.detectedMsg}>
+          <span className={cs.count}>
+            {size(samples)} samples detected.&nbsp;
+          </span>
+          Remove samples you do not want to upload.
+        </div>
+        <DataTable
```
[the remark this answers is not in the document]
MarkAZhang: Helpful to know. Will wait until styles are in sync. Shouldn't be a costly refactor in the future.

app/assets/src/components/views/SampleUploadFlow/UploadSampleStep.jsx
```diff
@@ -84,23 +95,93 @@ class UploadSampleStep extends React.Component {
...
+      client_id: basespaceClientId,
+      redirect_uri: basespaceOauthRedirectUri,
+      response_type: "code",
+      scope: `browse+global${projectReadPermissions}`,
+    });
+
+    this._window = openUrlInPopupWindow(
+      `https://basespace.illumina.com/oauth/authorize?${params}`,
+      "BASEPACE_OAUTH_WINDOW",
+      1000,
```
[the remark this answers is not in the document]
MarkAZhang: Moved to separate utility function.

app/controllers/basespace_controller.rb
MarkAZhang: The `v2` doesn't have a way to list all the projects that a user belongs to...I will check with Basespace about this.

app/controllers/basespace_controller.rb
MarkAZhang: If we don't set the limit, it defaults to 10. Basespace does alert if there are more projects.

I deferred implementing the additional fetch if we detect that there is additional data. I could see it being relevant for samples though. Maybe some users will have more than 1024 samples in a project.

app/controllers/basespace_controller.rb
MarkAZhang: Checked with our Basespace contact. They should both be usable, but he said he'd check on it for us and give us an update. Will update later if necessary.

MarkAZhang:
New confirmation page.

Screen Shot 2019-06-25 at 3 41 28 PM

@happyimadesignr 
I fixed (1) and (2). 
For (3), it's dark blue because the button is "focused". It's not disabled. If you click someone else on the page, the button loses focus and turns back to normal color.
For (4), I think it's nice to have the basespace project name so the user remembers which project they imported the sample from. Do you feel like it's unnecessary? Open to removing it.

app/assets/src/api/upload.js
```diff
@@ -4,10 +4,18 @@ import {
...
 import { bulkUploadWithMetadata } from "~/api/metadata";
 import { putWithCSRF } from "./core";
 
+// TODO(mark): Implement back-end for basespace sample uploading.
+export const bulkUploadBasespace = ({ samples, metadata }) => {
+  return Promise.resolve({
+    sample_ids: map("id", samples),
+  });
+};
+
```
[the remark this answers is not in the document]
MarkAZhang: Good catch!

app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx
```diff
@@ -43,10 +66,27 @@ class ReviewStep extends React.Component {
...
+      let bulkUploadFnName = "bulkUploadRemote";
+      let samples = this.props.samples;
+      let successLogEvent = "ReviewStep_remote-upload_succeeded";
+      let failureLogEvent = "ReviewStep_remote-upload_failed";
+
+      if (this.props.uploadType === "basespace") {
+        bulkUploadFn = bulkUploadBasespace;
+        bulkUploadFnName = "bulkUploadBasespace";
+        samples = map(pick(BASESPACE_SAMPLE_FIELDS), samples);
+        successLogEvent = "ReviewStep_basespace-upload_succeeded";
```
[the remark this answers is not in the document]
MarkAZhang: Not sure why attributes are better than events. It seems like events are slightly easier (since you only need to query on the event name, instead of the event AND the attribute). Happy to change if needed.

app/assets/src/components/utils/format.js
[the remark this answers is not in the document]
tfrcarvalho: My requested changes are simple and do not affect this implementation at all, so it would worth doing now, to incentivize people to use it... think if the change can incentivize future developers to use it.
To use this, they might have to change Upload code to adapt which my serve as deterrent. 

`mm` and `cm` is the opposite case (< 1) and so, that could be an extra generalization. Makes sense here, but since it is a bit more work my be worth to do later.

Anyway, will leave it up to you...

Gemfile
```diff
@@ -111,3 +111,5 @@ end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+
+gem "webmock", "~> 3.6"
```
tfrcarvalho: Is this for testing only? If so, add it within `group :test`.

app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx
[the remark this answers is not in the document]
tfrcarvalho: Depends how you aggregate them or think about analysis. Normally, I would be interested in seeing successful vs. failed uploads, first, and then slice by type of upload. In the end, I see the core event is sample uploaded. 
As I said, there are only two variations here (although there is the local upload one), and it could become more relevant if we added more upload types...
Think of a case where you might have more types... that could lead to an explosion of event with weirder queries (E.g. either `event IN ['ReviewStep_type1', 'ReviewStep_type2', ... 'ReviewStep_type3']` or `event LIKE 'ReviewStep_%'`... 
As with any decision is a trade-off... and should have the analysis you are planning into account.

app/assets/src/components/utils/format.js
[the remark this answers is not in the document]
MarkAZhang: Okay, factored it out into a separate function.

Gemfile
MarkAZhang: Yes, thanks!

MarkAZhang:
Okay, that makes sense regarding the log events. `event IN ['ReviewStep_type1', 'ReviewStep_type2', ... 'ReviewStep_type3']` is definitely unwieldy. Will change.

happyimadesignr:
Thanks @MarkAZhang!

4) Not sure if it's necessary, but I don't mind seeing the project twice, it does reassure me where it's from. Can the user choose multiple projects?

PR #2360, opened 2019-06-25
Clarify read labels in coverage viz.

MarkAZhang:
Add "loose" to read label and updated on-hover tooltip to clarify that only
unassembled reads were included.

# Description

Screen Shot 2019-06-24 at 5 55 48 PM


tfrcarvalho:
Maybe `individual` reads?

MarkAZhang:
I checked with Katrina, and `loose reads` is not an official science term. But she felt that the unknown term will nudge people to look at the docs, whereas `aligned reads` won't trigger any alarms.

"individual" reads feels somewhat more vague than "loose". "Loose" captures the notion that the read should be connected to something bigger but isn't.

PR #2363, opened 2019-06-25
Batch autoscaling (IDSEQ-237)

davidrissato:
# Description

Batch is often slow or unwilling to scale up the idseq compute environment.   The gsnap and raspearch tiers may be up and idle, with many jobs waiting in the batch queues. (IDSEQ-237)

# Tests

### DIFF
--- a/app/jobs/autoscaling.py
+++ b/app/jobs/autoscaling.py
@@ -9,6 +9,8 @@ from operator import itemgetter
 from functools import wraps
 from collections import defaultdict, Counter
 import dateutil.parser
+import batch_autoscaling
+
 
 DEBUG = False
 
@@ -144,6 +146,15 @@ def autoscaling_update(config):
             print "{num_to_discard} instances have finished draining and can be discarded: {instances_to_discard}.".format(num_to_discard=num_to_discard, instances_to_discard=instances_to_discard)
         else:
             print "No instances are ready to move from 'draining' to 'discarded' state."
+
+        print "Autoscaling batches:"
+        response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'])
+        print json.dumps(response_batch_autoscaling, default=_default_json_serializer)
+
+
+def _default_json_serializer(obj):
+    '''Default serializer that returns string <<non-sertializable: TYPE_NAME>> for that type'''
+    return "<<non-serializable: {type_name}>>".format(type_name=str(type(obj)))
 
 class ASG(object):
     r'''
--- a/app/jobs/autoscaling.py
+++ b/app/jobs/autoscaling.py
@@ -147,10 +147,10 @@ def autoscaling_update(config):
         else:
             print "No instances are ready to move from 'draining' to 'discarded' state."
 
-        if 'batch_configurations' in config:
-            print "Autoscaling batches:"
-            response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'], dry_run=True)
-            print json.dumps(response_batch_autoscaling, default=_default_json_serializer)
+    if 'batch_configurations' in config:
+        print "Autoscaling batches:"
+        response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'], dry_run=True)
+        print json.dumps(response_batch_autoscaling, default=_default_json_serializer)
 
 
 def _default_json_serializer(obj):
--- a/app/jobs/autoscaling.py
+++ b/app/jobs/autoscaling.py
@@ -147,9 +147,10 @@ def autoscaling_update(config):
         else:
             print "No instances are ready to move from 'draining' to 'discarded' state."
 
-        print "Autoscaling batches:"
-        response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'])
-        print json.dumps(response_batch_autoscaling, default=_default_json_serializer)
+        if 'batch_configurations' in config:
+            print "Autoscaling batches:"
+            response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'])
+            print json.dumps(response_batch_autoscaling, default=_default_json_serializer)
 
 
 def _default_json_serializer(obj):
--- a/app/jobs/autoscaling.py
+++ b/app/jobs/autoscaling.py
@@ -149,7 +149,7 @@ def autoscaling_update(config):
 
     if 'batch_configurations' in config:
         print "Autoscaling batches:"
-        response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'], dry_run=True)
+        response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'])
         print json.dumps(response_batch_autoscaling, default=_default_json_serializer)
 
 
--- a/app/jobs/autoscaling.py
+++ b/app/jobs/autoscaling.py
@@ -149,7 +149,7 @@ def autoscaling_update(config):
 
         if 'batch_configurations' in config:
             print "Autoscaling batches:"
-            response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'])
+            response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'], dry_run=True)
             print json.dumps(response_batch_autoscaling, default=_default_json_serializer)
 
 
--- a/app/jobs/batch_autoscaling.py
+++ b/app/jobs/batch_autoscaling.py
@@ -3,21 +3,21 @@ import os
 import boto3
 
 SCALING_PERMISSION_TAG = 'IDSeqEnvsThatCanScale'
-SCALING_PERMISSION_ENV_VAR = 'ID_SEQ_ENVS_THAT_CAN_SCALE'
 
 def get_compute_environment_configuration(compute_environment_name, region):
     batch_client = boto3.client('batch', region_name=region)
     response = batch_client.describe_compute_environments(computeEnvironments=[compute_environment_name])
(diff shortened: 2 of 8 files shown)


davidrissato:
> Cool! Once you deploy to staging, would be interesting to measure and post performance improvement stats from a real load test

One thing that I noticed about batch is that if the queue is empty and you send a burst of jobs, it will scale out immediately to the maximum adequate capacity. On the other hand, when you have only a few jobs running, and new jobs arrive, that's when it might take a while to scale out.

I don't know if we currently have any test script, but the performance gain might only be noticed if this test script is able to reproduce the second scenario.

cdebourcy:
> > Cool! Once you deploy to staging, would be interesting to measure and post performance improvement stats from a real load test
> 
> One thing that I noticed about batch is that if the queue is empty and you send a burst of jobs, it will scale out immediately to the maximum adequate capacity. On the other hand, when you have only a few jobs running, and new jobs arrive, that's when it might take a while to scale out.
> 
> I don't know if we currently have any test script, but the performance gain might only be noticed if this test script is able to reproduce the second scenario.

That's an interesting insight. Nice find. I wonder if we can tune for that

davidrissato:
I added a new commit to this PR to make Travis CI run python unit tests. It is already working on this branch, verified in Travis CI execution logs.

davidrissato:
@cdebourcy  I've just added a `dry-run` mode that we can deploy and collect some metrics before removing this flag. Since this code is storing all the values that would be applied, this is also a good opportunity to not only compare how often we would get benefits from this change, but also if it is behaving as expected.

app/jobs/autoscaling.py
```diff
@@ -145,6 +147,16 @@ def autoscaling_update(config):
         else:
             print "No instances are ready to move from 'draining' to 'discarded' state."
 
+        if 'batch_configurations' in config:
+            print "Autoscaling batches:"
+            response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'], dry_run=True)
```
boris-dimitrov: It would be nice if this dry_run option and similar could be configured based on tags attached to the Batch compute env.   That way we can turn this on or off without requiring a code push.   We already employ a similar technique for the rest of the autoscaling functionality, via a tag on the ASG called `IDSeqEnvsThatCanScale`.  The value of the tag is a list of environments whose autoscaler daemon has the right to modify that particular ASG.  You can entirely turn off the autoscaler (and turn it into a "dry run") by changing that tag to some value like "none".   Anyway, the point is that using tags we can make this config something that can be altered at runtime without a push.   It's valuable to have that ability when something isn't behaving quite right and we have lots of jobs running (when a push would be disruptive).

app/jobs/autoscaling.py
```diff
@@ -145,6 +147,16 @@ def autoscaling_update(config):
...
+            print json.dumps(response_batch_autoscaling, default=_default_json_serializer)
+
+
+def _default_json_serializer(obj):
+    '''Default serializer that returns string <<non-sertializable: TYPE_NAME>> for that type'''
+    return "<<non-serializable: {type_name}>>".format(type_name=str(type(obj)))
```
boris-dimitrov: I didn't know you could do that, cool.

app/jobs/batch_autoscaling.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def process_batch_configuration(batch_configuration, dry_run=False):
+    job_vcpus, queue_name, region = (batch_configuration[i] for i in ('vcpus', 'queue_name', 'region'))
+    compute_environment_name = find_compute_environment_name(queue_name, region)
+    current_capacity = get_compute_environment_capacity(compute_environment_name, region)
+    current_vcpu_min, current_vcpu_max = (current_capacity[i] for i in ('minvCpus', 'maxvCpus'))
+
+    pending_jobs_counts = check_pending_jobs_counts(queue_name, region)
+    total_jobs_count = sum(pending_jobs_counts.values())
```
boris-dimitrov: If we are doing a SUM here, rather then len(set union);  you want to be sure that you query the states in reverse order:  RUNNING, then STARTING, then RUNNABLE.   This will ensure you do not double-count jobs that have advanced from one state to the next as you do your separate requests.

app/jobs/batch_autoscaling.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def autoscale_compute_environments(batch_configurations, dry_run=False):
+    results = []
+    for bc in batch_configurations:
+        try:
+            process_results = process_batch_configuration(bc, dry_run)
+            results.append({'batch_configuration': bc, 'process_results': process_results})
+        except Exception as e:
+            error_type = str(type(e))
```
boris-dimitrov: Why is it necessary to suppress exceptions?

If the exception has to be suppressed, could you `traceback.print_exc()` or log `traceback.format_exc()`?   Having the stack trace reduces the time to identify the true cause of some bugs.

app/jobs/batch_autoscaling.py
davidrissato: Hmm, that's a good suggestion, I'll do this change

app/jobs/batch_autoscaling.py
davidrissato: I just want to reduce the noise in case that error `Cannot update, compute environment ... is being modified.` happens too often. I hope that's not the case, since autoscaler reevaluates every 60 seconds and it is probably time enough to make compute environment get back into a ready state. 
However I agree with you, it might be helpful to have the stack trace if something else goes wrong. I'm be adding a traceback to this results object.

app/jobs/autoscaling.py
davidrissato: I like this idea. Implementing it now.

lib/tasks/pipeline_monitor.rake
```diff
@@ -96,7 +98,19 @@ class CheckPipelineRuns
       'max_job_dispatch_lag_seconds' => PipelineRun::MAX_JOB_DISPATCH_LAG_SECONDS,
       'job_tag_prefix' => PipelineRun::JOB_TAG_PREFIX,
       'job_tag_keep_alive_seconds' => PipelineRun::JOB_TAG_KEEP_ALIVE_SECONDS,
-      'draining_tag' => PipelineRun::DRAINING_TAG
+      'draining_tag' => PipelineRun::DRAINING_TAG,
+      'batch_configurations' => [
+        {
+          'queue_name' => Sample::DEFAULT_QUEUE,
+          'vcpus' => Sample::DEFAULT_VCPUS,
```
boris-dimitrov: Technically we could add up the vcpus from each job description in those queues, and not need this param.   I'm fine with this, though.

app/jobs/batch_autoscaling.py
boris-dimitrov: This reminds me of something important we haven't yet had a chance to look into.

It used to be that the lowmem queue would run only alignment jobs.   When that was the case, it was easy to limit the max number of alignment jobs running concurrently using the max vcpus on the compute env.   If we are now using that queue and env to also run other types of jobs, it may be impossible to use the max vcpu param to only control the max number of alignment jobs.   It's nice to have that sort of control, because of the way they compete for chunks.

A similar problem occurs in the highmem queue that is shared by host filering, postprocessing, etc.

I think for both problems we should actually create separate queues for host filtering and alignment.   The host filtering queue would share the high mem compute env, and the alignment queue would share the lowmem compute env.  However, the priority of the host filtering queue would be higher than the priority of the himem queue, and similarly the priority of the alignment queue would be higher than the priority of the lowmme queue.  That way alignment and host filtering jobs would be given priority over other jobs and stages;  this would ensure better utilization of the rapsearch and gsnap tiers.

boris-dimitrov: However, if we did that, we would lose the independent control for max number of alignment jobs.   Perhaps, then, alignment would need its own separate queue AND separate compute environment.

test/jobs/test_batch_autoscaling.py
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env python
```
boris-dimitrov: This looks like a lot of carefully written test code.   However, I don't understand how it's supposed to improve my confidence in the correctness of the new functionality, considering it's harder (for me) to read or understand than the functionality it's testing.  What kind of bugs exactly is it supposed to catch?

boris-dimitrov: Also - what kinds of non-bug changes is it supposed to tolerate?

app/jobs/batch_autoscaling.py
davidrissato: Well, I would need to change the logic if we wanted to share the same compute environment for two different queues. This piece of code is assuming that each queue is connected to a single compute environment. Is that a correct assumption?

lib/tasks/pipeline_monitor.rake
davidrissato: Yes, I just reused this constant since it was already there. That is saving extra calls to get the job definitions.

davidrissato:
1) Dry_run is now done via tags, using the same tag name used in `autoscaler.py`
2) Order reversed
3) I suppose we have different philosophies. I don't see test code as burden, but as a valuable asset. It is self documenting code, and can protect us from a few basic mistakes during small refactorings. 
I have simplified the use cases by decoupling some logic and making most of the methods private. I also added better more explicit descriptions.

davidrissato:
I realized it is not easy to change tags for batch compute environments. The only way I found so far is by recreating the resource, which kind of defeats the purpose of doing that using a tag.
I'm adding an alternative way to set this parameter via aws ssm parameter store.

davidrissato:
Reading tags from the ASG would require an extra call to AWS. For simplicity reasons, this tag is being read from ECS Cluster, following this logic:
1) describe compute environment
2) read find ecsClusterArn
3) describe ecs cluster
4) read the tags from that cluster

PR #2365, opened 2019-06-27
[Maps] Update frontend to support Country and State level clustering with new styles

jshoe:
# Description
- Update frontend to have new Country/State clustering based on styles and behavior from Design.
- Zooming automatically changes the geographic level.
- Crux of the logic is in handleMapLevelChange which puts sample_ids/project_ids into ancestor bubbles. An unclustered copy of rawMapLocationData is kept (open to suggestions about optimizing this).
- Adjusts some of the arbitrary sizing parameters for bubbles.
- See mocks: https://chanzuckerberg.invisionapp.com/d/main#/console/17624454/368938232/preview
- Behavior: In Country or State mode, everything below should bubble up to the Country or State. In State mode, you should see a rectangle for the Country that still has everything below. Below State level, you should see rectangles for the State and Country still, and the Subdivision + City level bubbles will be disaggregated.
- TODO: The dimensions statistics may not be in sync (Tiago's PR has an updated version of this).

# Demo
Jun-26-2019 19-34-50
Jun-26-2019 19-35-43
Jun-26-2019 19-36-46
Jun-26-2019 19-37-14

# Tests
- Go to Data Discovery, go to the maps, play around with zooming in and out over different geographic samples. See behavior above.

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoveryView.jsx
+++ b/app/assets/src/components/views/discovery/DiscoveryView.jsx
@@ -6,14 +6,12 @@ import {
   at,
   capitalize,
   clone,
-  cloneDeep,
   compact,
   concat,
   defaults,
   escapeRegExp,
   find,
   findIndex,
-  indexOf,
   isEmpty,
   keyBy,
   map,
@@ -914,7 +912,8 @@ class DiscoveryView extends React.Component {
       }
 
       MAP_CLUSTER_ENABLED_LEVELS.forEach(ancestorLevel => {
-        // If you have ancestors higher than or at the map's level, add yourself to them.
+        // If you have ancestors higher than or at the map's level, add your samples/projects to
+        // their bubble.
         if (indexOfMapLevel(ancestorLevel) <= indexOfMap) {
           addToAncestor(entry, ancestorLevel);
         }
--- a/app/assets/src/components/views/discovery/DiscoveryView.jsx
+++ b/app/assets/src/components/views/discovery/DiscoveryView.jsx
@@ -34,7 +34,7 @@ import { logAnalyticsEvent } from "~/api/analytics";
 import { openUrl } from "~utils/links";
 import NarrowContainer from "~/components/layout/NarrowContainer";
 import { Divider } from "~/components/layout";
-import { GEO_LEVEL_ORDER } from "~/components/views/discovery/mapping/constants";
+import { MAP_LEVEL_ORDER } from "~/components/views/discovery/mapping/constants";
 
 import DiscoveryHeader from "./DiscoveryHeader";
 import ProjectsView from "../projects/ProjectsView";
@@ -99,15 +99,14 @@ class DiscoveryView extends React.Component {
         filteredSampleDimensions: [],
         filteredSampleStats: {},
         filters: {},
-        geoLevel: "country",
         loadingDimensions: true,
         loadingLocations: true,
         loadingProjects: true,
         loadingSamples: true,
         loadingStats: true,
         loadingVisualizations: true,
+        mapLevel: "country",
         mapLocationData: {},
-        rawMapLocationData: {},
         mapPreviewedLocationId: null,
         mapPreviewedProjects: [],
         mapPreviewedSampleIds: [],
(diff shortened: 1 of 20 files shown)


app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -866,24 +879,69 @@ class DiscoveryView extends React.Component {
     });
   };
 
+  handleMapLevelChange = mapLevel => {
+    const { rawMapLocationData, currentTab } = this.state;
+
+    const ids = currentTab === "samples" ? "sample_ids" : "project_ids";
+    let clusteredData = {};
```
tfrcarvalho: nit: This could actually be const...

app/assets/src/components/views/discovery/mapping/DiscoveryMap.jsx
```diff
@@ -88,33 +114,49 @@ class DiscoveryMap extends React.Component {
...
     // Scale based on the zoom and point count (zoomed-in = higher zoom)
-    // Log1.5 of the count looked nice visually for not getting too large with many points.
+    // Log1.3 of the count looked nice visually for not getting too large with many points.
     const markerSize = Math.max(
-      Math.log(pointCount) / Math.log(1.4) * (get("zoom", viewport) || 3),
+      Math.log(pointCount) / Math.log(1.3) * (get("zoom", viewport) || 3),
       minSize
     );
 
+    let ShapeMarker;
```
tfrcarvalho: Can you define a `<ShapeMarker rectangular />` component please? 
This gets much more confusing here.
The shape marker should all the numbers/sizing hidden inside and maybe configurable through props.

app/assets/src/components/views/discovery/mapping/DiscoveryMap.jsx
```diff
@@ -88,33 +114,49 @@ class DiscoveryMap extends React.Component {
...
     const lng = parseFloat(locationInfo.lng);
+    const geoLevel = locationInfo.geo_level;
+
     const idsField = currentTab === "samples" ? "sample_ids" : "project_ids";
     if (!locationInfo[idsField]) return;
+
     const pointCount = locationInfo[idsField].length;
-    const minSize = 10;
+    // Sizing parameters determined via eyeballing.
+    const minSize = 14;
```
tfrcarvalho: Can we move all the magic eyeballed numbers to props?

app/assets/src/components/views/discovery/mapping/DiscoveryMap.jsx
```diff
@@ -88,33 +114,49 @@ class DiscoveryMap extends React.Component {
...
     // Scale based on the zoom and point count (zoomed-in = higher zoom)
-    // Log1.5 of the count looked nice visually for not getting too large with many points.
+    // Log1.3 of the count looked nice visually for not getting too large with many points.
     const markerSize = Math.max(
-      Math.log(pointCount) / Math.log(1.4) * (get("zoom", viewport) || 3),
+      Math.log(pointCount) / Math.log(1.3) * (get("zoom", viewport) || 3),
```
tfrcarvalho: why `Math.log(1.3)` and not `0.26` or `4x`.

app/assets/src/components/views/discovery/DiscoveryView.jsx
jshoe: 👍

app/assets/src/components/views/discovery/mapping/DiscoveryMap.jsx
jshoe: refactored 👍

app/assets/src/components/views/discovery/mapping/DiscoveryMap.jsx
jshoe: 👍

app/assets/src/components/views/discovery/mapping/DiscoveryMap.jsx
jshoe: I was thinking something logarithmic of the pointCount would just keep it from ever getting too large (vs just multiplying), and then Math.log/Math.log is how to set log1.3(pointCount)

jshoe:
@tfrcarvalho split off a new ShapeMarker component 👍

app/assets/src/components/views/discovery/mapping/DiscoveryMap.jsx
tfrcarvalho: I guess I would prefer a simple version that would gives a way to customize it... Hard to understand the effect.
Would we ever have different scenario where we need to change this?

PR #2366, opened 2019-06-27
[AMR Heatmap] Add mouseover tooltips when hovering over heatmap nodes

lvreynoso:
# Description

As the title says, this PR adds a nifty little box next to the user's cursor when they hover over a heatmap node. The box is just like the one in the Taxon Heatmap, and it displays the sample name, gene name, allele name, coverage value, and depth value as reported in the AMR counts.

# Notes

* Tried to avoid extra logic or data structures where possible, but ended up adding an `alleleToGeneMap` to the state for `AMRHeatmapVis`.
* ~~Performance for the tooltip seems a bit slower for the AMR Heatmap than for the same feature for the Taxon Heatmap. I've thought about ways to speed it up, but leaving that for a future PR (if at all).~~ Fixed

# Tests

* No server side code is changed
* For client side code, go to All Data page, pick a bunch of samples with AMR Counts (like tb-flash), click on Heatmap Icon and select "AMR Heatmap" from the dropdown, when the heatmap loads hover over node to see the tooltip. Click on the dropdowns and select different options to see if the tooltip is responsive and accurate.

# Screenshot
Screen Shot 2019-06-27 at 2 33 56 PM

### DIFF
--- a/app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
+++ b/app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
@@ -37,8 +37,10 @@ export default class AMRHeatmapVis extends React.Component {
     });
   }
 
-  componentDidUpdate() {
-    this.updateHeatmap();
+  componentDidUpdate(prevProps) {
+    if (this.props !== prevProps) {
+      this.updateHeatmap();
+    }
   }
 
   // Sometimes, when presenting the tooltip popup as a user hovers over
@@ -212,11 +214,6 @@ export default class AMRHeatmapVis extends React.Component {
       this.initializeHeatmap(rows, columns, values);
     }
   }
-
-  colorFilter = (value, node, originalColor, _, colorNoValue) => {
-    // Leave zero values grey
-    return value > 0 ? originalColor : colorNoValue;
-  };
 
   initializeHeatmap(rows, columns, values) {
     const { selectedOptions } = this.props;
--- a/app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
+++ b/app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
@@ -16,12 +16,11 @@ export default class AMRHeatmapVis extends React.Component {
   }
 
   componentDidMount() {
-    const samplesWithAMRCounts = this.props.samplesWithAMRCounts;
+    const { samplesWithAMRCounts } = this.props;
     const [sampleLabels, geneLabels, alleleLabels] = this.extractLabels(
       samplesWithAMRCounts
     );
     this.setState({
-      samplesWithAMRCounts,
       sampleLabels,
       geneLabels,
       alleleLabels,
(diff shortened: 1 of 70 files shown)


lvreynoso:
On testing the tooltip again it is running unacceptably slow, so I'll rewrite the code then put up another PR when it's optimized.

lvreynoso:
Fixed the slowdown issue!

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -52,8 +84,92 @@ export default class AMRHeatmapVis extends React.Component {
...
+        amrCountIdentifier = alleleLabels[node.rowIndex].label;
+        allele = amrCountIdentifier;
+        break;
+      }
+      case VIEW_LEVEL_GENES: {
+        amrCountIdentifier = geneLabels[node.rowIndex].label;
+        gene = amrCountIdentifier;
+        break;
+      }
+      default: {
```
jshoe: looks like you don't need the `default break`

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -52,8 +84,92 @@ export default class AMRHeatmapVis extends React.Component {
...
+        break;
+      }
+    }
+
+    const sampleForColumn = samplesWithAMRCounts[node.columnIndex];
+    const amrCountForRow = sampleForColumn.amr_counts.find(
```
MarkAZhang: `amrCountForNode` might be a better name.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -20,15 +33,34 @@ export default class AMRHeatmapVis extends React.Component {
...
+    if (this.props !== prevProps || this.heatmap === null) {
+      this.updateHeatmap();
+    }
+  }
+
+  // Sometimes, when presenting the tooltip popup as a user hovers over
+  // a node on the heatmap, they will be over a node where the row is
+  // an allele and the column is a sample with no AMR count for the allele.
+  // With no AMR count, there's no easy way to grab the name of the gene
+  // for the allele. Hence the allele-to-gene mapping.
```
MarkAZhang: This makes sense! But I'm a little confused why there isn't also a gene-to-allele mapping. Isn't it possible that the sample has no AMR count for the gene as well?

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -1,17 +1,30 @@
...
+import { DataTooltip } from "~ui/containers";
+import { getTooltipStyle } from "~/components/utils/tooltip";
 
 import cs from "./amr_heatmap_vis.scss";
 
 const VIEW_LEVEL_ALLELES = "allele";
 const VIEW_LEVEL_GENES = "gene";
 
+const METRICS = [
+  { text: "Coverage", value: "coverage" },
```
MarkAZhang: I would call this "key" instead of "value". Since "value" implies a numerical value for the metric for a specific amr_count.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
MarkAZhang: Wait never mind. If there's no AMR count for the gene, there's no way to tell which allele it is.

app/assets/src/components/views/amr_heatmap/amr_heatmap_vis.scss
```diff
@@ -1,11 +1,26 @@
...
+
+  .tooltip {
+    position: absolute;
+    display: none;
+
+    &.visible {
+      display: block;
+    }
+
+    .data-tooltip__label {
```
MarkAZhang: Don't think this rule does anything.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -52,8 +84,92 @@ export default class AMRHeatmapVis extends React.Component {
...
+    const {
+      sampleLabels,
+      geneLabels,
+      alleleLabels,
+      alleleToGeneMap,
+    } = this.state;
+    const sampleName = sampleLabels[node.columnIndex].label;
+    let amrCountIdentifier = "";
+    let gene = "";
+    let allele = "";
```
MarkAZhang: Readability nit:
`amrCountIdentifier` feels a little vague. I wonder if it makes sense to name it "rowLabel".

You could do
`const rowLabel = this.getHeatmapLabels()[node.rowIndex].label`
before the `switch` statement, reusing your `getHeatmapLabels` function. (using getHeatmapLabels also makes the `rowLabel` name make more sense).

You can then also move the `sampleForColumn` and `amrCountForRow` assignments above the `switch` statement, and just assign `allele` and `gene` in the `switch` statement. This removes the need for the `if (gene === "")` conditions later on, which feel a little awkward.

All of this is to make the code a bit more readable.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: It's good practice though, no?

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: Exactly!

app/assets/src/components/views/amr_heatmap/amr_heatmap_vis.scss
lvreynoso: It does not it seems. Removed

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: That's much more sensible, thank you! I've changed the code to incorporate your suggestion.

PR #2367, opened 2019-06-27
[Maps] Filter by location

tfrcarvalho:
# Description

Change location filter to include all locations that have the same location at the geo-level of the chosen location.
Side-change: Since projects_controller_spec tests were broken locally (using `web`) decide to fix it by disabling mocks for those tests.

# Tests

Updated backend tests.
Frontend: make sure you have location v2 entries.
1) Verify that location v2 filter shows not only the original location but options for all the parent locations (e.g. "Redwood City, California, USA", "California, USA", "USA")
2) Verify that selecting a parent location shows all entries bellow (e.g. entries for "Redwood City, California, USA" and "Los Angeles, California, USA")

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,8 @@
 source 'https://rubygems.org'
 
 git_source(:github) do |repo_name|
-  repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?('/')
-  "https://github.com/#{repo_name}.git"
+  repo_name = '#{repo_name}/#{repo_name}' unless repo_name.include?('/')
+  'https://github.com/#{repo_name}.git'
 end
 
 gem 'airbrake', '~> 8'
@@ -16,7 +16,7 @@ gem 'bcrypt', '~> 3.1.7'
 gem 'coffee-rails', '~> 4.2', '>= 4.2.2'
 gem 'consul', '>= 0.13.1'
 gem 'devise', '>= 4.6.0'
-gem 'health_check', ">= 2.7.0"
+gem 'health_check', '>= 2.7.0'
 gem 'honeycomb-rails', '>= 0.8.1'
 gem 'mailgun_rails', '>= 0.9.0'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
@@ -74,7 +74,7 @@ gem 'analytics-ruby', '~> 2.0.0', require: 'segment/analytics'
 # Url shortener
 gem 'shortener'
 
-# Explicit load to avoid an "unable to load" warning
+# Explicit load to avoid an 'unable to load' warning
 gem 'http-2'
 
 gem 'deep_cloneable', '~> 2.4.0'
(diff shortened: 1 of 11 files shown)


app/assets/src/components/common/filters/LocationFilter.jsx
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { BaseMultipleFilter } from "~/components/common/filters";
+import { concat, forEach, values } from "lodash/fp";
```
jshoe: nit: move lodash up

app/assets/src/components/common/filters/LocationFilter.jsx
```diff
@@ -0,0 +1,42 @@
...
+            text: parent,
+            value: parent,
+            count: option.count,
+          };
+        } else {
+          parentOptions[parent].count += option.count;
+        }
+      }, option.parents || []);
+    }, options);
+    console.log(parentOptions);
```
jshoe: console.log

app/assets/src/components/views/discovery/DiscoveryFilters.jsx
```diff
@@ -150,6 +157,9 @@ class DiscoveryFilters extends React.Component {
       visibility,
     } = this.props;
 
+    console.log("Location V2: ", locationV2);
```
jshoe: console.log

app/helpers/location_helper.rb
```diff
@@ -75,31 +75,44 @@ def self.sample_dimensions(sample_ids, field_name, samples_count)
...
+                             .group_by { |(_, geo_level)| geo_level }
+                             .map { |field, values| [field, values.map { |id| id[0] }] }
+                             .to_h
+
     samples = samples_with_metadata.includes(metadata: :location)
-    # Plain text locations in string_validated_value
-    samples.where(metadata: { string_validated_value: query })
-           .or(samples.where(metadata: { locations: { name: query } }))
+    # Plain text locations in string_validated_value + multi-geo-level location search
+    samples.where(
```
jshoe: you might be able to use .where().or() or inject(:or) (but probably wouldn't be an improvement)

tfrcarvalho: No, I cannot, although I tried.
It complains of an error of non-concordance between includes of `string_validated_value` and the location ones, even if you have the same includes and/or references.
Since, I did not want to dive into ActiveRecord I wrote my own clause.
This is an issue of treating locations differently... we should get rid of free string for locations.

Gemfile
```diff
@@ -106,8 +106,8 @@ group :development do
 end
 
 group :test do
+  gem 'webmock', '~> 3.6'
```
jshoe: I'm assuming moving webmock on top didn't actually do anything?

spec/spec_helper.rb
```diff
@@ -98,4 +98,3 @@
 end
 
 require "rspec/json_expectations"
-require 'webmock/rspec'
```
jshoe: you may want to mention this change in the PR description in case we're ever trying to track down changes ;)

Gemfile
tfrcarvalho: Indeed it did not. I was exploring possible differences between web and webtest.

PR #2370, opened 2019-06-28
[Pipeline Viz] Add arrows between stages

emilyzhong:
# Add arrows between stages
Adds arrows between stages, + other misc. UI updates

image

### DIFF
--- a/app/assets/src/components/views/PipelineViz.jsx
+++ b/app/assets/src/components/views/PipelineViz.jsx
@@ -4,7 +4,9 @@ import PropTypes from "prop-types";
 import ReactPanZoom from "@ajainarayanan/react-pan-zoom";
 
 import RemoveIcon from "~/components/ui/icons/RemoveIcon";
-import NetworkGraph from "~/components/visualizations/NetworkGraph.js";
+import NetworkGraph from "~/components/visualizations/NetworkGraph";
+import PipelineStageArrowheadIcon from "~/components/ui/icons/PipelineStageArrowheadIcon";
+
 import cs from "./pipeline_viz.scss";
 
 const START_NODE_ID = -1;
@@ -256,8 +258,8 @@ class PipelineViz extends React.Component {
       },
       groups: {
         startEndNodes: {
-          widthConstraint: 8,
-          heightConstraint: 0,
+          shape: "dot",
+          size: 1,
           color: backgroundColor,
           fixed: {
             x: true,
@@ -311,7 +313,7 @@ class PipelineViz extends React.Component {
   }
 
   render() {
-    const stageContainers = this.stageNames.map((stageName, i) => {
+    let stageContainers = this.stageNames.map((stageName, i) => {
       const isOpened = this.state.stagesOpened[i];
 
       return (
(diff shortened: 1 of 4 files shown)


app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -326,10 +328,7 @@ class PipelineViz extends React.Component {
           <div className={isOpened ? cs.openedStage : cs.hidden}>
             <div className={cs.graphLabel}>
               {stageName}
-              <RemoveIcon
-                className={cs.closeButton}
-                onClick={() => this.toggleStage(i)}
-              />
+              <RemoveIcon onClick={() => this.toggleStage(i)} />
```
lvreynoso: I believe you can just write `onClick={this.toggleStage(i)}`

app/assets/src/components/views/pipeline_viz.scss
```diff
@@ -4,38 +4,62 @@
...
   border-radius: 8px;
   padding: 12px;
-  width: 120px;
+  width: 250px;
   text-align: center;
   font-weight: $font-weight-bold;
   cursor: pointer;
 }
 
+.stageArrow {
```
lvreynoso: Would this not be better as a React component that lets you pass in the width of the arrow as a prop? Or will stage arrows always be static?

emilyzhong: For now stage arrows are static, and are only used in the context of this specific part of the graph so not sure if it's worth creating a whole separate component

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -342,6 +341,19 @@ class PipelineViz extends React.Component {
       );
     });
 
+    stageContainers = stageContainers.reduce((containers, stage, i) => {
```
tfrcarvalho: nit: you can just call reduce on the output of the map and make `stageContainers` const again.

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -311,7 +313,7 @@ class PipelineViz extends React.Component {
   }
 
   render() {
-    const stageContainers = this.stageNames.map((stageName, i) => {
+    let stageContainers = this.stageNames.map((stageName, i) => {
```
tfrcarvalho: This is getting harder read. Can you split into
* a function that render container
* a function that render stage arrow
* render that then loops through the stages and calls `renderStageArrow` for the first stage and `renderContainer` for all the others
?

app/assets/src/components/views/pipeline_viz.scss
```diff
@@ -4,38 +4,62 @@
...
-  margin: 24px;
+  margin: 24px 0;
 }
 
 .stageButton {
   background-color: $light-grey;
   border-radius: 8px;
   padding: 12px;
-  width: 120px;
+  width: 250px;
```
tfrcarvalho: Can this be automatic width base on nowrap text?
Or is it supposed to have the same width for all the stages?
If the latter, can you make sure the text wraps if ever necessary `word-break: break-word`.

emilyzhong: The mocks have them all at the same width but I can ask jenn about it on Monday to confirm!

PR #2371, opened 2019-06-29
Add logging for sample upload entry points.

MarkAZhang:
# Description

The entry points to the sample upload flow aren't logged, so adding logging.

# Tests

* Verified that the logging is triggered when the elements are clicked.

### DIFF
--- a/app/assets/src/components/views/samples/ProjectUploadMenu.jsx
+++ b/app/assets/src/components/views/samples/ProjectUploadMenu.jsx
@@ -1,8 +1,11 @@
 import React from "react";
-import cs from "./project_upload_menu.scss";
-import BareDropdown from "~ui/controls/dropdowns/BareDropdown";
 import PropTypes from "prop-types";
+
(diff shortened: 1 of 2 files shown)


app/assets/src/components/views/samples/ProjectUploadMenu.jsx
```diff
@@ -1,8 +1,11 @@
 import React from "react";
-import cs from "./project_upload_menu.scss";
-import BareDropdown from "~ui/controls/dropdowns/BareDropdown";
 import PropTypes from "prop-types";
+
+import BareDropdown from "~ui/controls/dropdowns/BareDropdown";
+import { logAnalyticsEvent } from "~/api/analytics";
+
```
tfrcarvalho: nit: remove empty line

MarkAZhang: I believe our convention was to have empty lines between the import groups.

PR #2376, opened 2019-07-04
[Pipeline Viz] Node Hovering + Refactor

emilyzhong:
# Description
* Refactor code _(main change: instead of documenting `from` and `to` node for each file (redundant since many files share similar paths), store `from` and `to` per edges + list of files that follow the edge.)_
* Highlights corresponding input and output edges when hovering over a node

ezgif-2-382bdcf4356e

# Tests
Go to `samples/[SAMPLE_ID]/stage_results` with pipeline viz enabled account

### DIFF
--- a/app/assets/src/components/common/DetailsSidebar/PipelineStepDetailsMode/PipelineStepDetailsMode.jsx
+++ b/app/assets/src/components/common/DetailsSidebar/PipelineStepDetailsMode/PipelineStepDetailsMode.jsx
@@ -1,0 +1,90 @@
+import React from "react";
+import { groupBy } from "lodash/fp";
+import PropTypes from "prop-types";
+
+import cs from "./pipeline_step_details_mode.scss";
+
+class PipelineStepDetailsMode extends React.Component {
+  renderInputFiles() {
+    const { inputFiles } = this.props;
+    if (!inputFiles || !inputFiles.length) {
+      return null;
+    }
+
+    const fromStepNameToFile = groupBy("fromStepName", inputFiles);
+    const fileGroupList = Object.keys(fromStepNameToFile).map(fromStepName => {
+      const fileGroup = fromStepNameToFile[fromStepName];
+      const fileList = fileGroup.map((file, i) => {
+        return (
+          <div className={cs.fileLink} key={`${file.fileName}-${i}`}>
+            {file.fileName}
+          </div>
+        );
+      });
+
+      // TODO(ezhong): Figure out what to put as fileGroupHeader if input is
+      // provided by a user (instead of step output)
+      return (
+        <div className={cs.fileGroup} key={fromStepName}>
+          <div className={cs.fileGroupHeader}>{`From ${fromStepName ||
+            "Sample"}:`}</div>
+          {fileList}
+        </div>
+      );
+    });
+    return (
+      <div className={cs.stepFilesListBox}>
+        <div className={cs.stepFilesListBoxHeader}>Input Files</div>
+        {fileGroupList}
+      </div>
+    );
+  }
+
+  renderOutputFiles() {
+    const { outputFiles } = this.props;
+    if (outputFiles && outputFiles.length) {
+      const fileList = outputFiles.map((file, i) => {
+        return (
+          <div className={cs.fileLink} key={`${file.fileName}-${i}`}>
+            {file.fileName}
+          </div>
+        );
+      });
+      return (
+        <div className={cs.stepFilesListBox}>
+          <div className={cs.stepFilesListBoxHeader}>Output Files</div>
+          {fileList}
+        </div>
+      );
+    }
+  }
+
+  render() {
+    const { stepName, description } = this.props;
+    return (
+      <div className={cs.content}>
+        <div className={cs.stepName}>{stepName}</div>
+        <div className={cs.description}>{description}</div>
+        {this.renderInputFiles()}
+        {this.renderOutputFiles()}
+      </div>
+    );
+  }
+}
+
+PipelineStepDetailsMode.propTypes = {
+  stepName: PropTypes.string,
+  description: PropTypes.string,
+  inputFiles: PropTypes.arrayOf(
+    PropTypes.shape({
+      fileName: PropTypes.string,
+    })
+  ),
+  outputFiles: PropTypes.arrayOf(
+    PropTypes.shape({
+      fileName: PropTypes.string,
+    })
+  ),
+};
+
+export default PipelineStepDetailsMode;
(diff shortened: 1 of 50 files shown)


app/assets/src/components/common/DetailsSidebar/PipelineStepDetailsMode/PipelineStepDetailsMode.jsx
```diff
@@ -11,10 +10,8 @@ class PipelineStepDetailsMode extends React.Component {
       return null;
     }
 
-    const fromStepNameToFile = groupBy("fromStepName", inputFiles);
-    const fileGroupList = Object.keys(fromStepNameToFile).map(fromStepName => {
-      const fileGroup = fromStepNameToFile[fromStepName];
-      const fileList = fileGroup.map((file, i) => {
+    const fileGroupList = inputFiles.map((inputFileGroup, i) => {
+      const fileList = inputFileGroup.files.map((file, i) => {
```
tfrcarvalho: Is `.files` guaranteed to be defined?

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -25,12 +26,16 @@ class PipelineViz extends React.Component {
     ];
```
tfrcarvalho: I realized that props.admin does not have props validation...

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -49,154 +54,324 @@ class PipelineViz extends React.Component {
...
       },
     });
   }
 
+  handleNodeHover(stageIndex, info) {
+    const { highlightColor } = this.props;
+    const graph = this.graphs[stageIndex];
+    const updatedInterStageArrows = [...this.state.interStageArrows];
+
+    const inputColorOptions = {
```
tfrcarvalho: Would it make sense for this to be a prop or constant?

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -49,154 +54,324 @@ class PipelineViz extends React.Component {
...
+      from && this.getStepDataAt(from).outputEdges.push(edgeIndex);
+      to && this.getStepDataAt(to).inputEdges.push(edgeIndex);
+    });
+  }
+
+  getEdgesBetweenNodes(graph, fromId, toId, type) {
+    const allEdges = graph.getEdgesBetweenNodes(fromId, toId);
+    if (type) {
+      const edgeTypeRegex = new RegExp(`-${type}$`, "g");
+      return allEdges.filter(edgeId => edgeId.match(edgeTypeRegex));
```
tfrcarvalho: Have you thought of using a graph data structure to store everything? That would make this kind of retrieval O(1) instead of having to loop through all edges.
We can also define a map indexed by `from` and `to` nodes to make this a but more efficient.

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -321,37 +529,48 @@ class PipelineViz extends React.Component {
...
-  closeNonativeSteps() {
-    this.stagesData.forEach((stageData, i) => {
-      const graph = this.graphs[i];
-      if (stageData.jobStatus !== "STARTED") {
-        graph.afterDrawingOnce(() => {
-          this.toggleStage(i);
-        });
-      }
-    });
+  closeIfNonativeStage(graph, stageIndex) {
```
tfrcarvalho: Did you mean `Nonactive` ?
nit: also reads better if you capitalize the a, i.e. `...NonActive...`

app/assets/src/components/common/DetailsSidebar/PipelineStepDetailsMode/PipelineStepDetailsMode.jsx
```diff
@@ -77,7 +75,8 @@ PipelineStepDetailsMode.propTypes = {
   description: PropTypes.string,
   inputFiles: PropTypes.arrayOf(
     PropTypes.shape({
-      fileName: PropTypes.string,
+      fromStepName: PropTypes.string,
+      files: PropTypes.arrayOf(PropTypes.shape({ fileName: PropTypes.string })),
```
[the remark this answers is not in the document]
emilyzhong: This should be handled with `isRequired` in propTypes, right?

app/assets/src/components/views/PipelineViz.jsx
emilyzhong: I set the #000000 to a prop, but the entire object I left as a const since the format of it is pretty specific to visjs

app/assets/src/components/common/DetailsSidebar/PipelineStepDetailsMode/PipelineStepDetailsMode.jsx
emilyzhong: This should be handled with isRequired in propTypes, right?

app/assets/src/components/views/PipelineViz.jsx
tfrcarvalho: Sorry, I meant the entire object being a constant outside this function. Since it did not change it would be unnecessary to redefine it on the function every time...

app/assets/src/components/common/DetailsSidebar/PipelineStepDetailsMode/PipelineStepDetailsMode.jsx
tfrcarvalho: Yes, saw you did it below.

app/assets/src/components/views/PipelineViz.jsx
tfrcarvalho: Any comments on this?

app/assets/src/components/views/PipelineViz.jsx
emilyzhong: I think I'm going to leave it as a prop just to stay consistent with how the other edge color updates are written!

app/assets/src/components/views/PipelineViz.jsx
emilyzhong: I'm trying to refactor it currently based off the edgeIds I've created in `generateEdgeData`, which should make it more efficient

emilyzhong: Looking at the code again, I realized that I'm looking through connected edges to a node multiple times for one hover action (ie once in my call to `getEdgesBetweenNodes` and then again when looking through the results of `this.getEdgeInfoFor` — I've moved all the logic into looping through the result of `this.getEdgeInfoFor`
https://github.com/chanzuckerberg/idseq-web/pull/2376/commits/6cb2059cc181abf5855f3dada9c495067318fb85#diff-749f700b60f6b84cf7f1ce76d874bff2R284

PR #2377, opened 2019-07-04
[Pipeline Viz] Fix zoom in interaction bug

emilyzhong:
# Fix zoom in interaction bug
Fixes bug where graph interactions don't work properly after zooming in/out, due to CSS transforms + how canvases detect interactions based on coordinates.

# Notes
**Todo: Interactions to fix**
- [x] Click
- [x] Hover

# Tests
Go to `samples/[SAMPLE_ID]/stage_results` with pipeline viz enabled account


app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -220,12 +229,46 @@ class PipelineViz extends React.Component {
     return stepClassName.replace(/^(PipelineStep(Run|Generate)?)/, "");
   }
 
-  handleStepClick(stageIndex, info) {
-    const clickedNodeId = info.nodes[0];
+  inverseTransformDOMCoordinates(x, y) {
+    const panZoomContainerDOM = ReactDOM.findDOMNode(this.panZoomContainer)
```
jshoe: would it be possible to do this with a React ref instead?

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -276,11 +319,62 @@ class PipelineViz extends React.Component {
     });
   }
 
-  handleNodeHover(stageIndex, info) {
-    const { highlightColor } = this.props;
+  shouldUpdateMouseMove(x, y) {
```
jshoe: nit: this would sound more readable as shouldUpdateMoveMouse?

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -359,12 +451,36 @@ class PipelineViz extends React.Component {
   }
 
   handleNodeBlur = () => {
-    this.graphs.forEach(graph => {
-      const allEdges = this.getEdgesBetweenNodes(graph, null, null, "colored");
-      graph.updateEdges(allEdges, { hidden: true });
+
```
jshoe: nit: empty line

app/assets/src/components/views/PipelineViz.jsx
jshoe: if the goal is mostly to reduce the number of calls to handleMouseMove, you could also just use Lodash `throttle` https://css-tricks.com/debouncing-throttling-explained-examples/

It would save on this computation with every event firing

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -359,12 +451,36 @@ class PipelineViz extends React.Component {
...
+      color: {
+        background: nodeColor,
+        border: nodeColor,
+      },
+    };
+    this.graphs[graphIndex].updateNodes([nodeId], defaultNodeOptions);
+
+    alteredGraphs.forEach(i => {
+      const graph = this.graphs[i];
+      const allColoredEdgeIds = this.getEdgesBetweenNodes(
```
jshoe: nit: since you're often filing the params to getEdgesBetweenNodes with 1-2 nulls, you could consider making the params an object and checking for fromId, toId

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -276,11 +319,62 @@ class PipelineViz extends React.Component {
...
+    if (
+      stageIndex != this.lastMouseMoveInfo.graphIndex ||
+      nodeId != this.lastMouseMoveInfo.nodeId
+    ) {
+      this.handleNodeBlur();
+    }
+
+    Object.assign(this.lastMouseMoveInfo, {
+      graphIndex: stageIndex,
+      nodeId: nodeId,
```
jshoe: you could just do `nodeId`

app/assets/src/components/views/PipelineViz.jsx
emilyzhong: I don't think it's possible since it's trying to get the styling of the first child of the ref (instead of the ref itself)

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -220,12 +229,46 @@ class PipelineViz extends React.Component {
...
 
-  handleStepClick(stageIndex, info) {
-    const clickedNodeId = info.nodes[0];
+  inverseTransformDOMCoordinates(x, y) {
+    const panZoomContainerDOM = ReactDOM.findDOMNode(this.panZoomContainer)
+      .firstChild;
+    const cssMatrixString = window
+      .getComputedStyle(panZoomContainerDOM)
+      .getPropertyValue("transform");
+    const [a, b, c, d] = cssMatrixString
```
tfrcarvalho: Is it worth not running the rest if no zoom was applied?

app/assets/src/components/views/PipelineViz.jsx
```diff
@@ -220,12 +229,46 @@ class PipelineViz extends React.Component {
     return stepClassName.replace(/^(PipelineStep(Run|Generate)?)/, "");
   }
 
-  handleStepClick(stageIndex, info) {
-    const clickedNodeId = info.nodes[0];
+  inverseTransformDOMCoordinates(x, y) {
```
tfrcarvalho: Can you do a small refactor here? Move `PipelineViz.jsx` and `pipeline_vis.scss` to a folder named `PipelineViz` and move this generic transform conversion functions to a util file within that folder. 
Just trying to hide some of this complexity, since is not a core component and should not be a an instance method. It can be a generic function that receives a container and transforms the coordinates.

app/assets/src/components/views/PipelineViz.jsx
emilyzhong: `handleNodeBlur` doesn't update `lastMouseMoveInfo` and I want to keep that method to be clearing previous styling (as opposed to updating to the current values)

app/assets/src/components/views/PipelineViz.jsx
emilyzhong: I tried this at 100ms but it created a bit of a lag, especially when the mouse moves quickly — will look into it more in a separate PR since midpoint presentations are coming up soon 
throttle

jshoe:
LGTM 👍

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -438,37 +351,32 @@ class PipelineViz extends React.Component {
...
-        prevGraph.updateEdges(prevGraphEdgeIds, inputColorOptions);
-        updatedInterStageArrows[edgeInfo.from.stageIndex] = "from";
-        this.lastMouseMoveInfo.alteredGraphs.add(edgeInfo.from.stageIndex);
+    const intraStageInputEdges = inputEdgesInfo.reduce((edgeIds, edgeInfo) => {
+      if (edgeInfo.from) {
+        if (edgeInfo.isIntraStage) {
+          edgeIds.push(`${edgeInfo.from.stepIndex}-${nodeId}-colored`);
+        } else {
+          const prevGraph = this.graphs[edgeInfo.from.stageIndex];
+          const prevEdgeId = `${
```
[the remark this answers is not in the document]
emilyzhong: No idea — I'm not too familiar with how the linter works. I'm guessing that's just how it splits longer strings with formatting

PR #2379, opened 2019-07-08
[AMR Heatmap] Add sample details sidebar

lvreynoso:
# Description

Adds the ability for a user to click on a sample name in the heatmap to access the sample details sidebar, as in the Taxon heatmap.

# Notes

I worked on adding both the sample details sidebar and gene details sidebar since just the former didn't require much work, but the latter turned out to be much bigger than I anticipated so I've broken the features up into two PRs.

# Tests

* No server-side code changed (changed a comment)
* For client side code, go to All Data page, pick a bunch of samples with AMR Counts (like tb-flash), click on Heatmap Icon and select "AMR Heatmap" from the dropdown, when the heatmap loads hover over node to see the tooltip. Click on the sample names and watch the sidebar details bar pop up.


app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -91,6 +89,66 @@ export default class AMRHeatmapView extends React.Component {
...
+    let newOptions = Object.assign({}, selectedOptions, options);
+    this.setState({
+      selectedOptions: newOptions,
+    });
+  };
+
+  onSampleLabelClick = sampleId => {
+    const { sidebarVisible, sidebarMode, selectedSampleId } = this.state;
+    if (!sampleId) {
+      this.setState({
```
MarkAZhang: Nit: Can call `this.closeSidebar` here instead.

app/assets/src/api/amr.js
```diff
@@ -1,6 +1,6 @@
 import { get } from "./core";
 
-export const getAMRCounts = sampleIds =>
```
MarkAZhang: Don't think you need the `async` here, since you don't have an `await`.

Another property of async functions is they will always return a promise, and will wrap a non-promise return value (e.g. `return {}`) in a promise. Since you're returning a promise with `get()` I don't think the `async` makes any difference.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -68,9 +68,16 @@ export default class AMRHeatmapVis extends React.Component {
...
     sampleData.forEach(sample => {
-      sampleLabels.push({ label: sample.sample_name });
+      sampleLabels.push({ label: sample.sample_name, id: sample.sample_id });
       sample.amr_counts.forEach(amrCount => {
-        genes[amrCount.gene] = true;
+        // The following three lines are a kind of hacky workaround to the fact that
+        // the amr counts stored in the db have a gene name that includes the actual gene
+        // plus the drug class.
+        const geneNameExtractionRegex = /[^_]+/; // matches everything before the first underscore
+        const geneName = geneNameExtractionRegex.exec(amrCount.gene)[0];
```
MarkAZhang: In this case, would it be easier to do `geneName = amrCount.gene.split("_")[0]` instead?

MarkAZhang: Is this logic also needed in `mapAllelesToGenes`? If so, maybe it would make sense to factor out into a separate method `getGeneNameForAmrCount`.

MarkAZhang:
Looks great! A few comments.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: It actually modifies the name of the gene in each amr count in the sample data. The logic is kind of unclear though, so I'll factor it out.

MarkAZhang: Ahh I see. 
I was expecting code in `extractLabels` to generate new objects (for the labels) but not mutate any existing data.
You might put this logic in `AMRHeatmapView` in a post-processing function right after fetching the AMR counts. `processAMRCounts`.

PR #2381, opened 2019-07-08
Add bat host genome

jshoe:
### Description
- This adds the Bat host genome: https://jira.czi.team/browse/IDSEQ-1091

### Tests
- Sanity tested on staging: https://staging.idseq.net/samples/12892 and https://staging.idseq.net/samples/12893

### DIFF
--- a/db/migrate/20190708183858_add_bat_host.rb
+++ b/db/migrate/20190708183858_add_bat_host.rb
@@ -7,6 +7,9 @@ class AddBatHost < ActiveRecord::Migration[5.1]
     hg.s3_star_index_path = "s3://idseq-database/host_filter/bat/2019-07-02/bat_STAR_genome.tar"
     hg.s3_bowtie2_index_path = "s3://idseq-database/host_filter/bat/2019-07-02/bat_bowtie2_genome.tar"
 
+    # For lack of a better default, use Human background.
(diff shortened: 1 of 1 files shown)


db/migrate/20190708183858_add_bat_host.rb
```diff
@@ -0,0 +1,19 @@
...
+  def up
+    return if HostGenome.find_by(name: "Bat")
+
+    hg = HostGenome.new
+    hg.name = "Bat"
+    hg.s3_star_index_path = "s3://idseq-database/host_filter/bat/2019-07-02/bat_STAR_genome.tar"
+    hg.s3_bowtie2_index_path = "s3://idseq-database/host_filter/bat/2019-07-02/bat_bowtie2_genome.tar"
+
+    human_host = HostGenome.find_by(name: "Human")
+    hg.default_background_id = human_host.default_background_id if human_host
```
gregdingle: curious why human background should be the default for bat genome

jshoe: added a comment 👍 gist is that we don't have uninfected bat samples

Template: https://github.com/chanzuckerberg/idseq-web/blob/0efb9d50d08cd6f85ea9d1024c8b420bcb799565/db/migrate/20190417183014_add_carp_host.rb

PR #2382, opened 2019-07-08
Fix issue with autoscaler

davidrissato:
# Description

Autoscaler is being terminated with an exception during `count_running_alignment_jobs` execution.

# Notes

This error is not easy to reproduce because it depends on a combination of current time and specific services responses.

# Tests

* Created a unit test that can trigger the issue.

### DIFF
--- a/app/jobs/autoscaling.py
+++ b/app/jobs/autoscaling.py
@@ -391,7 +391,7 @@ class ASG(object):
         def count_running_alignment_jobs():
             ''' returns a map of draining instance IDs to number of jobs still running on the instance '''
             instance_ids = self.draining_instances
-            count_dict = {id: 0 for id in instance_ids}
+            count_dict = defaultdict(int, {id: 0 for id in instance_ids})
             expired_jobs = []
             for item in self.tags:
                 if item['Key'].startswith(self.job_tag_prefix):
--- a/.snyk
+++ b/.snyk
@@ -1,8 +1,71 @@
 # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
-version: v1.13.3
+version: v1.13.5
 ignore: {}
 # patches apply the minimum changes required to fix a vulnerability
 patch:
   SNYK-JS-AXIOS-174505:
     - axios:
         patched: '2019-05-06T06:40:07.049Z'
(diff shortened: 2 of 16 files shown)


app/jobs/autoscaling.py
```diff
@@ -391,7 +391,7 @@ def discard_if_safe(self):
         def count_running_alignment_jobs():
             ''' returns a map of draining instance IDs to number of jobs still running on the instance '''
             instance_ids = self.draining_instances
-            count_dict = {id: 0 for id in instance_ids}
+            count_dict = defaultdict(int, {id: 0 for id in instance_ids})
```
cdebourcy: Thanks for making this change! To follow up on our offline discussion: the function count_running_alignment_jobs will count across all instances. But only the counts from draining instances will actually be used to make discard decisions due to line https://github.com/chanzuckerberg/idseq-web/blob/acf90afc37423de2571ca8d97a7584e356cb54cc/app/jobs/autoscaling.py#L413. So the counts for non-draining instances are unnecessary. If you wanted, you could also fix the key error by wrapping line https://github.com/chanzuckerberg/idseq-web/blob/acf90afc37423de2571ca8d97a7584e356cb54cc/app/jobs/autoscaling.py#L402 into a "if key present" type condition. But the defaultdict solution seems less invasive

davidrissato: Actually I did exactly that in my first attempt, but I ended up amending my commit with the `defaultdict` approach to be more aligned to the method name.

davidrissato: There are some conflicts in this branch. Let me fix these issues, and maybe I'll add the filter "if key present" as well.

cdebourcy: Yes I agree I actually think your current fix with the defaultdict is better to satisfy the method name, no need to change!

davidrissato: This method actually does a little bit more than just counting. It removes expired tags from draining instances as well, and the method comment doesn't necessarily match its functionality anymore.
I wanted to refactor the entire method to decouple this tag deletion logic, but considering this code is somewhat stable, I'm postponing a refactoring on it to a moment that we need to change other things on that logic.

PR #2383, opened 2019-07-09
[Maps] Fix a bug when setting batch Locations with specificity restrictions

jshoe:
### Description
- Description of the bug: When we get a Human location with a City name, we discard the City name and do a geosearch for the subdivision (county). The "bug" is that this means when you set a batch to, say, Redwood City, each entry will be changed to San Mateo County and then save San Mateo County. Then every sample in the list would redo this because it never stopped and saved "Redwood City" (so it's not found). Then you'll hit service provider rate limits of a few requests per second and the batch operation will fail.
- Solution: The gist is that we just move the check_and_restrict_specificity line before the call to find_or_new_by_api_ids. This means the specificity is restricted before we do extra geosearches. In the batch case, the first entry will still do a geosearch for the subdivision/county, but then subsequent lines will re-use that entry without re-fetching from the provider.

# Tests
- See tests

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -123,7 +123,24 @@ class Location < ApplicationRecord
   end
 
   def self.check_and_restrict_specificity_v2(location_data, host_genome_name)
+    # We don't want Human locations with city
+    if host_genome_name == "Human" && location_data[:city_name].present?
+      # Return our existing entry if found
+      existing = Location.find_by(country_name: location_data[:country_name], state_name: location_data[:state_name], subdivision_name: location_data[:subdivision_name], city_name: "")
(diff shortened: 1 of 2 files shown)


davidrissato:
What happens if you find the same city name in different regions/countries?
For example, there are 3 or more places with this name: `Joinville`
- Joinville, Algeria, a village near Blida
- Joinville, Haute-Marne, a commune in Haute-Marne, France
- Joinville, Santa Catarina, Brazil

jshoe:
@davidrissato that's handled by the location hierarchy of country_name, state_name, subdivision_name, city_name :)

PR #2384, opened 2019-07-09
[AMR Heatmap] Add Gene Details Sidebar

lvreynoso:
# Description

Allows users to click on a gene (or allele) name in the AMR heatmap and display a sidebar with details for that gene. Pulling from the CARD Ontology, if a match for the gene name is found the sidebar will display the gene name as it appears in CARD, any synonyms for the gene, the gene family, the class of drugs the gene confers resistance to, the resistance mechanism, and peer-reviewed publications that give information about the gene.

At the bottom are links to the CARD Ontology entry, an NCBI AMR Reference Gene Catalog search, a Pubmed search, and a Google Scholar search.

# Notes

* To implement this functionality a request is made to the IDSeq backend in order to get around CORS issues. The backend sorts through the Ontology data and returns the needed information in JSON.
* Apologies for the long diff

# Tests

* Server-side code rspec tests added: `docker-compose run --rm web "RAILS_ENV=test rspec spec/controllers/amr_heatmap_controller_spec.rb"`
* For client side code, go to All Data page, pick a bunch of samples with AMR Counts (like tb-flash), click on Heatmap Icon and select "AMR Heatmap" from the dropdown, when the heatmap loads hover over node to see the tooltip. Click on the gene names and watch the sidebar details bar pop up.

Jul-08-2019 18-21-30


app/assets/src/api/amr.js
```diff
@@ -6,3 +6,13 @@ export const getAMRCounts = sampleIds =>
       sampleIds,
     },
   });
+
+export const getCARDIndex = accession =>
```
jshoe: this one doesn't use the param `accession`

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -0,0 +1,430 @@
...
+import PropTypes from "prop-types";
+
+import { getCARDIndex, getAroEntry } from "~/api/amr";
+
+import cs from "./gene_details_mode.scss";
+
+const SOURCE_CARD = "card";
+const SOURCE_PUBMED = "pubmed";
+const SOURCE_GOOGLE_SCHOLAR = "googlescholar";
+const SOURCE_NCBI_REF_GENE = "ncbirefgene";
```
jshoe: nit: up to you but the value would be more readable as ncbiRefGene?

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -0,0 +1,430 @@
...
+            </li>
+          </ul>
+          <ul className={cs.linksList}>
+            <li className={cs.link}>
+              <a
+                href={this.generateLinkTo(SOURCE_PUBMED)}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                Pubmed Search
```
jshoe: nit: Pubmed is branded as PubMed

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -0,0 +1,430 @@
...
+      this.getGeneInfo(this.props.geneName);
+    }
+  }
+
+  async getGeneInfo(geneName) {
+    const { cardIndex } = this.state;
+    const updatedOntology = {
+      accession: undefined,
+      label: "---",
+      synonyms: "---",
```
jshoe: I think it would look better just not to show the section for 'Synonym(s)' and these other fields if there's no info, instead of the `---`

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -0,0 +1,430 @@
...
+    const { ontology } = this.state;
+    return ontology.publications.map(publication => {
+      const citation = /.*(?=(\(PMID))/.exec(publication)[0];
+      const pmidText = /(PMID)\s[0-9]*/.exec(publication)[0];
+      const pubmedId = pmidText.split(" ")[1];
+      return (
+        <div className={cs.textInner} key={pubmedId}>
+          {citation}
+          <div className={cs.link}>
+            (<a
```
jshoe: you probably want to add another `span` in here because the entire row turns into the pointer cursor

jshoe: also maybe the parentheses shouldn't be part of the link

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -0,0 +1,430 @@
...
+        </div>
+        {!collapseOntology && (
+          <div>
+            <div className={cs.subtitle}>{CARD_FAMILY}</div>
+            <div className={cs.text}>
+              <div className={cs.textInner}>{ontology.geneFamily}</div>
+            </div>
+            <div className={cs.subtitle}>{CARD_CLASS}</div>
+            <div className={cs.text}>
+              <div className={cs.textInner}>{ontology.drugClass}</div>
```
jshoe: It might look nicer to have the first letter capitalized in the drug class and mechanisms

app/assets/src/api/amr.js
```diff
@@ -6,3 +6,13 @@ export const getAMRCounts = sampleIds =>
...
+  get("amr_heatmap/fetch_card_index.json");
+
```
jshoe: It looks like the error responses for getCARDIndex and getAroEntry are not being handled, but you can have a simple message like "Error loading blah". Over time the source data URLs will go down and change

jshoe:
I think you should move all the parsing into the backend controller. There are a few reasons for this: 1) moves the necessary processing upstream to be cleaner since you have the backend endpoints anyway, 2) keeps our backend contract of returning JSON API responses, 3) avoids parsing stored XML in the frontend (which can be brittle and error-prone anyway), 4) enhances security by allowing for backend validation and cleaning the response passed to the client browsers. You should just assume that all external sources will have malicious data and try to inject code into our backend and the client pages.

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -0,0 +1,430 @@
...
+      return;
+    }
+    return (
+      <div className={cs.geneContents}>
+        {cardEntryFound ? this.renderOntology() : this.renderNotFound()}
+        <div className={cs.subtitle}>Links</div>
+        <div className={cs.linksSection}>
+          <ul className={cs.linksList}>
+            {cardEntryFound && (
+              <li className={cs.link}>
```
jshoe: you could DRY this out by having a render function for the <li> blocks since they're all the same

lvreynoso:
Refactoring to add tests and move parsing to backend.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -1,3 +1,6 @@
+URL_CARD_ARO = "https://card.mcmaster.ca/aro/".freeze
+URL_CARD_INDEX = "https://raw.githubusercontent.com/arpcard/aro/master/aro.owl".freeze
```
jshoe: one more comment while you're making changes :O I noticed this file aro.owl is actually fairly large, so that means a lot of repeated calls to an external service. If it's static and not expected to change often, we should just vendor it and store a version in, say, idseq-database on S3.

lvreynoso: Vendorized!

app/assets/src/api/amr.js
lvreynoso: Errors are now shown in the sidebar!

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
lvreynoso: Fixed!

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
lvreynoso: Done!

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
lvreynoso: Done!

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
lvreynoso: Done!

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
lvreynoso: Made even more readable!

lvreynoso:
Tests added, parsing moved to the backend, polish added on the frontend. Open again for review!

Will NOT be submitting/merging this PR until legal issues are cleared up. Do not approve until then.

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -0,0 +1,282 @@
...
+    if (ontology.error !== "") {
+      cardEntryFound = false;
+    } else {
+      cardEntryFound = true;
+    }
+
+    this.setState({
+      ontology,
+      loading: false,
+      cardEntryFound,
```
MarkAZhang: Nit: You can just do `cardEntryFound: ontology.error === ""` here.

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/gene_details_mode.scss
```diff
@@ -0,0 +1,74 @@
...
+    color: $primary-light;
+    margin-top: 10px;
+    cursor: pointer;
+  }
+}
+
+.cardLicense {
+  padding-top: 10px;
+  color: $medium-grey;
+  font-size: $font-size-small;
```
MarkAZhang: We're pushing to use the mixins from `styles/themes/_typography` now. You can use them like `@include font-body-xs;`

`font-size-small` and all those variables are deprecated.

app/assets/src/components/utils/format.js
```diff
@@ -18,3 +18,11 @@ export const formatWithUnits = (number, unitFactor, units) => {
 
   return `${unitIndex > 0 ? number.toFixed(1) : number} ${units[unitIndex]}`;
 };
+
+export const formatSentenceCase = string => {
```
MarkAZhang: There's already a function in `StringHelper` called `capitalizeFirstLetter`. I would use that instead.

As a side note, I typically avoid using regex when simpler tools will work, because regex is less human-readable and it makes the whole function harder to understand. Would it be burdensome for you to use string functions instead of regex, where applicable?

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -110,6 +110,18 @@ export default class AMRHeatmapVis extends React.Component {
     this.setState({ nodeHoverInfo: null });
   };
 
+  // mediate the callback function for clicking a row label in the heatmap,
+  // so that clicking an allele returns the proper gene
+  onRowLabelClick = rowLabel => {
```
MarkAZhang: This is interesting! If the user clicks on an allele label, we show them information about the gene.

Is there any additional allele-specific information that we can show them as well? We might at least indicate in the sidebar that we are showing the gene for that allele.

We can handle this in a future PR, but worth thinking about.

PR #2385, opened 2019-07-09
[Maps] Add right pane filter functionality + a track for the selected samples counter

jshoe:
### Description
- This came out of the last map check-in meeting when discussing product analytics and opportunistically adds the right panel metadata filtering (instead of just adding a no-op track call to the metadata bars).
- Also adds a no-op track to the samples counter circle (e.g. "5 Selected"). Requested by Jenn to see if users are searching for a way to view their selected samples. "Fake door testing".

### Notes
Jul-09-2019 15-23-07
Jul-09-2019 15-24-02

### Tests
- Go to Data Discovery, try clicking on the right sidebars to add filters. Try it in the map view too.

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoverySidebar.jsx
+++ b/app/assets/src/components/views/discovery/DiscoverySidebar.jsx
@@ -3,7 +3,7 @@ import cx from "classnames";
 import { find, maxBy, orderBy, sumBy } from "lodash/fp";
 import moment from "moment";
 
-import { logAnalyticsEvent } from "~/api/analytics";
+import { withAnalytics } from "~/api/analytics";
 import PropTypes from "~/components/utils/propTypes";
 import { Accordion } from "~/components/layout";
 import BasicPopup from "~/components/BasicPopup";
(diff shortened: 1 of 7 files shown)


app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -111,7 +111,6 @@ export default class DiscoverySidebar extends React.Component {
                 key={entry.value}
                 style={{ height: percent + "px" }}
                 onClick={() => {
-                  onFilterClick && onFilterClick(entry);
```
jshoe: Date histogram filter not yet implemented, so I removed this

app/assets/src/components/views/discovery/discovery_sidebar.scss
```diff
@@ -63,7 +63,6 @@
       margin: 0 1px;
       flex: 1 0;
       background-color: $melrose;
-      cursor: pointer;
```
jshoe: date histogram bars aren't clickable so I removed this

app/assets/src/components/views/discovery/DiscoverySidebar.jsx
```diff
@@ -203,29 +202,33 @@ export default class DiscoverySidebar extends React.Component {
...
-                this.handleFilterClick(value);
-                logAnalyticsEvent("DiscoverySidebar_metadata-filter_clicked", {
-                  value,
-                  count,
-                  percent,
-                  rows: rows.length,
-                });
-              }}
-            >
+            <a onClick={onClick}>
```
tfrcarvalho: I think `withAnalytics` is the actual function defined for this.
Funny thing is that withAnalytics also comes with a lot more "event name" validation... 

I will leave up to you to update the code to use that function.

jshoe: ah yes, it looks like withAnalytics is preferred now so I switched to it 👍

PR #2390, opened 2019-07-11
[Metadata] Refresh project view after editing project-level metadata + Port over public sample notifications

jshoe:
### Description
Two changes bundled here:
- Refresh project data in the project view after editing in the batch metadata editor. Before the data actually wouldn't reload so the columns would be out of date. This was confusing on the maps because you would need to refresh before you see your newly inputted locations on the map.
- While making this change and looking at the project header, I noticed that the "public sample notifications" were never ported over to data discovery. I checked with Mark and just decided to copy them over to DiscoveryView.

### Notes
Jul-10-2019 17-54-12
Screen Shot 2019-07-10 at 6 04 57 PM

# Tests
- Go to a project, go to Upload Metadata in the upper-right, change the metadata and save, and see the data reload to have current column values.
- Load data discovery and watch the public samples notifications appear. You could force it by modifying checkPublicSamples.

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoveryView.jsx
+++ b/app/assets/src/components/views/discovery/DiscoveryView.jsx
@@ -957,9 +957,7 @@ class DiscoveryView extends React.Component {
 
   // Cloned from Samples.jsx (deprecated)
   checkPublicSamples = () => {
-    console.log("checkPublicSamples called");
     get("/samples/samples_going_public.json").then(res => {
-      console.log("res: ", res);
       if ((res || []).length) this.displayPublicSampleNotifications(res);
     });
   };
--- a/app/assets/src/components/views/discovery/DiscoveryView.jsx
+++ b/app/assets/src/components/views/discovery/DiscoveryView.jsx
@@ -1,5 +1,3 @@
-import axios from "axios/index";
-import { partition } from "lodash";
 import React from "react";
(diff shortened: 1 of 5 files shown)


app/assets/src/components/views/samples/notifications.jsx
```diff
@@ -35,10 +36,10 @@ const publicSampleNotificationsByProject = samples => {
         localStorage.setItem(
           "dismissedPublicSamples",
           JSON.stringify(
-            new Set([
```
jshoe: there was a bug here b/c JSON.stringify of a set will always be "{}" https://stackoverflow.com/questions/31190885/json-stringify-a-set

app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -951,6 +955,37 @@ class DiscoveryView extends React.Component {
     this.setState({ mapLocationData: clusteredData, mapLevel });
   };
 
+  // Cloned from Samples.jsx (deprecated)
```
tfrcarvalho: Why is this deprecated?

jshoe: @tfrcarvalho deleted this comment b/c it was confusing and added a comment to the setItem below 👍

PR #2387, opened 2019-07-10
Add link to pipeline viz 

emilyzhong:
# Add link to pipeline viz 
image
image

# Notes
- [x] Under "Downloads" dropdown
- [x] Under "Pipeline Info" section

# Tests
Go to sample view, click on "Sample Details", then "Pipeline" tab on sidebar

PR #2392, opened 2019-07-11
Fix sidebar bug for errored samples

emilyzhong:
# Fix sidebar bug for errored samples
From https://github.com/chanzuckerberg/idseq-web/pull/2387

# Tests
Go to an errored sample's sampleView, click on "See sample details" and switch to "Pipeline" tab on sidebar (previously would error)


emilyzhong:
> LGTM but maybe it should just return null if `!pipelineInfo[field.key]`

Before the breaking change the table shows "--" if `pipelineInfo[field.key]` doesn't exist, which is why didn't return early

PR #2393, opened 2019-07-11
[Locations] Fix bug with filtering by Unknown locations

jshoe:
### Description
- There was a bug where filtering by Unknown / "Not set" locations wouldn't work. The side effect was that the filter would still get stored in their URL state, so a refresh wouldn't work either and it would break the interface for the user unless they manually removed the filter from the URL.

### Notes
- The bug was in LocationHelper#filter_by_name where the SQL statement just had a dangling "OR" at the end if there were no locations. The statement would look like
```sql
SELECT `samples`.`id` FROM `samples` LEFT OUTER JOIN `metadata` ON `metadata`.`sample_id` = `samples`.`id` LEFT OUTER JOIN `metadata_fields` ON `metadata_fields`.`id` = `metadata`.`metadata_field_id` LEFT OUTER JOIN `locations` ON `locations`.`id` = `metadata`.`location_id` WHERE (project_id in (25,50,63)) AND `metadata`.`metadata_field_id` = 363 AND (`metadata`.`string_validated_value` IN ('not_set') OR )
```

### Tests
- Go to `http://localhost:3000/projects/dimensions.json?domain=my_data&locationV2[]=not_set` and see results (before would crash).

### DIFF
--- a/app/helpers/location_helper.rb
+++ b/app/helpers/location_helper.rb
@@ -106,8 +106,8 @@ module LocationHelper
                              .group_by { |(_, geo_level)| geo_level }
                              .map { |field, values| [field, values.map { |id| id[0] }] }
                              .to_h
-    samples = samples_with_metadata.includes(metadata: :location)
 
+    samples = samples_with_metadata.includes(metadata: :location)
     # Plain text locations in string_validated_value + multi-geo-level location search
     if locations_by_geo_level.present?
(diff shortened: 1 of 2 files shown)


jshoe:
> Are there unit tests to cover these SQL queries?

Looks like it was pending for just this method, so I added two!

PR #2396, opened 2019-07-12
IDSEQ-1069 - Fix issue with file upload triggering pipeline before uploads complete

davidrissato:
# Description

Fix issue with file upload triggering pipeline before uploads complete.

# Notes

`onUploadProgress` was using `Math.round` instead of `Math.floor`, which potentially could mark progress as being 100% in cases where more than 99.5% of the file had been uploaded. If that happened to one of the files, and the other one had already completed, `onFileUploadSuccess` would incorrectly assume both files as completed due during this check: `sampleFiles.every(f => fileNamesToProgress[f.name] === 100)`

# Tests

* Condition almost impossible to manually verify and mock libraries doesn't allow to mock uploadProgress.

### DIFF
--- a/app/assets/src/api/upload.js
+++ b/app/assets/src/api/upload.js
@@ -79,13 +79,16 @@ export const bulkUploadLocalWithMetadata = ({
 
           uploadFileToUrlWithRetries(file, url, {
             onUploadProgress: e => {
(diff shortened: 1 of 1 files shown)


jshoe:
nice find!

PR #2400, opened 2019-07-12
[Pipeline Viz] Visual bug fixes

emilyzhong:
# [Pipeline Viz] Visual bug fixes
Changes:
* Cursor on drag/pannable sections
* Window resizing handler to "redraw" graphs
* Fix hover effect on inter-stage arrows for files that "skip" stages
* Arrows for sample fastq input will also be highlighted on node hover
* Hovering on node switched to pointer cursor to indicate clicking
* Center graph vertically + horizontally on page

# Tests
Go to `samples/[sample id]/stage_results` on a pipeline_viz enabled account

### DIFF
--- a/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
+++ b/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
@@ -370,7 +370,15 @@ class PipelineViz extends React.Component {
             edgeInfo.from.stepIndex
           }-${END_NODE_ID}-colored`;
           prevGraph.updateEdges([prevEdgeId], inputColorOptions);
-          updatedInterStageArrows[edgeInfo.from.stageIndex] = "from";
+
+          for (
+            let arrowIndex = edgeInfo.from.stageIndex;
+            arrowIndex < stageIndex;
+            arrowIndex++
+          ) {
+            updatedInterStageArrows[arrowIndex] = "from";
+          }
+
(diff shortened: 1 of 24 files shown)


app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -742,10 +777,13 @@ class PipelineViz extends React.Component {
...
             {stageName}
-            <RemoveIcon onClick={() => this.toggleStage(i)} />
+            <RemoveIcon
+              onClick={() => this.toggleStage(i)}
+              className={cs.closeIcon}
+            />
           </div>
           <div
-            className={cs.graph}
+            className={cx(cs.graph, this.state.hovered ? cs.hovered : "")}
```
jshoe: you should just do `cx(cs.graph, this.state.hovered && cs.hovered)`

app/assets/src/components/views/PipelineViz/pipeline_viz.scss
```diff
@@ -36,6 +41,7 @@
   overflow: visible;
   position: relative;
   flex-shrink: 0;
+  pointer-events: none;
```
jshoe: what does `pointer-events: none` do?

emilyzhong: It makes it so that mouse interactions are ignored for the element / go straight to whatever's behind it. I added it in so dragging to pan still works on the spaces between stages

PR #2402, opened 2019-07-13
Add back-end for basespace integration.

MarkAZhang:
Add code to detect basespace files from dataset ids and download the files to
the correct folder in S3.

Ensure that sample fails gracefully if download is unsuccessful.

# Tests

* Verified that back-end tests pass.
* Verified that Basespace upload succeeds locally. (the sample files end up in S3 and host filtering begins)

### DIFF
--- a/app/assets/src/components/views/SampleView/SampleView.jsx
+++ b/app/assets/src/components/views/SampleView/SampleView.jsx
@@ -237,10 +237,11 @@ class SampleView extends React.Component {
   };
 
   pipelineInProgress = () => {
-    if (
-      this.props.pipelineRun &&
-      this.props.pipelineRun.results_finalized > 0
-    ) {
+    const { sample, pipelineRun } = this.props;
+    if (sample.upload_error) {
+      return false;
+    }
+    if (pipelineRun && pipelineRun.results_finalized > 0) {
       return false;
     }
     return true;
@@ -260,12 +261,15 @@ class SampleView extends React.Component {
   };
 
   renderPipelineWarnings = () => {
+    const { pipelineRun } = this.props;
+    if (!pipelineRun) return null;
+
     const warnings = [];
 
     if (
       !this.pipelineInProgress() &&
-      pipelineVersionHasAssembly(this.props.pipelineRun.pipeline_version) &&
-      this.props.pipelineRun.assembled !== 1
+      pipelineVersionHasAssembly(pipelineRun.pipeline_version) &&
+      pipelineRun.assembled !== 1
     ) {
       warnings.push("The reads did not assemble for this run.");
     }
@@ -299,7 +303,17 @@ class SampleView extends React.Component {
   renderSampleReportError = () => {
     const { pipelineRun, sample } = this.props;
     let status, message, linkText, issueType, link;
-    switch (pipelineRun.known_user_error) {
+    switch (
+      sample.upload_error || (pipelineRun && pipelineRun.known_user_error)
+    ) {
+      case "BASESPACE_UPLOAD_FAILED":
+        status = "SAMPLE FAILED";
+        message =
+          "Oh no! There was an issue uploading your sample file from Basespace.";
+        linkText = "Contact us for help.";
+        issueType = "error";
+        link = "mailto:help@idseq.net";
+        break;
       case "FAULTY_INPUT":
         status = "COMPLETE - ISSUE";
         message = `Sorry, something was wrong with your input file. ${
--- a/app/controllers/basespace_controller.rb
+++ b/app/controllers/basespace_controller.rb
@@ -1,13 +1,8 @@
-BASESPACE_OAUTH_URL = ["https://api.basespace.illumina.com/v1pre3/oauthv2/token",
-                       BASESPACE_CURRENT_PROJECTS_URL = "https://api.basespace.illumina.com/v1pre3/users/current/projects".freeze].freeze
-BASESPACE_PROJECT_DATASETS_URL = "https://api.basespace.illumina.com/v2/projects/%s/datasets".freeze
-# The maximum number of items to fetch from a Basespace API endpoint at one time.
-# If we don't set this, it defaults to 10.
-# TODO(mark): Implement a way to automatically fetch additional pages if necessary.
-BASESPACE_PAGE_SIZE = 1024
+BASESPACE_OAUTH_URL = "https://api.basespace.illumina.com/v1pre3/oauthv2/token".freeze
 
 class BasespaceController < ApplicationController
   include HttpHelper
+  include BasespaceHelper
 
   before_action :admin_required
 
@@ -44,32 +39,20 @@ class BasespaceController < ApplicationController
       return
     end
 
-    # 1024 is the maximum limit allowed by Basespace.
-    # If users reach this limit, we will need to implement multiple requests to fetch all the projects.
-    response = HttpHelper.get_json(
-      BASESPACE_CURRENT_PROJECTS_URL,
-      { limit: BASESPACE_PAGE_SIZE },
-      "Authorization" => "Bearer #{access_token}"
-    )
+    bs_projects = basespace_projects(access_token)
 
-    if response.nil? || response["Response"].nil? || response["Response"]["Items"].nil?
+    if bs_projects.nil?
       render json: {
         error: "unable to fetch data from basespace"
       }
       return
     end
 
-    # Just return selected fields.
-    formatted_projects = response["Response"]["Items"].map do |dataset|
-      {
-        id: dataset["Id"],
-        name: dataset["Name"]
-      }
-    end
-
-    render json: formatted_projects
+    render json: bs_projects
   end
 
+  # Actually fetches basespace DATASETS, which are collections of related fastq files (such as paired-end files)
+  # This roughly corresponds to "samples" in IDseq.
   def samples_for_project
     access_token = params[:access_token]
     basespace_project_id = params[:basespace_project_id]
@@ -88,51 +71,15 @@ class BasespaceController < ApplicationController
       return
     end
 
-    # 1024 is the maximum limit allowed by Basespace.
-    # If users reach this limit, we will need to implement multiple requests to fetch all the samples.
-    response = HttpHelper.get_json(
-      BASESPACE_PROJECT_DATASETS_URL % basespace_project_id,
-      { limit: BASESPACE_PAGE_SIZE },
-      "Authorization" => "Bearer #{access_token}"
-    )
+    bs_samples = samples_for_basespace_project(basespace_project_id, access_token)
 
-    if response.nil? || response["Items"].nil?
+    if bs_samples.nil?
       render json: {
         error: "unable to fetch data from basespace"
       }
       return
     end
 
-    # Just return selected fields.
-    formatted_samples = response["Items"].map do |dataset|
-      {
-        basespace_dataset_id: dataset["Id"],
-        name: dataset["Name"],
-        file_size: dataset["TotalSize"],
-        file_type: get_dataset_file_type(dataset),
-        basespace_project_id: basespace_project_id,
-        basespace_project_name: dataset["Project"]["Name"]
-      }
-    end
-
-    # Remove all non-FASTQ files.
-    # We will add other file types as we encounter them.
-    formatted_samples = formatted_samples.select { |dataset| dataset[:file_type].present? }
-
-    render json: formatted_samples
-  end
-
-  private
-
-  def get_dataset_file_type(dataset)
-    if dataset["DatasetType"].present? && dataset["DatasetType"]["Name"].downcase.include?("fastq")
-      if dataset["Attributes"].present? && dataset["Attributes"]["common_fastq"].present? && dataset["Attributes"]["common_fastq"]["IsPairedEnd"] == true
-        return "Paired-end FASTQ"
-      else
-        return "Single-end FASTQ"
-      end
-    end
-
-    return nil
+    render json: bs_samples
   end
 end
--- a/app/helpers/basespace_helper.rb
+++ b/app/helpers/basespace_helper.rb
@@ -1,0 +1,126 @@
+require 'open-uri'
+
+# TODO(mark): Investigate if there is a way to fetch the user's current projects with v2 API. No obvious way from the docs.
+BASESPACE_CURRENT_PROJECTS_URL = "https://api.basespace.illumina.com/v1pre3/users/current/projects".freeze
+BASESPACE_PROJECT_DATASETS_URL = "https://api.basespace.illumina.com/v2/projects/%s/datasets".freeze
+BASESPACE_DATASET_FILES_URL = "https://api.basespace.illumina.com/v2/datasets/%s/files?filehrefcontentresolution=true".freeze
+# 1024 is the maximum limit allowed by Basespace.
+# If users reach this limit, we will need to implement multiple requests to fetch all the samples.
+BASESPACE_PAGE_SIZE = 1024
+
+module BasespaceHelper
+  def fetch_from_basespace(url, access_token, params = {})
+    HttpHelper.get_json(
+      url,
+      params.merge(limit: BASESPACE_PAGE_SIZE),
+      "Authorization" => "Bearer #{access_token}"
+    )
+  end
+
+  def basespace_projects(access_token)
+    response = fetch_from_basespace(BASESPACE_CURRENT_PROJECTS_URL, access_token)
+
+    if response.nil? || response.dig("Response", "Items").nil?
+      if response.present? && response.dig("ResponseStatus", "Message").present?
+        LogUtil.log_err_and_airbrake("basespace_projects failed with error: #{response['ResponseStatus']['Message']}")
+      end
+      return nil
+    end
+
+    # Just return selected fields.
+    response["Response"]["Items"].map do |dataset|
+      {
+        id: dataset["Id"],
+        name: dataset["Name"]
+      }
+    end
+  end
+
+  def samples_for_basespace_project(project_id, access_token)
+    response = fetch_from_basespace(BASESPACE_PROJECT_DATASETS_URL % project_id, access_token)
+
+    if response.nil? || response["Items"].nil?
+      if response.present? && response["ErrorMessage"].present?
+        LogUtil.log_err_and_airbrake("samples_for_basespace_project failed with error: #{response['ErrorMessage']}")
+      end
+      return nil
+    end
+
+    # Just return selected fields.
+    formatted_samples = response["Items"].map do |dataset|
+      {
+        basespace_dataset_id: dataset["Id"],
+        name: dataset["Name"],
+        file_size: dataset["TotalSize"],
+        file_type: get_dataset_file_type(dataset),
+        basespace_project_id: project_id,
+        basespace_project_name: dataset["Project"]["Name"]
+      }
+    end
+
+    # Remove all non-FASTQ files.
+    # We will add other file types as we encounter them.
+    formatted_samples.select { |dataset| dataset[:file_type].present? }
+  end
+
+  def files_for_basespace_dataset(dataset_id, access_token)
+    response = fetch_from_basespace(BASESPACE_DATASET_FILES_URL % dataset_id, access_token, filehrefcontentresolution: "true")
+
+    if response.nil? || response["Items"].nil?
+      if response.present? && response["ErrorMessage"].present?
+        LogUtil.log_err_and_airbrake("files_for_basespace_dataset failed with error: #{response['ErrorMessage']}")
+      end
+      return nil
+    end
+
+    return response["Items"].map do |file|
+      {
+        name: file["Name"],
+        # Path to download the file. Includes all auth information to download the file.
+        download_path: file["HrefContent"],
+        # Store the file id for debugging purposes.
+        # Without a valid access token, the file cannot be accessed using the file id.
+        source_path: file["Href"]
+      }
+    end
+  end
+
+  def upload_from_basespace_to_s3(basespace_path, s3_path, file_name)
+    temp_file = Tempfile.new
+    success = false
+
+    begin
+      # Download the basespace file into a temp file.
+      download = open(basespace_path)
+      IO.copy_stream(download, temp_file.path.to_s)
+
+      # Upload the temp file to S3.
+      upload_success = Syscall.s3_cp(temp_file.path.to_s, "#{s3_path}/#{file_name}")
+      # Raise an error if the copy failed.
+      raise StandardError, "failed to copy to S3 #{stderr}" unless upload_success
+
+      success = true
+    rescue => exception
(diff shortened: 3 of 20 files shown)


cdebourcy:
@davidrissato @tfrcarvalho any thoughts?

app/helpers/basespace_helper.rb
```diff
@@ -0,0 +1,122 @@
+require 'open-uri'
+
+BASESPACE_CURRENT_PROJECTS_URL = "https://api.basespace.illumina.com/v1pre3/users/current/projects".freeze
```
jshoe: it looks a little odd to have one endpoint from `v1pre3` and one from `v2`. Is there not a `v2/users`?

app/helpers/basespace_helper.rb
```diff
@@ -0,0 +1,122 @@
...
+
+module BasespaceHelper
+  def basespace_projects(access_token)
+    response = HttpHelper.get_json(
+      BASESPACE_CURRENT_PROJECTS_URL,
+      { limit: BASESPACE_PAGE_SIZE },
+      "Authorization" => "Bearer #{access_token}"
+    )
+
+    if response.nil? || response["Response"].nil? || response["Response"]["Items"].nil?
```
jshoe: you could also do something like `response&.dig("Response", "Items").nil?`

app/helpers/basespace_helper.rb
```diff
@@ -0,0 +1,122 @@
...
+    response["Response"]["Items"].map do |dataset|
+      {
+        id: dataset["Id"],
+        name: dataset["Name"]
+      }
+    end
+  end
+
+  def samples_for_basespace_project(project_id, access_token)
+    response = HttpHelper.get_json(
```
jshoe: you could potentially have another function that wraps the:

```
    response = HttpHelper.get_json(
      { limit: BASESPACE_PAGE_SIZE },
      "Authorization" => "Bearer #{access_token}"
    )
```

parts

app/helpers/basespace_helper.rb
```diff
@@ -0,0 +1,122 @@
...
+        # Store the file id for debugging purposes.
+        # Without a valid access token, the file cannot be accessed using the file id.
+        source_path: file["Href"]
+      }
+    end
+  end
+
+  def download_from_basespace_to_s3(basespace_path, s3_path, file_name)
+    temp_file = Tempfile.new
+    success = true
```
jshoe: might be safer long-term to have a default of success: false?

app/helpers/basespace_helper.rb
```diff
@@ -0,0 +1,122 @@
...
+
+    begin
+      # Download the basespace file into a temp file.
+      download = open(basespace_path)
+      IO.copy_stream(download, temp_file.path.to_s)
+
+      # Upload the temp file to S3.
+      _stdout, stderr, status = Open3.capture3("aws", "s3", "cp", temp_file.path.to_s, "#{s3_path}/#{file_name}")
```
jshoe: you should use Syscall.s3_cp here instead (safer long-term)

db/schema.rb
```diff
@@ -367,6 +367,8 @@
     t.text "dag_vars"
     t.integer "max_input_fragments"
     t.datetime "client_updated_at"
+    t.string "basespace_access_token"
+    t.string "basespace_dataset_id"
```
gregdingle: Based on the name "dataset", I would expect there to be a many to many relationship between samples and datasets. If that's not correct, please add a comment somewhere explaining why there is a one to one relationship.

app/models/sample.rb
```diff
@@ -328,6 +337,62 @@ def initiate_s3_cp
     LogUtil.log_err_and_airbrake("Failed to upload S3 sample '#{name}' (#{id}): #{e}")
   end
 
+  def initiate_basespace_file_transfer
+    # Only continue if the sample status is CREATED and not yet UPLOADED.
+    return unless status == STATUS_CREATED
+
+    Rails.logger.info("HERE #{basespace_dataset_id} #{basespace_access_token}")
```
jshoe: was this for debugging?

app/models/sample.rb
```diff
@@ -328,6 +337,62 @@ def initiate_s3_cp
...
+      while try < max_tries
+        success = download_from_basespace_to_s3(file[:download_path], sample_input_s3_path, file[:name])
+
+        if success
+          break
+        else
+          Rails.logger.error("Try ##{try}: Upload of sample '#{name}' (#{id}) file '#{file[:name]}' from Basespace failed")
+          try += 1
+          if try < max_tries
+            Kernel.sleep(time_between_tries)
```
jshoe: generally I would suggest an exponential backoff to avoid the scenario of many workers overwhelming our servers or theirs

app/models/sample.rb
```diff
@@ -328,6 +337,62 @@ def initiate_s3_cp
...
+    self.status = STATUS_UPLOADED
+    save
+  rescue => e
+    LogUtil.log_err_and_airbrake("SampleUploadFailedEvent: #{e}")
+
+    pr = new_failed_pipeline_run(self)
+    pr.error_message = PipelineRun::ERROR_BASESPACE_UPLOAD_FAILED
+    pr.save!
+    self.status = STATUS_CHECKED
+    save
```
jshoe: removing the `self.save` here would save one extra db call (since the `update` below is ensured

spec/helpers/pipeline_runs_helper_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    def create_new_failed_pipeline_run(sample)
+      pr = helper.new_failed_pipeline_run(sample)
+      pr.error_message = "Failed Pipeline Run"
+      pr.save
+      return pr
+    end
+
+    before do
+      # Create a dummy alignment config for the pipeline run.
+      alignment_config = create(:alignment_config)
```
jshoe: I haven't done this but I think Tiago would recommend using spec/factories/alignment_configs factories?

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -299,7 +299,15 @@ class SampleView extends React.Component {
   renderSampleReportError = () => {
     const { pipelineRun, sample } = this.props;
     let status, message, linkText, issueType, link;
-    switch (pipelineRun.known_user_error) {
+    switch (pipelineRun.known_user_error || pipelineRun.error_message) {
```
davidrissato: I believe field `error_message` should be used to store a description of the issue, and the error code (`BASESPACE_UPLOAD_FAILED`) should be stored in the `known_user_error`.

spec/helpers/basespace_helper_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+RSpec.describe BasespaceHelper, type: :helper do
+  describe "#files_for_basespace_dataset" do
+    context "#basespace API call returns successfully" do
+      before do
+        allow(HttpHelper).to receive(:get_json)
+          .and_return(
+            "Items" => [
+              {
+                "Name" => "Item 1",
+                "HrefContent" => "Download Path 1",
```
davidrissato: I suggest using fake values that resemble the actual data, because that helps as a documentation. For example: `"HrefContent" => "v1pre3/projects/1234"` or something alike

spec/helpers/basespace_helper_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+              name: "Item 2",
+              download_path: "Download Path 2",
+              source_path: "Source Path 2"
+            }
+          ]
+        )
+      end
+    end
+
+    context "#basespace API call fails" do
```
davidrissato: It looks like this test is more related to "basespace api returning zero results", correct?

spec/models/sample_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    context "fetches metadata for and downloads files successfully" do
+      before do
+        @sample = create_basespace_sample
+
+        allow_any_instance_of(BasespaceHelper).to receive(:files_for_basespace_dataset).and_return(
+          [
+            {
+              name: "file 1",
+              download_path: "download_path 1",
```
davidrissato: Same note about download path here.

spec/models/sample_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    context "cannot download basespace files" do
+      before do
+        @sample = create_basespace_sample
+
+        allow_any_instance_of(BasespaceHelper).to receive(:files_for_basespace_dataset).and_return(
+          [
+            {
+              name: "file 1",
+              download_path: "download_path 1",
```
davidrissato: Same note about download path here. Since you are repeating this fake value, maybe it would be nice to extract it to a constant to avoid repeating it.

app/models/sample.rb
```diff
@@ -328,6 +337,62 @@ def initiate_s3_cp
...
+    return unless status == STATUS_CREATED
+
+    Rails.logger.info("HERE #{basespace_dataset_id} #{basespace_access_token}")
+
+    files = files_for_basespace_dataset(basespace_dataset_id, basespace_access_token)
+
+    raise "Could not fetch Basespace files for dataset #{basespace_dataset_id} for sample '#{name}' (#{id})" if files.blank?
+
+    files.each do |file|
+      max_tries = 3
```
gregdingle: did you see any cases so far where retries were necessary? it would be good to add a comment about this in the code

app/models/sample.rb
```diff
@@ -328,6 +337,62 @@ def initiate_s3_cp
     LogUtil.log_err_and_airbrake("Failed to upload S3 sample '#{name}' (#{id}): #{e}")
   end
 
+  def initiate_basespace_file_transfer
```
gregdingle: `initiate_basespace_file_transfer` is somewhat misleading as a name because it also completes the transfer... I think most readers would expect it to be named `transfer_basespace_files`

app/models/sample.rb
```diff
@@ -328,6 +337,62 @@ def initiate_s3_cp
...
+  rescue => e
+    LogUtil.log_err_and_airbrake("SampleUploadFailedEvent: #{e}")
+
+    pr = new_failed_pipeline_run(self)
+    pr.error_message = PipelineRun::ERROR_BASESPACE_UPLOAD_FAILED
+    pr.save!
+    self.status = STATUS_CHECKED
+    save
+  ensure
+    # Delete the basespace access token
```
gregdingle: is this for security?

app/helpers/error_helper.rb
```diff
@@ -29,6 +29,10 @@ def self.invalid_project_id(sample)
     def self.missing_required_metadata(sample, missing_metadata_fields)
       "Could not save sample '#{sample['name']}'. Missing required metadata: #{missing_metadata_fields.join(', ')}"
     end
+
+    def self.missing_input_files_or_token(sample_name)
+      "Neither basespace_access_token nor input_files_attributes found in sample #{sample_name}"
```
gregdingle: why not split this into two different error messages?

spec/helpers/pipeline_runs_helper_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      alignment_config.name = AlignmentConfig::DEFAULT_NAME
```
gregdingle: if `AlignmentConfig::DEFAULT_NAME` is truly a default why doesn't the config object set it on creation?

app/assets/src/components/views/SampleView/SampleView.jsx
gregdingle: 👍

app/controllers/basespace_controller.rb
```diff
@@ -44,30 +39,16 @@ def projects
...
     end
 
-    # 1024 is the maximum limit allowed by Basespace.
-    # If users reach this limit, we will need to implement multiple requests to fetch all the projects.
-    response = HttpHelper.get_json(
-      BASESPACE_CURRENT_PROJECTS_URL,
-      { limit: BASESPACE_PAGE_SIZE },
-      "Authorization" => "Bearer #{access_token}"
-    )
+    projects = basespace_projects(access_token)
```
gregdingle: to avoid confusion with idseq projects, I recommend basespace projects prefixed everywhere they appear, for example, `bs_projects = basespace_projects(access_token)`

app/controllers/basespace_controller.rb
```diff
@@ -44,30 +39,16 @@ def projects
...
-    # Just return selected fields.
-    formatted_projects = response["Response"]["Items"].map do |dataset|
-      {
-        id: dataset["Id"],
-        name: dataset["Name"]
-      }
-    end
-
-    render json: formatted_projects
+    render json: projects
```
gregdingle: as above

app/controllers/basespace_controller.rb
```diff
@@ -88,51 +69,15 @@ def samples_for_project
...
     end
 
-    # 1024 is the maximum limit allowed by Basespace.
-    # If users reach this limit, we will need to implement multiple requests to fetch all the samples.
-    response = HttpHelper.get_json(
-      BASESPACE_PROJECT_DATASETS_URL % basespace_project_id,
-      { limit: BASESPACE_PAGE_SIZE },
-      "Authorization" => "Bearer #{access_token}"
-    )
+    samples = samples_for_basespace_project(basespace_project_id, access_token)
```
gregdingle: as above, rename to `bs_samples = samples_for_basespace_project(basespace_project_id, access_token)`

app/models/sample.rb
```diff
@@ -246,8 +251,12 @@ def fastqs_folder_files
   end
 
   def initiate_input_file_upload
-    return unless input_files.first.source_type == InputFile::SOURCE_TYPE_S3
-    Resque.enqueue(InitiateS3Cp, id)
+    if !input_files.empty? && input_files.first.source_type == InputFile::SOURCE_TYPE_S3
+      Resque.enqueue(InitiateS3Cp, id)
+    end
+    if basespace_upload?
```
gregdingle: should this be `elif` ? and it seems like there should by symmetry with `InputFile::SOURCE_TYPE_S3` ... why do we branch on `basespace_dataset_id.present?` and not `InputFile::SOURCE_TYPE_BASESPACE` ?

app/models/sample.rb
```diff
@@ -328,6 +337,62 @@ def initiate_s3_cp
...
+
+      input_files << input_file
+    end
+
+    self.status = STATUS_UPLOADED
+    save
+  rescue => e
+    LogUtil.log_err_and_airbrake("SampleUploadFailedEvent: #{e}")
+
+    pr = new_failed_pipeline_run(self)
```
gregdingle: It's not totally clear to me why we need to create a failed pipeline object when there is a failed basepace transfer. Is it because the transfer happens asynchronously after the user clicks submit on their samples? And so the pipeline object is the most convenient channel for reporting an error? And if so, why don't we do the same for s3 transfers that also appear to happen after the user clicks submit?

spec/helpers/basespace_helper_spec.rb
gregdingle: 👍

db/schema.rb
```diff
@@ -367,6 +367,8 @@
     t.text "dag_vars"
     t.integer "max_input_fragments"
     t.datetime "client_updated_at"
+    t.string "basespace_access_token"
```
tfrcarvalho: Why do we need to store these fields?

app/helpers/basespace_helper.rb
MarkAZhang: Not that I can see :( I pinged our Illumina contact again. Will add a TODO for now.

app/helpers/basespace_helper.rb
MarkAZhang: Wow!

app/models/sample.rb
MarkAZhang: I had trouble stubbing out sleep in the rspec test without calling it with `Kernel`, strangely enough. I can do exponential backoff.

db/schema.rb
MarkAZhang: No longer storing these in the database and instead passing them into resque as parameters. Added a comment explaining what a dataset is.

app/models/sample.rb
MarkAZhang: Sounds good! I didn't realize `update` updated all hanged values, not just the value passed in `update`.

jshoe: oh nvm it doesn't. you could do `self.status = STATUS_CHECKED` and `self.basespace_access_token = nil` and `save` in the ensure

spec/helpers/pipeline_runs_helper_spec.rb
MarkAZhang: The naming isn't great.

It doesn't mean "the default name for a newly created alignment config", but rather, "the name of the default alignment config to use for newly created samples". Let me know if you have a suggestion for a better name (I can't think of one readily)

spec/helpers/samples_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        expect(created_sample.name).to eq("mock_sample")
+        expect(created_sample.host_genome.id).to be @host_genome.id
+        expect(created_sample.project.id).to be @project.id
+        expect(created_sample.basespace_dataset_id).to eq("mock_dataset_id")
+        expect(created_sample.basespace_access_token).to eq("mock_access_token")
+
+        expect(Metadatum.where(sample_id: created_sample.id).length).to be 4
+      end
+
+      it "fails if neither basespace_access_token nor input_files_attributes is provided" do
```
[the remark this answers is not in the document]
MarkAZhang: `input_file_attributes` has a very specific meaning in Rails. It's a [nested attribute](https://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html).

Since I don't want `basespace_access_token` to be inserted into a new InputFile model, I have to use a separate variable.

app/assets/src/components/views/SampleView/SampleView.jsx
MarkAZhang: That makes sense regarding `error_message`. But not sure about `known_user_error`.

In my mind, `known_user_error` means an error on the user's part, instead of a technical error. A basespace upload failing could be due to user error (giving us an invalid dataset ID...but this is unlikely since they are picking files in our upload flow, not giving us dataset IDs directly), but more likely it is a network error or maybe the Basespace API changed.

A new field like `error_type` might be more appropriate. `known_user_error` seems like it's the kind of error where we wouldn't send an alert in Datadog, but I do think we want to send it in the basespace_upload_failed case.

Thoughts? Should I just go ahead and use `known_user_error`?

davidrissato: Makes sense, but I believe in this case we need to create support for non-user errors, like the error_type field you suggested (or perhaps error_code)

MarkAZhang: I can do `error_code`. I'll add a comment that `error_code` is for non-user errors.

MarkAZhang: Actually I prefer `error_type`, which implies a string. I think it's inconvenient to pass around a non-human-readable number, and then convert it to a string. For example, you'd have to convert it in the code above when you do the switch statement.

spec/helpers/basespace_helper_spec.rb
MarkAZhang: It's possible for the basespace API to fail as well. For example, with a 401 not authorized.

Returning zero results is a different error case that I didn't test. Just added tests for them.

app/models/sample.rb
MarkAZhang: It's in case there are transient network failures. I now do exponential back-off, as Jonathan suggested above. Will add comment.

app/models/sample.rb
MarkAZhang: Yes. But now I don't store it in the database at all, as suggested by Tiago.

app/helpers/error_helper.rb
MarkAZhang: Because you need one OR the other. I've changed the wording of the message to be a little clearer.

app/models/sample.rb
MarkAZhang: It's a limitation of the basespace API.

I added this comment to the code:
```
    # The reason why we don't check input_files.first.source_type == InputFile::SOURCE_TYPE_BASESPACE here is:
    # We don't yet have the input files at this point.
    # The basespace API doesn't let us query for all input files for a project.
    # We can only query all the datasets (and get the dataset IDs).
    # We then need to query each dataset individually (with a separate API call) to get the input files.
    # We do this query inside the TransferBasespaceFiles resque task.
```

app/models/sample.rb
MarkAZhang: That's a good point. I did this to be consistent with how we do our error handling elsewhere. In all other cases besides file upload, the failure would happen during the pipeline.

But I could see an argument for adding an `upload_error` field to the sample, since this technically happens independent of any pipeline.

spec/helpers/basespace_helper_spec.rb
```diff
@@ -0,0 +1,60 @@
+require "rails_helper"
+require "webmock/rspec"
+
+RSpec.describe BasespaceHelper, type: :helper do
```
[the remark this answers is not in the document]
MarkAZhang: They are tested in `basespace_controller_spec.rb`, since they are exposed via endpoints. The methods tested here aren't exposed via controller endpoints.

app/models/sample.rb
MarkAZhang: I ended up using an `upload_error` field on the sample, which I think makes the most sense.

MarkAZhang:
Verified that:
* Upload from Basespace succeeds.
* Upload from Basespace fails if you modify the Resque task to not pass the access token...and the sample correctly displays an error message (upload_error is correctly set)
* Upload from local still succeeds.
* Upload from s3 still succeeds.
* All new rspec tests pass.

Screen Shot 2019-07-17 at 6 00 45 PM

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -299,7 +303,17 @@ class SampleView extends React.Component {
   renderSampleReportError = () => {
     const { pipelineRun, sample } = this.props;
     let status, message, linkText, issueType, link;
-    switch (pipelineRun.known_user_error) {
+    switch (
+      sample.upload_error || (pipelineRun && pipelineRun.known_user_error)
```
davidrissato: Out of the scope of this PR, but I suggest we should add a ticket and refactor there error fields to more generic versions, like:
```
error_type = enum(UPLOAD_ERROR, KNOWN_USER_ERROR, SYSTEM_ERROR)
error_code = "SOME_ERROR_CODE"
error_description = "An extended error description"
error_additional_details = {...} # object with additional details
error_internal_details = {...} # object with internal details that can be used internally (ex: stacktrace, ids, service error messages, etc.)
```
```

PR #2403, opened 2019-07-15
Add condition on user bucket to prevent error.

MarkAZhang:
# Description

Tiny change in response to an error thrown by prod.


MarkAZhang:
Occurs when a non-admin user enters a malformed S3 path into the "Upload from S3" bucket field.

PR #2404, opened 2019-07-15
[PipelineViz] Create new pipelineviz controller

emilyzhong:
# Create new pipelineviz controller
Moves existing rails code + tests for pipeline viz to its own controller

# Tests
* Rspec tests
* Go to `samples/[sample_id]/pipeline_viz` on a pipeline_viz enabled account


config/routes.rb
```diff
@@ -113,6 +112,8 @@
   get 'basespace/projects', to: 'basespace#projects'
   get 'basespace/samples_for_project', to: 'basespace#samples_for_project'
 
+  get 'pipeline_viz', to: 'pipeline_viz#index'
```
tfrcarvalho: Probably matches a show route better:
`get '/pipeline_viz/:id', to: 'pipeline_viz#show'`

tfrcarvalho: Or even a nested resource: `/samples/:sample_id/pipeline_viz`?
Wdyt?

emilyzhong: Sounds good to me!

app/controllers/pipeline_viz_controller.rb
```diff
@@ -0,0 +1,40 @@
+class PipelineVizController < ApplicationController
+  before_action :authenticate_user!
+
+  current_power do
+    Power.new(current_user)
+  end
```
lvreynoso: IIRC this shouldn't be necessary, since `current_power` is already defined in the superclass `ApplicationController`.

app/controllers/pipeline_viz_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  # GET /sample/:sample_id/pipeline_viz
+  # GET /sample/:sample_id/pipeline_viz.json
+  def show
```
tfrcarvalho: Just a note: as a show action in a nested controller this would be: `/sample/:sample_id/pipeline_viz/:pipeline_id`. You can use the pipeline version as the id, and thus you would have support for different pipeline viz within same sample.

emilyzhong: I added selecting pipeline version into the method!

PR #2405, opened 2019-07-16
Find missing records script

gregdingle:
# Description

This adds a script we can run to find out integrity issues. 

Output looks like 

```
Finding OutputState records that are missing a parent record by belongs_to assocs [:pipeline_run] ...
Model OutputState has missing parents in records: {
  "71331": [
    "pipeline_run"
  ],
  "71332": [
    "pipeline_run"
  ],
  "71333": [
    "pipeline_run"
  ],
  "71334": [
    "pipeline_run"
  ],
  "71335": [
    "pipeline_run"
  ]
}
Finished checking 10 most recent records out of 71340 in the database.
```

I also added another less important script `find_test_records` that looks like this:

```
Finding Background records that contain 'test' or 'demo' in their name ...
Record #111 'test bg 3-29-19' owned by unknown appears to be test data
Record #100 'test' owned by unknown appears to be test data
Record #73 '2018-07-23_test' owned by unknown appears to be test data
Finished checking 100 most recent records out of 29 in the database.
```

# Notes

* To use effectively, we need to configure the `must_have_parents` and `must_have_children` correctly, and we need to keep them up to date. Please advise on what the values of these should be. 

* This could take hours to run on prod data. 

# Tests

* Ran `rake find_records_missing_children[10000]`
* Ran `rake find_records_missing_parents[10000]`
* Ran `rake find_test_records[10000]`


lib/tasks/find_records_missing.rake
```diff
@@ -0,0 +1,158 @@
...
+    :InputFile => [:sample],
+    :OutputState => [:pipeline_run],
+    :PipelineRun => %i[sample alignment_config],
+    :TaxonCount => [:pipeline_run],
+    :Background => [:project],
+    :Contig => [:pipeline_run],
+    :ErccCount => [:pipeline_run],
+    :FavoriteProject => %i[project user],
+    :JobStat => [:pipeline_run],
+    # :Location => %i[country state subdivision city],
```
jshoe: For Locations: I can't think of any must-have relationships. They don't even have to be matched to a sample/metadatum b/c we can have higher-level Locations just for clustering (e.g. "USA" even if we don't have samples that are just at "USA")

lib/tasks/find_records_missing.rake
```diff
@@ -0,0 +1,158 @@
...
+    # :Metadatum => %i[sample metadata_field location],
```
jshoe: Metadatum: All pieces of Metadata must belong to a type of MetadataField and one Sample 👍

lib/tasks/find_records_missing.rake
```diff
@@ -0,0 +1,158 @@
...
+def must_have_children
+  {
+    :AlignmentConfig => [:pipeline_runs],
+    :HostGenome => [:samples],
+    :MetadataField => [:metadata],
+    :PipelineRun => [
+       :pipeline_run_stages,
+       :output_states,
+       # :taxon_counts,
+       :job_stats,
```
jshoe: PipelineRun: I think job_stats are not required either (if a run fails really early). See pipeline_run#monitor_results (compiles the stats file after starting to check output states)

lib/tasks/find_records_missing.rake
```diff
@@ -0,0 +1,158 @@
...
+       # :taxon_byteranges,
+       # :ercc_counts,
+       # :amr_counts,
+       # :contigs
+    ],
+    # :User => %i[samples favorite_projects favorites visualizations phylo_trees],
```
jshoe: User: looks right that none of these are required

lib/tasks/find_records_missing.rake
```diff
@@ -0,0 +1,158 @@
...
+       :samples_unsafe, 
+       # :samples, 
+       # :favorite_projects, 
+       # :favorited_by, 
+       # :phylo_trees
+    ],
+    :Sample => [
+       :pipeline_runs,
+       :input_files,
+       # :metadata
```
jshoe: this is right that `metadata` is not required on the backend (from the user's perspective it is required to submit the sample but, for example, admins could skip adding metadata in the legacy upload page)

PR #2406, opened 2019-07-16
[WIP] Force-terminate zombie instances

cdebourcy:
Goes with https://github.com/chanzuckerberg/idseq-infra/pull/227
TODO:
- remove precondition: do terminate command regardless (will be no-op if unnecessary)
- test if terminate command works if instance has finished terminating
- shuffle instance order randomly in terminate command

### DIFF
--- a/app/jobs/autoscaling.py
+++ b/app/jobs/autoscaling.py
@@ -63,6 +63,8 @@ def delete_tags_from_instances(instance_ids, tag_keys):
     aws_command(cmd)
 
 def terminate_instances(instance_ids):
+    if not instance_ids:
+        return
     aws_command("aws ec2 terminate-instances --instance-ids {list_instances}".format(list_instances=' '.join(instance_ids)))
 
 def get_asg_list():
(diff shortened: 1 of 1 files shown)


app/jobs/autoscaling.py
```diff
@@ -102,6 +105,15 @@ def autoscaling_update(config):
         if not is_valid_classification:
             print "WARNING: some instances were classified into multiple states or none. This should never happen."
 
+        if num_discarded > 0 and num_desired > num_available + num_draining:
+            # Case where zombie instances exist.
+            # The "discarded" instances will not in fact get killed, because num_desired is higher than the number
```
tfrcarvalho: I guess I would like to understand why a new instance is not launched in this case... Shouldn't `desired` be compared with `available` only? Seems to me that draining should not count as active machines anymore. 
Also, yesterday it looked as if the script was considering the instance still active but busy. Wouldn't this be out of this case? 

Feel free to merge this if you think it mitigates the problem. We can discuss this offline.

cdebourcy: It doesn't launch a new instance, because from AWS perspective, DesiredCapacity is X and X instances already exist... the AWS autoscaling mechanism itself doesn't know about our draining logic. Conceptually, we would also not want to spin up new instances when there are draining instances because we would rather "rescue" the draining instances into the `available` state to save on init cost. 

However, you have a good point that our scaling mechanism here is a little too complicated.

cdebourcy:
Closing because the need for this change has not reoccurred and the change has the potential to lead to new problems down the line unless we invest in further investigation.

PR #2407, opened 2019-07-16
[PipelineViz] Move data-restructuring logic from react to rails

emilyzhong:
# Move data-restructuring logic from react to rails
Mainly logic from `createEdgeFileData` and `createStageStepData`

# Notes
~Built on top of branch `ezhong/pipeline_viz_controller` which moves existing rails code to a separate controller (hence the base being that branch). Will update when that branch gets merged :) 
PR here: https://github.com/chanzuckerberg/idseq-web/pull/2404~ Updated to master

# Tests
* Updated rspec tests
* Go to `samples/[sample_id]/pipeline_viz` on a pipeline_viz enabled account and it should look the same.


app/controllers/pipeline_viz_controller.rb
```diff
@@ -13,21 +13,47 @@ def show
     pipeline_run = sample && sample.first_pipeline_run
 
     if feature_allowed && pipeline_run
-      stage_info = {}
-      pipeline_run.pipeline_run_stages.each do |stage|
+      @stages = []
+      @edges = []
+      all_stage_dag_jsons = []
+      pipeline_run.pipeline_run_stages.each_with_index do |stage, stage_index|
```
tfrcarvalho: A few high-level comments:
- *To keep this controller thin*, we should move this logic to a `PipelineRunGraphService` (you can change the name) that takes care of all the data structure rearrangement.
- Add comments at the start to describe the structure you are going for. This will allow us to discuss the structure before discussing the actual code.
- Should we return a single structure, e.g. graph with edges and stages, as opposed to separate params?

app/controllers/pipeline_viz_controller.rb
```diff
@@ -37,4 +63,107 @@ def show
              }, status: status)
     end
   end
+
+  def stage_node_scaffolding(stage_dag_json)
```
tfrcarvalho: These and below should private... (but we should move it to the service anyway).

app/controllers/pipeline_viz_controller.rb
```diff
@@ -12,21 +12,47 @@ def show
...
+                       job_status: stage.job_status)
+          all_stage_dag_jsons.push(stage_dag_json)
+        end
+      end
+
+      between_stage_edges = inter_stage_edges(all_stage_dag_jsons, pipeline_run.pipeline_version)
+      between_stage_edges.each_with_index do |edge, edge_index|
+        # Offset index for correct value after concatination later
+        if edge[:from]
+          @stages[edge[:from][:stage_index]][:steps][edge[:from][:step_index]][:output_edges].push(edge_index + @edges.length)
```
lvreynoso: I think this could be more readable if you pull out `edge[:from][:stage_index]` as its own var.  So then this line could be something like

`@stages[edge_from][:steps][edge_from][:output_edges].push...`

Is it necessary to have the same key for to levels of your data structure here though? I know you're dealing with a lot of nested data here, but I'm wondering if it isn't redundant.

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -61,171 +56,29 @@ class PipelineViz extends React.Component {
...
+        return stepData.input_edges.map(edgeId => edgeFileInfo[edgeId]);
       case "output":
-        return stepData.outputEdges.map(edgeId => this.edgeFileData[edgeId]);
+        return stepData.output_edges.map(edgeId => edgeFileInfo[edgeId]);
       default:
-        return stepData.inputEdges
-          .concat(stepData.outputEdges)
-          .map(edgeId => this.edgeFileData[edgeId]);
+        return stepData.input_edges
+          .concat(stepData.output_edges)
```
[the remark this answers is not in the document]
emilyzhong: I switched over since the object created in rails using snake case fields 🐍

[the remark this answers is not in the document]
lvreynoso: According to the Dev Guidelines: "Use camelCase for properties in JSON object (even if writing in Ruby)."

app/controllers/pipeline_viz_controller.rb
emilyzhong: I use the same key mainly so that (on the javascript side) the same accessor methods can be used regardless of input/output edges, but I will pull out the indices into separate variables in this part of the code to make it readable 👍

app/controllers/pipeline_viz_controller.rb
emilyzhong: I created `RetrievePipelineVizGraphDataService`, added the comment, and changed it to a single structure so the service return value is easier / cleaner

app/controllers/pipeline_viz_controller.rb
```diff
@@ -1,29 +1,37 @@
...
 
     if feature_allowed && pipeline_run
-      stage_info = {}
-      pipeline_run.pipeline_run_stages.each do |stage|
-        if stage.name != "Experimental" || current_user.admin?
-          stage_info[stage.name] = JSON.parse(stage.dag_json || "{}")
-          stage_info[stage.name][:job_status] = stage.job_status
-        end
-      end
+      @results = RetrievePipelineVizGraphDataService.call(pipeline_run.id, current_user.admin?)
```
tfrcarvalho: Results does not need to be an instance variable

app/controllers/pipeline_viz_controller.rb
```diff
@@ -1,29 +1,37 @@
...
+  # edges: An array of edges, each edge object having the following structure:
+  #     - from: An object containing a stageIndex and stepIndex, denoting the originating node it is from
+  #     - to: An object containing a stageIndex and stepIndex, denoating the node it ends after
+  #     - files: An array of files that get passed between the from and to nodes. Currently each file is a string,
+  #       but it may become an object containing download url and potentially other information.
+
   # GET /sample/:sample_id/pipeline_viz/:pipeline_version
   # GET /sample/:sample_id/pipeline_viz/:pipeline_version.json
   def show
     feature_allowed = current_user.allowed_feature_list.include?("pipeline_viz")
```
tfrcarvalho: We should return here if feature not allowed, otherwise we will still do unnecessary extra DB queries.

app/controllers/pipeline_viz_controller.rb
```diff
@@ -1,29 +1,37 @@
 class PipelineVizController < ApplicationController
   before_action :authenticate_user!
 
+  # Structures dag_json of each stage of the pipeline run into the following in @results for drawing
```
tfrcarvalho: Would this description make more sense on the service?

app/services/retrieve_pipeline_viz_graph_data_service.rb
```diff
@@ -0,0 +1,157 @@
+class RetrievePipelineVizGraphDataService
```
tfrcarvalho: Would it make sense to have tests for this service?

app/controllers/pipeline_viz_controller.rb
emilyzhong: I thought I needed the @ symbol to make it available to the html.erb file?

app/services/retrieve_pipeline_viz_graph_data_service.rb
emilyzhong: We could, but they would look pretty similar to the tests for the pipeline_viz_controller, which essentially tests the outcome of this service along with some permission-ing.

app/controllers/pipeline_viz_controller.rb
tfrcarvalho: You do... Sorry, I thought you were only returning only JSON.

PR #2408, opened 2019-07-16
Surface upload limit in UI

cdebourcy:


app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
```diff
@@ -112,6 +112,17 @@ class LocalSampleFileUpload extends React.Component {
                 Paired files must be labeled with &quot;_R1&quot; or
                 &quot;_R2&quot; at the end of the basename.
               </li>
+              <li>
+                Each file must be &lt; 5 GB. To upload larger files, please use
```
gregdingle: where is the 5GB determined?

cdebourcy: It's an S3 limitation; it accepts 5GB as a single file, beyond that you need to implement multipart upload yourself (which we did for the CLI). Good point though, I'm extracting this number into a constant.

cdebourcy:
Closing this because I found out the window already raises an alert when files are bigger than 5GB.

PR #2409, opened 2019-07-16
IDSEQ-1133 - Improve permissions on results folders

jshoe:
### Description
- Only the original uploader should be able to get to raw_results_folder (otherwise they would never get a link to it from results_folder either).
- See https://jira.czi.team/browse/IDSEQ-1133

### Tests
- See tests

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -947,13 +947,15 @@ class SamplesController < ApplicationController
   end
 
   def raw_results_folder
-    can_see_stage1_results = (current_user.id == @sample.user_id)
-    if can_see_stage1_results
+    if current_user.id != @sample.user_id
+      render json: {
+        message: "Only the uploader can view the raw results folder."
+      }, status: :unauthorized
+    else
       @file_list = @sample.results_folder_files
       @file_path = "#{@sample.sample_path}/results/"
+      render template: "samples/raw_folder"
     end
-
-    render template: "samples/raw_folder"
   end
 
   def results_folder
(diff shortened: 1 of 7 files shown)


spec/controllers/pipeline_viz_controller_spec.rb
```diff
@@ -154,5 +154,21 @@
         expect(response).to have_http_status 401
       end
     end
+
+    describe "GET raw_results_folder (nonadmin)" do
```
jshoe: going to move to a different file..

app/controllers/samples_controller.rb
```diff
@@ -947,9 +947,15 @@ def unidentified_fasta
...
-    @file_list = @sample.results_folder_files
-    @file_path = "#{@sample.sample_path}/results/"
-    render template: "samples/raw_folder"
+    if current_user.id == @sample.user_id
+      @file_list = @sample.results_folder_files
+      @file_path = "#{@sample.sample_path}/results/"
+      render template: "samples/raw_folder"
+    else
+      render json: {
+        message: "Only the original uploader can view the sample's raw results folder."
```
gregdingle: how will this be communicated to the user? or should the never have a link to the raw results if they are not allowed to see them?

gregdingle: nevermind I see

app/controllers/samples_controller.rb
```diff
@@ -1265,4 +1268,13 @@ def sort_by(samples, dir = nil)
     end
     samples
   end
+
+  def check_owner
```
jshoe: this is here instead of ApplicationController b/c it's samples-specific and to access @sample

jshoe:
Updated to add OWNER_ACTIONS and add back admin access

PR #2411, opened 2019-07-16
Fix bug where date to become public was not displayed correctly.

tfrcarvalho:
### Description

Fix issue where date when next sample will become public was incorrect on `ProjectSettingsModal`.

### Tests

* Open a private project > Settings
* Verify that date on top header of the modal is correct.


jshoe:
I see
```
  const nextPublicSampleTimestamp = min(
    fetchedSamples
      .map(sample => moment(sample.privateUntil))
      .filter(timestamp => timestamp >= currentTimestamp)
  );
```
but I'm having trouble seeing why the final result had the current timestamp being displayed?

tfrcarvalho:
> I see
> 
> ```
>   const nextPublicSampleTimestamp = min(
>     fetchedSamples
>       .map(sample => moment(sample.privateUntil))
>       .filter(timestamp => timestamp >= currentTimestamp)
>   );
> ```
> 
> but I'm having trouble seeing why the final result had the current timestamp being displayed?

Because `moment(null)` Gives the current timestamp.

PR #2412, opened 2019-07-16
[AMR Heatmap] Sample metadata integration

lvreynoso:
# Description

This PR allows users to select sample metadata fields and sort samples by their metadata, just like in the Taxon heatmap.

# Notes

I found that in the test db, all the `MetadataField` objects already present (like `collection_location`) have `default_for_new_host_genome` set to `0`, which caused some trouble when trying to update the AMR controller tests.

# Tests

* Rspec tests updated to test for sample metadata
* For client side code, go to All Data page, pick a bunch of samples with AMR Counts (like tb-flash), click on Heatmap Icon and select "AMR Heatmap" from the dropdown, when the heatmap loads hover over metadata label to see the tooltip. Click on the "Add Metadata" button to add different metadata fields to display. Click on a metadata label to sort samples by metadata field.

Screen Shot 2019-07-16 at 3 25 47 PM


spec/factories/samples.rb
```diff
@@ -34,6 +34,16 @@
...
 
     after :create do |sample, options|
+      if MetadataField.table_exists?
+        # So it turns out that common metadata fields like collection_location
+        # and sample_type that are already present in the test db
+        # have 'default_for_new_host_genome' set to '0' for some reason.
+        # so trying to update the host genome with
+        # 'sample.host_genome.metadata_fields = MetadataField.where(default_for_new_host_genome: 1)'
+        # only works if you make up a new field. So just shove them all in for now.
+        sample.host_genome.metadata_fields = MetadataField.all
```
jshoe: I'm a little confused by these comments. Is the point just to ensure that there's some MetadataField on the sample?

app/assets/src/components/views/amr_heatmap/amr_heatmap_vis.scss
```diff
@@ -19,4 +19,28 @@
       display: block;
     }
   }
+
+  .legend {
+    padding: 10px;
+    background-color: $white;
+    border: 1px solid $light-grey;
+    border-radius: 5px;
+    box-shadow: 2px 2px 5px rgba(0, 0, 0, 0.2);
```
jshoe: you should use one of the box-shadows in _elements.scss or ask Jenn if you need to define a new one

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -50,28 +55,39 @@ export default class AMRHeatmapView extends React.Component {
...
 
-  async requestAMRCountsData(sampleIds) {
+  async requestSampleData(sampleIds) {
     const rawSampleData = await getAMRCounts(sampleIds);
     const filteredSamples = rawSampleData.filter(
       sampleData => sampleData.error === ""
     );
-    const samplesWithAMRCounts = this.processAMRCounts(filteredSamples);
+    const samplesWithKeyedMetadata = filteredSamples.map(sample =>
+      Object.assign({}, sample, {
```
MarkAZhang: Nit: I think 
```
({
  ...sample,
  metadata: processMetadata(sample.metadata, true)
})
```

is a bit cleaner.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -50,28 +55,39 @@ export default class AMRHeatmapView extends React.Component {
...
+        metadata: processMetadata(sample.metadata, true),
+      })
+    );
+    const samplesWithAMRCounts = this.processSampleAMRCounts(
+      samplesWithKeyedMetadata
+    );
     const maxValues = this.findMaxValues(samplesWithAMRCounts);
+    const rawSamplesMetadataTypes = await getSampleMetadataFields(
```
MarkAZhang: I would recommend using `Promise.all` to do the two HTTP requests in parallel. You can see examples in other files. This means less time waiting on the network.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -16,19 +17,41 @@ const METRICS = [
   { text: "Depth", key: "depth" },
 ];
 
+const DEFAULT_METADATA_TYPE = ["collection_location"];
```
MarkAZhang: Maybe `DEFAULT_SELECTED_METADATA` instead.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -16,19 +17,41 @@ const METRICS = [
...
     this.heatmap = null;
   }
 
   componentDidMount() {
+    this.processSampleData();
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    const { selectedMetadata } = this.state;
+    if (this.props !== prevProps || this.heatmap === null) {
```
MarkAZhang: I didn't catch this before, but comparing all props is bad practice. You should just compare the props that actually affect the rendering of the heatmap.

MarkAZhang: It's bad practice because you might run extra code unnecessarily.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -248,11 +307,86 @@ export default class AMRHeatmapVis extends React.Component {
...
+        onAddColumnMetadataClick: this.onMetadataAddButtonClick,
+        marginLeft: 75, // our gene names are very short, so this is to prevent metadata names from disappearing
       }
     );
     this.heatmap.start();
   }
 
+  //*** Render functions ***
+
+  renderMetadataLegend() {
```
MarkAZhang: These two components look nearly identical to the ones in `SamplesHeatmapVis` (both the JS and the CSS). Could you extract them into a shared component that both heatmaps can use?

MarkAZhang: Metadata legend and metadata selector.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -42,21 +42,23 @@ def amr_counts
         end
       end
       amr_data << {
-        sample_name: sample.name,
-        sample_id: sample.id,
-        amr_counts: amr_counts,
-        error: ""
+        "sampleName" => sample.name,
```
MarkAZhang: We seem to have a convention to pass snake_case fields back from Rails Controllers. Even though it leads to an awkward mix in React. I would follow convention and change to snake_case here, so we're at least consistent.

spec/factories/samples.rb
MarkAZhang: Yeah, not sure what problem you're trying to solve here. Were your tests failing without this line?

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -235,7 +294,7 @@ export default class AMRHeatmapVis extends React.Component {
       // in the rightmost box, and vice versa.
       {
         rowLabels: rows,
-        columnLabels: columns,
+        columnLabels: columns, // needs to contain selected metadata *values*
```
MarkAZhang: Can you change the comment to `Each columnLabel object contains the sample name, id, and all metadata values`? (I believe it's all metadata values, not selected?)

This part of the heatmap implementation is a little weird right now, so it's nice to have a thorough comment.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -248,11 +307,86 @@ export default class AMRHeatmapVis extends React.Component {
         onNodeHoverOut: this.onNodeHoverOut,
         onColumnLabelClick: onSampleLabelClick,
         onRowLabelClick: this.onRowLabelClick,
+        columnMetadata: this.getSelectedMetadataFields(), // gets the selected metadata *fields*
+        onColumnMetadataLabelMove: this.onNodeHoverMove,
+        onColumnMetadataLabelHover: this.onMetadataLabelHover,
+        onColumnMetadataLabelOut: this.onMetadataLabelOut,
+        onAddColumnMetadataClick: this.onMetadataAddButtonClick,
+        marginLeft: 75, // our gene names are very short, so this is to prevent metadata names from disappearing
```
MarkAZhang: Not sure you need this. I think the `rowLabelsWidth` in `Heatmap` already takes the metadata name lengths into account.

app/controllers/amr_heatmap_controller.rb
lvreynoso: I was going by https://github.com/chanzuckerberg/idseq-web/blob/master/DEV_GUIDELINES.md - see the last line.

spec/factories/samples.rb
lvreynoso: Yeah, so before when I tried to create a sample with metadata fields in the Rspec tests that use common metadata fields like `collection_location`, the test would not work. When the sample factory tried to create metadatum (the lines after this one) it would fail with the error `Invalid field for host genome Invalid field for metadata key`. I found through some debugging that this was because of some validation code in `metadata.rb`:

```
    # Check if the key is valid. Metadata_field was supposed to be set.
    valid_keys = sample.host_genome.metadata_fields.pluck(:name, :display_name).flatten
    unless key && valid_keys.include?(key) && metadata_field
      errors.add(:invalid_field_for_host_genome, MetadataValidationErrors::INVALID_FIELD_FOR_HOST_GENOME)
      return
    end
```

And when I looked, the `host_genome` being created in the `samples` factory did not have any `metadata_fields`, thus failing the validation. Passing brand new metadata fields in the test would work because they are created with `default_for_new_host_genome: 1` and so are added to the `host_genome` created in the factory. However all the existing metadata fields like `collection_location` and `sample_type` had `default_for_new_host_genome` set to `0`, causing tests using those fields to fail.

Updating them in the rails console worked, but since a new test db created with `docker-compose run --rm web "RAILS_ENV=test rake db:drop db:create db:migrate db:seed"` I was worried that they wouldn't pass the Travis CI tests either.

Admittedly this is not the most elegant solution, especially looking at it now. It might be more correct to look at the transient options and just add the metadata fields passed in. Or even better, change the test db so that common metadata fields don't have `default_for_new_host_genome: 0` by default

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: I'll see what I can do!

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: It does...until you click on them to sort the samples and the metadata names get cut off.

app/controllers/amr_heatmap_controller.rb
MarkAZhang: Talked to Tiago. This is good. You're totally right, I missed the dev guidelines.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
MarkAZhang: Ahh I see, good catch. The sort icon adds some additional width. Will "50" work? (it might based on the width of the sort icon plus the sort icon's left margin). Don't want to add extra unnecessary margin.

lvreynoso: Changed it to 50

app/assets/src/components/common/Heatmap/metadata_selector.scss
```diff
@@ -0,0 +1,33 @@
...
+      padding-top: 5px;
+      padding-bottom: 5px;
+    }
+  }
+
+  .hasBackground:nth-child(odd) {
+    background-color: $off-white;
+  }
+
+  // See SampleDetailsMode/metadata_section.scss
```
MarkAZhang: Not sure if we need this comment.

app/assets/src/components/common/Heatmap/MetadataLegend.jsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
```
MarkAZhang: Can you add a comment about this component, since it is now a shared component?

Something like "This component is used on both the AMR and Samples Heatmap. When the user hovers over a label for a metadata field, this component shows a tooltip legend for that field."

It's good to add comments for shared components.

app/assets/src/components/common/Heatmap/MetadataLegend.jsx
```diff
@@ -0,0 +1,41 @@
...
+              </div>
+            ))}
+        </div>
+      </div>
+    );
+  }
+}
+
+MetadataLegend.propTypes = {
+  columnMetadataLegend: PropTypes.object,
```
MarkAZhang: Can we rename this to metadataColors and add a comment saying "A map containing metadata values and their color in the heatmap."

app/assets/src/components/common/Heatmap/MetadataSelector.jsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
```
MarkAZhang: Same with this component. Add a comment explaining what it does and that it's used in both heatmaps.

PR #2414, opened 2019-07-17
Improve db queries, reduce Brakeman warnings, and add Brakeman [High] as blocking

jshoe:
### Description
- None of these changes were exploitable (as far as I know) so this is not sensitive.
- This PR address all the Brakeman High confidence warnings. I made minimal changes and verified the parameters already had whitelisted inputs or controlled params, but we can add even more parameterization and sanitization. We should also dive into the Medium confidence warnings later.
- As discussed, this PR sets up Travis such that Brakeman High confidence warnings will return a non-zero exit code and fail the build. Devs should run brakeman locally if they want to view a full report.
- I plan to write up more notes on preferred mitigations in general.

#### Before
Screen Shot 2019-07-17 at 3 12 15 PM
[and more]

#### After
Screen Shot 2019-07-17 at 3 12 59 PM
[and more]

### Tests
- Ran commands manually in local console to verify results.

phylo_tree#viewable:
- Ran `PhyloTree.viewable(User.find_by(role: 0))` (or other user with trees).
- SQL statement and output were correct.

phylo_tree#job_command:
- Ran `PhyloTree.find(114).send(:job_command)`
- SQL statement and output were correct.

pipeline_run#load_taxons and load_job_stats:
- Ran Open3.capture3 manually. Output was correct.

pipeline_run#update_names:
- Ran `PipelineRun.last.update_names`
- SQL statement and output were correct.

samples#resync_prod_data_to_staging:
- Ran with filler values (ex: sample 2000). SQL statement and output were correct.
- Note: there will be another task to remove this functionality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,10 +25,8 @@ script:
   - docker-compose run --rm webtest "rspec"
   - docker-compose run --rm webtest "npm test"
   - docker-compose run --rm webtest "pip install mock; python -m unittest discover -v -s /app/test/ -p 'test_*.py'"
-  # Run brakeman once to get a full report.
-  - brakeman --no-pager || true
-  # Run again on High so exit code will fail the build. Hide output to avoid confusion.
-  - brakeman --no-pager -w3 >/dev/null 2>&1
+  # Block on High Confidence warnings
+  - brakeman --no-pager -w3
 env:
   global:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -25,7 +25,10 @@ script:
   - docker-compose run --rm webtest "rspec"
   - docker-compose run --rm webtest "npm test"
   - docker-compose run --rm webtest "pip install mock; python -m unittest discover -v -s /app/test/ -p 'test_*.py'"
+  # Run brakeman once to get a full report.
```
cdebourcy: Do we still need this?

jshoe: I was thinking it'd be easier for people to notice their Low/Medium warnings. but I also doubt people check it so.. 👍/👎?

app/models/pipeline_run.rb
```diff
@@ -645,7 +645,7 @@ def load_taxons(downloaded_json_path, refined = false)
     update_is_phage
 
     # rm the json
-    _stdout, _stderr, _status = Open3.capture3("rm -f #{downloaded_json_path}")
+    _stdout, _stderr, _status = Open3.capture3("rm", "-f", downloaded_json_path)
```
cdebourcy: Did your tests include checking whether this file actually got removed?

jshoe: yep 👍

.travis.yml
jshoe: ok I just removed the full run. Non-actioning reports don't really belong in CI

PR #2418, opened 2019-07-18
[AMR Heatmap] Add option to download amr data in csv format

lvreynoso:
# Description

Adds a dropdown menu to the AMR Heatmap header that allows users to download the AMR data used in the heatmap as a csv file.

# Notes

Although the UI appearance is identical to the same functionality implemented in the taxon heatmap, only the frontend is used to build the csv file.

# Tests

* No server side code changed
* For client side code, go to All Data page, pick a bunch of samples with AMR Counts (like tb-flash), click on Heatmap Icon and select "AMR Heatmap" from the dropdown. When the heatmap loads, click on the "Download" button, then click "Download CSV" on the dropdown menu. Then open up the downloaded csv file to verify.

Screen Shot 2019-07-18 at 3 22 55 PM

Screen Shot 2019-07-18 at 3 23 19 PM

### DIFF
--- a/app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
+++ b/app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
@@ -7,6 +7,11 @@ import AMRHeatmapVis from "~/components/views/amr_heatmap/AMRHeatmapVis";
 import DetailsSidebar from "~/components/common/DetailsSidebar";
 import ErrorBoundary from "~/components/ErrorBoundary";
 import { getAMRCounts } from "~/api/amr";
+import { getSampleMetadataFields } from "~/api/metadata";
+import {
+  processMetadata,
+  processMetadataTypes,
+} from "~/components/utils/metadata";
 import LoadingIcon from "~ui/icons/LoadingIcon";
 import { ViewHeader, NarrowContainer } from "~/components/layout";
 import { DownloadButtonDropdown } from "~ui/controls/dropdowns";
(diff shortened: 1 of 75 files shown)


app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -185,10 +206,38 @@ export default class AMRHeatmapView extends React.Component {
...
+        rel="noopener noreferrer"
+        key={`${csvOption.text}_csv_link`}
+      >
+        {csvOption.text}
+      </a>
+    );
+  }
+
+  getDownloadOptions() {
+    const downloadOptions = DOWNLOAD_OPTIONS;
```
MarkAZhang: I think you can make this function cleaner by just doing:
```
return {
  value: "csv",
  text: this.getDownloadCSVLink()
}
```

Don't think you need the additional variables.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -185,10 +206,38 @@ export default class AMRHeatmapView extends React.Component {
     }
   }
 
+  getDownloadCSVLink(csvOption) {
+    const dataString = this.computeHeatmapValuesForCSV();
+    const dataBlob = new Blob([dataString], { type: "text/csv" });
```
MarkAZhang: Would it be possible to factor some of this into a separate method `createCSVObjectURL` in `utils/csv.js`?

Maybe `createCSVObjectURL`  can take the headers and rows that you compute in `computeHeatmapValuesForCSV` and handle the logic of creating the data string, turning it into a blob, and turning it into an object URL. This feels like a method that could be useful in the future.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -185,10 +206,31 @@ export default class AMRHeatmapView extends React.Component {
     }
   }
 
+  getDownloadCSVLink() {
+    const csvOption = DOWNLOAD_OPTIONS[0];
```
MarkAZhang: I don't think you need `DOWNLOAD_OPTIONS`. You can replace `{csvOption.text}` below with the string value.

PR #2419, opened 2019-07-18
 Add foreign key constraints to selected tables

gregdingle:
# Description

This add FK constraints to tables that we expect to always have parent records, so we don't get any nasty surprises, and we always delete dependent data when requested, or raise an exception. 

# Notes

* I used the `immigrant` gem to generate the migration which generates migration for all possible tables. 
* I'm not sure whether constraints on many-to-many junction tables are a good idea. I hadn't thought about it so I ignored them all. 
* I went through all the affected assocs and updated the `dependent` directive to either `destroy` or `restrict_with_exception` so the rails behavior reflects the db behavior

# Tests

* rails db:migrate in dev, see no errors
* run in staging, verify no valid rows dropped
* verify there are no violations in prod before deploying -- some db cleanup will be required


db/migrate/20190718233448_add_foreign_key_constraints.rb
```diff
@@ -0,0 +1,24 @@
...
+    add_foreign_key "output_states", "pipeline_runs", name: "output_states_pipeline_run_id_fk"
+    add_foreign_key "phylo_trees", "projects", name: "phylo_trees_project_id_fk"
+    add_foreign_key "phylo_trees", "users", name: "phylo_trees_user_id_fk"
+    add_foreign_key "pipeline_run_stages", "pipeline_runs", name: "pipeline_run_stages_pipeline_run_id_fk"
+    add_foreign_key "pipeline_runs", "alignment_configs", name: "pipeline_runs_alignment_config_id_fk"
+    add_foreign_key "pipeline_runs", "samples", name: "pipeline_runs_sample_id_fk"
+    add_foreign_key "samples", "host_genomes", name: "samples_host_genome_id_fk"
+    add_foreign_key "samples", "projects", name: "samples_project_id_fk"
+    add_foreign_key "samples", "users", name: "samples_user_id_fk"
+    add_foreign_key "taxon_summaries", "backgrounds", name: "taxon_summaries_background_id_fk"
```
gregdingle: I can't remember, is this optional?

cdebourcy:
Could you include comprehensive performance research and testing of all mission critical flows in the app? In both staging and then prod? Recall how in the past you found that performance research in dev/staging was not representative of what happens in prod. I think two people should be in charge of carefully validating this and related PRs. Looking at @davidrissato for his analytical prowess and thoroughness of testing :)

gregdingle:
I did some quick research into mysql perf and foreign keys:

* http://johanandersson.blogspot.com/2013/06/benchmarking-performance-impact-of.html

* https://www.percona.com/blog/2010/09/20/instrumentation-and-the-cost-of-foreign-keys/

Conclusion: FK perf cost is small (as expected) but it can be too much depending on the situation

Then I talked with @davidrissato and @cdebourcy about performance benchmarking. we decided to go ahead with the stage 1 only rollout plan, with some follow-ups:

* Email YF and ask her about what table was locking

* Monitor slow query logs (https://aws.amazon.com/about-aws/whats-new/2018/05/amazon-aurora-publishes-general-slow-query-and-error-logs-to-amazon-cloudwatch/)

* Keep monitoring endpoint performance: https://chanzuckerberg.airbrake.io/projects/158403/performance?environment=prod&sortColumn=total.p95&sortDesc=true&page=1

gregdingle:
Another question raised by @davidrissato is whether we should delete all samples of a project if a project is deleted? What if the sample is public? 

Currently samples are *not* deleted. 

Likewise, if a user is deleted, should their samples all be deleted? And all the dependent data? 

Does the privacy policy say? @cdebourcy ?

gregdingle:
> I'm not sure whether constraints on many-to-many junction tables are a good idea. I hadn't thought about it so I ignored them all.

@davidrissato convinced me this is a good idea so I will add them here as well, and clean up and bad data in the junction tables.

db/migrate/20190718233448_add_foreign_key_constraints.rb
```diff
@@ -0,0 +1,23 @@
+class AddForeignKeyConstraints < ActiveRecord::Migration[5.1]
+  def change
+    add_foreign_key "amr_counts", "pipeline_runs", name: "amr_counts_pipeline_run_id_fk", on_delete: :cascade
+    add_foreign_key "ercc_counts", "pipeline_runs", name: "ercc_counts_pipeline_run_id_fk", on_delete: :cascade
```
davidrissato: I think we shouldn't enable this one yet (ercc_counts has 2.5M records). I would add foreign keys in batches (one per week) and monitor to see if we have any big performance issue.

db/migrate/20190718233448_add_foreign_key_constraints.rb
```diff
@@ -0,0 +1,23 @@
...
+    add_foreign_key "output_states", "pipeline_runs", name: "output_states_pipeline_run_id_fk", on_delete: :cascade
+    add_foreign_key "phylo_trees", "projects", name: "phylo_trees_project_id_fk", on_delete: :cascade
+    add_foreign_key "phylo_trees", "users", name: "phylo_trees_user_id_fk", on_delete: :cascade
+    add_foreign_key "pipeline_run_stages", "pipeline_runs", name: "pipeline_run_stages_pipeline_run_id_fk", on_delete: :cascade
+    add_foreign_key "pipeline_runs", "alignment_configs", name: "pipeline_runs_alignment_config_id_fk", on_delete: :cascade
+    add_foreign_key "pipeline_runs", "samples", name: "pipeline_runs_sample_id_fk", on_delete: :cascade
+    add_foreign_key "samples", "host_genomes", name: "samples_host_genome_id_fk", on_delete: :cascade
+    add_foreign_key "samples", "projects", name: "samples_project_id_fk", on_delete: :cascade
+    add_foreign_key "samples", "users", name: "samples_user_id_fk", on_delete: :cascade
+    add_foreign_key "taxon_summaries", "backgrounds", name: "taxon_summaries_background_id_fk", on_delete: :cascade
```
davidrissato: Same for `taxon_summaries`. 2.5M. I suggest adding this FK during a second iteration.

gregdingle:
Some research on the question of how to delete:

Privacy policy:
image

Sample.rb
```  def deletable?(user)
    if user.admin?
      true
    elsif user_id == user.id
      # Sample belongs to the user
      # Allow deletion if no pipeline runs, or report failed.
      unless pipeline_runs.empty?
        pipeline_runs.each do |prun|
          return false unless prun.report_failed?
        end
      end
      true
    end
  end
```

So it looks like 
1) the uploader can delete a sample and all visualizations even if it is public
2) the uploader cannot delete a sample if it has successful pipelines... this looks in contradiction to privacy policy

cdebourcy:
The policy is about users requesting to have their data deleted via email, not about the sample Delete button in the app

> Some research on the question of how to delete:
> 
> Privacy policy:
> image
> 
> Sample.rb
> 
> ```
> def deletable?(user)
>     if user.admin?
>       true
>     elsif user_id == user.id
>       # Sample belongs to the user
>       # Allow deletion if no pipeline runs, or report failed.
>       unless pipeline_runs.empty?
>         pipeline_runs.each do |prun|
>           return false unless prun.report_failed?
>         end
>       end
>       true
>     end
>   end
> ```
> 
> So it looks like
> 
> 1. the uploader can delete a sample and all visualizations even if it is public
> 2. the uploader cannot delete a sample if it has successful pipelines... this looks in contradiction to privacy policy

The policy is

gregdingle:
Note I went through all the affected assocs and updated the `dependent` directive to either `destroy` or `restrict_with_exception` so the rails behavior reflects the db behavior. `destroy` when it is safe in my judgement and `restrict_with_exception` when it is unclear.

cdebourcy:
Beyond checking for errors, do you mind verifying that the data migration produces the expected results in staging and no valid rows are dropped? Thanks!

gregdingle:
Yes, my plan was to look carefully at staging before prod. I updated the test plan.

While I wrote everything to be executed in the rails migration, I would actually prefer to run each sql `DELETE` from the console first with someone else beside to be in maximum control. 

We could then even run the migration with all the `DELETE`s commented out.

cdebourcy:
> Yes, my plan was to look carefully at staging before prod. I updated the test plan.
> 
> While I wrote everything to be executed in the rails migration, I would actually prefer to run each sql `DELETE` from the console first with someone else beside to be in maximum control.
> 
> We could then even run the migration with all the `DELETE`s commented out.

Awesome!

PR #2420, opened 2019-07-19
[Pipeline Viz] Sidebar File Download Links

emilyzhong:
# Description
Adds download links for files on sidebar

*Sample owner can download any files*
sample_owner_can_download_all

*Non sample owner can download only files after host filtering*
not sample owner

# Notes
* Only admin will be able to see experimental results
* Host filtering results only visible by sample owner
* Refactor of how edges are constructed in pipeline viz graph service

# Tests

* **Backend**:  New `retrieve_pipeline_viz_graph_data_service_spec.rb` tests!
* **Frontend**: Go to `samples/[sample id]/pipeline_viz`, click on a node, then click on file on sidebar to download.


app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -134,10 +133,10 @@ class PipelineViz extends React.Component {
       .map(edgeInfo => {
         // Remove duplicate output file listings
         return edgeInfo.files.reduce((files, fileInfo) => {
-          if (!seenFiles.has(fileInfo)) {
-            seenFiles.add(fileInfo);
-            // TODO(ezhong): Include file url for downloading
-            files.push({ fileName: fileInfo });
+          let fileInfoAsString = JSON.stringify(fileInfo);
```
tfrcarvalho: `let` -> `const`

app/assets/src/components/common/DetailsSidebar/PipelineStepDetailsMode/PipelineStepDetailsMode.jsx
```diff
@@ -77,13 +79,17 @@ PipelineStepDetailsMode.propTypes = {
     PropTypes.shape({
       fromStepName: PropTypes.string,
       files: PropTypes.arrayOf(
```
tfrcarvalho: You can define 
`
const FileList = PropTypes.arrayOf(
    PropTypes.shape({
      fileName: PropTypes.string.isRequired,
      url: PropTypes.string,
    }).isRequired
  )
`
and avoid the repetition.
If you use it other components, you can event extract it to an util file.

PR #2421, opened 2019-07-19
[Pipeline Viz] Step descriptions

emilyzhong:
# Description
Adds step descriptions to the sidebar. Descriptions are from `pipeline_runs_helper.rb` (originally used for `results_folder` view).

image

# Notes
* Temporary — planning to include more dynamic step descriptions to ensure they are included for pipeline updates/changes 

# Tests
Go to `samples/[sample id]/pipeline_viz` and click on a node to see description on the sidebar (Not all nodes may have samples since descriptions are hardcoded / not necessarily updated with each pipeline change)

### DIFF
--- a/app/services/retrieve_pipeline_viz_graph_data_service.rb
+++ b/app/services/retrieve_pipeline_viz_graph_data_service.rb
@@ -1,5 +1,6 @@
 class RetrievePipelineVizGraphDataService
   include Callable
+  include PipelineRunsHelper # For step descriptions
 
   # Structures dag_json of each stage of the pipeline run into the following in @results for drawing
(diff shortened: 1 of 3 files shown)


app/services/retrieve_pipeline_viz_graph_data_service.rb
```diff
@@ -1,5 +1,6 @@
 class RetrievePipelineVizGraphDataService
   include Callable
+  include PipelineRunsHelper # For step descriptions
```
tfrcarvalho: Can you add a TODO here that we will move to server-fed descriptions.

PR #2422, opened 2019-07-19
[Pipeline Viz] Remove ReactDOM.findDOMNode

emilyzhong:
# Description
Replaces `ReactDOM.findDOMNode` with refs (https://reactjs.org/docs/strict-mode.html#warning-about-deprecated-finddomnode-usage) when transforming DOM coordinates to detect interaction

# Notes
* Uses `React.createRef` instead of callback refs in order for `.current` to work

# Tests
Go to `samples/[sample id]/pipeline_viz' on a pipeline viz enabled account and everything should work the same as before

### DIFF
--- a/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
+++ b/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
@@ -1,5 +1,4 @@
 import React from "react";
-import ReactDOM from "react-dom";
 import PropTypes from "prop-types";
 import { PanZoom } from "react-easy-panzoom";
 import cx from "classnames";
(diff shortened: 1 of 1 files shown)


app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -719,9 +719,7 @@ class PipelineViz extends React.Component {
           minZoom={zoomMin}
           maxZoom={zoomMax}
           zoomSpeed={3}
-          ref={ref => {
```
tfrcarvalho: Do you know why setting this panZoomContainer like this does not work?
Is it null whenever you request current the first time?

emilyzhong: Yeah current is null if I use a callback ref

PR #2425, opened 2019-07-22
Change rspec db setup to be same as rails test

gregdingle:
# Description

In working on https://github.com/chanzuckerberg/idseq-web/pull/2419, I found that rspec was running off of `schema.rb` version of our db.  Following https://github.com/chanzuckerberg/idseq-web/pull/2419 , `schema.rb` will no longer execute because the foreign key constraints require creating the tables in a specific order.  `schema.rb` is alphabetical.

In any case, we should not use `schema.rb` for more than documentation because we use the migrated db in local, staging and prod. 

# Tests

* Run `RAILS_ENV=test rake db:drop db:create db:migrate db:seed`
* Run rspec, see no errors

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ script:
   - docker-compose run --rm webtest "rubocop -R --config .rubocop_todo.yml"
   - docker-compose run --rm webtest "rake db:drop db:create db:migrate db:seed"
   - docker-compose run --rm webtest "rails test"
-  - docker-compose run --rm webtest "rake db:drop db:create db:migrate db:reset"
+  - docker-compose run --rm webtest "rake db:drop db:create db:migrate db:seed"
   - docker-compose run --rm webtest "rspec"
   - docker-compose run --rm webtest "npm test"
   - docker-compose run --rm webtest "pip install mock; python -m unittest discover -v -s /app/test/ -p 'test_*.py'"
--- a/spec/factories/samples.rb
+++ b/spec/factories/samples.rb
@@ -25,11 +25,14 @@ FactoryBot.define do
     # ensure the metadata field is added to the host genome
     before(:create) do |sample, options|
       options.metadata_fields.each_key do |metadata_field_name|
-        unless MetadataField.exists?(name: metadata_field_name)
-          metadata_field = create(:metadata_field, name: metadata_field_name, default_for_new_host_genome: 1)
-          # This is needed to make metadata pass validation below
-          sample.host_genome.metadata_fields << metadata_field
+        metadata_field = MetadataField.find_by(name: metadata_field_name)
+        unless metadata_field
+          metadata_field = create(:metadata_field, name: metadata_field_name)
         end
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -21,7 +21,7 @@ script:
   - docker-compose run --rm webtest "rubocop -R --config .rubocop_todo.yml"
   - docker-compose run --rm webtest "rake db:drop db:create db:migrate db:seed"
   - docker-compose run --rm webtest "rails test"
-  - docker-compose run --rm webtest "rake db:drop db:create db:migrate db:reset"
+  - docker-compose run --rm webtest "rake db:drop db:create db:migrate db:seed"
```
tfrcarvalho: Do not run `db:seed`. Database should be empty for tests.

gregdingle: Why is `db:seed` listed before `rails test` then? From the rails docs, `db:seed` is for loading data that should appear in prod. It is not the same as fixtures.

tfrcarvalho: Because the old tests assume existing data, which they should not.

tfrcarvalho: We should run seed nor should we have loaded data in the migrations (which is why the reset is there).

spec/factories/samples.rb
```diff
@@ -23,13 +23,11 @@
...
-    # COuld find out why the host genome metadata fields could not be updated.
-    before(:create) do |_sample, options|
+    before(:create) do |sample, options|
       options.metadata_fields.each_key do |metadata_field_name|
-        unless MetadataField.exists?(name: metadata_field_name)
-          create(:metadata_field, name: metadata_field_name, default_for_new_host_genome: 1)
-        end
+        metadata_field = create(:metadata_field, name: metadata_field_name, default_for_new_host_genome: 1)
+        # This is needed to make metadata pass validation below
+        sample.host_genome.metadata_fields << metadata_field
```
gregdingle: I reviewed this briefly with @MarkAZhang

gregdingle: I don't see why it should... but then the test is failing now with a single error related to metadata: 

image

Any ideas @MarkAZhang ?

gregdingle: The test passes for me locally.

tfrcarvalho: Because it will create a the same metadata field twice. That should fail if we have a restriction on unique names. Thats why the unless is there. Do you mind trying to associate the same metadata field with two samples?
BTW, the above error does not seem related to the this particular potential problem.

gregdingle: @tfrcarvalho I see now. I took out the `unless` for testing and I forgot to put it back. Now fixed. Good catch!

.travis.yml
tfrcarvalho: As discussed offline, I was wrong about the seeding... `db:reset`, is intended to remove all data from the data from the database to avoid data added by migrations. 
Since the test are not failing and I do not remember a case that was failing because of this I will approve the change.

PR #2427, opened 2019-07-23
[PipelineViz] Move dag_json creation to beginning of pipeline run

emilyzhong:
# Description
* Moves creation of all `dag_json`s to the beginning of the pipeline run (instead of the beginning of the corresponding pipeline run *stage*)
* ~Fixes frontend so that older erroring pipeline versions without this change will still be able to render a pipeline visualization~ Done in separate PR (see conversation)

# Notes
* Frontend will probably change / is temporary (right now, stages without dag_json is faded out)
Screen Shot 2019-07-22 at 6 01 41 PM

* Moving dag json creation earlier will allow user to view future stages while pipeline is still running or if the pipeline has errored on a previous stage 
   * Keeping creation of the dag json where it currently was (at the beginning of stage execution, not pipeline execution) could be useful in the case where the pipeline becomes more dynamic and future stages are dependent on the output of previous steps. For the time being, however, since there doesn't seem to be any plans of that, moving the dag_json to be created earlier seems okay
  * Previously, the dag_json was created when the pipeline stage is kicked off, in the same method that creates the job command. This PR splits the pipeline creation and job command creation into two separate methods, and calls the pipeline creation method when the pipeline stage is initially created

# Tests
* Updated pipeline_run rspec tests to check that dag_jsons exist for each stage despite erroring
* Ran sample through both original pipeline + pipeline with dag_jsons all created in beginning, and dag_jsons were the same
* Ran sample through pipeline with changes and it works!
* Frontend — go to pipeline viz on an error-ed file

### DIFF
--- a/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
+++ b/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
@@ -1,5 +1,4 @@
 import React from "react";
-import ReactDOM from "react-dom";
 import PropTypes from "prop-types";
 import { PanZoom } from "react-easy-panzoom";
 import cx from "classnames";
@@ -36,6 +35,7 @@ class PipelineViz extends React.Component {
       y: 0,
       alteredGraphs: new Set(),
     };
+    this.panZoomContainer = React.createRef();
 
     this.state = {
       stagesOpened: [true, true, true, true],
@@ -84,7 +84,7 @@ class PipelineViz extends React.Component {
 
   getNodeIdAtCoords(graph, xCoord, yCoord) {
     const { x, y } = inverseTransformDOMCoordinates(
-      ReactDOM.findDOMNode(this.panZoomContainer).firstChild,
+      this.panZoomContainer.current.dragContainer.current,
       xCoord,
       yCoord
     );
@@ -718,9 +718,7 @@ class PipelineViz extends React.Component {
           minZoom={zoomMin}
           maxZoom={zoomMax}
           zoomSpeed={3}
-          ref={ref => {
-            this.panZoomContainer = ref;
-          }}
+          ref={this.panZoomContainer}
         >
           <div className={cs.pipelineViz}>{stageContainers}</div>
         </PanZoom>
--- a/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
+++ b/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
@@ -523,6 +523,7 @@ class PipelineViz extends React.Component {
     const {
       graphData: { stages },
     } = this.props;
+
     stages.forEach((_, i) => {
       this.drawStageGraph(i);
     });
--- a/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
+++ b/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
@@ -655,7 +655,8 @@ class PipelineViz extends React.Component {
     const {
       graphData: { stages },
     } = this.props;
-    const isOpened = this.state.stagesOpened[i];
+    const { stagesOpened, hovered } = this.state;
+    const isOpened = stagesOpened[i];
 
     // Stages without dag_json recorded are not toggleable
     const toggleable = i < stages.length;
@@ -670,7 +671,7 @@ class PipelineViz extends React.Component {
           />
         </div>
         <div
-          className={cx(cs.graph, this.state.hovered && cs.hovered)}
+          className={cx(cs.graph, hovered && cs.hovered)}
           onMouseMove={e => this.handleMouseMove(i, e)}
           ref={ref => {
             this.graphContainers[i] = ref;
@@ -684,7 +685,7 @@ class PipelineViz extends React.Component {
         <div
           className={cx(
             isOpened && toggleable ? cs.hidden : cs.stageButton,
-            !toggleable && cs.notToggableStageButton
+            !toggleable && cs.disabled
           )}
           onClick={() => this.toggleStage(i)}
         >
--- a/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
+++ b/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
@@ -636,7 +636,6 @@ class PipelineViz extends React.Component {
       options
     );
     this.graphs.push(currStageGraph);
-
     currStageGraph.minimizeSizeGivenScale(1.0);
     this.centerEndNodeVertically(currStageGraph);
     this.closeIfNonActiveStage(index);
@@ -661,7 +660,7 @@ class PipelineViz extends React.Component {
     // Stages without dag_json recorded are not toggleable
     const toggleable = i < stages.length;
 
-    const openedStageContainer = toggleable && (
+    const stageContainer = toggleable && (
       <div className={isOpened ? cs.openedStage : cs.hidden}>
         <div className={cs.graphLabel}>
           {stageName}
@@ -691,7 +690,7 @@ class PipelineViz extends React.Component {
         >
           {stageName}
         </div>
-        {openedStageContainer}
+        {stageContainer}
       </div>
     );
   }
--- a/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
+++ b/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
@@ -38,7 +38,10 @@ class PipelineViz extends React.Component {
     this.panZoomContainer = React.createRef();
 
     this.state = {
-      stagesOpened: [true, true, true, true],
+      // Set stages that can be rendered to open, and others to closed.
+      stagesOpened: this.stageNames.map(
+        (_, i) => i < this.props.graphData.stages.length
+      ),
       interStageArrows: ["", "", ""],
       sidebarVisible: false,
       sidebarParams: {},
@@ -505,18 +508,22 @@ class PipelineViz extends React.Component {
     graph.moveNodeToPosition(END_NODE_ID, xEndNodePos, yStartNodePos);
   }
 
-  closeIfNonActiveStage(graph, stageIndex) {
+  closeIfNonActiveStage(stageIndex) {
     const {
       graphData: { stages },
     } = this.props;
     const stageData = stages[stageIndex];
+    const graph = this.graphs[stageIndex];
     if (stageData.jobStatus != "STARTED") {
       graph.afterDrawingOnce(() => this.toggleStage(stageIndex));
     }
   }
 
   drawGraphs() {
-    this.stageNames.forEach((_, i) => {
+    const {
+      graphData: { stages },
+    } = this.props;
+    stages.forEach((_, i) => {
       this.drawStageGraph(i);
     });
   }
@@ -628,11 +635,11 @@ class PipelineViz extends React.Component {
       edgeData,
       options
     );
+    this.graphs.push(currStageGraph);
+
     currStageGraph.minimizeSizeGivenScale(1.0);
     this.centerEndNodeVertically(currStageGraph);
-    this.closeIfNonActiveStage(currStageGraph, index);
-
-    this.graphs.push(currStageGraph);
+    this.closeIfNonActiveStage(index);
   }
 
   handleWindowResize = () => {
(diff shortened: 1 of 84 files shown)


jshoe:
could you add a blurb to the description about the tradeoff discussed yesterday between generating them all at once vs. being able to change later graphs dynamically?

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -645,32 +652,45 @@ class PipelineViz extends React.Component {
...
+      <div className={isOpened ? cs.openedStage : cs.hidden}>
+        <div className={cs.graphLabel}>
+          {stageName}
+          <RemoveIcon
+            onClick={() => this.toggleStage(i)}
+            className={cs.closeIcon}
+          />
+        </div>
+        <div
+          className={cx(cs.graph, this.state.hovered && cs.hovered)}
```
jshoe: nit: could you pull out the state variables at the beginning?

cdebourcy:
This is an invasive change to our core functionality (the pipeline), so the test plan needs to include comprehensive testing of new sample submissions (both success and failure cases) and phylo_tree creations. Should get reviews from @davidrissato and @boris-dimitrov as well

app/assets/src/components/views/PipelineViz/pipeline_viz.scss
```diff
@@ -33,6 +33,11 @@
   word-break: break-word;
 }
 
+.notToggableStageButton {
```
jshoe: I think you can move this under .stageButton and change it to something like `&.disabled`

jshoe:
I agree that this diff is a much bigger change than I was expecting.. I think there should be a more minimal change possible, or if it's not possible then I'd recommend leaving off the future viewable steps functionality for now..

app/models/pipeline_run_stage.rb
```diff
@@ -171,12 +188,11 @@ def prepare_dag(dag_name, attribute_dict, key_s3_params = nil)
                            attribute_dict,
                            pipeline_run.parse_dag_vars)
     self.dag_json = dag.render
-    copy_done_file = "echo done | aws s3 cp - #{sample.sample_output_s3_path}/\\$AWS_BATCH_JOB_ID.#{JOB_SUCCEEDED_FILE_SUFFIX}"
-    upload_dag_json_and_return_job_command(dag_json, dag_s3, dag_name, key_s3_params, copy_done_file)
+    save
```
jshoe: why was the `save` not needed before?

spec/models/pipeline_run_spec.rb
```diff
@@ -32,6 +32,15 @@
                                                   **mutates_model_attributes)
         end
 
+        it "contains all dag_jsons for pipeline run stages" do
+          pipeline_run.update_job_status
+
+          pipeline_run.pipeline_run_stages.each do |prs|
+            puts prs.dag_json
```
jshoe: nit: remove the puts

emilyzhong:
I've opened a separate PR (https://github.com/chanzuckerberg/idseq-web/pull/2429) for the backwards compatibility changes (making sure the pipeline works with stages without a dag_json) so that it doesn't get held up by the larger changes on the actual pipeline

tfrcarvalho:
> could you add a blurb to the description about the tradeoff discussed yesterday between generating them all at once vs. being able to change later graphs dynamically?

The configuration should be static. If there are optional pathways, that should be captured in the configuration and in the visualization. When analyzing a pipeline we should be able to look at the configuration/visualization and understand the logic of what was run.
The results of the pipeline however, should show if each step was run and if not, why.

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -38,7 +38,10 @@ class PipelineViz extends React.Component {
     this.panZoomContainer = React.createRef();
 
     this.state = {
-      stagesOpened: [true, true, true, true],
+      // Set stages that can be rendered to open, and others to closed.
+      stagesOpened: this.stageNames.map(
+        (_, i) => i < this.props.graphData.stages.length
```
tfrcarvalho: Maybe use `!!` to convert the value to a boolean?

emilyzhong:
This PR should now only contain code that affects the pipeline run directly (moving the creation of `dag_json`s to the beginning of the pipeline run instead of pipeline run stage)

app/helpers/pipeline_runs_helper.rb
```diff
@@ -109,11 +109,11 @@ def fetch_pipeline_version(s3_file = pipeline_version_file)
     Regexp.last_match(1)
   end
 
-  def upload_dag_json_and_return_job_command(dag_json, dag_s3, dag_name, key_s3_params = nil, copy_done_file = "")
```
emilyzhong: I split this method into two: `upload_dag_json` and `dag_job_commands`, since there doesn't seem to be a reason why they are combined in one method (other than sharing some variables)

app/models/phylo_tree.rb
```diff
@@ -253,7 +253,8 @@ def prepare_dag(dag_name, attribute_dict)
                            attribute_dict,
                            parse_dag_vars)
     self.dag_json = dag.render
-    upload_dag_json_and_return_job_command(dag_json, dag_s3, dag_name)
```
emilyzhong: Since this method is split into two (see comment in changes for `pipeline_run_helpers`), I just call the two methods in the same order they initially appeared in the original function.

app/models/pipeline_run_stage.rb
```diff
@@ -171,12 +188,11 @@ def prepare_dag(dag_name, attribute_dict, key_s3_params = nil)
...
+    upload_dag_json(dag_json, dag_s3)
   end
 
-  def host_filtering_command
```
emilyzhong: These `[stage name]_command` are called when the pipeline stage is kicked off (https://github.com/chanzuckerberg/idseq-web/blob/master/app/models/pipeline_run_stage.rb#L69), creating both the attributes for the dag json and job command. I split each of these `[stage name]_command` into two methods: `[stage name]_dag_attributes` and `[stage_name]_command`. The `_dag_attributes` methods creates the fields needed to create the dag, and `_command` methods creates the fields needed to run the dag.

app/models/pipeline_run_stage.rb
```diff
@@ -160,7 +162,22 @@ def log_url
   end
 
   ########### STAGE SPECIFIC FUNCTIONS BELOW ############
-  def prepare_dag(dag_name, attribute_dict, key_s3_params = nil)
+  def create_dag
```
emilyzhong: This just decides which attributes to create the stage dag json from

emilyzhong:
I added a bunch of comments on the diffs to explain the changes + hopefully make reviewing easier!

jshoe:
It looks like some Rspec tests are failing btw

spec/models/pipeline_run_spec.rb
```diff
@@ -32,6 +32,14 @@
                                                   **mutates_model_attributes)
         end
 
+        it "contains all dag_jsons for pipeline run stages" do
+          pipeline_run.update_job_status
+
+          pipeline_run.pipeline_run_stages.each do |prs|
+            expect(prs.dag_json).to be_truthy
```
davidrissato: Ideally we should check the actual value, but since it might be too hard, I would say we should at least test if this is a valid json.

app/models/pipeline_run_stage.rb
```diff
@@ -26,6 +26,8 @@ class PipelineRunStage < ApplicationRecord
   # Max number of times we resubmit a job when it gets killed by EC2.
   MAX_RETRIES = 5
 
+  after_create :create_dag
```
davidrissato: Our system currently abuses using events, which I consider very harmful, because everything has a "side effect".

For example, I noticed method `create_dag` has a `save` method, which means the record will be saved twice in a row. Not exactly harmful in this specific situation, but just to show what kind of side effects we are talking about. For example: if someone decides adding a `after_save` later on, this person will have a hard time to trying to understand why it is being triggered twice.

I would rather make `create_dag` method to be stateless (i.e. receiving all parameters that it needs to create), and probably rename it to something like `generate_dag_json` (or even better, extract it to a helper class), and then invoke it explicitly in this block:
https://github.com/chanzuckerberg/idseq-web/blob/ca0b34ab1e68ae8c1d3455860c0b2408e531148e/app/models/pipeline_run.rb#L317-L349

app/models/pipeline_run_stage.rb
davidrissato: `save` here is not necessarily a good idea, see my notes above

app/models/pipeline_run_stage.rb
```diff
@@ -171,12 +188,11 @@ def prepare_dag(dag_name, attribute_dict, key_s3_params = nil)
                            attribute_dict,
                            pipeline_run.parse_dag_vars)
     self.dag_json = dag.render
-    copy_done_file = "echo done | aws s3 cp - #{sample.sample_output_s3_path}/\\$AWS_BATCH_JOB_ID.#{JOB_SUCCEEDED_FILE_SUFFIX}"
-    upload_dag_json_and_return_job_command(dag_json, dag_s3, dag_name, key_s3_params, copy_done_file)
+    save
+    upload_dag_json(dag_json, dag_s3)
```
davidrissato: I also notice that we are uploading the dag json at this point (this is part of the side effect). 
In the previous logic, that file would be uploaded immediately before running it (in this block https://github.com/chanzuckerberg/idseq-web/blob/ca0b34ab1e68ae8c1d3455860c0b2408e531148e/app/models/pipeline_run_stage.rb#L69, which is invoking by reflection one of these methods: https://github.com/chanzuckerberg/idseq-web/blob/ca0b34ab1e68ae8c1d3455860c0b2408e531148e/app/models/pipeline_run_stage.rb#L69)

I think it is a good idea preserving the behavior of uploading this file right before executing it. The existing logic can be modified to upload the contents of this `dag_json` field instead of calculating it (as it currently happens)

app/models/pipeline_run_stage.rb
emilyzhong: Updated with `dag_json_helper`

app/models/pipeline_run.rb
```diff
@@ -345,6 +346,11 @@ def create_run_stages
       job_command_func: 'experimental_command'
     )
 
+    dag_names = [DAG_NAME_HOST_FILTER, DAG_NAME_ALIGNMENT, DAG_NAME_POST_PROCESS, DAG_NAME_EXPERIMENTAL]
+    run_stages.each_with_index do |stage, i|
+      stage.dag_json = generate_dag_json(self, dag_names[i])
```
davidrissato: Positional loops iterating two arrays simultaneously can be dangerous if someone change the order in any of these arrays (like inserting an item in only one of them).

You could use a dictionary, but I believe it is easier to actually invoke them directly in the commands above, like:

```
    run_stages << PipelineRunStage.new(
      step_number: 1,
      name: PipelineRunStage::HOST_FILTERING_STAGE_NAME,
      job_command_func: 'host_filtering_command',
      dag_json: generate_dag_json(self, DagJsonHelper::DAG_NAME_HOST_FILTER)
    )
```

davidrissato: Actually, if you make the change this way, you don't actually need this `generate_dag_json` method. You could directly invoke the methods with the correct names (like `DagJsonHelper#host_filtering_dag_attributes`)

app/helpers/dag_json_helper.rb
```diff
@@ -0,0 +1,119 @@
+module DagJsonHelper
+  DAG_NAME_HOST_FILTER = "host_filter".freeze
+  DAG_NAME_ALIGNMENT = "non_host_alignment".freeze
+  DAG_NAME_POST_PROCESS = "postprocess".freeze
+  DAG_NAME_EXPERIMENTAL = "experimental".freeze
+
+  def generate_dag_json(pipeline_run, dag_name)
```
davidrissato: Consider removing this method and direct invoking `host_filtering_dag_attributes`, `alignment_dag_attributes`... directly.

You can create a private method in this module that contains this code:
```
def _render_dag_json(attribute_dict):
    sample = pipeline_run.sample
    dag = DagGenerator.new("app/lib/dags/#{dag_name}.json.erb",
                           sample.project_id,
                           sample.id,
                           sample.host_genome_name.downcase,
                           attribute_dict,
                           pipeline_run.parse_dag_vars)
    dag.render
```

app/helpers/dag_json_helper.rb
```diff
@@ -0,0 +1,119 @@
...
+    when DAG_NAME_ALIGNMENT
+      attribute_dict = alignment_dag_attributes(pipeline_run)
+    when DAG_NAME_POST_PROCESS
+      attribute_dict = postprocess_dag_attributes(pipeline_run)
+    when DAG_NAME_EXPERIMENTAL
+      attribute_dict = experimental_dag_attributes(pipeline_run)
+    end
+
+    sample = pipeline_run.sample
+    attribute_dict[:bucket] = SAMPLES_BUCKET_NAME
```
davidrissato: Not sure where this constant  `SAMPLES_BUCKET_NAME` is coming from.

I would move this code to each method below, to avoid mutating the attribute dict here.

spec/factories/samples.rb
```diff
@@ -46,6 +46,8 @@
         create(:metadatum, key: key, raw_value: value, sample: sample, metadata_field: MetadataField.find_by(name: key))
       end
 
+      # Pipline runs create pipeline_run_stages automatically, which require a human host genome record
+      HostGenome.find_by(name: "Human") || create(:host_genome, name: "Human")
```
tfrcarvalho: THis should be in the pipeline run factory, no?

emilyzhong: Oops — didn't realize that FactoryBot has a before create hook. Will update!

app/models/pipeline_run_stage.rb
```diff
@@ -160,129 +161,41 @@ def log_url
...
-                           sample.id,
-                           sample.host_genome_name.downcase,
-                           attribute_dict,
-                           pipeline_run.parse_dag_vars)
-    self.dag_json = dag.render
+    upload_dag_json(dag_json, dag_s3)
+
     copy_done_file = "echo done | aws s3 cp - #{sample.sample_output_s3_path}/\\$AWS_BATCH_JOB_ID.#{JOB_SUCCEEDED_FILE_SUFFIX}"
-    upload_dag_json_and_return_job_command(dag_json, dag_s3, dag_name, key_s3_params, copy_done_file)
+    dag_job_commands(dag_s3, dag_name, key_s3_params, copy_done_file)
```
davidrissato: I suggest splitting this method in two:
    `upload_dag_json` and `dag_job_commands`, so they can be more intuitive. You can invoke both here in sequence.

app/helpers/dag_json_helper.rb
```diff
@@ -0,0 +1,115 @@
+module DagJsonHelper
+  DAG_NAME_HOST_FILTER = "host_filter".freeze
```
davidrissato: NIT: If you add this magic comment as very first line of this file:
```
# frozen_string_literal: true
```
You can actually remove all these `.freeze` method calls.

That comment basically says: "-every literal string declared in this file is 'frozen' by default"

emilyzhong:
On hold as dag json creation requires pipeline version, which currently isn't available until the first stage is completed.

jshoe:
Closing due to inactivity/blockers

PR #2428, opened 2019-07-23
Filter out test data from analytics

gregdingle:
# Description

It looks like robot data was inserted into our active users by one of our tests because the user names were "Admin User" and the traffic only came from ruby. This protects against that happening in future while preserving the ability to test analytics manually in dev environment. 

# Tests

* rails t

### DIFF
--- a/app/lib/metric_util.rb
+++ b/app/lib/metric_util.rb
@@ -151,9 +151,9 @@ class MetricUtil
       end
     end
 
-    def a_test?
-      Rails.env == "test" ||
-        (request && request.user_agent == "Rails Testing")
+    def a_test?(request = nil)
+      Rails.env.test? ||
+        (!request.nil? && request.user_agent == "Rails Testing")
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/lib/metric_util.rb
```diff
@@ -150,5 +150,10 @@ def https_post(uri, data)
         end
       end
     end
+
+    def a_test?(request = nil)
```
jshoe: nit: maybe `is_test?` or `is_a_test?` for a boolean function? similar to `is_a?` https://apidock.com/ruby/Object/is_a%3F

app/lib/metric_util_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class HeatmapHelperTest < ActiveSupport::TestCase
```
jshoe: I think you forgot to change the name HeatmapHelperTest, but also we should use Rspec instead and have it in the /spec folder

jshoe: or at least the /test directory with Minitest

app/lib/metric_util.rb
gregdingle: I originally wrote `is_a_test?` but rubocop complained.

app/lib/metric_util_test.rb
gregdingle: 👍

PR #2429, opened 2019-07-23
[PipelineViz] Works without dag json in some stages

emilyzhong:
# Description
Fixes frontend so that older erroring pipeline versions without this change will still be able to render a pipeline visualization

# Notes
This is half of https://github.com/chanzuckerberg/idseq-web/pull/2427 — I'm planning on submitting the overall change as two, this one (backwards compatibility) and the rest (changing where the dag_json is generated for later versions)

# Tests

* Added a test for the service for stages with no dag_json


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20_190_717_221_024) do
+ActiveRecord::Schema.define(version: 20_190_719_221_024) do
```
tfrcarvalho: You remove this file from the PR.

spec/services/retrieve_pipeline_viz_graph_data_service_spec.rb
```diff
@@ -109,9 +109,9 @@
     before do
       # Project and sample required to create pipeline run.
       project = create(:public_project)
-      sample = create(:sample, project: project)
+      @sample = create(:sample, project: project)
```
tfrcarvalho: Why do you need the `@` ?

app/assets/src/components/views/report/utils/download.js
```diff
@@ -65,7 +65,8 @@ const getDownloadLinkInfoMap = (sampleId, pipelineRun) => ({
     newPage: true,
   },
   [PIPELINE_VIZ_LABEL]: {
```
tfrcarvalho: Shouldn't we remove the entry if there is nothing to download?
Link won't be disabled right?

spec/services/retrieve_pipeline_viz_graph_data_service_spec.rb
emilyzhong: In order to create a new pipeline run with modified stage data in the new test, I need to provide a sample as the argument; the @ makes the sample available for the entire instance so that I can use it in the new test below

PR #2432, opened 2019-07-24
surface informative error if too many pairs are broken

cdebourcy:
Goes with https://github.com/chanzuckerberg/idseq-dag/pull/171.

# Tests
- Run a samples whose read pairs don't match up and see the pipeline fail gracefully with an informative error.
- Run a normal test sample and see it succeed.

image

### DIFF
--- a/app/assets/src/components/views/SampleView/SampleView.jsx
+++ b/app/assets/src/components/views/SampleView/SampleView.jsx
@@ -335,8 +335,7 @@ class SampleView extends React.Component {
         status = "COMPLETE - ISSUE";
         message =
           "Sorry, something was wrong with your input files. " +
-          "Either the paired reads did not follow the same order in both files, " +
-          "or they were not named using the same identifiers in both files, " +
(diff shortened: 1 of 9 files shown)


app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -331,6 +331,16 @@ class SampleView extends React.Component {
         issueType = "warning";
         link = `/samples/${sample.id}/results_folder`;
         break;
+      case "BROKEN_PAIRS":
+        status = "COMPLETE - ISSUE";
+        message =
+          "Sorry, something was wrong with your input files. Either paired reads did not follow the same order in both files, " +
+          "or they followed an unconventional naming scheme.";
+        linkText =
+          "Please ensure mates have the same identifier and follow the same order in both input files, then reupload";
```
gregdingle: change "mates" to "paired reads" for consistency

cdebourcy: ok done

PR #2434, opened 2019-07-25
Improve basespace uploading logic.

MarkAZhang:
# Description

Different approach that uses curl and streaming to remove the need for storing the file in disk on the resque worker.

# Notes

Also added a new function in Syscall that wraps around the shell commands to make testing easier. Renamed the existing `pipe` method in `Syscall` to `pipe_with_output` to distinguish it from the new pipe.

# Tests
* Verify that Basespace uploading still works.
* Verify that tests pass.
* On staging, will verify that the samples still run through the entire pipeline correctly (and all results are loaded).

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -325,7 +325,7 @@ class ProjectsController < ApplicationController
   end
 
   def project_reports_csv_status
-    stdout = Syscall.pipe(["aws", "s3", "ls", @project.report_tar_s3(current_user.id)], ["wc", "-l"])
+    stdout = Syscall.pipe_with_output(["aws", "s3", "ls", @project.report_tar_s3(current_user.id)], ["wc", "-l"])
     return if stdout.blank?
     final_complete = stdout.to_i == 1
     if final_complete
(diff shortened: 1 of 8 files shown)


jshoe:
looks correct to me 👍

cdebourcy:
> Does it do the same input truncation that the S3 to S3 upload does?

https://github.com/chanzuckerberg/idseq-web/blob/4374e56f41ec13d92cc6a58e6193267a5c3b92c4/app/models/sample.rb#L287

MarkAZhang:
@cdebourcy It does not do the truncation. I think we discussed at some point that since the validate_input dag step also does the truncation, doing it here seems redundant. Unzipping, truncating, and then re-zipping can take a significant amount of time. It does mean uploading a smaller file, but overall it might actually take more time. Let me know if you think the truncation is worth it.

tfrcarvalho:
> Let me know if you think the truncation is worth it.

Is there any limit in file size? We should not be exposed to files that are extremely large. Even if we accept basespace limit we are letting another company do that restriction for us.

davidrissato:
I don't think truncation should happen here. If we do not agree with the file size, we should reject when the user is selecting it.

MarkAZhang:
Ready to be looked at. I will do the file size restriction in a separate PR. The restriction code is unrelated to the upload code in this PR.

PR #2437, opened 2019-07-26
[Pipeline Viz] Use output target names for step instead of class name

emilyzhong:
# Description
Uses titleized output target name for the title of each step instead of using the class name.
 
# Notes
Multiple steps can use the same python class, while targets are unique per stage, which will be important when passing step-specific information from the backend. 

# Tests
* Modified rspec tests for backend
* Go to /samples/sample_id/pipeline_viz on a pipeline_viz enabled account to see new names

### DIFF
--- a/app/services/retrieve_pipeline_viz_graph_data_service.rb
+++ b/app/services/retrieve_pipeline_viz_graph_data_service.rb
@@ -10,11 +10,12 @@ class RetrievePipelineVizGraphDataService
   #     - jobStatus: The job status of the stage
   #     - steps: Array of objects, each one defining a node in that stage's graph. This object
   #       is composed of:
-  #           - stepName: The name to be displayed for this step
-  #               - inputEdges: An array of indices that map to edges in the @edges array (see below), each
-  #                 edge being an input edge to the node.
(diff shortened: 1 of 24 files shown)


app/services/retrieve_pipeline_viz_graph_data_service.rb
```diff
@@ -10,11 +10,12 @@ class RetrievePipelineVizGraphDataService
   #     - jobStatus: The job status of the stage
   #     - steps: Array of objects, each one defining a node in that stage's graph. This object
   #       is composed of:
-  #           - stepName: The name to be displayed for this step
-  #               - inputEdges: An array of indices that map to edges in the @edges array (see below), each
-  #                 edge being an input edge to the node.
-  #               - ouputEdges: An array of indices that map to edges in the @edges array, each edge being an
-  #                 output edge from the node
+  #           - name: The name to be displayed for this step. This is the out target name.
```
tfrcarvalho: To make it more clear: `This is the out target name` -> `Currently, this corresponds to the name of the output target (works as there is a single output target per step).`

PR #2438, opened 2019-07-26
Improve error message if user forgets to select project.

MarkAZhang:
# Description

Before, the error message is ambiguous, especially for Basespace upload (since the user also has to pick a Basespace project)

Screen Shot 2019-07-25 at 7 41 57 PM

Screen Shot 2019-07-25 at 7 42 08 PM

With this change, the error message is more detailed, and the erred Dropdown is also highlighted in red.

Screen Shot 2019-07-25 at 7 29 56 PM

Also changed the error messages in S3 upload tab to use <Notification> to make it more consistent.

Screen Shot 2019-07-25 at 7 26 50 PM

Screen Shot 2019-07-25 at 7 27 02 PM

# Notes

Also moved the placeholder prop from Dropdown to DropdownTrigger, so that it's easier to style them both at once during the erred state.

# Tests

* Verified that all error states behave as expected (see screenshots)
* Verified that dropdowns on other pages, such as the Sample Upload page, are unaffected. (since we made a change to Dropdown)

### DIFF
--- a/app/assets/src/components/views/SampleUploadFlow/BasespaceSampleImport.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/BasespaceSampleImport.jsx
@@ -13,6 +13,7 @@ import {
 import { withAnalytics } from "~/api/analytics";
 
 import { openBasespaceOAuthPopup } from "./utils";
+import { NO_TARGET_PROJECT_ERROR } from "./constants";
 import cs from "./basespace_sample_import.scss";
 
 export default class BasespaceSampleImport extends React.Component {
@@ -29,6 +30,17 @@ export default class BasespaceSampleImport extends React.Component {
     window.addEventListener("message", this.handleMessageEvent);
     if (accessToken) {
       this.fetchBasespaceProjects(accessToken);
+    }
+  }
+
+  componentDidUpdate() {
+    if (
+      this.state.error === NO_TARGET_PROJECT_ERROR &&
+      this.props.project !== null
+    ) {
+      this.setState({
+        error: "",
+      });
     }
   }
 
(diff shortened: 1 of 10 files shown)


app/assets/src/components/views/SampleUploadFlow/constants.js
```diff
@@ -0,0 +1,2 @@
+export const NO_PROJECT_ERROR =
+  "Please select a project to upload your samples to.";
```
tfrcarvalho: I would still make it more clear: `a project` -> `an IDseq project`

app/assets/src/components/views/SampleUploadFlow/BasespaceSampleImport.jsx
```diff
@@ -13,6 +13,7 @@ import {
 import { withAnalytics } from "~/api/analytics";
 
 import { openBasespaceOAuthPopup } from "./utils";
+import { NO_PROJECT_ERROR } from "./constants";
```
tfrcarvalho: This can also be confusing for future developers. I would do something like `NO_TARGET_PROJECT_ERROR`.

happyimadesignr:
Hey @MarkAZhang, The error looks unclear, b/c it looks like it's associated with the bottom input even when you highlight the top one in read. We should move toward a more contextual errors for some of these specific cases.

This is contextual error for the new phylotree flow:
Screen Shot 2019-07-29 at 9 55 24 AM

MarkAZhang:
Hey @happyimadesignr! I totally get what you mean. The reason why I had the error be associated with both the top and the bottom input is because this error happens when the user clicks on the button input. If you just have the error appear on the top input, this can also be confusing to the user because they clicked on the bottom input (and aren't expecting this to cause an error elsewhere on the page). They might not even notice the error appear on the top input because they are focused on the bottom input.

I can add an additional message underneath the top input if you think that'd be helpful. But I don't think we should remove the error from the bottom, because it's important to have something change in close proximity to where the user clicked.

PR #2440, opened 2019-07-26
Add names to dag_json templates

emilyzhong:
# Description
Adds names to dag_json (see https://github.com/chanzuckerberg/idseq-dag/pull/172) needed to ensure that pipeline step-level updates are stored in unique files (don't override each other)

**Needs to be merged in before approving corresponding PR in idseq-dag**

### DIFF
--- a/app/lib/dags/host_filter.json.erb
+++ b/app/lib/dags/host_filter.json.erb
@@ -1,4 +1,5 @@
 {
+  "name": "host_filtering",
   "output_dir_s3":
     "s3://<%= @attribute_dict[:bucket] %>/samples/<%= @project_id %>/<%= @sample_id %>/results",
   "targets": {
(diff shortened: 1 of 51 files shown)


app/lib/dags/host_filter.json.erb
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "host_filtering",
```
tfrcarvalho: We should use the same names that we defined in constants as dag stage names, e.g. DAG_NAME_HOST_FILTER. Can we passed them to the template?

emilyzhong: Sure — since they're created in https://github.com/chanzuckerberg/idseq-web/pull/2427, I'll wait for that to merge in first

emilyzhong: Added in constants!

PR #2441, opened 2019-07-27
Alert on uploads that take more than 3 hours

gregdingle:
# Description

This will help us react to user uploads that stall because their connection fails or other unknown causes. The check is done in results monitor every 5 seconds.

# Notes

Please check the logic of the check. I'm not 100% sure that it is what we want. For example, are there valid cases where a user will create a sample, then wait to begin uploading? 

Also, I'm not sure how reliable `client_updated_at` because I only saw around 300 of them in samples in prod database. 

# Tests

Run locally from CLI
`rake alert_stalled_uploads`

See output 
```
LongRunningUploadsEvent: Samples [11734, 11815, 12313, 12314, 12315, 12316,
12340, 12341, 12353, 12436, 12440, 12461, 12462, 12526, 12527, 12528, 12582,
12805, 12806, 12828, 12841, 12907] by ["admin user", "non-admin user"] were
created 5497.1 hours ago. See: ["https://idseq.net/samples/11734/pipeline_runs",
"https://idseq.net/samples/11815/pipeline_runs",
"https://idseq.net/samples/12313/pipeline_runs",
"https://idseq.net/samples/12314/pipeline_runs",
"https://idseq.net/samples/12315/pipeline_runs",
"https://idseq.net/samples/12316/pipeline_runs",
"https://idseq.net/samples/12340/pipeline_runs",
"https://idseq.net/samples/12341/pipeline_runs",
"https://idseq.net/samples/12353/pipeline_runs",
"https://idseq.net/samples/12436/pipeline_runs",
"https://idseq.net/samples/12440/pipeline_runs",
"https://idseq.net/samples/12461/pipeline_runs",
"https://idseq.net/samples/12462/pipeline_runs",
"https://idseq.net/samples/12526/pipeline_runs",
"https://idseq.net/samples/12527/pipeline_runs",
"https://idseq.net/samples/12528/pipeline_runs",
"https://idseq.net/samples/12582/pipeline_runs",
"https://idseq.net/samples/12805/pipeline_runs",
"https://idseq.net/samples/12806/pipeline_runs",
"https://idseq.net/samples/12828/pipeline_runs",
"https://idseq.net/samples/12841/pipeline_runs",
"https://idseq.net/samples/12907/pipeline_runs"]
```


app/models/sample.rb
```diff
@@ -848,4 +853,13 @@ def metadata_fields_info
   def first_pipeline_run
     pipeline_runs.order(created_at: :desc).first
   end
+
+  def status_url
+    base_url = if Rails.env == 'staging'
```
tfrcarvalho: What happens in dev?

app/jobs/check_upload_status_after_delay.rb
```diff
@@ -0,0 +1,20 @@
...
+    Rails.logger.info("Checking upload status of sample #{sample.id}")
+
+    if sample.status == 'created' &&
+       (Time.current - created_at).minutes > minutes
+      duration_hrs = ((Time.current - sample.created_at) / 60 / 60).round(2)
+      input_files = sample.input_files ? sample.input_files.length : 0
+      msg = "LongRunningUploadEvent: Sample #{sample.id} by #{sample.user.role_name} " \
+        "was created #{duration_hrs} hours ago. " \
+        "#{input_files} input files have been uploaded. " \
+        "Last client ping was at #{sample.ccccccjttcgulcihtkiitkjlllfbncdvuuutujnjtdlc}. See: #{status_url}"
```
tfrcarvalho: I am assuming you touched your yubikey?

app/models/sample.rb
tfrcarvalho: I know this was copied from pipeline run, but this still seems like the wrong place to define the url. This can probably be a constant defined in a shareable place. Can this be obtained from routes?

tfrcarvalho: If you still have access to the request when triggering the job, you can send it to the resque job from the original request using something like `request.base_url`...

MarkAZhang:
Seems like this overlaps with this issue: https://jira.czi.team/secure/RapidBoard.jspa?rapidView=12&view=planning&selectedIssue=IDSEQ-1191&quickFilter=41&issueLimit=100

`client_updated_at` is updated whenever the user's browser calls a "upload heartbeat" endpoint for local sample uploads. This was lost during the new sample upload, and we need to add it back. Maybe it makes sense to look at `client_updated_at` instead of `created_at`, once we add it back?

Also, should we set the sample to "Failed" after a certain point? Down the line, we can even give the user an option to re-upload their files.

jshoe:
^ Actually the client_updated_at was added in April but it seems to have been broken between then and now :|

gregdingle:
@cdebourcy:

> We mustn’t overload the resque tier by adding 1 job for every single sample even though most of them are going to be no-ops. 

Instead, you should make it a demon that just lists Sample.where(status: Sample::STATUS_CREATED).where("created_at < ?", Time.now - 3.hours) periodically. That will be much more efficient. 

> You can piggyback on the pipeline_monitor code (https://github.com/chanzuckerberg/idseq-web/blob/master/lib/tasks/pipeline_monitor.rake),

Will do. I'll add a comment following @tfcarvalho's point. 

> Time.current will not be consistent with the created_at column here. You should use Time.now.

Good catch! But what about all the other instances of Time.current, as in PipelineRun: 

```
def run_time
  Time.current - created_at
end
```

app/jobs/check_upload_status_after_delay.rb
gregdingle: oops 😊

app/models/sample.rb
gregdingle: > What happens in dev? 

The link will be to prod which may be inaccurate. I could update it to point to localhost. 

> Can this be obtained from routes?

Probably. I would just need to research it.

gregdingle:
@markzhang @jsheu

> Maybe it makes sense to look at `client_updated_at` instead of created_at, once we add it back?

Thanks for the extra context. However, I don't think we should use `client_updated_at` for the alert because 1) the client may not update at all, 2) the CLI does not set `client_updated_at`

Also, should we set the sample to "Failed" after a certain point? Down the line, we can even give the user an option to re-upload their files.

> Actually the client_updated_at was added in April but it seems to have been broken between then and now :|

Is this tracked anywhere? It seems like it would be worth fixing.

cdebourcy:
> run_time

Thanks for catching that! Looks like `runtime` is only used in admin-facing pages, not user-facing ones (phew), but we should fix it! Thanks!

gregdingle:
@cdebourcy I piggybacked on the result monitor code instead of pipeline monitor because the code was simpler and it has a min `sleep_quantum` of 5 seconds, which seems appropriate.

gregdingle:
See new PR https://github.com/chanzuckerberg/idseq-web/pull/2450

cdebourcy:
> @cdebourcy I piggybacked on the result monitor code instead of pipeline monitor because the code was simpler and it has a min `sleep_quantum` of 5 seconds, which seems appropriate.

Thanks for making the change! One more question:
Is it going to keep sending alerts for the same sample every 5 seconds? Should we bundle this change with the "upload failed" status update, so that we can stop sending more alerts after the status is changed to "upload failed"? Of course, in that case the time-out needs to be longer than 3 hours (some users do report letting uploads run overnight)...

If you did want to do this alert in the pipeline_monitor -- which seems conceptually more logical to me, but opinions can differ -- I believe this would be the place to put it: https://github.com/chanzuckerberg/idseq-web/blob/220ac23c8863b09af1a2985646df1e6a88e97c5e/lib/tasks/pipeline_monitor.rake#L306

cdebourcy:
Oh I see this PR is superseded by a new one.

gregdingle:
Rubocop says:

> lib/tasks/result_monitor.rake:56:8: C: Avoid using update_all because it skips validations.

😂

PR #2443, opened 2019-07-30
[Maps] Fix geosearch box and Location V2 switch in development environments

jshoe:
### Description
(1) Fixes the geosearch box in development environments without the location provider API keys. Before it would fail and appear to spin forever:
Jul-29-2019 17-06-20

(2) Swaps collection_location => collection_location_v2 in the migration and makes the field required. For devs b/c I already toggled these flags in the deployed environments.

### Tests
See pic + verified database state locally.

### DIFF
--- a/app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
+++ b/app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
@@ -53,6 +53,7 @@ class GeoSearchInputBox extends React.Component {
       console.log(e);
       logAnalyticsEvent("GeoSearchInputBox_request_erred", {
         query,
+        message: e.message,
       });
     }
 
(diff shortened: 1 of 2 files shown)


app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
```diff
@@ -28,23 +28,31 @@ class GeoSearchInputBox extends React.Component {
...
+            r.description = nameParts.slice(1).join(", ");
+            r.key = `loc-${r.locationiq_id}`;
+            return r;
+          }),
+        };
+      }
+    } catch (e) {
+      // In the case of an error (e.g. no API key in dev), just catch and show the plain text option.
+      // eslint-disable-next-line no-console
+      console.log(e);
```
tfrcarvalho: Can you log this particular case somehow? 
It would be important if we are alerted for this and verify the entry... Maybe for now just analytics and include the free text value...

jshoe: @tfrcarvalho added 👍

PR #2445, opened 2019-07-30
[BenchDash] Scaffold code: Benchmark Controller and action to return benchmark list as JSON

tfrcarvalho:
# Description

* This PR introduces backend scaffold for retrieving benchmarks
* Defines an index action to return list of current benchmarks for the environment

# Tests

* Unit and integration tests
* To retrieve the JSON: `/benchmarks.json`

### DIFF
--- a/app/helpers/benchmarks_helper.rb
+++ b/app/helpers/benchmarks_helper.rb
@@ -1,0 +1,2 @@
+module BenchmarksHelper
+end
--- a/app/helpers/benchmarks_helper.rb
+++ b/app/helpers/benchmarks_helper.rb
@@ -1,2 +1,62 @@
 module BenchmarksHelper
+
+  IDSEQ_BENCH_BUCKET = "idseq-bench".freeze
+  IDSEQ_BENCH_KEY = "config.json".freeze
+
+  def get_benchmarks_list
+    benchmarks = {}
+
+    # Read benchmarks from configuration store in S3
+    benchmark_config = get_config()
+    if benchmark_config
+      benchmarks = parse_config(benchmark_config)
+
+      benchmarks[:active_benchmarks].each do |benchmark|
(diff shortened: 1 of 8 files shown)


app/helpers/benchmarks_helper.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    return benchmarks
+  end
+
+  def benchmarks_config
+    s3_response = S3_CLIENT.get_object(bucket: IDSEQ_BENCH_BUCKET, key: IDSEQ_BENCH_KEY)
+    return JSON.parse(s3_response.body.read, symbolize_names: true)
+  rescue Aws::S3::Errors::NoSuchKey
+    LogUtil.log_err_and_airbrake("[BenchmarkController] Config file not found - '#{IDSEQ_BENCH_KEY}'")
```
jshoe: nit: actually BenchmarksController

tfrcarvalho: This was actually in the helper now! Removed the prefix.

app/helpers/benchmarks_helper.rb
```diff
@@ -0,0 +1,63 @@
+module BenchmarksHelper
+  IDSEQ_BENCH_BUCKET = "idseq-bench".freeze
+  IDSEQ_BENCH_KEY = "config.json".freeze
+
+  def benchmarks_list
+    benchmarks = {}
+
+    # Read benchmarks from configuration store in S3
+    benchmark_config = benchmarks_config()
```
jshoe: this was a little confusing to read and I read it as `benchmark_config = benchmark_config` at first

tfrcarvalho: oh yeah... rubocop forced me to change the name and end up being the same. Renamed.

PR #2449, opened 2019-07-30
Add upload heartbeat to bulk upload

gregdingle:
# Description

This adds the existing heartbeat feature to the new bulk upload code path. See https://jira.czi.team/browse/IDSEQ-1237 . 

# Notes

* I also changed the heartbeat function to send a first heartbeat immediately for instant feedback. 

# Tests

* Open http://localhost:3000/samples/upload 
* Upload some files
* See heartbeat request in debug console

### DIFF
--- a/app/assets/src/api/upload.js
+++ b/app/assets/src/api/upload.js
@@ -84,6 +84,9 @@ export const bulkUploadLocalWithMetadata = ({
         const sampleName = sample.name;
         const files = sampleNamesToFiles[sampleName];
 
+        // Start pinging server to monitor uploads server-side
+        startUploadHeartbeat(sample.id);
+
         sample.input_files.map(inputFileAttributes => {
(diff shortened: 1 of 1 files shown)


MarkAZhang:
If the user is uploading 20 samples, this will send 20 separate requests. Would it be worth it to consolidate all of these into one request?

You might also stop sending the heartbeat for a sample once it's done uploading.

gregdingle:
> You might also stop sending the heartbeat for a sample once it's done uploading.

Good catch! That's what I always intended. 😊 

> If the user is uploading 20 samples, this will send 20 separate requests.

That doesn't seem like a lot in the context of all the upload requests that are happening concurrently (they are not queued are they?) And I'm not sure how to make `startUploadHeartbeat` work only for the batch of samples that have `createSample` succeed.

PR #2450, opened 2019-07-30
Alert then fail on uploads that take too long

gregdingle:
# Description

This supersedes https://github.com/chanzuckerberg/idseq-web/pull/2441/files by combining alerting and failing. The check is done in results monitor every 5 seconds.

All samples that were created more than 3 hours ago that have not finished uploading will be put in  an error message for us to see. Their `upload_error` field will be set to `LOCAL_UPLOAD_STALLED`.

All samples that were created more than 18 hours ago that have not finished uploading will have their `upload_error` set to `LOCAL_UPLOAD_FAILED`. 

The `LOCAL_UPLOAD_FAILED` will cause an error message to be shown on the sample page. 

image

# Notes

* The thresholds of 3 and 18 hours could need some tweaking. Input welcome. 

* I fixed `client_updated_at` in https://github.com/chanzuckerberg/idseq-web/pull/2449 . We could try to make the check smarter by incorporating that, though it still wouldn't cover the CLI. 

# Tests

Run locally from CLI
`rake alert_stalled_uploads`

See output 
```
LongRunningUploadsEvent: Samples [11734, 11815, 12313, 12314, 12315, 12316,
12340, 12341, 12353, 12436, 12440, 12461, 12462, 12526, 12527, 12528, 12582,
12805, 12806, 12828, 12841, 12907] by ["admin user", "non-admin user"] were
created 5497.1 hours ago. See: ["https://idseq.net/samples/11734/pipeline_runs",
"https://idseq.net/samples/11815/pipeline_runs",
"https://idseq.net/samples/12313/pipeline_runs",
"https://idseq.net/samples/12314/pipeline_runs",
"https://idseq.net/samples/12315/pipeline_runs",
"https://idseq.net/samples/12316/pipeline_runs",
"https://idseq.net/samples/12340/pipeline_runs",
"https://idseq.net/samples/12341/pipeline_runs",
"https://idseq.net/samples/12353/pipeline_runs",
"https://idseq.net/samples/12436/pipeline_runs",
"https://idseq.net/samples/12440/pipeline_runs",
"https://idseq.net/samples/12461/pipeline_runs",
"https://idseq.net/samples/12462/pipeline_runs",
"https://idseq.net/samples/12526/pipeline_runs",
"https://idseq.net/samples/12527/pipeline_runs",
"https://idseq.net/samples/12528/pipeline_runs",
"https://idseq.net/samples/12582/pipeline_runs",
"https://idseq.net/samples/12805/pipeline_runs",
"https://idseq.net/samples/12806/pipeline_runs",
"https://idseq.net/samples/12828/pipeline_runs",
"https://idseq.net/samples/12841/pipeline_runs",
"https://idseq.net/samples/12907/pipeline_runs"]
```

Run `rake alert_stalled_uploads` again, see no more message for those samples

Run `rake fail_stalled_uploads`, see error message in UI for a sample

Deploy to staging 
* see existing stalled samples cause alert in slack and then be marked as failed
`Sample.where(upload_error: "LOCAL_UPLOAD_FAILED").count`
* start an upload then disconnect from the internet
* check returned by `stalled_samples` in staging console
* see alert 3 hours later
* see fail status 18 hours later
* see no other samples accidentally failed


app/models/sample.rb
```diff
@@ -856,4 +872,13 @@ def metadata_fields_info
   def first_pipeline_run
     pipeline_runs.order(created_at: :desc).first
   end
+
+  def status_url
+    base_url = if Rails.env == 'staging'
```
tfrcarvalho: You did not find a way to get rid of the domain here? In theory, the rails app could move to any domain.
I still think it should at least move to a generic helper/concern/config.

lib/tasks/result_monitor.rake
```diff
@@ -32,6 +32,51 @@ class MonitorPipelineResults
...
+    rescue => exception
+      LogUtil.log_err_and_airbrake("Failed to fail stalled uploads: #{exception.message}")
+      LogUtil.log_backtrace(exception)
+    end
+  end
+
+  def self.fail_stalled_uploads!
+    # Delay determined by doubling the time after which an alert is sent to admins.
+    Sample
+      .stalled_uploads(6.hours)
```
cdebourcy: I think 6 hours runs a risk of interrupting uploads that are still going. How about 18 hours for people with slow networks that leave the browser open overnight?

lib/tasks/result_monitor.rake
```diff
@@ -32,6 +32,51 @@ class MonitorPipelineResults
...
+      .each { |sample| sample.fail_local_upload!("Stalled for more than 6 hours") }
```
cdebourcy: Could we do the status update in batch using `update_all`? Adding O(N) synchronous operations to result monitor is a scalability no-go.

app/models/pipeline_run.rb
```diff
@@ -1433,14 +1433,7 @@ def self.viewable(user)
...
 
-  def status_url
-    base_url = if Rails.env == 'staging'
-                 "https://staging.idseq.net"
-               else
-                 "https://idseq.net"
-               end
-    base_url + "/samples/#{sample.id}/pipeline_runs"
-  end
+  delegate :status_url, to: :sample
```
tfrcarvalho: nit: Can you move delegate to the top?

lib/tasks/result_monitor.rake
```diff
@@ -32,6 +32,51 @@ class MonitorPipelineResults
...
+
+  def self.alert_stalled_uploads!
+    # Delay determined based on query of historical upload times, where 80%
+    # of successful uploads took less than 3 hours by client_updated_at.
+    samples = Sample.stalled_uploads(3.hours).where(upload_error: nil)
+    if samples.empty?
+      return
+    end
+
+    samples.each { |sample| sample.update(upload_error: Sample::UPLOAD_ERROR_LOCAL_UPLOAD_STALLED) }
```
cdebourcy: same comment about update_all

cdebourcy: With update_all, you'll also avoid triggering all these before/after_save callbacks that have no business being triggered here: https://github.com/chanzuckerberg/idseq-web/blob/master/app/models/sample.rb#L78

lib/tasks/result_monitor.rake
gregdingle: 👍

lib/tasks/result_monitor.rake
gregdingle: We could. I didn't think it would be an issue because it is in the error path. If we have so many upload failures that the performance of the result monitor is affected, then we have bigger problems 🙂

lib/tasks/result_monitor.rake
gregdingle: Same comment re performance.

Thanks for pointing out those callbacks. It seems like if they are not generally applicable, they should be restricted to certain types of events.

app/models/sample.rb
gregdingle: No I didn't do that yet. I'll look into it.

app/models/pipeline_run.rb
gregdingle: 👍

lib/tasks/result_monitor.rake
cdebourcy: I still think you should. The cycle is already a bottleneck when there many samples and we need every perf gain we can get, especially when it’s so easy.

gregdingle: Updated

lib/tasks/result_monitor.rake
gregdingle: Updated

app/models/sample.rb
gregdingle: Updated

gregdingle:
@cdebourcy 

> If you did want to do this alert in the pipeline_monitor -- which seems conceptually more logical to me, but opinions can differ -- I believe this would be the place to put it:
> https://github.com/chanzuckerberg/idseq-web/blob/220ac23c8863b09af1a2985646df1e6a88e97c5e/lib/tasks/pipeline_monitor.rake#L306

I don't think either monitor is that good a fit. Why not create more scheduled rake tasks? Why was the auto benchmark code not put in its own task?

gregdingle:
Rubocop says:

> lib/tasks/result_monitor.rake:56:8: C: Avoid using update_all because it skips validations.

😂

cdebourcy:
> @cdebourcy
> 
> > If you did want to do this alert in the pipeline_monitor -- which seems conceptually more logical to me, but opinions can differ -- I believe this would be the place to put it:
> > https://github.com/chanzuckerberg/idseq-web/blob/220ac23c8863b09af1a2985646df1e6a88e97c5e/lib/tasks/pipeline_monitor.rake#L306
> 
> I don't think either monitor is that good a fit. Why not create more scheduled rake tasks? Why was the auto benchmark code not put in its own task?

Putting it in its own scheduled rake task would be the best!! You can look at the infra piece here: https://github.com/chanzuckerberg/idseq-infra/blob/master/terraform/envs/prod/resque/main.tf#L11

app/models/sample.rb
```diff
@@ -520,6 +522,13 @@ def check_status
     end
   end
 
+  # Delay determined based on query of historical upload times, where 80%
+  # of successful uploads took less than 3 hours by client_updated_at.
+  def self.stalled_uploads(delay = 3.hours)
+    where(status: STATUS_CREATED)
```
cdebourcy: Sorry to be adding more comments here, but in order to make sure we're only dealing with local uploads, we also need to check that `sample.input_files.pluck(:source_type)` are both set to "local". 

Context: We often notice lots of S3 uploads waiting for a resque worker (this will happen more often once people upload from Basespace), and it would be wrong to fail those samples, regardless of how long they have to wait for their upload to come through.

cdebourcy:
Left a couple more comments. I think this change also needs some tests of to make sure real, longer-running but successful uploads are not disrupted. Can test in staging before merging. Will approve once the comment about `source_type == "local"` and the test plan are addressed ;)

cdebourcy:
> Rubocop says:
> 
> > lib/tasks/result_monitor.rake:56:8: C: Avoid using update_all because it skips validations.
> 
> 😂

haha

gregdingle:
> Putting it in its own scheduled rake task would be the best!! You can look at the infra piece here: https://github.com/chanzuckerberg/idseq-infra/blob/master/terraform/envs/prod/resque/main.tf#L11

Thanks. I'm looking into that now. Would you recommend delaying this PR until that is ready? 

> Left a couple more comments. I think this change also needs some tests of to make sure real, longer-running but successful uploads are not disrupted. Can test in staging before merging. Will approve once the comment about source_type == "local" and the test plan are addressed ;)

Good ideas.

app/models/sample.rb
gregdingle: That's good context, and I will restrict to local files, but wouldn't we want an alert if an s3 upload takes longer than 3 hours?

cdebourcy: > That's good context, and I will restrict to local files, but wouldn't we want an alert if an s3 upload takes longer than 3 hours?

We're specifically trying to detect cases where user gave up or their network got interrupted, right? In case of S3 or basespace upload, this doesn't really happen. It would be either just waiting in the queue or it's a really huge file that just takes a long time to transfer but will complete eventually. Maybe we don't want to mix all these different types of concerns.

gregdingle:
Here is my staging test plan: 

Deploy to staging
* start an upload then disconnect from the internet, see alert 3 hours later
* see fail status 18 hours later
* see no other samples accidentally failed

gregdingle:
@cdebourcy  I tested the following in staging:
* see existing stalled samples cause alert in slack and then be marked as failed
* Sample.where(upload_error: "LOCAL_UPLOAD_FAILED").count
* start an upload then disconnect from the internet
* check returned by stalled_samples in staging console
* see no other samples accidentally failed

And I will keep my eye out for this:
* see alert 3 hours later
* see fail status 18 hours later

I'd like to merge in now to unblock staging.

PR #2452, opened 2019-07-30
[Pipeline Viz] Add step-level status information to graph service

emilyzhong:
# Description

* Adds step-level job status information to the pipeline viz graph service (based on step information from https://github.com/chanzuckerberg/idseq-dag/pull/172)

# Tests
* Added rspec tests for service
* Ran a sample using idseq-dag branch with changes, then go to `samples/[sample id]/pipeline_viz/[version].json` to see the json, which includes the status for each step.


app/models/pipeline_run_stage.rb
```diff
@@ -41,6 +48,19 @@ def stage_status_file(status)
...
+  def dag_json_name
+    STEP_NUMBER_TO_DAG_JSON_NAMES[step_number]
+  end
+
+  def step_status_file_path
+    path_beginning = if step_number <= 2
+                       pipeline_run.sample.sample_output_s3_path
+                     else
+                       pipeline_run.sample.sample_postprocess_s3_path
+                     end
```
jshoe: is there going to be one in the `expt` folder too?

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -190,4 +190,6 @@ def status_display_helper(states_by_output_hash, results_finalized_var)
       "HOST FILTERING"
     end
   end
+
+  module_function :get_s3_file
```
jshoe: is this for the purpose of avoiding the `include PipelineOutputsHelper`?

app/models/pipeline_run_stage.rb
```diff
@@ -29,6 +29,13 @@ class PipelineRunStage < ApplicationRecord
   DAG_NAME_POSTPROCESS = "postprocess".freeze
   DAG_NAME_EXPERIMENTAL = "experimental".freeze
 
+  STEP_NUMBER_TO_DAG_JSON_NAMES = {
```
jshoe: I think we should avoid redefining the ordering here because the dag model is meant to (take steps to) decouple the stage orderings from an exact sequence. So PipelineRunStage.step_number is meant to handle the ordering. So as an alternative you could have a mapping of the stage names to the "dag JSON" names like "Host Filtering" => "host_filter".

app/services/retrieve_pipeline_viz_graph_data_service.rb
```diff
@@ -49,14 +49,19 @@ def call
   private
 
   def create_stage_nodes_scaffolding
+    all_step_status = step_status
```
jshoe: s/all_step_status/all_step_statuses ?

spec/services/retrieve_pipeline_viz_graph_data_service_spec.rb
```diff
@@ -120,6 +127,23 @@
       expect(results.keys).to contain_exactly(*expected_stage_results.keys)
     end
 
+    it "should include step-level status updates" do
+      allow(PipelineOutputsHelper).to receive(:get_s3_file).and_return(step_status_data.to_json)
+      results = RetrievePipelineVizGraphDataService.call(@pr.id, true, false)
+
+      results[:stages][0][:steps].each do |step|
+        # Status should be nil of no job status data exists for the stage
```
jshoe: nil of no job status?

app/models/pipeline_run_stage.rb
```diff
@@ -41,6 +48,19 @@ def stage_status_file(status)
...
 
+  def dag_json_name
+    STEP_NUMBER_TO_DAG_JSON_NAMES[step_number]
+  end
+
+  def step_status_file_path
+    path_beginning = if step_number <= 2
+                       pipeline_run.sample.sample_output_s3_path
+                     else
+                       pipeline_run.sample.sample_postprocess_s3_path
```
tfrcarvalho: What about experimental?

app/services/retrieve_pipeline_viz_graph_data_service.rb
tfrcarvalho: all_steps_...?

app/services/retrieve_pipeline_viz_graph_data_service.rb
```diff
@@ -83,6 +88,12 @@ def create_edges
     return edges
   end
 
+  def step_status
+    @pipeline_run.pipeline_run_stages.map do |prs|
+      JSON.parse(PipelineOutputsHelper.get_s3_file(prs.step_status_file_path) || "{}")
+    end
```
tfrcarvalho: Is it guaranteed to be correct JSON?
Maybe we should rescue parser error, just in case

app/models/pipeline_run_stage.rb
tfrcarvalho: I would actually just merge this with a structure that defines ALL properties (dag to STAGES.
Would then change `create_run_stages` to use it.
Didnt we agree on this for the currently blocked PR?

app/helpers/pipeline_outputs_helper.rb
emilyzhong: It's so I can stub the method in the test — otherwise, it says that `PipelineOutputsHelper` doesn't have the method

app/models/pipeline_run_stage.rb
emilyzhong: Hm I haven't seen a sample with a `expt` folder — it seems like the experimental dag json template also outputs to the postprocess directory

app/models/pipeline_run_stage.rb
emilyzhong: Mentioned it below but it seems like all the experimental results are also being placed in the `postprocess` folder

spec/services/retrieve_pipeline_viz_graph_data_service_spec.rb
emilyzhong: Oops, should be "if"

app/models/pipeline_run_stage.rb
jshoe: ah looks like it's been updated. guess we don't need that backwards compatibility

app/services/retrieve_pipeline_viz_graph_data_service.rb
emilyzhong: It should be correct unless the file is written incorrectly, but I'll add the rescue parser error too 👍

app/services/retrieve_pipeline_viz_graph_data_service.rb
```diff
@@ -17,6 +16,7 @@ class RetrievePipelineVizGraphDataService
   #             edge being an input edge to the node.
   #           - ouputEdges: An array of indices that map to edges in the @edges array, each edge being an
   #             output edge from the node
+  #           - status: A number representing the current status of the step at time of retrieval.
```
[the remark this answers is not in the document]
emilyzhong: I was thinking about having the frontend do the number => status, but it makes sense to put it here too

app/models/pipeline_run_stage.rb
tfrcarvalho: I guess we missed the extra map that is done on the `create_run_stages`. We should consolidate that one too.

emilyzhong: Do you mean a mapping from stage number => stage name?

app/helpers/pipeline_outputs_helper.rb
tfrcarvalho: There are some alternatives: 
1. Make the method static, if safe
2. Mock the S3 service: 
	``` 
		s3 = Aws::S3::Client.new(stub_responses: true)
      	s3.stub_responses(:get_object,
	```
3. Mock any instance with `allow_any_instance_of(PipelineOutputsHelper).to receive(: get_s3_file).and_return(step_status_data.to_json)`

Since this looks like an integration test... I would go with option 2.

app/models/pipeline_run_stage.rb
tfrcarvalho: I meant defining a unique structure like:
```
STAGE_INFO = {
	1 => {
		name: "Host Filtering".freeze,
		dag_name: "host_filtering".freeze,
		job_command_func: "host_filtering_command".freeze
	},
	...
```

And then always using this same structure to feed all places where you need to access this info.

emilyzhong: I added in a map like above, but also left the original constant definitions above since some parts of the code reference the name or dag_name directly (for example, the dag_json templates or the mapping of descriptions in `pipeline_runs_helper`), and going along with Jonathan's comment, I don't want to directly relate those (especially the dag_json template) with ordering.

tfrcarvalho: Jonathan meant that you should always use the same order that is injected into the DB. And that order should be defined in a unique place.
At this point it is defined it in two different places that you need to keep in sync.

app/helpers/pipeline_outputs_helper.rb
emilyzhong: I went with the second option (should be reflected in the code now)

emilyzhong:
Ran this branch on a couple samples and everything on this side worked as expected 👍

PR #2453, opened 2019-07-31
Add checkboxes to the sample upload table in Sample Upload flow.

MarkAZhang:
# Description

User can now select, unselect, and remove unselected samples for all three upload types during the upload flow.

Screen Shot 2019-07-30 at 5 30 14 PM

Screen Shot 2019-07-30 at 5 30 27 PM

Screen Shot 2019-07-30 at 5 30 39 PM

We are now also using the new virtualized table component for these uploads. This means the table headers are sortable.

Screen Shot 2019-07-30 at 5 32 30 PM

# Notes

Overall, this PR:
* Adds the checkbox/selection feature.
* Refactors the upload tables to use the new virtual table component.
* Simplifies the code flow in `<UploadSampleStep>`, a large 800-line file. The file, while not perfect, is much more readable now.

This commit is large, so I've listed the major things I did for this commit:
* For local uploads, white-list the fields that we pass to the bulk-upload endpoint. (so we exclude `files`, `_selectId`, etc.)
* Add a larger `<SortIcon>` to `<Table>`.
* Consolidate all three SampleUploadTables into a single component `<SampleUploadTable>`. Different sample fields are displayed based on the upload type. (previously, local and remote uploading used `<BulkSampleUploadTable>`)
* `<SampleUploadTable>` wraps around `<Table>` and adds functionality for removing unselected samples. 
* Some modifications to the `<Table>`: allow it to respond to new columns props, change the sortBy prop to defaultSortBy, and simplify the signature of onSelectAllRows.

In `UploadSampleStep.jsx` (the biggest refactor):
* Got rid of the `sampleNamesToFiles` variable, and store the files in the sample instead. This made it easier to consolidate the local upload code with the S3/basespace code in UploadSampleStep. The `sampleNamesToFiles` object is reconstructed in `bulkLocalUploadWithMetadata` because we need it there.
* Previously, local samples were treating differently and had separate functions such as `handleLocalSampleChange`. These are now all merged into single functions for all upload types.
* Re-ordered the functions in the file by functionality. (tab-related, project-related, sample-related, basespace-related, etc.)
* Verify that analytics still work.

# Tests

Verified that:
* Sample upload still works for Basespace, including samples from multiple basespace projects.
* Sample upload still works for S3.
* Sample upload still works for local uploads.
* Select, unselect, select all, unselect all, and remove unselected work for all three tabs.
* Table headers sort properly in all three tabs.
* Samples renaming (for duplicate sample names) works properly in all three tabs, and selection still works as well. Including the case where local samples are first uploaded, then the project is selected.
* If returning to Step 1 from a later step, unselecting a sample causes the header navigation for future steps to disable (requiring the user to proceed via the Continue button)
* Project creation still works.
* Tables on Data Discovery work as before (default columns are correct, adding/removing columns via dropdown still works).


.eslintrc
```diff
@@ -28,6 +28,7 @@
     "react/jsx-key": "warn",
     "react/no-string-refs": "warn",
     "react/prop-types": "warn",
+    "react/display-name": "off",
```
MarkAZhang: I added this because it was giving me what I thought was an unreasonable error. (I was trying to create a cell renderer in SampleUploadTable, which involved a function that returned JSX. eslint interpreted this as a component class). Based on what I read about the rule, we don't seem to use it.

app/assets/src/components/views/SampleUploadFlow/SampleUploadTable.jsx
```diff
@@ -0,0 +1,176 @@
...
+    if (sampleUploadType === "basespace") {
+      return [
+        NAME_COLUMN,
+        BASESPACE_PROJECT_NAME_COLUMN,
+        FILE_SIZE_COLUMN,
+        FILE_TYPE_COLUMN,
+      ];
+    }
+
+    if (sampleUploadType === "remote") {
```
tfrcarvalho: Why not join the two conditions?
If you do not like the `&&` you can go with `["remote", "local"].includes`

.eslintrc
tfrcarvalho: I believe the rule just asks  you to defined the renderer as a named function vs. inline. 
In general, closures are considered bad practice, because this means that everytime you call it you are creating a new function. 
You should probably use a class method for the components.

tfrcarvalho: IMO, i would keep this rule.

app/assets/src/components/views/SampleUploadFlow/UploadSampleStep.jsx
```diff
@@ -57,19 +63,23 @@ class UploadSampleStep extends React.Component {
...
     createProjectOpen: false,
     projects: [],
-    currentTab: LOCAL_UPLOAD_TAB,
+    currentTab: LOCAL_UPLOAD,
     localSamples: [],
     remoteSamples: [],
     basespaceSamples: [],
+    // We generate a unique "selectId" for each sample, which we use to store which samples are selected.
+    // This simplifies the logic, because sample names can change (they can get renamed when de-duped)
+    localSelectedSampleIds: new Set(),
```
tfrcarvalho: Do we allow for sample upload from multiple sources ?
Why do we need to keep track of all? To go back and forth?

app/assets/src/components/views/SampleUploadFlow/UploadSampleStep.jsx
```diff
@@ -370,52 +286,123 @@ class UploadSampleStep extends React.Component {
...
+        ...pick(["name", SELECT_ID_KEY], sample),
+        file_names: map(
+          file => file.name || file.source,
+          sample.input_files_attributes
+        ).sort(),
+      }));
+    }
+  };
+
+  areAllSamplesSelected = sampleType => {
```
tfrcarvalho: This would have been taken care of by the table... is this an issue due to the fact you were trying to use the same table for all upload types?

app/assets/src/components/visualizations/table/BaseTable.jsx
```diff
@@ -14,11 +13,21 @@ import BasicPopup from "~/components/BasicPopup";
 import Checkbox from "~ui/controls/Checkbox";
 import MultipleDropdown from "~ui/controls/dropdowns/MultipleDropdown";
 import PlusIcon from "~ui/icons/PlusIcon";
+import SortIcon from "~ui/icons/SortIcon";
 import { humanize } from "~/helpers/strings";
 import { logAnalyticsEvent } from "~/api/analytics";
 
 import cs from "./base_table.scss";
 
+const addDefaultsToColumns = (columns, defaultColumnWidth) => {
```
tfrcarvalho: Why define this outside? 
YOu can just keep the setDefaults function but change its signature.

app/assets/src/components/visualizations/table/BaseTable.jsx
```diff
@@ -29,21 +38,28 @@ class BaseTable extends React.Component {
...
+      return {
+        activeColumns: props.initialActiveColumns,
+        columns: addDefaultsToColumns(props.columns, props.defaultColumnWidth),
+        prevPropsColumns: props.columns,
+      };
+    }
+    return null;
+  }
+
   setDefaults = columns => {
```
tfrcarvalho: Makes not sense to define a whole different function when you can just redefined the signature of this one.

app/assets/src/components/visualizations/table/base_table.scss
```diff
@@ -34,13 +34,25 @@
...
     max-width: 100%;
-    white-space: nowrap;
-    text-overflow: ellipsis;
-    overflow: hidden;
-    vertical-align: middle;
+    display: flex;
+    align-items: center;
+
+    .label {
+      // Shrink once you're at the maximum width (e.g. don't overflow). Do not grow unless needed.
```
tfrcarvalho: I wonder if it is better to wrap an just hide the overflow... I would have liked the ellipsis, but maybe it is better to use the space better. Maybe check with Jenn.

.eslintrc
MarkAZhang: Okay, sounds good.

app/assets/src/components/views/SampleUploadFlow/UploadSampleStep.jsx
MarkAZhang: I'm not sure I understand your question.

We need to track which samples are selected somehow. We could track by name, but sample names can change (if a sample is named Test, and the user switches to a project that already has a sample Test, the new sample will be re-named to Test_1). So we track by a special id we generate here.

Even if there were only local uploads we would need this logic.

tfrcarvalho: I guess my question is why do we need to keep track of all 3 (i.e. local, basespace, s3) vs. a generic selectedSampleIds that gets re-used based on the selected upload type.

I was assuming that people could go back and forth between s3, basespace and local, and you do not want to reload the samples every time they change. Is that correct?

app/assets/src/components/views/SampleUploadFlow/UploadSampleStep.jsx
MarkAZhang: I missed that this was in the table. Thanks!

app/assets/src/components/visualizations/table/BaseTable.jsx
MarkAZhang: Done. Renamed to `setColumnDefaults` for clarity.

MarkAZhang: Actually when I ran it, it failed, and now I remember why I did this.

You can't call `this.setDefaults` in a static function (`getDerivedStateFromProps`)

tfrcarvalho: yeah, i meant to make it static. I still like within the class though...

app/assets/src/components/visualizations/table/base_table.scss
MarkAZhang: I'll leave it for now, since I haven't changed the overflow behavior of the text. What this code does is make sure that if the text has ellipsis, the sort icon is still showing. Previously, the sort icon was considered part of the text and would disappear when the ellipsis shows. We can follow up on the text overflow.

app/assets/src/components/views/SampleUploadFlow/UploadSampleStep.jsx
MarkAZhang: Ah that makes sense. Yeah, the other approach can also work, but this felt a little easier. For example, to figure out how many basespace samples are selected, you can just do basespaceSelectedSampleIds.length, instead of doing the extra of step of figuring out which sample ids correspond to basespace samples. Both approaches can work though.

tfrcarvalho: Yeah, i get the logic.

app/assets/src/components/visualizations/table/BaseTable.jsx
MarkAZhang: Done.

PR #2454, opened 2019-07-31
Make rubocop require trailing comma in multiline expression

gregdingle:
# Description

This makes the linting rules in ruby consistent with eslint. We decided to change the equivalent rule there because it makes diffs more accurate and reduces annoying syntax errors. 

# Tests

`rubocop -a -R`


gregdingle:
> Are you sure you ran this rule (using the right config)? This does not fail?
There seems like there are plenty of places where this is done yet.

Sorry you got here too soon! I just committed the fix.

PR #2455, opened 2019-07-31
Optimize report precaching

gregdingle:
# Description

See https://jira.czi.team/browse/IDSEQ-1011 . This diff shrinks the number of backgrounds we precache for on every sample run. Precache runtime should improve proportionally.  

The selected backgrounds are:
* all public access
* default host background
* backgrounds created by the sample owner

# Notes

* I added active record `belongs_to` for background to user because it was absent, and `has_many` in the other direction. 

* The Background `has_and_belongs_to_many :samples, through: :pipeline_runs` appears to be broken

* Most of the background do not have a owner. It's not clear why or whether we should fix that. 

# Tests

run `precache_report_info!` in rails console, see no error


app/models/background.rb
```diff
@@ -11,10 +11,30 @@ class Background < ApplicationRecord
   DEFAULT_BACKGROUND_MODEL_NAME = "default".freeze
   TAXON_SUMMARY_CHUNK_SIZE = 100
 
+  # This set of ten backgrounds was generated by SQL analysis of sample page
+  # requests for the purpose of precache_report_info. The top 1 accounts for
+  # 52% of unique users, top 5 accounts for 71%, and top 10, 83%.
+  TOP_BY_USAGE = [
```
tfrcarvalho: I disagree with this approach a static on the moment analysis.
This is not a scalable approach, and is not something we can generalize to other pages.
If we want to do something like this, then we need to at least generate this list dynamically, e.g. cron task that computes this stats and adds them to our database.

gregdingle: Yes, this static list might fall out of sync with actual usage over time. There are other ways we can rank backgrounds by usage such as:

* host_genomes default_background_id
* pipeline_run background_id
* sample background_id

Do any of those make sense? 

We could also limit the backgrounds to those that have `public_access`

cdebourcy: `public_access` seems to be a good one! I believe the default_backgroud_id would also be included in that set  -- it's required to be public or the default view would break.

Not sure what you mean by "background_id of the sample or pipeline being cached". There is an association between backgrounds and pipeline_runs, but it works the other way around: the background is generated from a set of pipeline_runs

cdebourcy: Could add to that any backgrounds created by the user who uploaded the sample (maybe top 10 starting with the most recent ones?). I.e. where background.user_id == sample.user_id

tfrcarvalho: Your last paragraph seems to contain the best two alternatives. 
Agree, with setting a limit as @cdebourcy suggests, but would probably set a shorted limited.
For public samples, would make sense as long as we keep making a background public restricted (i.e. AFAIK there is no way now for a user to a make a background public).

A bit more convoluted third option, to cover team projects with multiple uploaders or where uploader != analyst: backgrounds that contain samples of the project that the samples below belong too.

Of course, the alternatives are not mutually exclusive.

Are we already measuring the effectiveness of the cache? Could be good to keep track of the backgrounds/samples for which it misses.

gregdingle: Thanks for the input. I pushed some changes. See the updated PR description.

tfrcarvalho:
Do you have the analytics in place to study the effectiveness of this approach?

gregdingle:
Yes, we have the dashboard that shows cache hit rates: https://app.datadoghq.com/dashboard/h6d-92u-nw4/reportinfo-caching-by-gdingle <https://app.datadoghq.com/dashboard/h6d-92u-nw4/reportinfo-caching-by-gdingle> . We should expect it to remain unchanged if the optimization is effective. 

And we should not see as much waiting on https://idseq.net/resque/overview <https://idseq.net/resque/overview>, though I don't know of any summary stats for our Resque. Please point them out if they exist.

gregdingle:
Hi, could I get a review of this?

PR #2459, opened 2019-08-01
[AMR Heatmap] Change AMR Heatmap Controller to use precomputed ontology information

lvreynoso:
# Description

This PR adds a rake task that gathers information on genes in the ARG-ANNOT database and stores them in json format on S3. The AMR heatmap controller's method to populate the gene details sidebar is modified to use S3's object select to get only the relevant entry for a requested gene. This should result in a noticeable speedup.

# Notes

# Tests

* Updated Rspec tests
* For client side code, go to All Data page, pick a bunch of samples with AMR Counts (like tb-flash), click on Heatmap Icon and select "AMR Heatmap" from the dropdown. When the heatmap loads, click on a gene name and verify that correct information is being presented in the sidebar.


app/controllers/amr_heatmap_controller.rb
```diff
@@ -68,169 +69,67 @@ def amr_counts
...
-      search_result["error"] = format(err, gene_name)
-    else
-      search_result["matching_node"] = matching_entry
-    end
-
-    return search_result
+  def fetch_current_ontology_file_key
+    ontology_folder = @client.list_objects_v2(bucket: S3_JSON_BUCKET,
+                                              prefix: S3_JSON_PREFIX).to_h
+    target_key = ontology_folder[:contents][-1][:key] # contents already sorted by key
```
MarkAZhang: This is interesting. It looks like you automatically take the last file in the folder. An alternative approach would be to make this an environment variable (via Parameter store, I can show you how), or store it in the database somehow.

For our nt/nr indices, we have AlignmentConfig objects in our Rails model that store the paths to the index files. 

Will defer to others about whether it's worth it to take an alternative approach.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -68,169 +69,67 @@ def amr_counts
...
+      expression_type: "SQL",
+      expression: "SELECT * FROM S3Object[*].#{gene_name} LIMIT 1",
+      input_serialization: {
+        json: {
+          type: "DOCUMENT",
+        },
+      },
+      output_serialization: {
+        json: {},
+      },
```
MarkAZhang: Awesome use of the S3 SDK! I can see this being useful in other situations as well. Can you generalize this function and put it in the lib folder?

lib/tasks/precompute_gene_ontology.rake
```diff
@@ -0,0 +1,209 @@
...
+    return
+  end
+
+  split_lines = arg_annot_fasta.split("\n")
+  arg_entries = []
+  split_lines.each do |line|
+    if line[0] == ">"
+      split_entry = line.split(";")
+      gene_name = split_entry[2]
+      drug_class = if DRUG_CLASSES.key?(drug_class)
```
MarkAZhang: Is there a line missing here? `drug_class` doesn't seem to be defined before this line.

lib/tasks/precompute_gene_ontology.rake
```diff
@@ -0,0 +1,209 @@
+MASTER_OWL_URI = "https://raw.githubusercontent.com/arpcard/aro/master/aro.owl".freeze
+URL_PUBMED = "https://www.ncbi.nlm.nih.gov/pubmed/".freeze
+S3_JSON_BUCKET = "idseq-database".freeze
+S3_ARGANNOT_FASTA = "s3://idseq-database/amr/ARGannot_r2.fasta".freeze
+
+DRUG_CLASSES = {
+  "agly" => ["Aminoglycosides", "0000016"],
+  "bla" => ["Beta-lactamases", "3000001"],
+  "colistin" => ["Colistins", "0000067"],
+  # "fcd" => [], # Unclear which one this refers to
```
MarkAZhang: Is `fcd` a drug class that appears in ARGannot? Could you add a comment about where this one came from?

lib/tasks/precompute_gene_ontology.rake
lvreynoso: That's a Ruby conditional assignment! This is where `drug_class` is defined. If you don't like it, take it up with Rubocop 🚨haha

app/controllers/amr_heatmap_controller.rb
lvreynoso: Thanks! It was Charles' idea. I'll see if I can generalize it.

app/controllers/amr_heatmap_controller.rb
lvreynoso: Yeah, each time the rake task is run the json file is put in a folder `amr/ontology/YYYY-MM-DD/aro.json` so the latest run of the rake task will be the last key listed here. I like the simplicity of this way, but I don't mind changing it.

lib/tasks/precompute_gene_ontology.rake
lvreynoso: Yes it is. Will do.

app/controllers/amr_heatmap_controller.rb
jshoe: it sounds fine to me to assume the last one (altho maybe I would sort it again just to be sure)

app/controllers/amr_heatmap_controller.rb
```diff
@@ -68,169 +69,67 @@ def amr_counts
...
-  def get_drug_classes(accession, xml_dom)
-    accession_class = xml_dom.at_xpath(".//owl:Class[@rdf:about='http://purl.obolibrary.org/obo/ARO_#{accession}']")
-    drug_class_node_markers = accession_class.xpath("./rdfs:subClassOf/owl:Restriction/owl:onProperty[@rdf:resource='http://purl.obolibrary.org/obo/RO#_confers_resistance_to_drug']")
-    drug_class_family_node_markers = accession_class.xpath("./rdfs:subClassOf/owl:Restriction/owl:onProperty[@rdf:resource='http://purl.obolibrary.org/obo/RO#_confers_resistance_to']")
-    drug_markers = drug_class_node_markers + drug_class_family_node_markers
-    drug_class_nodes = []
-    drug_markers.each do |node_marker|
-      drug_class_nodes.push(node_marker.at_xpath("./following-sibling::owl:someValuesFrom"))
+    rescue
+      Rails.logger.info("Error retrieving ontology entry for #{gene_name} from s3")
```
jshoe: you probably want to LogUtil.log_err_and_airbrake or at least Rails.logger.error these two errors, right?

lib/tasks/precompute_gene_ontology.rake
```diff
@@ -0,0 +1,209 @@
...
+  owl_uri = URI(MASTER_OWL_URI)
+  response = Net::HTTP.get_response(owl_uri)
+  unless response.is_a?(Net::HTTPSuccess)
+    Rails.logger.error("Unable to fetch OWL file")
+    return
+  end
+  owl_doc = Nokogiri::XML(response.body.force_encoding("utf-8"))
+
+  arg_annot_fasta = nil
+  split_uri = S3_ARGANNOT_FASTA.split("/", 4)
```
jshoe: it looks like you could dry it out by using S3_JSON_BUCKET and just key `amr/ARGannot_r2.fasta`

app/controllers/amr_heatmap_controller.rb
lvreynoso: Yeah, will fix this.

lvreynoso:
Generalized S3 select function and incorporated PR comments.

app/controllers/amr_heatmap_controller.rb
```diff
@@ -1,10 +1,13 @@
...
+S3_JSON_BUCKET = "idseq-database".freeze
+S3_JSON_PREFIX = "amr/ontology/".freeze
+DEFAULT_S3_REGION = "us-west-2".freeze
 
 class AmrHeatmapController < ApplicationController
-  include PipelineOutputsHelper
+  include S3Util
 
   before_action :admin_required
+  before_action :set_aws_client, only: [:fetch_ontology]
```
MarkAZhang: Don't need aws_client here any more.

app/controllers/amr_heatmap_controller.rb
MarkAZhang: Can you add a comment for this function explaining why the last key is the most recently generated file? (as you just explained in your Github comment)

app/controllers/amr_heatmap_controller.rb
```diff
@@ -66,171 +69,56 @@ def amr_counts
...
-  def get_publication_name(pmid)
-    pubmed_uri = URI(URL_PUBMED + pmid)
-    response = Net::HTTP.get_response(pubmed_uri)
-    unless response.is_a?(Net::HTTPSuccess)
-      return pmid
+  def fetch_ontology_entry(s3_key, gene_name)
+    sql_expression = "SELECT * FROM S3Object[*].#{gene_name.dump} LIMIT 1"
+    entry = s3_select_json(S3_JSON_BUCKET, s3_key, sql_expression)
+    whole_entry = entry.join.chomp(",")
+    Rails.logger.info(whole_entry)
```
MarkAZhang: Remove logger?

app/controllers/amr_heatmap_controller.rb
lvreynoso: Done!

spec/controllers/amr_heatmap_controller_spec.rb
```diff
@@ -1,4 +1,32 @@
...
+  "description" => "RND efflux pump conferring resistance to fluoroquinolone",
+  "geneFamily" => [
+    {
+      "label" => "subunit of efflux pump conferring antibiotic resistance",
+      "description" => "Subunits of efflux proteins that pump antibiotic out of a cell to confer resistance.",
+    },
+  ],
+  "drugClass" => {
+    "label" => "Fluroquinolones",
+    "description" => "The fluoroquinolones are a family of synthetic broad-spectrum antibiotics that are 4-quinolone-3-carboxylates. These compounds interact with topoisomerase II (DNA gyrase) to disrupt bacterial DNA replication, damage DNA, and cause cell death.",
```
MarkAZhang: A lot of these long strings are duplicated twice. Can you extract them into constants?

app/lib/s3_util.rb
```diff
@@ -0,0 +1,36 @@
+require "aws-sdk-s3"
+
+module S3Util
+  def self.s3_select_json(bucket, key, expression)
```
MarkAZhang: Can you add a comment "select a particular part of a JSON file using Amazon S3 Select SQL syntax"? To help future users of this function.

spec/controllers/amr_heatmap_controller_spec.rb
```diff
@@ -229,64 +257,79 @@
...
+            prefix: context.params[:prefix].to_s,
+            max_keys: 1000,
+            key_count: 1,
+          }
+        })
+        stub_const("S3_CLIENT", s3)
+      end
+      context "when a matching gene name is found" do
+        before do
+          allow(S3Util).to receive(:s3_select_json).and_return(OQXB_ONTOLOGY.to_json + ",")
```
MarkAZhang: Can you add a comment explaining why you need to add the comma to the end? I wouldn't know why otherwise.

app/lib/s3_util.rb
```diff
@@ -0,0 +1,36 @@
...
+    begin
+      S3_CLIENT.select_object_content(s3_select_params) do |stream|
+        stream.on_records_event do |event|
+          entry.push(event.payload.read)
+        end
+      end
+    rescue Aws::S3::Errors => e
+      Rails.logger.error("Error retrieving entry #{expression} from #{bucket}/#{key} from s3")
+      Rails.logger.error(e.message)
+      return []
```
MarkAZhang: Should this return an empty array, or an empty string? (it looks entry.join below is a string)

MarkAZhang:
A few more small changes. Thanks for making this into a rake task!!

app/lib/s3_util.rb
lvreynoso: You're right, fixed

lvreynoso:
Incorporated requested changes

PR #2460, opened 2019-08-01
[Maps] Add interface for setting collection locations after uploading CSVs

jshoe:
### Description
- This PR adds an interface for setting collection locations after uploading CSVs. We agreed to invest in this (with automatic geosearching) before officially announcing the maps feature.
- I synced up with @happyimadesignr on the design and interactions but she's welcome to review more.

### Notes
- MetadataCSVLocationsMenu: new component that shows up under the CSV box. We'll automatically geosearch for what the user has in their CSV. The results show up in the menu, and they can fix them from there.
- MetadataInput: Renamed 'locationWarning' to just 'warning'. Moved `processLocationSelection` into GeoSearchInputBox for modularity.
- MetadataUpload: Added step for kicking off geosearchCSVLocations and added MetadataCSVLocationsMenu below all the issues based on Mark's recommendations. Added `metadata` as a prop to access `metadata` in this component.
- LocationsController: Added a results limit option for efficiency when we only want 1 geosearch result.
- Added a LOCATION_UNRESOLVED_WARNING.

## Tests
- Go through the flows shown above at http://localhost:3000/samples/upload


cdebourcy:
Nice!!!

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
```diff
@@ -0,0 +1,192 @@
...
+              key={metadataType.key}
+              className={cs.input}
+              value={row[metadataType.name]}
+              metadataType={metadataType}
+              onChange={onChange}
+              withinModal={true}
+              isHuman={isRowHuman(row)}
+              warning={CSVLocationWarnings[sampleName]}
+            />
+            {this.renderApplyToAll(sampleName)}
```
tfrcarvalho: I suggest making the condition here `applyToAllSample && this.renderApplyToAll(sampleName)` instead of inside the function. More clear when you are reading the outside component and the function `renderApplyToAll` actually renders `ApplyToAll`.

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
```diff
@@ -0,0 +1,192 @@
...
+  if (!(metadata && metadata.rows)) return;
+
+  // For each unique plain text value, get the #1 search result, if any.
+  const rawNames = uniq(metadata.rows.map(r => r[metadataType.name]));
+  const matchedLocations = {};
+  const requests = rawNames.map(async query => {
+    const res = await getGeoSearchSuggestions(query, 1);
+    if (res.length > 0) matchedLocations[query] = res[0];
+  });
+  await Promise.all(requests);
```
tfrcarvalho: Can you have request limit issues?

app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
```diff
@@ -1,11 +1,43 @@
...
 import { logAnalyticsEvent } from "~/api/analytics";
 import { getGeoSearchSuggestions } from "~/api/locations";
 import LiveSearchPopBox from "~ui/controls/LiveSearchPopBox";
 
+export const LOCATION_PRIVACY_WARNING =
+  "Changed to county/district level for personal privacy.";
+export const LOCATION_UNRESOLVED_WARNING =
+  "Unresolved plain text, not shown on maps.";
+
+// For human samples, drop the city part of the name and show a warning.
```
tfrcarvalho: Shouldn't this be enforced on the server? Clients are vulnerable.
Remind me again, but we still narrows it down when saving, right?

MarkAZhang:
Could we change "Location Matches" to "Confirm your Collection Locations"? Slightly clearer title.

Also, for the warnings such as "unresolved plain text", "changed to state", can we do yellow warning text with a small alert icon? Similar to (without the border):

62066832-29e5d280-b1e7-11e9-8b42-ca78d8e0d572

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
```diff
@@ -0,0 +1,192 @@
...
+
+      return {
+        [NAME_COLUMN]: (
+          <div className={cs.sampleName} key={NAME_COLUMN}>
+            {sampleName}
+          </div>
+        ),
+        [metadataType.key]: (
+          <div>
+            <MetadataInput
```
MarkAZhang: Yay for reuse!

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
```diff
@@ -0,0 +1,192 @@
...
+import { getGeoSearchSuggestions } from "~/api/locations";
+import MetadataInput from "~/components/common/MetadataInput";
+import { processLocationSelection } from "~/components/ui/controls/GeoSearchInputBox";
+import PropTypes from "~/components/utils/propTypes";
+import DataTable from "~/components/visualizations/table/DataTable";
+
+import cs from "./metadata_csv_locations_menu.scss";
+
+// Batch geosearch CSV locations for matches
+export const geosearchCSVlocations = async (metadata, metadataType) => {
```
MarkAZhang: Mega-nit: capitalize "l" in "locations"

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
```diff
@@ -0,0 +1,192 @@
...
+          data={this.getManualInputData()}
+        />
+      </React.Fragment>
+    );
+  }
+}
+
+MetadataCSVLocationsMenu.propTypes = {
+  CSVLocationWarnings: PropTypes.object,
+  metadata: PropTypes.object,
```
MarkAZhang: Can you do `PropTypes.shape` here to document better what the structure of this object is? For readability.

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
```diff
@@ -0,0 +1,192 @@
...
+  metadataType: PropTypes.object,
```
MarkAZhang: Similar here with PropTypes.shape. Also, maybe rename to `locationMetadataType`, since it's just the metadata type for the collection location.

MarkAZhang:
I wonder if it would make sense to wrap the locations menu in something like this:

Screen Shot 2019-08-01 at 9 46 41 AM

Except light blue, with the "Confirm your Collection Locations" title outside (similar to Warnings and Errors).

The reason for this is that then, all the different elements follow a consistent styling. Right now, you have the collapsible warnings/errors, but then the locations menu is styled differently. If you have both warnings and also the csv location menu, I think the inconsistent styling will look a bit jarring.

Can you try that and see how it looks?

happyimadesignr:
Those are good ideas Mark, to wrap in in the same notification background in light blue and the input warning style. Your title suggestion is nice too, more clear.

Other than that, can we move the location column closer to the sample name? And still keep the table full width.

I will mock it up!

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
jshoe: 👍 done

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
jshoe: it's possible yeah.. is there a way I should catch errors while using Promise.all?

app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
jshoe: Yes and yes. I clarified the comment to emphasize that.

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
jshoe: 👍 thank you!

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
jshoe: 👍 done

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
jshoe: 👍 done

jshoe:
> * What happens when there is no match? Adds free text?

Yes, it "stays" as free text.

> * Do we have plans to remove free text? It is a way to circumvent our privacy controls.

No current plans AFAIK. We've discussed this before but we can talk about it more in some docs. I added some more notes to the mapping doc: https://czi.quip.com/gzWdAafIX8Yb/LocationMapping-Sprint-Plan

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
jshoe: 👍

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
tfrcarvalho: I have not used it... but I believe the promise will have `then` functions that will return an array of results (success) and a `catch` that will be called for each(or first?) error. This pattern on the web seems interesting: 
```
Promise.all([
  p1.catch(error => { return error }),
  p2.catch(error => { return error }),
]).then(values => { 
  console.log(values[0]) // "p1_delayed_resolution"
  console.log(values[1]) // "Error: p2_immediate_rejection"
})
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

I guess you might be able to guess when you are going to have the issue depending on how many promises you have and introduce some random delays to desync the requests?

jshoe:
> Could we change "Location Matches" to "Confirm your Collection Locations"? Slightly clearer title.
> 
> Also, for the warnings such as "unresolved plain text", "changed to state", can we do yellow warning text with a small alert icon? Similar to (without the border):

Done and done 👍 

Screen Shot 2019-08-01 at 4 45 45 PM

(altho I just realized you didn't ask for the border :o)

happyimadesignr:
@MarkAZhang  @jshoe 

Here's what I was imaging from our convo:
Screen Shot 2019-08-01 at 5 16 07 PM

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
jshoe: 👍 added batching 20 requests at a time as discussed offline

jshoe:
Turned it into an IssueGroup, as discussed ;) 

Screen Shot 2019-08-01 at 9 27 32 PM

PR #2461, opened 2019-08-01
[IDSEQ-1230] Send email on account request

tfrcarvalho:
# Description

* Send email on account request to confirm contact and inform that the team will be in touch soon.

# Tests

Request a new account on IDseq's landing page:
* Action mailer reports email sent in logs.
* Will confirm in staging that the email is correctly sent.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -86,7 +86,11 @@ class HomeController < ApplicationController
       return
     end
 
-    UserMailer.account_request_reply(home_params[:email])
+    begin
+      UserMailer.account_request_reply(home_params[:email]).deliver_now
+    rescue
+      LogUtil.log_err_and_airbrake("account_request_reply(#{home_params[:email]} failed")
+    end
 
     body = ""
     home_params.each do |k, v|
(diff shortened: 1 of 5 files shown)


app/controllers/home_controller.rb
```diff
@@ -86,6 +86,12 @@ def sign_up
       return
     end
 
+    begin
+      UserMailer.account_request_reply(home_params[:email]).deliver_now
+    rescue
```
jshoe: could you throw in the err.message as well? or at least Rails.logger that one

tfrcarvalho: Did backtrace instead.

app/lib/url_util.rb
```diff
@@ -4,7 +4,9 @@ class UrlUtil
   # absolute URLs to be the same in emails and in other places.
   def self.absolute_base_url
     options = Rails.application.config.action_mailer.default_url_options
-    host, port, protocol = options[:host], options[:port], options[:protocol] # rubocop:disable Performance/ParallelAssignment
```
gregdingle: thanks for the fix 😄

PR #2469, opened 2019-08-02
Bump LongRunningSampleEvent threshold to 10 hours

jshoe:
### Description
- Raise the LongRunningSampleEvent threshold from 8 hours to 10 hours to fit better with the distribution of normally running samples.

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -901,7 +901,7 @@ class PipelineRun < ApplicationRecord
     MetricUtil.put_metric_now("samples.running.run_time", run_time, tags, "gauge")
 
     if alert_sent.zero?
-      threshold = 8.hours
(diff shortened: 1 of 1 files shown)


app/models/pipeline_run.rb
```diff
@@ -901,7 +901,7 @@ def check_and_log_long_run
     MetricUtil.put_metric_now("samples.running.run_time", run_time, tags, "gauge")
 
     if alert_sent.zero?
-      threshold = 8.hours
+      threshold = 10.hours
```
gregdingle: it's always nice to add a comment for constant such as these as to how the value was determined. for example "queries last 1000 pipline runs, found 95% finished in less than 10 hours"

jshoe: 👍

PR #2474, opened 2019-08-02
[Pipeline Viz] Step-level updates frontend

emilyzhong:
# Description

Frontend for step-level updates, along with node visual design

Error state:

image
image

Completed + In Progress:
image
image

# Notes

Includes the `hover-state` sass function, along with using `:export` to access sass variable strings directly, both which I enjoyed learning about very much. 🎉 

# Tests

Run with pipeline_viz enabled account by uploading a sample with idseq-dag branch `revert-173-revert-172-ezhong/step_info`, then go to `/samples/[sample id]/pipeline_viz` (+ refresh to see updates as pipeline executes)


app/helpers/pipeline_outputs_helper.rb
```diff
@@ -190,4 +190,6 @@ def status_display_helper(states_by_output_hash, results_finalized_var)
       "HOST FILTERING"
     end
   end
+
+  module_function :get_s3_file
```
tfrcarvalho: Didnt you remove this?

app/services/retrieve_pipeline_viz_graph_data_service.rb
```diff
@@ -64,26 +55,66 @@ def create_stage_nodes_scaffolding
...
+    end
+  end
+
+  def redefine_job_status(status)
+    case status
+    when "instantiated", nil
+      "notStarted"
+    when "running", "finished_running"
+      "inProgress"
+    when "uploaded"
```
tfrcarvalho: Can you add a comment explaining this apparent mismatch between uploaded and finished?

app/services/retrieve_pipeline_viz_graph_data_service.rb
```diff
@@ -64,26 +55,66 @@ def create_stage_nodes_scaffolding
...
+        {}
+      end
+    end
+  end
+
+  def redefine_job_status(status)
+    case status
+    when "instantiated", nil
+      "notStarted"
+    when "running", "finished_running"
```
tfrcarvalho: finished_running = inProgress?

app/helpers/pipeline_outputs_helper.rb
emilyzhong: Oops forgot to remove it — will do 👍

app/services/retrieve_pipeline_viz_graph_data_service.rb
emilyzhong: Just added a comment, but `finished_running` occurs when the outputs have been created, but hasn't been uploaded to aws yet. Since the file is not available to download yet, I marked as "inProgress"

PR #2475, opened 2019-08-03
Rake task to identify columns that have bad NULLs

gregdingle:
# Description

For data cleanup, this script identifies all database columns that have `NULL` values. In particular, it can identify columns that are configured in Rails to validate `presence: true`, which then clearly should not have `NULL`s. 

# Notes

* In local testing only visualizations table had nulls in presence true cols. 

```
Model Visualization has NULLs in presence: true columns: {
  "data": 0.0,
  "name": 0.00023
}
```

* My intention is to examine the rows that this script turns up in prod and decide whether to delete them or fix them, then make the cols `NOT NULL`.

* My next intention, with `only_presence_true` false, is consider making `NOT NULL` columns that have zero `NULL`s in prod. This will eliminate the need for checking `nil` on those col values in ruby, and `null` in javascript. 

* Here is the output locally for that:
https://czi.quip.com/LWbJAyLuX7fn/2019-08-02-NULL-values-in-database

# Tests

`rake find_cols_that_have_nulls`

verify non-zero values are in the database with manual sql query


lib/tasks/find_null_values.rake
```diff
@@ -0,0 +1,57 @@
+
+desc 'Finds database columns that are have NULL values'
```
jshoe: nit: "that are have"

lib/tasks/find_null_values.rake
```diff
@@ -0,0 +1,57 @@
...
+task 'find_cols_that_have_nulls', [:max_per_model, :only_presence_true] => :environment do |_t, args|
+  ActiveRecord::Base.logger.level = :info
+  Rails.application.eager_load!
+  models = ApplicationRecord.descendants
+
+  args.with_defaults(max_per_model: 1_000_000)
+  args.with_defaults(only_presence_true: false)
+
+  models.each do |model|
+    # puts "\nFinding #{model} cols that have NULL values ..."
```
jshoe: uncomment?

PR #2476, opened 2019-08-03
Update PipelineRun and experimental dag template for pipeline version 3.9

lvreynoso:
# Description

To make the AMR Heatmap more scientifically useful, I've submitted a PR in idseq-dag ([#175](https://github.com/chanzuckerberg/idseq-dag/pull/175)) that would add total reads, reads per million, and depth per million to the results table for each resistance gene SRST2 identifies in a sample. The gene given in the ARG-ANNOT annotation is also added since this name is much more closely aligned with the CARD Ontology (>90%), as well as the NCBI GeneBank accession number for easy linking to detailed genome information.

This PR updates the dag template to take in account the new requirements for the srst2 step, and adds the new columns in the results to a pipeline run's `amr_counts`.

# Tests

* Ran a benchmark test sample and a giant AMR NOVA test sample on staging and local, hit the `amr_heatmap/amr_counts.json` endpoint to make sure the new data was being returned.
Screen Shot 2019-08-08 at 11 35 29 AM


app/lib/dags/experimental.json.erb
```diff
@@ -38,7 +38,11 @@
       "taxid_locations_combined.json"
     ],
     "alignment_viz_out": ["align_viz.summary"],
+    <% if attribute_dict[:pipeline_version].to_f >= 3.8 %>
```
cdebourcy: I think if we start adding pipeline version conditions to this file, it will quickly become unmanageable. In the past, we've usually just made a clean break and deployed the new idseq-dag version at the same time as the new idseq-web version to keep the two in sync. There is currently no expectation that the web will be backwards-compatible with older idseq-dag versions, so you don't need to allow for it here.

In other words: please get rid of the if-else here and just make sure to coordinate the idseq-dag and idseq-web deployments.

Ignore me if you're just writing this condition for some testing in staging. My comment applies to the final code that goes into master after you have experimented and tested it.

MarkAZhang: @cdebourcy I had advised her to do this because it seemed like a safer way to deploy. If you have this if-else wrapper, then the dag will be valid both before and after the idseq_dag feature branch is merged into master. The plan is to then remove the if-else wrapper once the dag change is deployed and stable, so it will not be there long-term.

The reason it feels safer is that you don't have to coordinate the deploy of the two services, with a risk that if a user submits a sample at a bad time, it might fail.

My approach does involve deploying the if-else to production and then removing it later.

Let me know if you prefer your way, and I can change the idseq dag quick start docs.

app/models/pipeline_run.rb
```diff
@@ -563,7 +563,12 @@ def db_load_amr_counts
                               allele: amr_result_fields[1],
                               coverage: amr_result_fields[2],
                               depth:  amr_result_fields[3],
-                              drug_family: amr_result_fields[12], }
+                              annotation_gene: amr_result_fields[11].split(";")[2],
+                              genbank_accession: amr_result_fields[11].split(";")[4],
+                              drug_family: amr_result_fields[12],
+                              total_reads: amr_result_fields[16],
+                              rpm: amr_result_fields[17],
+                              dpm: amr_result_fields[18].strip, } # Remove dangling newline character
```
MarkAZhang: Not a big deal, but is this something we can easily get rid off in the pipeline output, so we don't need the special case here?

app/lib/dags/experimental.json.erb
```diff
@@ -38,7 +38,7 @@
       "taxid_locations_combined.json"
     ],
     "alignment_viz_out": ["align_viz.summary"],
-    "srst2_out": ["output.log", "output__genes__ARGannot_r2__results.txt", "output__fullgenes__ARGannot_r2__results.txt", "amr_processed_results.csv", "amr_summary_results.csv"],
+    "srst2_out": ["out.log", "out__genes__ARGannot_r2__results.txt", "out__fullgenes__ARGannot_r2__results.txt", "amr_processed_results.csv", "amr_summary_results.csv", "output__.ARGannot_r2.sorted.bam"],
```
MarkAZhang: Out of curiosity, what's the rollout plan? Are you going to push to prod and push the dag change to master at the same time?

app/models/pipeline_run.rb
lvreynoso: No, since it's the last column in the csv file the entry will have a newline character.

I *could* rewrite the function to use Ruby's builtin CSV tools, which is honestly probably better...I think I'll do that.

app/lib/dags/experimental.json.erb
MarkAZhang: I notice that this file still contains the guards from a past rollout (`<% if attribute_dict[:pipeline_version].to_f >= 3.6 %>`). That approach seems safer to me, but if you talked with Charles and feel that the coordinated rollout is better, please feel free.

db/migrate/20190808181844_add_rpm_data_to_amr_counts.rb
```diff
@@ -0,0 +1,9 @@
+class AddRpmDataToAmrCounts < ActiveRecord::Migration[5.1]
+  def change
+    add_column :amr_counts, :annotation_gene, :string
+    add_column :amr_counts, :genbank_accession, :string
+    add_column :amr_counts, :total_reads, :bigint
```
MarkAZhang: Nit: Maybe prefer `:integer` to `:bigint`? I believe `:integer` takes up fewer bytes (4 instead of 8) and still stores very big numbers.

app/lib/dags/experimental.json.erb
lvreynoso: That's the plan, yeah. I could deploy it to staging for a short bit, as long as people are aware that samples won't run unless they specify the `lvreynoso/amr_reads` pipeline branch.

lvreynoso:
> LGTM. Please coordinate this with the `idseq_dag` change.

Thanks! Someone does need to approve the PR in idseq_dag first though ;)

PR #2477, opened 2019-08-03
[Pipeline Viz] Scrolling doesn't Zoom

emilyzhong:
# Description

Makes use of new `disableScrollZoom` attribute in `react-easy-panzoom` library.

# Notes
* Some _slight_ buggy behavior that I haven't been able to fully figure out -- sometimes, if you are zoomed in and drag to the left, then try to scroll to the beginning of the pipeline, it won't let you scroll to the beginning of the pipeline (you will have to drag to see it again). I think this has to do with the way CSS transforms work but it's a relatively small edge case (I think)

# Tests

Go to `samples/[sample id]/pipeline_viz` to see it work!


cdebourcy:
Nice!

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -720,6 +720,16 @@ class PipelineViz extends React.Component {
     );
   }
 
+  handleZoom = isIn => {
+    return () => {
+      if (this.panZoomContainer && this.panZoomContainer.current) {
+        isIn
+          ? this.panZoomContainer.current.zoomIn(3)
```
tfrcarvalho: Make zoom speed into a prop?

package.json
```diff
@@ -106,7 +106,7 @@
     "react-autocomplete": "^1.7.2",
     "react-dom": ">=16.4.2",
     "react-dropzone": "^6.2.4",
-    "react-easy-panzoom": "^0.3.0",
+    "react-easy-panzoom": "^0.4.2",
```
tfrcarvalho: 👍

PR #2478, opened 2019-08-05
Change columns to NOT NULL

gregdingle:
# Description

This uses the output of https://github.com/chanzuckerberg/idseq-web/pull/2475 to change columns that have only non-NULL values to accept only non-NULL values. This will reduce the complexity of the app and help downstream programmers. 

See also https://czi.quip.com/LWbJAyLuX7fn/2019-08-02-NULL-values-in-database . 

# Tests

run `rails db:migrate`, see no errors


db/migrate/20190805170949_change_columns_to_not_null.rb
```diff
@@ -0,0 +1,121 @@
+class ChangeColumnsToNotNull < ActiveRecord::Migration[5.1]
+  def change
+    change_column_null :pipeline_run_stages, :pipeline_run_id, false
```
cdebourcy: For my own education, is this going to cause problems because create_run_stages happens before_create for new pipeline_runs?

https://github.com/chanzuckerberg/idseq-web/blob/3e72558c8538c43fbe7681ba3e9a885e79b62c2b/app/models/pipeline_run.rb#L191

db/migrate/20190805170949_change_columns_to_not_null.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    change_column_null :metadata_fields, :display_name, false
+    change_column_null :metadata_fields, :force_options, false
+    change_column_null :metadata_fields, :is_core, false
+    change_column_null :metadata_fields, :is_default, false
+    change_column_null :metadata_fields, :is_required, false
+    change_column_null :metadata_fields, :default_for_new_host_genome, false
+
+    change_column_null :output_states, :output, false
+    change_column_null :output_states, :pipeline_run_id, false
```
cdebourcy: Same question about `create_output_states` happening `before_create` of `pipeline_run`

db/migrate/20190805170949_change_columns_to_not_null.rb
```diff
@@ -0,0 +1,121 @@
...
+    change_column_null :metadata, :sample_id, false
+    change_column_null :metadata, :metadata_field_id, false
+
+    change_column_null :phylo_trees, :taxid, false
+    change_column_null :phylo_trees, :tax_level, false
+    change_column_null :phylo_trees, :tax_name, false
+    change_column_null :phylo_trees, :user_id, false
+    change_column_null :phylo_trees, :project_id, false
+    change_column_null :phylo_trees, :status, false
+    change_column_null :phylo_trees, :dag_json, false
```
cdebourcy: This is going to break phylo_tree creation: phylo_trees are always created with dag_json being NULL (https://github.com/chanzuckerberg/idseq-web/blob/9716a80a847ed62a182b3a283146ad81884dac69/app/controllers/phylo_trees_controller.rb#L229). The dag_json column only gets populated once the phylo_tree pipeline is kicked off: https://github.com/chanzuckerberg/idseq-web/blob/9716a80a847ed62a182b3a283146ad81884dac69/app/models/phylo_tree.rb#L255

db/migrate/20190805170949_change_columns_to_not_null.rb
```diff
@@ -0,0 +1,121 @@
...
+    change_column_null :phylo_trees, :command_stdout, false
```
cdebourcy: same comment for command_stdout, command_stderr, job_id

cdebourcy:
> This needs some more thought -- just because retrospective analysis shows that a column is populated _eventually_ doesn't mean it cannot be null transiently. See comments for examples of parts of the app that are going to break if you require non-null.

Given the above, the test plan needs to be more comprehensive than "run rails db:migrate, see no errors" :) We need to test that the functionality related to all the models that are being touched still works.

gregdingle:
I agree with all the comments so far. This is a WIP. Sorry, I didn't add that before. I didn't expect a review so quickly without adding reviewers 😂 

I created a quip doc here so that we could comment more easily on each column with some stats in view:  https://czi.quip.com/LWbJAyLuX7fn/2019-08-02-NULL-values-in-database .

cdebourcy:
> I agree with all the comments so far. This is a WIP. Sorry, I didn't add that before. I didn't expect a review so quickly without adding reviewers 😂
> 
> I created a quip doc here so that we could comment more easily on each column with some stats in view: https://czi.quip.com/LWbJAyLuX7fn/2019-08-02-NULL-values-in-database .

Cool!

PR #2479, opened 2019-08-05
[AMR Heatmap] Show message indicating that samples have no amr data if none are found

lvreynoso:
# Description

Instead of showing a blank page, display the message "No Antimicrobial Resistance data for selected samples." if selected samples do not have any `amr_counts` associated with them, or if they are empty.

Also updates the header to emphasize "Antimicrobial Resistance Heatmap".

# Tests

* No server side code changed
* For client side code, go to All Data, pick a project with samples that have no AMR data associated with them, create an AMR Heatmap with them, and verify that the above message displays.
Screen Shot 2019-08-08 at 12 01 46 PM


app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -124,6 +124,17 @@ export default class AMRHeatmapView extends React.Component {
     return maxValues;
   }
 
+  hasDataToDisplay(maxValues) {
+    let hasData = false;
+    Object.entries(maxValues).forEach(metric => {
```
tfrcarvalho: Not sure what is happenings here. I believe Object.entries is used for objects... 
Is maxValues an Object? If so you should use `keys()` or `values()` to be more clear.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -286,6 +295,7 @@ export default class AMRHeatmapView extends React.Component {
       samplesWithAMRCounts,
       selectedOptions,
       samplesMetadataTypes,
+      maxValues,
```
tfrcarvalho: You should initialize maxValues at the beginning of the components.
Helps with readability and understand variables used through out the code.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -247,11 +258,9 @@ export default class AMRHeatmapView extends React.Component {
       <ViewHeader className={cs.viewHeader}>
         <ViewHeader.Content>
           <ViewHeader.Pretitle>
-            Antimicrobial Resistance Heatmap
+            {`Comparing ${sampleIds ? sampleIds.length : ""} Samples`}
```
MarkAZhang: Curious about this change since the other heatmap has the two labels flipped (Comparing X samples is the title). Was this based on a design discussion? Should we flip the one in SamplesHeatmapHeader as well?

lvreynoso: Yes, this was a request from Jenn @happyimadesignr . I think she would like to flip the other one as well, but I would double check with her.

happyimadesignr: Yes! Would love to switch it on both heatmaps, the large text across the product should be the title of the page. 

A later thing: I'm also trying to make the small text into breadcrumb across the board too, like we're doing for pipeline viz:
Screen Shot 2019-08-08 at 5 05 10 PM

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -124,6 +124,17 @@ export default class AMRHeatmapView extends React.Component {
...
 
+  hasDataToDisplay(maxValues) {
+    let hasData = false;
+    Object.entries(maxValues).forEach(metric => {
+      const value = metric[1];
+      if (value > 0) {
+        hasData = true;
+      }
+    });
+    return hasData;
```
MarkAZhang: I think you can do `return some(maxValue => maxValue > 0, maxValues)`

But this implicitly assumes that all metrics only allow for positive numbers. This is true right now, but if for some reason we allow a metric with negative values in the future, a bug can sneak in that's hard to catch at that point.

Would `return some(sample => size(sample.amrCount) > 0, samplesWithAmrCounts)` work instead?

This checks if there's at least one sample that has a non-empty amrCount object (more than zero keys in the object).

(granted, for negative numbers the Legend would probably also break)

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -294,6 +304,12 @@ export default class AMRHeatmapView extends React.Component {
           Loading...
         </p>
       );
+    } else if (!this.hasDataToDisplay(maxValues)) {
+      return (
+        <p className={cs.loadingIndicator}>
```
MarkAZhang: Even though they have the same styling, can we do `cs.noDataMsg` here instead?

PR #862, opened 2018-03-09
resume from a failed stage and see report when ready

yunfangjuan:
# Description

WIP
1. Make a new status RESUME_PR to reuse the existing pipeline run 
2. Make report viewable explicit after ready step 
## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix a feature that knowingly causes existing functionality to not work as expected)

# Test and Reproduce
Outline the steps to test or reproduce the PR here.

1. 

## Impacted Areas in Application
List general components of the application that this PR will affect:

Which pages?

- [ ] Login Screen
- [ ] Single Upload Page
- [ ] Batch Upload Page
- [ ] All Projects Page
- [ ] Sample Details Page
- [ ] Report Page

Specific components 

* 

# Testing Script:

###Login Page
* Successfully login to the application and get taken to the project page

###Single Upload
* Using an AWS path that doesn't exist throws an error
* Succesfully uploaded files from a folder (use test host)
* Uploading a duplicate file throws an error
* Pipeline gets initiated successfully 

###Batch Upload
* Using an AWS path that doesn't exist throws an error
* Succesfully uploaded files from a folder (use test host)
* Navigate to batch upload settings page after upload 
* Host selection from previous page still selected 
* Uploading a duplicate file throws a warning
* Pipeline gets initiated successfully 

###All Project Page
* Switch between projects 
* Search for a sample
* Navigate through different pages of samples
* Select Upload new sample button
* Select sample to see generated report

###Sample Details Page
* Information is consistent from Project Page
* Notes are editable and save successfully 
* Can navigate to report page 

###Report Page
* Data loads succesfully 
* Collapse and Uncollapse button works
* Clicking on the taxonomy name gives you total reads for that taxon
* Category filter works
* Genus Search works
* Filters work as expected
* Disabling Filters works
* Filtering by column works 

# Checklist:

- [ ] I have run through the testing script to make sure current functionality is unchanged
- [ ] I have done relevant tests that prove my fix is effective or that my feature works
- [ ] I have spent time testing out edge cases for my feature
- [ ] I have updated the test script or pull request template if necessary
- [ ] New and existing unit tests pass locally with my changes

PR #2480, opened 2019-08-05
Drop deprecated columns

gregdingle:
# Description

Following the analysis of NULLs and last created_at detailed in https://czi.quip.com/LWbJAyLuX7fn/2019-08-02-NULL-values-in-database, this drops columns that are effectively dead. 

# Notes

* See also https://github.com/chanzuckerberg/idseq-web/pull/2478 and https://github.com/chanzuckerberg/idseq-web/pull/2475

# Tests

`rails db:migrate`, see no errors
`rails t && rspec` 

search codebase for references to removed columns


db/migrate/20190805204523_drop_deprecated_columns.rb
```diff
@@ -0,0 +1,47 @@
...
+    remove_column :taxon_counts, :species_total_concordant
+    remove_column :taxon_counts, :genus_total_concordant
+    remove_column :taxon_counts, :family_total_concordant
+
+    remove_columns :backgrounds, :project_id
+
+    remove_columns :job_stats, :reads_before
+
+    # TODO: (gdingle): ??
+    remove_column :metadata, :specificity
```
jshoe: `:metadata, :specificity`: this was a "nice to have" column in the schema but we never ended up using it so I would support removing it 👍

db/migrate/20190805204523_drop_deprecated_columns.rb
```diff
@@ -0,0 +1,47 @@
+class DropDeprecatedColumns < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :pipeline_run_stages, :failed_jobs
+
+    remove_column :host_genomes, :skip_deutero_filter
+
+    remove_column :metadata_fields, :validation_type
```
jshoe: - `validation_type` was a nice-to-have that we haven't used yet.. could remove 👍

db/migrate/20190805204523_drop_deprecated_columns.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    remove_column :samples, :subsample
+    remove_column :samples, :sample_detection
+    remove_column :samples, :max_input_fragments
+    # TODO: (gdingle): after data migration drop sample_organism, sample_*
+
+    remove_column :taxon_byteranges, :tax_level
+
+    # TODO: (gdingle): ???
+    remove_column :taxon_scoring_models, :user_id
```
jshoe: `:taxon_scoring_models, :user_id` I believe this was with an eye towards user customization of the aggregate scoring model. YF started this work but I don't see this being further developed anytime soon so could remove

db/migrate/20190805204523_drop_deprecated_columns.rb
```diff
@@ -0,0 +1,47 @@
...
+    remove_column :pipeline_runs, :command_stdout
+    remove_column :pipeline_runs, :command_error
+    remove_column :pipeline_runs, :command_status
+    remove_column :pipeline_runs, :job_description
+    remove_column :pipeline_runs, :job_log_id
+    remove_column :pipeline_runs, :postprocess_status
+    remove_column :pipeline_runs, :assembled_taxids
+
+    # TODO: (gdingle): check the 2% where this is present
+    remove_column :taxon_counts, :percent_concordant
```
jshoe: not sure about these *_concordant ones

gregdingle: @katrinakalantar @cdebourcy  these concordance values have not set in `taxon_counts` since `2018-03-15`. Do you know what happened and whether we should bring them back?

cdebourcy: percent_concordant used to be a value in the report, but we started hiding it (https://github.com/chanzuckerberg/idseq-web/blob/0b9078947e757302ccbd8c746dc9c3f3da2b429d/app/assets/src/components/PipelineSampleReport.jsx#L129), and now it's no longer being computed in the backend either. I believe the column can be removed from the taxon_counts table, but in that case all references to it need to be removed throughout the codebase:

image

db/migrate/20190805204523_drop_deprecated_columns.rb
```diff
@@ -0,0 +1,47 @@
+class DropDeprecatedColumns < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :pipeline_run_stages, :failed_jobs
```
gregdingle: @cdebourcy this blames to your PR #560. the col was last updated 2018-06-24.

db/migrate/20190805204523_drop_deprecated_columns.rb
cdebourcy: > This list looks right to me (unless otherwise commented). I think the column values speak for themselves and :failed_jobs, :skip_deutero_filter, etc. are surely legacy fields..

I disagree. `skip_deutero_filter` is a column we need. It tells you something about the host genome. For mammals, deuterostome filtering should be performed. For mosquitoes, it should not be performed, because mosquitoes are not deuterostomes. That's why Mosquito host is currently the only one that has skip_deutero_filter == 1. This value is being actively used in the pipeline.

`failed_jobs` is being used in the codebase: https://github.com/chanzuckerberg/idseq-web/blob/cd4311e3b3d1e42eacdc267c261f84331ad8cc56/app/models/pipeline_run_stage.rb#L145
It's possible that that code path is broken. In that case, we first need to get to the bottom of that behavior before we can remove the column.

cdebourcy:
> This list looks right to me (unless otherwise commented). I think the column values speak for themselves and :failed_jobs, :skip_deutero_filter, etc. are surely legacy fields..

I disagree. `skip_deutero_filter` is a column we need. It tells you something about the host genome. For mammals, deuterostome filtering should be performed. For mosquitoes, it should not be performed, because mosquitoes are not deuterostomes. That's why Mosquito host is currently the only one that has skip_deutero_filter == 1. This value is being actively used in the pipeline.

`failed_jobs` is being used in the codebase: https://github.com/chanzuckerberg/idseq-web/blob/cd4311e3b3d1e42eacdc267c261f84331ad8cc56/app/models/pipeline_run_stage.rb#L145
It's possible that that code path is broken. In that case, we first need to get to the bottom of that behavior before we can remove the column.

db/migrate/20190805204523_drop_deprecated_columns.rb
cdebourcy: It's possible that the code path is broken. The column is "supposed" to be in use, so we would need to investigate that first

cdebourcy: see https://github.com/chanzuckerberg/idseq-web/blob/cd4311e3b3d1e42eacdc267c261f84331ad8cc56/app/models/pipeline_run_stage.rb#L145
and
https://github.com/chanzuckerberg/idseq-web/blob/cd4311e3b3d1e42eacdc267c261f84331ad8cc56/app/models/pipeline_run_stage.rb#L185

cdebourcy: Upon further thought, I think the reason this code path is not being used anymore is because we no longer is spot instances for the Batch jobs. Now we use on-demand instances. If for whatever reason (e.g. cost) we decided to go back to spot instances, this code would ensure that terminated Batch jobs get resubmitted (just the failed stage, unlike the more recent "retry from top"). So I'm in favor of keeping it

gregdingle: Sounds good. I'll add the comment into the code.

cdebourcy: Thank you! I agree it's good to have that context in the code

cdebourcy:
> `skip_deutero_filter` is a column we need. It tells you something about the host genome. For mammals, deuterostome filtering should be performed. For mosquitoes, it should not be performed, because mosquitoes are not deuterostomes. That's why Mosquito host is currently the only one that has skip_deutero_filter == 1. This value is being actively used in the pipeline.

I'm adding appropriate instructions to the wiki on making new host genomes. Whenever we had a new host genome, we need to check if it's a deuterostome or not and set `skip_deuterostome_filter` accordingly.

gregdingle:
> Whenever we had a new host genome, we need to check if it's a deuterostome or not and set skip_deuterostome_filter accordingly.

@cdebourcy It appears that only mosquitoes have this set currently. Should we create a task to update existing host genomes?

cdebourcy:
> > Whenever we had a new host genome, we need to check if it's a deuterostome or not and set skip_deuterostome_filter accordingly.
> 
> @cdebourcy It appears that only mosquitoes have this set currently. Should we create a task to update existing host genomes?

Working on checking existing ones!

db/migrate/20190805204523_drop_deprecated_columns.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    remove_column :host_genomes, :skip_deutero_filter
+
+    remove_column :metadata_fields, :validation_type
+
+    remove_column :pipeline_runs, :version
```
gregdingle: @boris-dimitrov, the `version` last blames to #1189, and this col has not been set since 2018-07-09... should we drop it?

db/migrate/20190805204523_drop_deprecated_columns.rb
```diff
@@ -0,0 +1,47 @@
...
+    remove_column :pipeline_runs, :ready_step
```
gregdingle: @boris-dimitrov , similarly, this blame to #862 which you reviewed. the col was last updated 2018-06-09. should we drop it?

cdebourcy: I'll let Boris comment, but regardless of the values, removing columns will cause pages to break if the ruby code references them, even if the column no longer serves a real purpose, e.g. https://github.com/chanzuckerberg/idseq-web/blob/df35dc7ec2c2ef2e84ade08efd4d096ae005b2d4/app/helpers/samples_helper.rb#L253

So I recommend following your investigation on the NULLs up with a comprehensive evaluation of all the code paths in which the column is referenced

db/migrate/20190805204523_drop_deprecated_columns.rb
```diff
@@ -0,0 +1,47 @@
...
+    remove_column :taxon_counts, :family_total_concordant
+
+    remove_columns :backgrounds, :project_id
+
+    remove_columns :job_stats, :reads_before
+
+    # TODO: (gdingle): ??
+    remove_column :metadata, :specificity
+
+    remove_column :samples, :subsample
```
gregdingle: @cdebourcy, see #485. The last time this was set was 2018-07-18.

db/migrate/20190805204523_drop_deprecated_columns.rb
boris-dimitrov: I don't recall precisely how the design has evolved since the time I wrote this in April 2018.     Nevertheless, the code line
```
```
is what used to set this.   If that line is no longer executed, then you can probably remove that columns.   I am sure you will modify any references to it in the codebase to assume NULL.

db/migrate/20190805204523_drop_deprecated_columns.rb
boris-dimitrov: I would ask @davidrissato comment since, if I recall correctly, David recently removed the functionality to rerun from the failed stage, and now everything reruns from top.   Without this functionality, there may be no need for this particular column.  This doesn't mean, however, that the column is safe to remove.  It's entirely possible the state machine logic is needlessly entangled with that column in a way that could be time consuming to disentangle.

In general, for columns pertaining to the execution state machine, my suggestion would be to refrain from making any changes to this needlessly complex piece of machinery until such a time as someone can carefully examine, simplify, and re-test it all.   I do believe that would be a valuable project.

gregdingle: > no need for this particular column

So this column was only intended for the "rerun stage" feature of admins? Then I think we should remove it assuming all references to it are deleted.

cdebourcy: I believe (at least part of) the intent was to mark from which point onwards a report could be shown. This used to be after stage 2. Now it's after stage 3. It was like a flag that could be passed to the frontend to decide between "not ready" and "ready" views.

gregdingle: @cdebourcy  Do you think we should drop the column?

db/migrate/20190805204523_drop_deprecated_columns.rb
gregdingle: I replaced all callsites of `version` with `pipeline_version`.

db/migrate/20190805204523_drop_deprecated_columns.rb
gregdingle: thanks, I removed all the concordance.

db/migrate/20190805204523_drop_deprecated_columns.rb
jshoe: @gregdingle I'm fine with dropping `ready_step` (and removing references to it) since it's easy enough to add back if someone wants to use it

jshoe:
^ Please also run samples on staging and spot check the results :)

cdebourcy:
Another suggestion: as you click around staging, give some attention to the oldest samples as well as issues are more likely to pop up there :)

gregdingle:
The benchmark succeeded in staging and the I didn't see any errors clicking around a new sample or old one (more than a year) . 

https://staging.idseq.net/samples/4524?pipeline_version=3.3&background_id=26&search_taxon_id=0&rows_passing_filters=2966&rows_total=3019&pagesRendered=1&sort_by=nt_aggregatescore&includedCategories[]=Bacteria&includedCategories[]=Viruses&includedCategories[]=Eukaryota&rendering=false&loading=false&countType=NT&readSpecificity=0&treeMetric=aggregatescore&phyloTreeModalOpen=true&minContigSize=4&hoverRowId=173784

https://staging.idseq.net/samples/13012?pipeline_version=3.9&background_id=26&search_taxon_id=0&rows_passing_filters=1945&rows_total=1998&pagesRendered=1&sort_by=nt_aggregatescore&includedCategories[]=Bacteria&includedCategories[]=Viruses&includedCategories[]=Eukaryota&rendering=false&loading=false&countType=NT&readSpecificity=0&treeMetric=aggregatescore&phyloTreeModalOpen=true&minContigSize=4&hoverRowId=null

PR #2482, opened 2019-08-07
[Maps] Improve API rate limit errors

jshoe:
### Description
- This adjusts the error handling for the geosearch API. I didn't properly distinguish between API rate limit errors and "0 results" HTTPNotFound errors before.
- For the sake of accurate alerting and monitoring if users ever encounter rate limits.

### Tests
- See tests

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -19,14 +19,14 @@ class LocationsController < ApplicationController
     if query.present?
       success, resp = Location.geosearch(query, limit)
       if success && resp.is_a?(Array)
+        # Successful request with results. Just keep the first if you get duplicate locations.
         results = resp.map { |r| LocationHelper.adapt_location_iq_response(r) }
-        # Just keep the first if you get duplicate locations
         results = results.uniq { |r| [r[:name], r[:geo_level]] }
-      elsif success && resp.is_a?(Hash) && resp["error"].present? && resp["error"] == "Unable to geocode"
-        # Successful but 0 results
+      elsif success && resp.is_a?(Hash) && resp["error"] == "Unable to geocode"
+        # Successful request but 0 results
         Rails.logger.info("No geosearch results for: #{query}")
       else
-        # Likely Net::HTTPTooManyRequests. Monitor if users run up against geosearch API rate limits.
+        # Unsuccessful request. Likely Net::HTTPTooManyRequests. Monitor if users run up against geosearch API rate limits.
         msg = "Geosearch failed. Check API rate limits."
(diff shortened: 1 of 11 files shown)


app/controllers/locations_controller.rb
```diff
@@ -18,13 +19,17 @@ def external_search
...
-      if success
+      if success && resp.is_a?(Array)
+        # Successful request with results. Just keep the first if you get duplicate locations.
         results = resp.map { |r| LocationHelper.adapt_location_iq_response(r) }
-        # Just keep the first if you get duplicate locations
         results = results.uniq { |r| [r[:name], r[:geo_level]] }
+      elsif success && resp.is_a?(Hash) && resp["error"] == "Unable to geocode"
+        # Successful request but 0 results
+        Rails.logger.info("No geosearch results for: #{query}")
       else
```
tfrcarvalho: Is there no error message from the API? Nothing more specific we can log?

app/models/location.rb
```diff
@@ -46,7 +46,9 @@ def self.location_api_request(endpoint_query)
     resp = Net::HTTP.start(uri.host, uri.port, use_ssl: true) do |http|
       http.request(request)
     end
-    [resp.is_a?(Net::HTTPSuccess), JSON.parse(resp.body)]
+    # Search with 0 results will return HTTPNotFound. Consider it a successful request for our handling.
+    success = resp.is_a?(Net::HTTPSuccess) || resp.is_a?(Net::HTTPNotFound)
```
tfrcarvalho: Could HTTPNotFound be an error on our part ? Or if for some reason something changes on our config?

app/controllers/locations_controller.rb
jshoe: the full response is actually `{ "error": "Unable to geocode" }`  :3

app/models/location.rb
jshoe: sounds like it's possible yeah. what would you suggest for this?

app/controllers/locations_controller.rb
tfrcarvalho: I meant if there arent other errors... Can you guarantee that `success && resp.is_a?(Hash) && resp["error"] != "Unable to geocode"` does not happen?

jshoe: updated as discussed 👍

app/models/location.rb
tfrcarvalho: So, with your changes above any other errors will be logged, right?

jshoe: right, they will fall through and go to log and Airbrake

PR #2484, opened 2019-08-08
[Pipeline Viz] Visual tweaks + Front end updates

emilyzhong:
# Description

**Visual tweaks:**
* Add some space to the left of the stage node (8px L/R padding)
* Hiding the icon space when no icon on stage node
* Space after list item on sidebar
* Add link to pipeline version @ top of header
* Make “See/ View Pipeline Viz” copy consistent
* Change header breadcrumb “>” to an icon (currently a character)
* Lighten step nodes that haven’t run
* Feature flag for experimental
* Shorter edges between stages

**Front-end changes**
Differentiated between user errors and pipeline errors. User errors look like this:
image

Added status box on sidebar for completed steps, which looks like this:
image

Added resources section in sidebar, if resources available
image

# Tests

Go to `samples/[sample id]/pipeline_viz` to see changes on a pipeline viz enabled account


app/assets/src/components/views/PipelineViz/pipeline_viz.scss
```diff
@@ -116,38 +126,43 @@
...
   .inProgressIcon {
     color: $primary-light;
-    margin: 0;
     display: flex;
     align-items: center;
     justify-content: center;
     font-size: 15px;
   }
-  .erroredIcon {
+  .pipelineErroredIcon {
```
tfrcarvalho: nit: spaces between the selectors

app/assets/src/components/views/SampleView/SampleView.jsx
```diff
@@ -561,12 +561,25 @@ class SampleView extends React.Component {
...
             <ViewHeader.Content>
-              <div className={cs.pipelineInfo}>
+              <div
+                className={cx(
+                  cs.pipelineInfo,
+                  pipelineVersion && cs.linkToPipelineViz
+                )}
+                onClick={() =>
+                  pipelineVersion &&
+                  window.open(
```
tfrcarvalho: Why not use an anchor tag on the version display below?
Does this interfere with the select box?

emilyzhong: Switched to an anchor tag — I think I just forgot that they existed yesterday 😅

emilyzhong: Actually, I switched back (I forgot to start npm when developing). The anchor tag interferes with the select box.

PR #2487, opened 2019-08-08
[AMR Heatmap] Add initial analytics

lvreynoso:
# Description

Adds analytics logging to AMR Heatmap components.

# Tests

* No server side code changed
* Go to Discovery View, select samples, create AMR Heatmap; play around with all the options, click on links; used `console.log` for sanity check


app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -295,6 +316,7 @@ export default class AMRHeatmapView extends React.Component {
         </p>
       );
     }
+    logAnalyticsEvent("AMRHeatmapView_heatmap_rendered");
```
gregdingle: normally we don't log passive events and page views... if the same event is captured by clicking an already instrumented submit button or by loading the page then it is not needed.

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -146,6 +161,11 @@ export default class GeneDetailsMode extends React.Component {
     const {
       ontology: { error },
     } = this.state;
+    const { geneName } = this.props;
```
MarkAZhang: We generally don't log renders. Unless it's truly a special case.

You can often deduce renders from page views (which we log elsewhere) and api calls.

One reason why logging renders is tricky is that with React, it's best practice to assume that a render function can be called any number of times. Which is bad here because it means your log function can be called a bunch.

It might be better here to do a `Rails.logger.error` in the `fetch_ontology` controller function.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
MarkAZhang: Same as comment above, with logging a render. I think we can infer the page views by searching for the matching http get requests for `/amr_heatmap`.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -54,6 +61,12 @@ export default class AMRHeatmapVis extends React.Component {
     ) {
       this.heatmap.updateColumnMetadata(this.getSelectedMetadataFields());
     }
+    if (nodeHoverTriggered !== prevState.nodeHoverTriggered) {
+      logAnalyticsEvent("AMRHeatmapVis_heatmap-node-hover_triggered");
```
MarkAZhang: `AMRHeatmapVis_heatmap-node_hovered` and same below.

Also, maybe it should be `state.nodeHovered` for consistency.

PR #2488, opened 2019-08-08
[AMR Heatmap] Add realtime metadata update

lvreynoso:
# Description

Updates the heatmap in realtime when sample metadata is edited in the SampleDetailsMode sidebar.

# Tests

* No server side code changed
* For client side, go to Discovery View, select some samples and make an AMR heatmap. Click on a sample name to open up the sample details sidebar. Edit a metadata field and watch the magic happen.


app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -41,16 +41,17 @@ export default class AMRHeatmapVis extends React.Component {
...
       this.heatmap === null
     ) {
       this.updateHeatmap();
     }
     if (
-      selectedMetadata !== prevState.selectedMetadata &&
-      this.heatmap !== null
+      this.heatmap !== null &&
+      (selectedMetadata !== prevState.selectedMetadata ||
+        lastDataUpdate !== prevProps.lastDataUpdate)
```
MarkAZhang: It's not React best-practice to pass a parameter that specifies whether another prop was updated. You should be able to tell by directly comparing the old and new props.

The code should read something like:
```
if (samplesWithAMRCounts !== prevProps.samplesWithAMRCounts) {
  // Re-process the data.
  this.processSampleData()
  // Update the heatmap with the new data.
  this.updateHeatmap()
}
```

If you try to do this code directly, you'll run into a gotcha though. The issue is that `processSampleData` does `setState({ sampleLabels })` and then `updateHeatmap` retrieves `sampleLabels` from `this.state`. This seems fine, but actually `setState` is asynchronous and doesn't run until after the current function is completely finished, so `updateHeatmap` will get the old version of `sampleLabels`. This is probably the worst gotcha of React. (congrats for building something sufficiently complex to encounter it!) 

You might be able to get away with doing a two-step thing in this case:
```
if (samplesWithAMRCounts !== prevProps.samplesWithAMRCounts) {
  // Re-process the data.
  this.processSampleData()
}

// This will then run in the NEXT componentDidUpdate.
if (sampleLabels !== prevState.sampleLabels || geneLabels !== prevState.geneLabels || alleleLabels !== prevState.geneLabels || alleleToGeneMap !== prevState.alleleToGeneMap) {
  // Update the heatmap with the new data.
  this.updateHeatmap()
}
```

Otherwise, you'd have to refactor the code a bit (we can talk this over in person if it comes to that).

It might seem like a lot of work for such a small change, but I think it's a sign that this code has gotten quite complex and might be due for a bit of re-organization.

You can also do `some(key => this.state[key] !== prevState[key], ["sampleLabels", "geneLabels", "alleleLabels", "alleleToGeneMap"])` if the above seems too repetitive.

lvreynoso: Thanks for pointing this out, I'll see what I can do!

lvreynoso: So I ended up just lifting the state to `AMRHeatmapView`, that way `AMRHeatmapVis` can just compare `sampleLabels` directly as a prop and update the heatmap if it changes. Not too bad.

lvreynoso:
Lifted the state from `AMRHeatmapVis` to `AMRHeatmapView` to solve the problem.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -327,6 +383,10 @@ export default class AMRHeatmapView extends React.Component {
         </p>
       );
     }
+    const [sampleLabels, geneLabels, alleleLabels] = this.extractLabels(
+      samplesWithAMRCounts
+    );
+    const alleleToGeneMap = this.mapAllelesToGenes(samplesWithAMRCounts);
```
MarkAZhang: Since you're calling these methods in the render function (which can potentially get called a lot), it might be good to memoize them. A memoized function will save the last value it returns, and if the next time it's called, the input parameters haven't changed, it will short-circuit and just return the last value again. 

We use `memoize-one` in`CoverageVizBottomSidebar` so you can see how it's used.

PR #2492, opened 2019-08-09
[Pipeline Viz] Node Layout

emilyzhong:
# Description

Places nodes in a better position so they match mocks better (while still being flexible to pipeline updates!)

image

# Tests

All front-end changes; view a sample's pipeline viz to see changes!

### DIFF
--- a/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
+++ b/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
@@ -406,19 +406,20 @@ class PipelineViz extends React.Component {
   }
 
   addPositionToNodes(nodeData, edgeData) {
-    const {
-      xLayoutInterval,
-      yLayoutInterval,
-      staggerLayoutMultiplier,
-    } = this.props;
+    const fromToToEdgeMap = groupBy("from", edgeData);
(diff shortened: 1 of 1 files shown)


app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -406,41 +405,76 @@ class PipelineViz extends React.Component {
     };
   }
 
-  addHierarchicalLevelsToNodes(nodeData, edgeData) {
+  addPositionToNodes(nodeData, edgeData) {
```
tfrcarvalho: Lets make this into props with default values.

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -406,41 +405,78 @@ class PipelineViz extends React.Component {
...
           }
-        }
-        bfs.push(toNodeId);
-      });
+          bfs.push(toNodeId);
+        });
+      }
     }
 
+    const maxLevel = Math.max(...Object.values(nodeToCurrentLevel));
```
tfrcarvalho: As discussed offline, please break this into 2.

PR #2495, opened 2019-08-10
[Pipeline Viz] Persist opened stages across reloads

emilyzhong:
# Description

Use `window.sessionStorage` to keep opened stages open when refreshing.
persist

# Tests

* Go to `samples/[sample id]/pipeline_viz` on a `pipeline_viz` enabled account, click on a couple of the stages to open/close them, then refresh.


app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -347,6 +347,11 @@ class PipelineViz extends React.Component {
     const updatedStagesOpened = [...this.state.stagesOpened];
     updatedStagesOpened[index] = !updatedStagesOpened[index];
     this.setState({ stagesOpened: updatedStagesOpened });
+
+    window.sessionStorage.setItem(
```
tfrcarvalho: We should probably use url for this (using `history.replaceState`)... allows sharing the state with other too. Wdyt?

PR #2497, opened 2019-08-12
Force index in heatmap query

gregdingle:
# Description

Heatmaps are still slow, taking 0.5s * n samples on initial render. A 50 sample heatmap takes 25s. The most expensive part is getting all the taxons for the samples, about 50% of the total time. `taxon_counts` is now more than 114M rows. 

After more investigation with @davidrissato, I discovered that we can get a 2x speed-up on a 50 sample query by forcing mysql to use an existing index. We don't know why mysql was not using the index automatically. There does not seem to ever be any disadvantage. 

# Tests

* Run modified query in mysql console, see 2x speedup
* Run heatmap locally, see same results

### DIFF
--- a/app/helpers/heatmap_helper.rb
+++ b/app/helpers/heatmap_helper.rb
@@ -217,6 +217,8 @@ module HeatmapHelper
       #{rpm_sql} AS rpm,
       #{zscore_sql} AS zscore
     FROM taxon_counts
+    -- Needed to avoid a full table scan
+    FORCE INDEX (index_pr_tax_hit_level_tc)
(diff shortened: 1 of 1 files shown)


davidrissato:
I'm still wondering why mysql query engine wasn't using this index. I saw your tests in Staging, but I still think we might need to optimize something else.
We should check this behavior in prod first, because it has a different number of lines.

gregdingle:
Thanks @davidrissato . I did check in prod and the index *is* being used already, so this PR will probably have no effect.

PR #2498, opened 2019-08-13
[Pipeline Viz] Fetch updates

emilyzhong:
# Description

Fetches graph information + visualizes changes every second to mimic a "live update" visualization

# Tests
* Upload a sample, then go to `samples/[sample id]/pipeline_viz` to see changes update live (may take a couple minutes, since the pipeline itself is inherently slow 😢 )

### DIFF
--- a/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
+++ b/app/assets/src/components/views/PipelineViz/PipelineViz.jsx
@@ -86,34 +86,36 @@ class PipelineViz extends React.Component {
     );
     const updates = diff(oldGraphData, newGraphData);
     if (updates.stages) {
-      Object.entries(updates.stages).forEach((stageIndexStr, stageChanges) => {
-        const stageIndex = parseInt(stageIndexStr);
+      Object.entries(updates.stages).forEach(
+        ([stageIndexStr, stageChanges]) => {
+          const stageIndex = parseInt(stageIndexStr);
 
-        if (stageIndex < this.graphs.length) {
-          // Graph already exists.
-          const graph = this.graphs[stageIndex];
-          const steps = stageChanges.steps;
-          Object.entries(steps).forEach((stepIndexStr, stepChanges) => {
-            const stepIndex = parseInt(stepIndexStr);
-            const status = stepChanges.status;
-            graph.updateNodes([stepIndex], {
-              group: status,
-              ...this.getNodeStatusOptions(status),
+          if (stageIndex < this.graphs.length) {
+            // Graph already exists.
+            const graph = this.graphs[stageIndex];
+            const steps = stageChanges.steps;
+            Object.entries(steps).forEach(([stepIndexStr, stepChanges]) => {
+              const stepIndex = parseInt(stepIndexStr);
+              const status = stepChanges.status;
+              graph.updateNodes([stepIndex], {
+                group: status,
+                ...this.getNodeStatusOptions(status),
+              });
             });
-          });
-        } else {
-          // A new stage has started and needs to be drawn.
-          if (stageIndex > 0) {
-            // Modify previous stage's graph to create edges to new stage
-            const prevGraph = this.graphs[stageIndex - 1];
-            this.generateEdgeData(stageIndex - 1).forEach(edge => {
-              const { id, ...options } = edge;
-              prevGraph.updateEdges([id], options);
-            });
+          } else {
+            // A new stage has started and needs to be drawn.
+            if (stageIndex > 0) {
+              // Modify previous stage's graph to create edges to new stage
+              const prevGraph = this.graphs[stageIndex - 1];
+              this.generateEdgeData(stageIndex - 1).forEach(edge => {
+                const { id, ...options } = edge;
+                prevGraph.updateEdges([id], options);
+              });
+            }
+            this.drawStageGraph(stageIndex);
           }
-          this.drawStageGraph(stageIndex);
         }
-      });
+      );
     }
   }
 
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -280,6 +309,31 @@
     t.index ["pipeline_run_id", "step_number"], name: "index_pipeline_run_stages_on_pipeline_run_id_and_step_number"
   end
 
+  create_table "pipeline_run_stages__deleted_rows", id: false, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
```
lvreynoso: What's up with all the changes in schema.rb? Were these automatically added or did you edit it by hand? I think all these changes should be in a migration, no?

emilyzhong: Oops I accidentally committed these — will remove

config/routes.rb
```diff
@@ -113,8 +113,8 @@
   get 'basespace/projects', to: 'basespace#projects'
   get 'basespace/samples_for_project', to: 'basespace#samples_for_project'
 
-  get '/samples/:sample_id/pipeline_viz(/:pipeline_version)', to: 'pipeline_viz#show',
-                                                              constraints: { pipeline_version: /\d+\.\d+/ } # To allow period in pipeline version parameter
+  get 'samples/:sample_id/pipeline_viz(/:pipeline_version)', to: 'pipeline_viz#show',
+                                                             constraints: { pipeline_version: /\d+\.\d+/ } # To allow period in pipeline version parameter
```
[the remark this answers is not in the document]
emilyzhong: Hm I'm pretty sure the pipeline is only ever stored as the first two numbers on `idseq-web`, since updates to the third number should indicate no changes to the results of `idseq-dag`, + we use the pipeline_version parameter here to construct part of the s3 path, which currently is only the first two numbers given the no changes to results thing above.

Not sure if this is something to guard against since it doesn't seem like it's a possibility (?)

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -48,29 +50,98 @@ class PipelineViz extends React.Component {
...
+            this.generateEdgeData(stageIndex - 1).forEach(edge => {
+              const { id, ...options } = edge;
+              prevGraph.updateEdges([id], options);
+            });
+          }
+          this.drawStageGraph(stageIndex);
+        }
+      });
+
+      if (this.pipelineIsFinished()) {
```
tfrcarvalho: You should add this as callback of `setState`. This would probably be called before the state is updated with the new graph.

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -48,29 +50,98 @@ class PipelineViz extends React.Component {
...
+      graphData: props.graphData,
     };
   }
 
   componentDidMount() {
     this.drawGraphs();
     window.addEventListener("resize", this.handleWindowResize);
+    // Only set update loop if the pipeline is still running.
+    if (!this.pipelineIsFinished()) {
+      this.updateLoop = setInterval(() => this.updateGraphs(), 1000);
```
tfrcarvalho: 1 second seems excessive. 1 minute seems enough given pipeline speed.
Also, can you add this to constant or default props.

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -48,29 +50,98 @@ class PipelineViz extends React.Component {
...
+        }
+      });
+
+      if (this.pipelineIsFinished()) {
+        clearInterval(this.updateLoop);
+      }
+    }
+  }
+
+  pipelineIsFinished() {
```
tfrcarvalho: What do you think of having a pipeline status from the server? 
It is just moving the logic, but seems a good information to have.

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -48,29 +50,95 @@ class PipelineViz extends React.Component {
...
+    );
+    this.setState(
+      {
+        graphData: newGraphData,
+      },
+      () => this.pipelineIsFinished() && clearInterval(this.updateLoop)
+    );
+    const updates = diff(oldGraphData, newGraphData);
+    if (updates.stages) {
+      Object.keys(updates.stages).forEach(stageIndexStr => {
```
tfrcarvalho: Why is the stageIndex a str?

tfrcarvalho: Can you use entries to loop through keys and values at the same time?>

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
```diff
@@ -48,29 +50,95 @@ class PipelineViz extends React.Component {
...
+      });
+    }
+  }
+
+  pipelineIsFinished() {
+    const {
+      graphData: { status },
+    } = this.state;
+    return (
+      status == "finished" ||
```
tfrcarvalho: You could use `lodash/fp`'s includes to check this and avoid the multiple iff.

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
emilyzhong: The diff library I'm using stores changes in arrays as objects where the key is the string of the index https://www.npmjs.com/package/deep-object-diff

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
emilyzhong: I ended up using JavaScript's own `includes` method to reduce imports

app/assets/src/components/views/PipelineViz/PipelineViz.jsx
tfrcarvalho: blah... but ok

PR #2501, opened 2019-08-13
Fix loading amr counts for empty result values

lvreynoso:
# Description

Prod currently errors on loading empty results for amr counts, which are ~1 byte CSV files. This fixes it by only trying to read the CSV if the file is of a reasonable size for non empty results.

# Notes

Part of my change for updated results

# Tests

* Should write an automated Rspec test for this
Screen Shot 2019-08-13 at 10 31 41 AM

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -553,8 +553,13 @@ class PipelineRun < ApplicationRecord
 
   def db_load_amr_counts
     amr_results = PipelineRun.download_file(s3_file_for("amr_counts"), local_amr_full_results_path)
-    unless File.zero?(amr_results)
-      amr_counts_array = []
+    if amr_results.nil?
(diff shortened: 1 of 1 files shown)


app/models/pipeline_run.rb
```diff
@@ -553,8 +553,13 @@ def db_load_contigs(contig2taxid)
 
   def db_load_amr_counts
     amr_results = PipelineRun.download_file(s3_file_for("amr_counts"), local_amr_full_results_path)
-    unless File.zero?(amr_results)
-      amr_counts_array = []
+    if amr_results.nil?
+      Rails.logger.error("No AMR results file found for PipelineRun ##{id}. Is the pipeline okay?")
```
jshoe: nit: I would remove the `Is the pipeline okay?` probably

PR #2504, opened 2019-08-13
[WIP] Remove slow query of taxon_counts

gregdingle:
# Description

This removes two methods in `HeatmapHelper` to achieve a 5s to 10s speedup in a 50 sample heatmap. 

The speedup is mostly from removing the SQL to `taxon_counts`. It's not clear why the second query followed the first. It refetched from the same tables using the taxon ID results of the first query. 

Similarly, there were two similar methods that transformed results after each SQL query. As the second one returns more info and was already the final transformation before returning to the client, I chose to keep it. 

# Notes

* This branch is based on my `not-nulls` migration because it was a nightmare to develop otherwise

* I had to change the values of some of the unit test assertions but it's never been clear what should be the correct values

* The speedup is based on local execution, but I expect similar speedup  in prod by measuring the execution time of the SQL in isolation

* Working with large heatmaps locally has been painful because they regularly lock up my docker container or chrome browser or both. Sadly I must predict consistent pain for users if they do 50+ samples despite the improvements of this PR. 

# Tests

* `rails t`
* turn off caching, open a few heatmaps locally and see no difference before and after

image

image

image

image


app/helpers/heatmap_helper.rb
```diff
@@ -216,7 +150,7 @@ def self.fetch_top_taxons(
       #{rpm_sql} AS rpm,
       #{zscore_sql} AS zscore
     FROM taxon_counts
-    LEFT OUTER JOIN pipeline_runs pr ON pipeline_run_id = pr.id
+    JOIN pipeline_runs pr ON pipeline_run_id = pr.id
```
gregdingle: I thought this was needed earlier but now I know that the pipeline runs should exist

app/helpers/heatmap_helper.rb
```diff
@@ -267,38 +203,33 @@ def self.fetch_top_taxons(
   def self.samples_taxons_details(
     results_by_pr,
     samples,
-    taxon_ids,
```
gregdingle: filter is done upstream

app/helpers/heatmap_helper.rb
```diff
@@ -346,7 +277,7 @@ def self.fetch_samples_taxons_counts(samples, taxon_ids, parent_ids, background_
           taxon_counts.e_value IS NOT NULL,
           (0.0 - taxon_counts.e_value),
           #{ReportHelper::DEFAULT_SAMPLE_NEGLOGEVALUE}
-        )                                AS  neglogevalue,
```
gregdingle: repeat of fix by @MarkAZhang

app/helpers/heatmap_helper.rb
tfrcarvalho: This is guaranteed by the foreign keys?

gregdingle: No this is not guaranteed by the foreign keys because we explicitly decided to exclude `taxon_counts` but I know now that we don't intentionally delete pipeline_runs and leave behind their taxon_counts.

app/helpers/heatmap_helper.rb
```diff
@@ -267,38 +203,33 @@ def self.fetch_top_taxons(
...
+    results_by_pr.each do |_pr_id, res|
+      pr = res["pr"]
+      taxon_counts = res["taxon_counts"]
+      sample_id = pr.sample_id
+      tax_2d = ReportHelper.taxon_counts_cleanup(taxon_counts)
+      HeatmapHelper.only_species_or_genus_counts!(tax_2d, species_selected)
+
+      rows = []
+      tax_2d.each { |_tax_id, tax_info| rows << tax_info }
+      HeatmapHelper.compute_aggregate_scores_v2!(rows)
```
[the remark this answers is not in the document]
gregdingle: will do

gregdingle:
> Was this running the similar code for clean up and filter twice?

Yes... it looks like the code was copied and pasted at some point then modified slightly. 

> How do you feel about the coverage of tests you added last time?

I'm confident that the different code paths are exercised but I'm not confident that the results are "correct" because it's not clear what "correct" is. Beyond the unit tests, I always tried to make sure the heatmap result was visually the same.

gregdingle:
By the way I always get a ton of messages
```
Missing taxon_counts for genus ids #{missing_genera.to_a} corresponding to taxon ids #{taxids_with_missing_genera.to_a}.
```

I'm not sure if that's a side effect of working on the local data set or what.

app/helpers/heatmap_helper.rb
tfrcarvalho: In case of human/bug error, I guess the effect here would be to remove all taxon counts that do not have a pipeline run, which seems fine. We should probably report any such findings somehow, no?

Can replace JOIN by INNER JOIN ? I had to look it up because I was not fully sure which one was the default :)

tfrcarvalho:
> By the way I always get a ton of messages
> 
> ```
> Missing taxon_counts for genus ids #{missing_genera.to_a} corresponding to taxon ids #{taxids_with_missing_genera.to_a}.
> ```
> 
> I'm not sure if that's a side effect of working on the local data set or what.

I remember seeing something like this, but do not remember why. We should revisit the cause... Could be worse for being local as you say. Are you working with an old pipeline run version?

app/helpers/heatmap_helper.rb
gregdingle: Yes, sure thing

gregdingle:
> I remember seeing something like this, but do not remember why. We should revisit the cause... Could be worse for being local as you say. Are you working with an old pipeline run version?

I'm working with pipeline runs that are less than a year old. The implication of the error message is that all species should be summarized at a genus level, I guess by the loader process. Is that correct? To be clear, I was seeing this warning well before I started working on this PR.

app/helpers/heatmap_helper.rb
```diff
@@ -67,92 +67,22 @@ def self.sample_taxons_dict(params, samples, background_id)
...
-        break if count >= num_results
-        count += 1
-      end
-    end
-
-    candidate_taxons.values.sort_by { |taxon| -1.0 * taxon["max_aggregate_score"].to_f }
-  end
-
-  def self.fetch_top_taxons(
+  def self.fetch_top_taxons( # rubocop:disable Metrics/ParameterLists
```
tfrcarvalho: Can you add instead configure this rule with a higher threshold, say 10 ?

gregdingle: I actually think this is a good rule as-is, but it is outside the scope of this PR to refactor to fewer params, hence the exception. What do you think?

tfrcarvalho: I see it is already 10. So keep it.

gregdingle:
I discovered that `rails dev:cache` does not actually disable the controller caching, so I was always getting identical results. With caching code commented out, I get slightly different results with this PR compared to master. After some time, debugging, I'm still not sure why, or what is correct, so I'm going to put this PR back to a "work in progress".

PR #2506, opened 2019-08-13
Ensure idseq-cli gets `upgrade_required` status

jshoe:
### Description
- This ensures that the Python `requests` library gets a `426 Upgrade Required` status code on the returned request. Otherwise the deprecation warning may not display properly (on previous versions of idseq-cli).
- Originally I was going to raise the min version to 0.6.0 (a few months ago) but I noticed this bug.

### Tests
- Try to upload with a package version below 0.5.0, see messages:

Screen Shot 2019-08-12 at 6 15 40 PM

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -54,6 +54,7 @@ class SamplesController < ApplicationController
   PAGE_SIZE = 30
   MAX_PAGE_SIZE_V2 = 100
   MAX_BINS = 34
+  MIN_CLI_VERSION = '0.6.0'.freeze
 
   # GET /samples
   # GET /samples.json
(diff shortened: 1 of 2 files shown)


app/controllers/samples_controller.rb
```diff
@@ -54,6 +54,7 @@ class SamplesController < ApplicationController
   PAGE_SIZE = 30
   MAX_PAGE_SIZE_V2 = 100
   MAX_BINS = 34
+  MIN_CLI_VERSION = '0.5.0'.freeze
```
tfrcarvalho: We should force it to be at least 0.7.3 which fixes the long files bug.

jshoe: 👍 changed to 0.7.3.

PR #2508, opened 2019-08-14
IDSEQ-985 Add better styling for grayed-out heatmap cells.

MarkAZhang:
# Description

For heatmap cells that were grayed out due to the threshold filter, we now gray out the tooltip, and add an explanation notification at the top.

Screen Shot 2019-08-13 at 6 19 50 PM

We also change the styling for taxons that aren't present in a sample, for consistency.

Screen Shot 2019-08-13 at 6 19 53 PM

We also now show a caption at the bottom of the heatmap if threshold filters are present ONLY when the user is downloading the heatmap as a PNG or SVG.

Screen Shot 2019-08-13 at 6 21 31 PM

Screen Shot 2019-08-13 at 6 21 59 PM

# Notes

Discussed with Jenn about design, in particular to put the caption text on one line.

Since there isn't support for text wrapping in svg, had to write a custom utility function to separate a string into multiple lines. Right now, we split into lines based on character count (as opposed to measuring the rendered dimensions of the text) as it's a simpler approach.

Also refactored the code in the placeContainers method in Heatmap.js to be easier to read.

# Testing
* Verified that caption doesn't show normally, and only shows when user is downloading an image.
* Verified that tooltips behave correctly.
* Verified that splitIntoMultipleStrings behaves correctly if passed a really long word.

### DIFF
--- a/app/assets/src/components/ui/containers/data_tooltip.scss
+++ b/app/assets/src/components/ui/containers/data_tooltip.scss
@@ -1,11 +1,12 @@
 @import "~styles/themes/colors";
+@import "~styles/themes/typography";
 
 .dataTooltip {
+  @include font-body-xxs;
   background-color: $white;
   border: 1px solid $light-grey;
   border-radius: 5px;
   box-shadow: 2px 2px 5px rgba(0, 0, 0, 0.2);
-  font-size: 8pt;
   height: auto;
   min-width: 200px;
   max-width: 400px;
@@ -13,26 +14,35 @@
   pointer-events: none;
 
   &Title {
+    @include font-label-xs;
     border-bottom: 1px solid $lightest-grey;
     color: $medium-grey;
     margin-bottom: 10px;
-    text-transform: uppercase;
   }
 
   &Subtitle {
-    font-weight: 700;
+    @include font-body-xxs;
+    margin-bottom: 15px;
   }
 
   &SectionName {
+    @include font-label-xs;
     border-bottom: 1px solid $lightest-grey;
     color: $medium-grey;
     margin-bottom: 10px;
-    text-transform: uppercase;
   }
 
   &Section {
     &:not(:first-child) {
       margin-top: 20px;
+    }
+
+    &.disabled {
+      color: $light-grey;
+
+      .dataTooltipSectionName {
+        color: $light-grey;
+      }
     }
   }
 
@@ -41,7 +51,7 @@
   }
 
   &Label {
-    font-weight: 700;
+    @include font-header-xxs;
     margin-right: 10px;
     width: 160px;
     flex-shrink: 0;
--- a/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
@@ -1,7 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import cx from "classnames";
-import { keyBy } from "lodash/fp";
+import { size, map, keyBy } from "lodash/fp";
 
 import { logAnalyticsEvent } from "~/api/analytics";
 import { DataTooltip } from "~ui/containers";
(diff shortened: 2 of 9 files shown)


app/assets/src/components/ui/containers/data_tooltip.scss
```diff
@@ -20,7 +20,9 @@
   }
 
   &Subtitle {
-    font-weight: 700;
+    margin-bottom: 15px;
+    line-height: 1.5em;
+    font-size: 12px;
```
tfrcarvalho: Could we migrate to a typography mixin?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -146,6 +149,13 @@ export default class Heatmap {
     this.processData("parse");
   }
 
+  updateCaption(caption) {
+    this.options.printCaption = caption;
+    // We need to call placeContainers() to resize the height of the heatmap based on the number of captions.
+    // And update() to actually change the captions.
+    this.processData("placeContainers");
```
tfrcarvalho: We should not force the whole update function just for adding captions that might not even be necessary (if user does not trigger download).

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -253,6 +263,7 @@ export default class Heatmap {
       .append("g")
       .attr("class", cx(cs.dendogram, "columnDendogram"));
     this.gColumnMetadata = this.g.append("g").attr("class", cs.columnMetadata);
+    this.gCaption = this.g.append("g").attr("class", cs.captionContainer);
```
tfrcarvalho: This does not seem like the right place for this.

If it is for svg only we should add just before export and remove it afterwords. 
We could do this in `download` and `downloadAsPng`.
Regarding size below,  we could just extend the svg by the need totalCaptionHeight again before saving.
Do you foresee any problems with this approach?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -282,19 +293,38 @@ export default class Heatmap {
       ),
     };
 
+    const totalCellWidth = this.cell.width * this.columnLabels.length;
```
tfrcarvalho: Nice refactor for clarity!

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -116,6 +124,28 @@ class SamplesHeatmapVis extends React.Component {
...
+      filter => `${filter.metricDisplay} ${filter.operator} ${filter.value}`,
+      thresholdFilters
+    );
+
+    const fullString = `${numFilters} filter${
+      numFilters > 1 ? "s were" : " was"
+    } applied to the above heatmap: ${filterStrings.join(", ")}.
+      Non-conforming cells have been hidden or grayed out.`;
+
+    return splitIntoMultipleLines(fullString, CAPTION_LINE_WIDTH);
```
tfrcarvalho: Any idea if we could set the max number of characters having the font-size (could be exported) and width of heatmap (can be retrieved from the svg) into account? Would be cool to adapt to wider and shorter heatmaps.

tfrcarvalho: In any case, make sure current width works well with less or more samples.

app/assets/src/components/ui/containers/data_tooltip.scss
MarkAZhang: Yes, I've converted all fonts in this file to mixins.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
MarkAZhang: Since we're not using a monospace font and characters can have varying widths, it's hard to take the font-size and width of the heatmap into account.

The heatmap has min-width 1240 (even if there are fewer samples, the sample columns will get wider), so the caption line width of 100 seems to work well for that.

app/assets/src/components/visualizations/heatmap/Heatmap.js
MarkAZhang: This is a great point. I've made the code change and it seems to be working. I now have functions `showPrintCaption` and `hidePrintCaption`.

MarkAZhang:
How it now looks in the PNG and SVG (can't zoom into the SVG easily on my browser which is why it's small)

PNG:
Screen Shot 2019-08-14 at 1 05 05 PM

SVG:
Screen Shot 2019-08-14 at 1 05 36 PM

happyimadesignr:
Looks awesome! 

One thing: can you change make the line length of the caption longer it's cut off a bit early right now. Thank you!

I was going to make a comment about "grey" vs "gray" but it looks like you already changed to gray.

MarkAZhang:
Addressed Jenn's comment.

happyimadesignr:
Thanks Mark, LGTM

PR #2509, opened 2019-08-14
[Pipeline Viz] Matching error messages

emilyzhong:
# Description

Matches error messages on pipeline viz sidebar with error message on sample view

image

image

# Notes

Will push another commit on styling the link in the sidebar once that gets hashed out

# Tests

* No server-side changes
* Go to an errored sample run on pipeline version 3.8 or above, compare error message on sample page with error message on sidebar of pipeline viz after clicking on errored node.


app/assets/src/components/utils/sample.js
```diff
@@ -75,3 +75,83 @@ export const joinServerError = response => {
   });
   return joined;
 };
+
+export const errorMessageInfo = (sample, pipelineRun) => {
```
tfrcarvalho: Lets name this `sampleErrorInfo` ...

PR #2510, opened 2019-08-14
Add hover details on the single metadata square

gregdingle:
# Description

See https://jira.czi.team/browse/IDSEQ-1264.

# Tests

Load a heatmap
Try hovering over a metadata square
See tooltip
Hover out, see tooltip disappear

image

Add a metadata field that has no data
See "unknown"

image


app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -161,7 +163,24 @@ class SamplesHeatmapVis extends React.Component {
...
   }
 
+  getMetadataTooltipData(node) {
+    const metadata = node.metadata;
+    return [
+      {
+        name: "Metadata",
+        // TODO (gdingle): display name
+        data: Object.keys(metadata).map(k => [
+          k.replace("_", " ").replace("_v2", ""),
```
tfrcarvalho: Wouldn't the second replace fail? if you already replaced '`_`' with '` `'?
What is the use case here? Metadata should have a humanized name, no?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -968,7 +968,17 @@ export default class Heatmap {
       let columnMetadataCellEnter = columnMetadataCell
         .enter()
         .append("rect")
-        .attr("class", "columnMetadataCell");
+        .attr("class", "columnMetadataCell")
+        .on("mouseover", d => {
+          this.options.onNodeHover && this.options.onNodeHover(d);
```
tfrcarvalho: You can also leverage/replicate part of the code in `getColumnMetadataLegend`, which returns the aggregated metadata information.
So, i guess you could just return the label here.

gregdingle: good idea, thanks

PR #2512, opened 2019-08-14
[Pipeline Viz] Remove pipeline_viz feature flag

emilyzhong:
# Description

Removes feature flag for pipeline viz

# Notes

* Also renamed some test names

# Tests

* Updated server-side rspec tests to remove the flag, and all of them pass
* Removed feature flag for myself, then clicked around and everything was the same (in terms of where the link should appear and accessing the pipeline viz itself)


app/controllers/pipeline_viz_controller.rb
```diff
@@ -6,33 +6,31 @@ class PipelineVizController < ApplicationController
...
+      }
+      respond_to do |format|
+        format.html
+        format.json { render json: @results }
       end
+      return
     end
-    status = !feature_allowed ? :unauthorized : :not_found
+
+    status = pipeline_run ? :unauthorized : :not_found
```
jshoe: shouldn't this just be like "if !pipeline_run, status :not_found` and not have the message "Cannot access feature" ?

emilyzhong: Oh yes that makes sense — will update

emilyzhong: Actually — I just realized that it'll only ever return `:unauthorized` since if `pipeline_run` is true, it'll render something

PR #2513, opened 2019-08-14
[AMR Heatmap] Display new AMR metrics

lvreynoso:
# Description

In idseq-dag version 3.9, the pipeline added a few more metrics that get stored in a sample's amr_counts: reads per million, depth per million, and mapped reads (i.e. total reads for a gene). This PR updates the AMR Heatmap to display the new data, while keeping backwards compatibility for samples that were uploaded before the change.

# Notes

Fixed a few misc bugs I caught as well.

# Tests

* Server side code test updated & tested
* For client side code, select samples with amr counts and create an AMR heatmap. Check that the following work: changing the metric dropdown, the tooltip popup over the node, clicking on the gene name.
Screen Shot 2019-08-14 at 3 19 27 PM
Screen Shot 2019-08-14 at 3 19 56 PM


app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -65,10 +67,12 @@ export default class GeneDetailsMode extends React.Component {
 
   async getGeneInfo(geneName) {
     const ontology = await getOntology(geneName);
-    const cardEntryFound = ontology.error !== "" ? false : true;
+    const ontologyInfoFound = ontology.error === "" ? true : false;
```
tfrcarvalho: `ontology.error === ""` should suffice.
Same for line below.

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -238,11 +228,12 @@ export default class GeneDetailsMode extends React.Component {
             )}
           </div>
         )}
-        {collapseOntology && (
-          <div className={cs.expandLink} onClick={this.expandOntology}>
-            Show More
-          </div>
-        )}
+        {collapseOntology &&
+          (geneFamily.length > 0 || publications.length > 0) && (
```
MarkAZhang: Nice catch!

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -65,10 +67,12 @@ export default class GeneDetailsMode extends React.Component {
 
   async getGeneInfo(geneName) {
     const ontology = await getOntology(geneName);
-    const cardEntryFound = ontology.error !== "" ? false : true;
+    const ontologyInfoFound = ontology.error === "" ? true : false;
+    const cardEntryFound = ontology.label !== "" ? true : false;
```
MarkAZhang: I noticed that you changed `label` to `drugClass` in `fetch_ontology` below. In that case, should this just be `cardEntryFound = ontology.drugClass !== ""`?

`cardEntryFound` looks like it's only used to decide whether to show `SOURCE_CARD`. Just wanted to confirm whether this was correct.

app/assets/src/components/ui/containers/DataTooltip.jsx
```diff
@@ -25,8 +25,10 @@ const DataTooltip = ({ data, subtitle, title, singleColumn }) => {
                 <React.Fragment>
                   <div className={cs.dataTooltipLabel}>{keyValuePair[0]}</div>
                   <div className={cs.dataTooltipValue}>
-                    {/*Use .name if value is an object (e.g. location object)*/}
-                    {isObject(keyValuePair[1])
+                    {/*Use .name if value is an object (e.g. location object) AND it's
+                      an object with a name property */}
+                    {isObject(keyValuePair[1]) &&
+                    keyValuePair.hasOwnProperty("name")
```
MarkAZhang: I'm assuming that you have an object here which doesn't have a name property. Wouldn't directly displaying the object render something like `[Object object]`? I'm a little confused what's going on here. Maybe you should hide the key value pair if it's an object without a name property?

I don't like this convention of passing in objects and then inferring the name. (I know this wasn't you). A better approach is to explicitly pass the string that you want displayed into `<DataTooltip>`.

For example, instead of passing `data={["Key", { "name": "Value", "foo": "bar" }]}`, just pass `data={["Key", "Value"]}`

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -225,18 +153,38 @@ export default class AMRHeatmapVis extends React.Component {
...
+    const { selectedOptions } = this.props;
+    switch (selectedOptions.viewLevel) {
+      case VIEW_LEVEL_ALLELES: {
+        return sample.amrCounts.find(
+          amrCount => amrCount["allele"] === rowName
+        );
+      }
+      case VIEW_LEVEL_GENES: {
+        let amrCountForRow = sample.amrCounts.find(
+          amrCount => amrCount["annotation_gene"] === rowName
```
MarkAZhang: Can you add a comment somewhere explaining the difference between an `annotation_gene` and a `gene`?

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
```diff
@@ -193,11 +104,28 @@ export default class AMRHeatmapVis extends React.Component {
       }
     }
 
-    let values = METRICS.map(metric => {
-      const value = amrCountForNode ? amrCountForNode[metric.key] : 0;
+    let values = metrics.map(metric => {
```
MarkAZhang: So we show `N/A` if and only if the sample is too old to have the evaluated metrics, correct? Can you add a comment explicitly explaining that?

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
lvreynoso: Yes, that's correct.

app/assets/src/components/ui/containers/DataTooltip.jsx
lvreynoso: The "object" that I pass in `AMRHeatmapVis` is either just a plain string or a short piece of React. `keyValuePair` is an array of what is displayed in the tooltip, e.g. in this image:

Screen Shot 2019-08-14 at 5 55 20 PM

The names on the left are the first entry in the array, and the values on the right are the second entry in the array. The code to generate the bottom section in `AMRHeatmapView` for example is 

```
return [
    metric.text,
    metric.value === selectedOptions.metric ? <b>{value}</b> : value,
];
```

So it sometimes is an object, that object is sometimes a bit of React, and sometimes that object is something else (like with the phylo tree I think?). Since multiple files (I count at least 5) use `DataTooltip` changing the approach would require quite a bit of refactoring. So for an easier to read fix (and I agree it should be rewritten to be easier to read and handle) that would take some time. I don't have much, so perhaps someone else could implement a longer term fix.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: So it's kind of technical, but what gets passed as the gene name by the pipeline is a locus. A locus is a site in a DNA or RNA sequence where a cluster of different alleles may be found, but perform essentially the same function. When srst2 identifies an AMR gene, it compares a sequence to all the reference sequences that match to the same locus and gives only 1 best match between all the reference sequences. This is so that srst2 does not report closely related sequences as different genes. The locus name is what is used by the pipeline for "gene", and that's what gets stored under the `gene` key for an `AmrCount` in a sample.

However, the ARG ANNOT database also defines a gene name in the annotation for each reference sequence. This gene name is *much* more closely aligned with how genes are listed in CARD. I believe the issue of naming AMR genes is something that's still in the process of being solved by the research community. I could report the locus name as well (and in fact I think I'll update this to add that).

lvreynoso: Added the above as a comment

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
lvreynoso: Added.

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
MarkAZhang: Awesome. To make the code clearer, I think it'd be helpful to add a couple of comments.

In `fetch_ontology`, maybe add a comment explaining that `drugClass` appears in every ontology (which is why you use it as a shortcut to determine whether there any fields to extract). Otherwise, future readers will be confused why you are using `drugClass` in particular.

I'm assuming that whenever `label` is in the ontology, so is `accession`. SOURCE_CARD seems to rely on `accession`. Could you change to `cardEntryFound = ontology.accession !== ""`? Otherwise you would want to add a comment explaining the relationship of `label` and `accession`.

app/assets/src/components/ui/containers/DataTooltip.jsx
MarkAZhang: Okay, I understand. I forgot that you can also pass React elements, which are interpreted as objects. Thanks for clarifying!

Could you change the comment to something like: `Use .name if value is an object (e.g. location object) with a name property. Display React elements normally.`? So it's easier to make the connection.

app/assets/src/components/views/amr_heatmap/AMRHeatmapVis.jsx
MarkAZhang: This is super helpful! Now that I realize how deep the explanation is, I think this paragraph would be better to put in your AMR documentation file, so that we don't have large technical explanations in the middle of the code. Thanks for writing this up.

I do have a question (which we can discuss offline). So the gene is the "locus", and the annotation_gene is the reference sequence which had the best match. I would have expected the different reference sequences to be "alleles". So what's the difference between an allele and an annotation_gene?

MarkAZhang:
Just a couple more comments. Also, looks like you still need to address Tiago's changes.

lvreynoso:
@MarkAZhang How does it look now?

app/assets/src/components/common/DetailsSidebar/GeneDetailsMode/GeneDetailsMode.jsx
```diff
@@ -65,10 +67,13 @@ export default class GeneDetailsMode extends React.Component {
 
   async getGeneInfo(geneName) {
     const ontology = await getOntology(geneName);
-    const cardEntryFound = ontology.error !== "" ? false : true;
+    const ontologyInfoFound = ontology.error === "";
+    // either the CARD accession is 7 or it is useless
```
MarkAZhang: Maybe change this comment to `All CARD accession ids are 7 in length. Check if a card entry was found.`

PR #2518, opened 2019-08-15
Add check for nil annotation in amr results

lvreynoso:
# Description

Meant to handle this error in `pipeline_run.rb`, method `db_load_amr_counts`:

```
irb(main):003:0> Sample.find(26995).pipeline_runs[0].db_load_amr_counts
Traceback (most recent call last):
        3: from (irb):3
        2: from app/models/pipeline_run.rb:552:in `db_load_amr_counts'
        1: from app/models/pipeline_run.rb:558:in `block in db_load_amr_counts'
NoMethodError (undefined method `split' for nil:NilClass)
```

It appears that there are sequences without annotations in the ARG Annot database. This pr handles that case.

# Tests

* Rpsec test to be added. Should add one for the whole function.

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -560,11 +560,11 @@ class PipelineRun < ApplicationRecord
           rpm: row["rpm"],
           dpm: row["dpm"],
         }
-        if row["annotation"].present?
-          if row["annotation"].split(";").length >= 5
-            amr_count_for_sequence[:annotation_gene] = row["annotation"].split(";")[2]
-            amr_count_for_sequence[:genbank_accession] = row["annotation"].split(";")[4]
-          end
+        annotation_data = /(?:.+);(?:.+);(.+);(?:.+);(.+)/.match(row["annotation"])
+        if annotation_data.present?
+          annotation_captures = annotation_data.captures
+          amr_count_for_sequence[:annotation_gene] = annotation_captures[0]
+          amr_count_for_sequence[:genbank_accession] = annotation_captures[1]
         end
         amr_counts_array << amr_count_for_sequence
       end
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -561,8 +561,10 @@ class PipelineRun < ApplicationRecord
           dpm: row["dpm"],
         }
         if row["annotation"].present?
-          amr_count_for_sequence[:annotation_gene] = row["annotation"].split(";")[2]
-          amr_count_for_sequence[:genbank_accession] = row["annotation"].split(";")[4]
+          if row["annotation"].split(";").length >= 5
+            amr_count_for_sequence[:annotation_gene] = row["annotation"].split(";")[2]
+            amr_count_for_sequence[:genbank_accession] = row["annotation"].split(";")[4]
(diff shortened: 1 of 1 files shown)


app/models/pipeline_run.rb
```diff
@@ -555,13 +555,17 @@ def db_load_amr_counts
...
           coverage: row["coverage"],
           depth: row["depth"],
-          annotation_gene: row["annotation"].split(";")[2],
-          genbank_accession: row["annotation"].split(";")[4],
           drug_family: row["gene_family"],
           total_reads: row["total_reads"],
           rpm: row["rpm"],
           dpm: row["dpm"],
         }
+        if row["annotation"].nil? == false
```
tfrcarvalho: Do not compare too true/false... just use the result: 
```
if row["annotation"]
	...
```

app/models/pipeline_run.rb
```diff
@@ -555,13 +555,17 @@ def db_load_amr_counts
...
           coverage: row["coverage"],
           depth: row["depth"],
-          annotation_gene: row["annotation"].split(";")[2],
-          genbank_accession: row["annotation"].split(";")[4],
           drug_family: row["gene_family"],
           total_reads: row["total_reads"],
           rpm: row["rpm"],
           dpm: row["dpm"],
         }
+        if row["annotation"]
```
jshoe: It looks like there's another bug here. You're doing `amr_results_table.each row` and then adding an entry to amr_counts_array for each row. But then `if annotation` you would add another row to amr_counts_array.

lvreynoso: You're right. I'm tired lol. I'm glad I have you all to do code review.

app/models/pipeline_run.rb
```diff
@@ -550,18 +550,21 @@ def db_load_amr_counts
...
```
jshoe: I think you might want to still check if `row["annotation"].present?` unless you're sure that CSV.read wouldn't generate `{"annotation": ""}`

tfrcarvalho: In that case, should we also check if split has at least len 5? What can we assume about the output?
(As a side note/reference, when we expect a certain format we are better off using a regex to map and extract the content we need. We would just need to look at the regex comparison result to immediately know if the output was expected)

lvreynoso: I think it might almost be easier to write a python script to clean up the arg-annot fasta file.

Speaking of, srst2 updated their arg-annot file to r3 10 months ago. The chanzuckerberg fork we use is still on r2. I wonder what other features or fixes we might be missing out on :(

lvreynoso: In any case, I wrote the regex. Now either we get nil or something we can work with. For sure

cdebourcy:
What was the test plan on this PR?

lvreynoso:
> What was the test plan on this PR?

@cdebourcy I tested it manually in `irb` but I regretted merging this PR as soon as I did, because I realized that it fixes the problem but introduces another bug. You're right to bring up test plan, will make sure the fix for this fix is comprehensive

cdebourcy:
Ok, can you revert this commit?

PR #2520, opened 2019-08-15
Add check for a substantial annotation when loading amr counts

lvreynoso:
# Description

Meant to handle this error in `pipeline_run.rb`, method `db_load_amr_counts`:

```
irb(main):003:0> Sample.find(26995).pipeline_runs[0].db_load_amr_counts
Traceback (most recent call last):
        3: from (irb):3
        2: from app/models/pipeline_run.rb:552:in `db_load_amr_counts'
        1: from app/models/pipeline_run.rb:558:in `block in db_load_amr_counts'
NoMethodError (undefined method `split' for nil:NilClass)
```

It appears that there are sequences without annotations in the ARG Annot database. This pr handles that case and adds error checking for _every_ key it tries to read.

Also, too, I added more Rspec tests specifically to make sure `db_load_amr_counts()` can handle what we can throw at it.

# Tests

* Run rspec test with `docker-compose run --rm web "RAILS_ENV=test rspec spec/models/pipeline_run_spec.rb"`

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -550,18 +550,23 @@ class PipelineRun < ApplicationRecord
     unless File.size?(amr_results) < 10
       amr_results_table = CSV.read(amr_results, headers: true)
       amr_results_table.each do |row|
-        amr_counts_array << {
+        amr_count_for_gene = {
           gene: row["gene"],
           allele: row["allele"],
           coverage: row["coverage"],
           depth: row["depth"],
-          annotation_gene: row["annotation"].split(";")[2],
-          genbank_accession: row["annotation"].split(";")[4],
           drug_family: row["gene_family"],
           total_reads: row["total_reads"],
           rpm: row["rpm"],
           dpm: row["dpm"],
         }
+        if row["annotation"].present?
+          if row["annotation"].split(";").length >= 5
+            amr_count_for_gene[:annotation_gene] = row["annotation"].split(";")[2]
+            amr_count_for_gene[:genbank_accession] = row["annotation"].split(";")[4]
+          end
+        end
+        amr_counts_array << amr_count_for_gene
       end
     end
     update(amr_counts_attributes: amr_counts_array)
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -561,9 +561,10 @@ class PipelineRun < ApplicationRecord
           dpm: row["dpm"],
         }
         if row["annotation"].present?
-          if row["annotation"].split(";").length >= 5
-            amr_count_for_gene[:annotation_gene] = row["annotation"].split(";")[2]
-            amr_count_for_gene[:genbank_accession] = row["annotation"].split(";")[4]
+          split_annotation = row["annotation"].split(";")
+          if split_annotation.length >= 5
+            amr_count_for_gene[:annotation_gene] = split_annotation[2]
+            amr_count_for_gene[:genbank_accession] = split_annotation[4]
           end
         end
         amr_counts_array << amr_count_for_gene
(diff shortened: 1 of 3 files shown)


app/models/pipeline_run.rb
```diff
@@ -550,18 +550,23 @@ def db_load_amr_counts
...
-          genbank_accession: row["annotation"].split(";")[4],
           drug_family: row["gene_family"],
           total_reads: row["total_reads"],
           rpm: row["rpm"],
           dpm: row["dpm"],
         }
+        if row["annotation"].present?
+          if row["annotation"].split(";").length >= 5
+            amr_count_for_gene[:annotation_gene] = row["annotation"].split(";")[2]
+            amr_count_for_gene[:genbank_accession] = row["annotation"].split(";")[4]
```
davidrissato: suggestion: apply DRY concepts here
```
   if row["annotation"].present?
          split_annotation=row["annotation"].split(";")
          if split_annotation.length >= 5
            amr_count_for_gene[:annotation_gene] = split_annotation[2]
            amr_count_for_gene[:genbank_accession] = split_annotation[4]
```

spec/models/pipeline_run_spec.rb
```diff
@@ -1,6 +1,31 @@
 require 'rails_helper'
 
-describe PipelineRun, type: :model do
+MALFORMED_AMR_RESULTS = [
+  ["Sample", "DB", "gene", "allele", "coverage", "depth", "diffs", "uncertainty", "divergence", "length", "maxMAF", "clusterid", "seqid", "annotation", "gene_family", "total_gene_hits", "total_coverage", "total_depth", "total_reads", "rpm", "dpm"],
+  ["", "ARGannot_r2", "Aph3''Ia_AGly", "Aph3''Ia_1217", "4.779", "1.875", "1snp777holes", "edge0.0", "2.5639999999999996", "816", "0.0", "251", "1217", "", "AGly", "1", "4.779", "1.875", "3", "0.02481738581632464", "0.0155108661352029"],
+  ["", "ARGannot_r2", "MeowZ", "", "4.779", "1.875", "1snp777holes", "edge0.0", "2.5639999999999996", "816", "0.0", "251", "1217", "", "", "1", "4.779", "1.875", "3", "0.02481738581632464", ""],
+].freeze
+MALFORMED_AMR_COUNTS = [
```
tfrcarvalho: For readability, you could define each line as a constant whose name identifies the type of malformation you are testing for. And then you can set: 
```
MALFORMED_AMR_COUNTS = [
	VALUES_WITH_SINGLE_QUOTES,
	...
]
```

I'm actually not sure what is wrong with the second entry...

spec/models/pipeline_run_spec.rb
```diff
@@ -158,4 +183,68 @@
       end
     end
   end
+  context "ensure that loading amr counts works properly" do
```
tfrcarvalho: Do these tests trigger the behavior you were seeing with the annotation string and the ';'s?

spec/models/pipeline_run_spec.rb
```diff
@@ -158,4 +183,68 @@
...
+        allow(Rails.logger).to receive(:error)
+      end
+      it "should log an error and simply return" do
+        pipeline_run.db_load_amr_counts()
+
+        expect(Rails.logger).to have_received(:error).with(match(/No AMR results file found for PipelineRun #/))
+        expect(pipeline_run.amr_counts).to eq([])
+      end
+    end
+    context "AMR counts file exists, but is empty (results found no amr counts)" do
```
jshoe: nit: could you add blank lines between the blocks for readability?

spec/models/pipeline_run_spec.rb
lvreynoso: Nah cause I took out the regex

tfrcarvalho: But they should rigger that use case, to make sure it is fixed...

lvreynoso: Yeah it's in there!

MarkAZhang:
LGTM! But I'll leave the final approval to someone who's familiar with why the first fix didn't work.

PR #2522, opened 2019-08-16
Heatmap zoom

gregdingle:
# Description

See https://jira.czi.team/browse/IDSEQ-1233

This adds zoom to heatmap so that users can see more or less of a heatmap as they wish. In addition, it adds shrink-to-fit so that all samples of a heatmap are visible on initial load. 

To keep the scrollbars in view on zoom, I changed the heatmap container to have a fixed height. A side effect of this is to lose the semi-fixed page header on scroll. Now scrolling controls the content of the heatmap only, and the content above does not move, so it is effectively the same, except that the black idseq header and other content above the filters does not disappear. This change also fixes the visual issue where the fixed filter bar did not cover the whole width of the screen, leaving awkward patches of heatmap on either side of it. If any of this is not clear enough, please ask for me a demo. 

Lastly, the fixed height container revealed a latent issue where tooltips at the bottom got clipped on hover. I fixed this by extending @MarkAZhang 's fix for the same issue on the right side of page -- the tooltips now show above when close to the bottom. Note that the fix assumes two constant heights of the tooltips. I'm not sure how to get the height of the content before it has rendered. 

## Zoom out
image

## Zoom in
image

## Zoom control
image

## Flip tooltip at bottom of page
image

image

# Notes

* While the text becomes too small to read, the hover tooltip is still at normal size, allowing a user to read the contents of a cell
* The +/- control is the same one used on the map and pipeline viz

# Tests

**Zoom in and out**
Load a heatmap

Click on the plus 
See the heatmap expand up to the max

Click on the minus
See the heatmap shrink
See the heatmap stop shrinking at the smallest size

**Shrink-to-fit**
Shrink the size of your browser to <1000px wide
Load a large heatmap
See the heatmap fit in the width
Enlarge the browser to full screen
Reload
See the heatmap at the default size (1300px)

### DIFF
--- a/app/assets/src/components/views/PipelineViz/pipeline_viz.scss
+++ b/app/assets/src/components/views/PipelineViz/pipeline_viz.scss
@@ -257,6 +257,9 @@
   bottom: 50px;
   right: 15px;
   box-shadow: $box-shadow-dropdown-menu;
+  svg:hover {
+    background-color: darken(white, 5%);
+  }
 }
 
 .hovered {
--- a/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
@@ -363,11 +363,11 @@ class SamplesHeatmapVis extends React.Component {
       <div className={cs.samplesHeatmapVis}>
         <PlusMinusControl
           onPlusClick={withAnalytics(
-            this.handleZoom("in"),
+            () => this.handleZoom("in"),
             "SamplesHeatmapVis_zoom-in-control_clicked"
           )}
           onMinusClick={withAnalytics(
-            this.handleZoom("out"),
+            () => this.handleZoom("out"),
             "SamplesHeatmapVis_zoom-out-control_clicked"
           )}
           className={cs.plusMinusControl}
--- a/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
@@ -341,8 +341,8 @@ class SamplesHeatmapVis extends React.Component {
 
   handleZoom(increment) {
     const newZoom = Math.min(
-      3,
-      Math.max(0.1, this.heatmap.options.zoom + increment)
+      3, // max zoom factor
+      Math.max(0.2, this.heatmap.options.zoom + increment)
     );
     this.heatmap.updateZoom(newZoom);
   }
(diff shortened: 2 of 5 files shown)


app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -81,6 +83,9 @@ class SamplesHeatmapVis extends React.Component {
         // only display colors to positive values
         scaleMin: 0,
         printCaption: this.generateHeatmapCaptions(),
+        // Shrink to fit the viewport width
+        // See https://developer.mozilla.org/en-US/docs/Web/API/window/innerWidth
+        maxWidth: window.innerWidth,
```
tfrcarvalho: We discuss this sometime ago for other visualization. This should be size of the div that includes the heatmap svg.

app/assets/src/components/views/compare/samples_heatmap_vis.scss
```diff
@@ -1,6 +1,17 @@
 @import "~styles/themes/colors";
+@import "~styles/themes/typography";
+@import "~styles/themes/elements";
 
 .samplesHeatmapVis {
+  .plusMinusControl {
+    position: absolute;
```
tfrcarvalho: I am not particular fond of the position of the zoom... 
But in any case we should not use absolute position. If design decides that it should be next to the header, it should be part of the header component.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -322,6 +320,15 @@ export default class Heatmap {
 
     this.svg.attr("width", this.width).attr("height", this.height);
 
+    const zoom = Math.min(this.width, this.options.maxWidth) / this.width;
+
+    // Not standard but it works for downloads and svgsaver. See:
+    // https://stackoverflow.com/questions/11293026/default-background-color-of-svg-root-element
+    this.svg.attr(
+      "style",
+      `background-color: ${this.options.svgBackgroundColor}; zoom: ${zoom}`
```
tfrcarvalho: More d3ish like code is something like:
```
this.svg
	.style('background-color', this.options.svgBackgroundColor)
	.style('zoom', zoom)
```

MarkAZhang:
Hey, wanted to point out that `zoom` is a "non-standard" CSS property.

https://caniuse.com/#feat=css-zoom

It won't work on Firefox. CSS3 transform scale is much safer. https://medium.com/@sai_prasanna/simulating-css-zoom-with-css3-transform-scale-461d1b9762d6

app/assets/src/components/views/PipelineViz/pipeline_viz.scss
```diff
@@ -257,6 +257,9 @@
   bottom: 50px;
   right: 15px;
   box-shadow: $box-shadow-dropdown-menu;
+  svg:hover {
+    background-color: darken(white, 5%);
```
gregdingle: this is just to make the hover state consistent across pages

gregdingle:
Thanks @MarkAZhang . I'm leaving it in your hands now.

PR #2523, opened 2019-08-16
Create migration to add descriptions to projects

j-x-han:
# Description

Create migration file adding description column to projects.

# Tests

Run `rails db:migrate`, see no errors.


jshoe:
I just want to point this out as a reminder, separating the migration from the code change will only make a difference if the migration goes out in an earlier deploy

MarkAZhang:
!!! That is a great point

PR #2527, opened 2019-08-17
Revoke Basespace access token after upload is complete.

MarkAZhang:
# Note

Revoke the Basespace access token after we're done with it for maximum security.

# Tests

* Added rspec tests.
* Verified via local basespace upload and Resque logging that access token DELETE call is successful and calling an endpoint with the access token works before the DELETE call and returns 401 after the call.

### DIFF
--- a/app/jobs/transfer_basespace_files.rb
+++ b/app/jobs/transfer_basespace_files.rb
@@ -10,6 +10,11 @@ class TransferBasespaceFiles
     Rails.logger.info("Start TransferBasespaceFiles for sample id #{sample_id}")
     sample = Sample.find(sample_id)
     sample.transfer_basespace_files(basespace_dataset_id, basespace_access_token)
+
+    # Revoke the access token, so that it can no longer be used.
+    BasespaceHelper.revoke_access_token(basespace_access_token)
+
+    BasespaceHelper.verify_access_token_revoked(basespace_access_token)
   rescue => e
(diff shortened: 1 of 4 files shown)


app/jobs/transfer_basespace_files.rb
```diff
@@ -10,6 +10,11 @@ def self.perform(sample_id, basespace_dataset_id, basespace_access_token)
     Rails.logger.info("Start TransferBasespaceFiles for sample id #{sample_id}")
     sample = Sample.find(sample_id)
     sample.transfer_basespace_files(basespace_dataset_id, basespace_access_token)
+
+    # Revoke the access token, so that it can no longer be used.
+    BasespaceHelper.revoke_access_token(basespace_access_token)
```
jshoe: not a blocker: you could still handle for the success/failure of `revoke_access_token` (e.g. if the revocation request fails, we could throw an error and don't need to verify it was revoked) but it looks like HttpHelper.delete wouldn't return the success/failure

MarkAZhang: I do throw a warning in HttpHelper.delete if the request fails so that's one way of knowing.

I'm also a little worried that Basespace could eventually deprecate or modify the endpoint I'm hitting, since it doesn't appear in the official developer docs. So I wanted to verify directly that the token was indeed revoked. For maximum security.

PR #2533, opened 2019-08-20
Add sample sort control to heatmap

gregdingle:
# Description

See https://jira.czi.team/browse/IDSEQ-1259. This enables users to turn off clustering and sort by sample name. 

image

image

# Notes

* We are also going to add sorting taxa rows by genera. See https://jira.czi.team/browse/IDSEQ-1258 .
* We could add reverse alpha order if we want
* I left the current default to "cluster" though I think there is a good argument to make the default "alpha". 
* I adjusted the size of some of controls to make the new one fit in the row. I'm waiting on @happyimadesignr for a final visual design. 
* I think we should explain better what "clustering" does. I don't think the average user would guess that it is clustering by euclidean distance on the selected metric. 

# Tests

* Open a heatmap
* See default sort is "cluster" 
* Select "alpha"
* See heatmap change to correct order 
* Select "cluster"
* See heatmap return to previous order


app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -369,14 +394,14 @@ export default class SamplesHeatmapControls extends React.Component {
       <div className={cs.menu}>
         <Divider />
         <div className={`${cs.filterRow} row`}>
-          <div className="col s3">{this.renderTaxonLevelSelect()}</div>
+          <div className="col s2">{this.renderTaxonLevelSelect()}</div>
```
MarkAZhang: Is there a reason this was changed to `s2`? It leaves some empty space on the right of the filters.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -510,18 +515,19 @@ class SamplesHeatmapView extends React.Component {
...
+    sortSamplesOptions: SORT_SAMPLES_OPTIONS,
     taxonsPerSample: TAXONS_PER_SAMPLE_RANGE,
     specificityOptions: SPECIFICITY_OPTIONS,
   });
 
   handleSelectedOptionsChange = newOptions => {
-    const refetchData = !isEqual(keys(newOptions), ["dataScaleIdx"]);
-
+    const excluding = ["dataScaleIdx", "sortSamples"];
+    const refetchData = difference(keys(newOptions), excluding);
```
MarkAZhang: Could we do `shouldRefetchData = difference(keys(newOptions), excluding).length === 0`? Makes the naming clearer.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -81,6 +81,7 @@ class SamplesHeatmapVis extends React.Component {
         // only display colors to positive values
         scaleMin: 0,
         printCaption: this.generateHeatmapCaptions(),
+        sortColumns: this.props.sortSamples === "alpha", // else cluster
```
MarkAZhang: Nit: Can we name this `shouldSortColumns`? The reason being that sortSamples and sortColumns sound very similar, yet one is a boolean and the other is a string.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -588,6 +593,7 @@ class SamplesHeatmapView extends React.Component {
           taxonDetails={this.state.taxonDetails}
           taxonFilterState={this.state.taxonFilterState}
           thresholdFilters={this.state.selectedOptions.thresholdFilters}
+          sortSamples={this.state.selectedOptions.sortSamples}
```
MarkAZhang: Nit: Maybe call this `sampleSort`? Since this contains a sort type like "alpha" or "cluster", `sampleSort` (or `sampleSortType`) sounds slightly better.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
gregdingle: No good reason. I'll change it back.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
gregdingle: 👍

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
gregdingle: 👍

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
gregdingle: 👍

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
gregdingle: 👍

PR #2536, opened 2019-08-20
Add pipeline_run_id to cache key

gregdingle:
# Description

A new pipeline run of a sample that was run with the same version on the same day would not create a new cache key. This fixes that issue by including the run id in the cache key. 

I also added a `skip_cache` param. 

# Tests

cache test
* open a report in local
* see `pipeline_run_id` in the cache key
* see cache hit on second load

precache test
* open a pipeline run in the console
* run `precache_report_info`
* see pipeline_run_id in cache key 
* see success

skip_cache
* open a report
* add the param
* see no x-idseq headers in response
* remove the param
* see x-idseq headers in response

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -750,20 +750,20 @@ class SamplesController < ApplicationController
         .merge(pipeline_run_id: pipeline_run.id)
     )
 
-    # This allows 304 Not Modified to be returned so that the client can use its
-    # local cache and avoid the large download.
-    httpdate = Time.at(report_info_params[:report_ts]).utc.httpdate
-    response.headers["Last-Modified"] = httpdate unless skip_cache
-    # This is a custom header for testing and debugging
-    response.headers["X-IDseq-Cache"] = 'requested' unless skip_cache
-    response.headers["X-IDseq-Cache-Key"] = cache_key unless skip_cache
-    Rails.logger.info("Requesting report_info #{cache_key}") unless skip_cache
-
     pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
     json =
       if skip_cache
         ReportHelper.report_info_json(pipeline_run, params[:background_id])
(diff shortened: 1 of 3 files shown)


app/controllers/samples_controller.rb
```diff
@@ -744,7 +744,9 @@ def report_info
 
     cache_key = ReportHelper.report_info_cache_key(
       request.path,
-      params.permit(report_info_params.keys)
+      params
+        .permit(report_info_params.keys)
```
tfrcarvalho: Can't you add the id to report_info_params?

tfrcarvalho: I see, the `pipeline_run_id` is not in params?

tfrcarvalho: Should we do 
```
report_info_params.update(params.permit(report_info_params.keys))
```
Would use all the defaults of `report_info_params` even if not set by params...

gregdingle: yes that would work too. it would be nice to use the defaults of `report_info_params`, but I don't think it's worth the time to retest the PR.

app/controllers/samples_controller.rb
```diff
@@ -744,23 +745,31 @@ def report_info
...
+        .merge(pipeline_run_id: pipeline_run.id)
     )
 
     # This allows 304 Not Modified to be returned so that the client can use its
     # local cache and avoid the large download.
     httpdate = Time.at(report_info_params[:report_ts]).utc.httpdate
-    response.headers["Last-Modified"] = httpdate
+    response.headers["Last-Modified"] = httpdate unless skip_cache
```
tfrcarvalho: Can you add all these to the else ?

app/controllers/samples_controller.rb
tfrcarvalho: ok, i understand your concern...

PR #2538, opened 2019-08-21
IDSEQ-1178 - Stronger validation for input filenames

davidrissato:
# Description

Add stronger validation for input file names. 

File names must be no longer than 120 characters and can only contain letters from English alphabet (A-Z, upper and lower case), numbers (0-9), periods (.), hyphens (-) and underscores (_). Spaces are not allowed.

I'm also replacing a few aws cli commands by their AWS ruby SDK counterparts.

# Tests

* Manual tests

### DIFF
--- a/app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
@@ -131,7 +131,7 @@ class LocalSampleFileUpload extends React.Component {
               </li>
               <li>
                 File names must be no longer than 120 characters and can only
-                contain letters from English alphabet (A-Z, upper and lower
+                contain letters from the English alphabet (A-Z, upper and lower
                 case), numbers (0-9), periods (.), hyphens (-) and underscores
                 (_). Spaces are not allowed.
               </li>
--- a/app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
@@ -129,6 +129,12 @@ class LocalSampleFileUpload extends React.Component {
                 Paired files must be labeled with &quot;_R1&quot; or
                 &quot;_R2&quot; at the end of the basename.
               </li>
+              <li>
+                File names must be no longer than 120 characters and can only
+                contain letters from English alphabet (A-Z, upper and lower
+                case), numbers (0-9), periods (.), hyphens (-) and underscores
+                (_). Spaces are not allowed.
+              </li>
             </ul>
           </div>
         )}
--- a/app/assets/src/components/views/SampleUploadFlow/RemoteSampleFileUpload.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/RemoteSampleFileUpload.jsx
@@ -135,7 +135,7 @@ class RemoteSampleFileUpload extends React.Component {
               </li>
               <li>
                 File names must be no longer than 120 characters and can only
-                contain letters from English alphabet (A-Z, upper and lower
+                contain letters from the English alphabet (A-Z, upper and lower
                 case), numbers (0-9), periods (.), hyphens (-) and underscores
                 (_). Spaces are not allowed.
               </li>
--- a/app/assets/src/components/views/SampleUploadFlow/RemoteSampleFileUpload.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/RemoteSampleFileUpload.jsx
@@ -139,7 +139,6 @@ class RemoteSampleFileUpload extends React.Component {
                 case), numbers (0-9), periods (.), hyphens (-) and underscores
                 (_). Spaces are not allowed.
               </li>
-              <li>File names must be no longer than 140 characters.</li>
             </ul>
           </div>
         )}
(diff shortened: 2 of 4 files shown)


app/helpers/samples_helper.rb
```diff
@@ -142,14 +143,26 @@ def parsed_samples_for_s3_path(s3_path, project_id, host_genome_id)
     default_attributes = { project_id: project_id.to_i,
                            host_genome_id: host_genome_id,
                            status: 'created', }
-    s3_path.chomp!('/')
-    s3_bucket = 's3://' + s3_path.sub('s3://', '').split('/')[0]
-    s3_output, _stderr, status = Open3.capture3("aws", "s3", "ls", "--recursive", "#{s3_path}/")
-    return unless status.exitstatus.zero?
-    s3_output.chomp!
-    entries = s3_output.split("\n").reject { |line| line.include? "Undetermined" }
```
jshoe: it looks the removal of "Undetermined" files was dropped. was this intentional? I think it's still useful to ignore "Undetermined" files if we've been using this

davidrissato: Are you sure this is about a file name pattern? I mean, it gives me the impression this is something related to parsing the result lines of `aws s3 ls` that could contain such thing, but I might be totally wrong. Could you give me an example in which cases this is needed?

davidrissato: Thank you for the explanation. I believed it was related to AWS cli outputs. I'm adding it back.

app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
```diff
@@ -129,6 +129,12 @@ class LocalSampleFileUpload extends React.Component {
                 Paired files must be labeled with &quot;_R1&quot; or
                 &quot;_R2&quot; at the end of the basename.
               </li>
+              <li>
+                File names must be no longer than 120 characters and can only
```
tfrcarvalho: Can you make this text a constant for reuse?

app/helpers/samples_helper.rb
```diff
@@ -142,14 +145,25 @@ def parsed_samples_for_s3_path(s3_path, project_id, host_genome_id)
...
+
+    parsed_uri = URI.parse(s3_path)
+    return unless parsed_uri.scheme == "s3"
```
tfrcarvalho: Can we use if like below? `if parsed_uri.scheme != "s3"`

davidrissato: I don't like unless either, but I'm following conventions from here: https://github.com/rubocop-hq/ruby-style-guide#unless-for-negatives

tfrcarvalho: ok, if it must...

davidrissato: I was trying to follow the unless mantra, but in this case I think it makes sense

app/assets/src/components/views/SampleUploadFlow/RemoteSampleFileUpload.jsx
```diff
@@ -133,6 +133,13 @@ class RemoteSampleFileUpload extends React.Component {
                 Paired files must be labeled with &quot;_R1&quot; or
                 &quot;_R2&quot; at the end of the basename.
               </li>
+              <li>
+                File names must be no longer than 120 characters and can only
+                contain letters from the English alphabet (A-Z, upper and lower
+                case), numbers (0-9), periods (.), hyphens (-) and underscores
+                (_). Spaces are not allowed.
+              </li>
+              <li>File names must be no longer than 140 characters.</li>
```
tfrcarvalho: This line is not included above? Any particular reason?

app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
davidrissato: You mean, the entire text or just this new sentence? Because it seems there are other lines that can be reused as well.
Do you think this is a blocker for this PR or is it an sanitization that we should reconsider in a PR dedicated to remove such redundancies?

app/assets/src/components/views/SampleUploadFlow/RemoteSampleFileUpload.jsx
davidrissato: No, actually that line should have been removed, since the previous item already includes it. Thank you for pointing it out.

app/assets/src/components/views/SampleUploadFlow/LocalSampleFileUpload.jsx
tfrcarvalho: I would prefer to do it since it is a new sentence. It is not the problem of redundancy but keeping it in sync. 
I see your point but I prefer to start doing it and set an example than keep following a wrong pattern.

davidrissato: Considering this is not only about this sentence, but the entire block of instructions (the previous 2 `<li>` items are exactly the same), would it make sense to create a new react component with these instructions instead of using a constant?

davidrissato: Do we have any pre-existing pattern that I should follow in this case?

tfrcarvalho: Ok, you got me convinced. Let's just make a low priority jira task to just consolidate this

tfrcarvalho: Ok, you got me convinced. Let's just make a low priority jira task to just consolidate this

tfrcarvalho: Ok, you got me convinced. Let's just make a low priority jira task to just consolidate this

davidrissato:
> Tiago wrote:
> Should we do this verification on the cli too, before trying to upload?
Created a new ticket for that: IDSEQ-1330

PR #2540, opened 2019-08-22
IDSEQ-1136 - Improve exception pages

jshoe:
- See description and screenshots here: https://docs.google.com/document/d/1bLCvQ862MMXZC_yO-vj-ehkm7O_pETJIY7LiYEKoSLs/edit#


config/environments/staging.rb
```diff
@@ -11,7 +11,7 @@
   config.eager_load = true
 
   # Full error reports are disabled and caching is turned on.
```
jshoe: this comment was incorrect before (see prod.rb)

PR #2542, opened 2019-08-23
Add datadog frontend error monitoring.

MarkAZhang:
# Notes

It's not possible to overwrite the "source" field for browsers. It will always be "source:browser". Instead, we store the current environment ("staging", "prod") in `DATADOG_BROWSER_ENV`. We can create a facet on Datadog and use it to search and create monitors (see screenshot).

Screen Shot 2019-08-23 at 12 57 59 PM

# Tests
Verified that logs are captured in Datadog in dev when env variables are passed in, and NOT sent to Datadog without env variables.
Will verify in staging (assuming I can figure out how to trigger an error) when it rolls out.
At the very least, will verify that the embedded Javascript looks correct on staging.


app/views/layouts/_datadog_analytics.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <script>
+    // Snippet from https://docs.datadoghq.com/logs/log_collection/javascript/
+    // Client Token is designed to be public: https://docs.datadoghq.com/account_management/api-app-keys/#client-tokens
+
+    DD_LOGS.init({
+      clientToken: "<%= ENV["DATADOG_CLIENT_TOKEN"] %>",
+      forwardErrorsToLogs: true,
+    });
+
+    DD_LOGS.addLoggerGlobalContext('browser-env', "<%= ENV["DATADOG_BROWSER_ENV"] %>")
```
jshoe: you can just use `Rails.env` instead of creating DATADOG_BROWSER_ENV

app/views/layouts/_datadog_analytics.html.erb
```diff
@@ -0,0 +1,15 @@
+<% if ENV["DATADOG_CLIENT_TOKEN"] && current_user %>
+  <!-- Global site tags for Segment analytics-->
```
jshoe: I think you forgot to change the `Segment` comment

PR #2544, opened 2019-08-24
Add AppConfig table and site maintenance page.

MarkAZhang:
# Description

Added AppConfig table, so that we can modify production flags on the fly when needed.

Also added a screen for site maintenance, as a first use case for the AppConfig table.

When the flag is set, the following screen is shown. (notice that the header no longer shows any navigation)

Screen Shot 2019-08-23 at 5 32 09 PM

# Notes

Also stored the header height as a sass variable, since some apps depend on that value for layout.

# Tests

* Verified that when maintenance flag is off, things behave normally.
* Verified that when the flag doesn't exist, there are no errors.
* Verified that when the flag is on, the landing page does not direct to the maintenance page.
* Verified that when the flag is on, all pages re-direct to the maintenance page.
* Verified that there are no weird infinite re-direct loops.


app/helpers/app_config_helper.rb
```diff
@@ -0,0 +1,7 @@
+module AppConfigHelper
+  def get_app_config(key)
+    app_config = AppConfig.where(key: key).first
```
jshoe: Shouldn't we just make 'key' unique?

jshoe: It still feels not great for this particular flag to add another db call on each request, but we can re-evaluate 😬

PR #2545, opened 2019-08-26
Add project descriptions to sample upload flow pages

j-x-han:
# Description

Project creation during sample upload flow now has a field where users can add a description.

Screen Shot 2019-08-26 at 3 07 25 PM

The description will also be displayed on the review page. It will follow a Show More/Show Less pattern if the description takes up many lines.

Screen Shot 2019-08-26 at 3 08 47 PM

Screen Shot 2019-08-27 at 5 48 19 PM

Screen Shot 2019-08-27 at 5 48 53 PM

# Notes

Project descriptions are currently limited to 700 characters.

# Tests

* Verified that project descriptions are added to the database upon project creation.
* Verified that project descriptions of newly created projects will be displayed on the review page.
* Verified that users cannot exceed 700 characters.
* Verified that HTML tags are removed from project descriptions (tested using [these](https://www.owasp.org/index.php/Testing_for_HTML_Injection_(OTG-CLIENT-003)) examples)


app/controllers/projects_controller.rb
```diff
@@ -576,7 +578,7 @@ def set_project
 
   # Never trust parameters from the scary internet, only allow the white list through.
   def project_params
-    result = params.require(:project).permit(:name, :public_access, user_ids: [])
+    result = params.require(:project).permit(:name, :public_access, :description, user_ids: [])
     result[:name] = sanitize(result[:name]) if result[:name]
```
cdebourcy: Before merging, we need to check with @davidrissato what we should do for sanitization of the “description” input (for security)

davidrissato:
During your tests, copy and paste some examples from this page:
https://www.owasp.org/index.php/Testing_for_HTML_Injection_(OTG-CLIENT-003)

app/helpers/application_helper.rb
```diff
@@ -2,11 +2,16 @@
 require 'json'
 
 module ApplicationHelper
-  def sanitize(user_input_text)
+  def sanitize_project_name(user_input_text)
```
MarkAZhang: Could we move these functions into a new `ProjectsHelper`? Since they are related to projects only.

app/helpers/application_helper.rb
```diff
@@ -2,11 +2,16 @@
...
     # Allow letters, numbers, underscores, dashes, and spaces
     user_input_text.gsub(/[^A-Za-z0-9_\- ]/, '-')
   end
 
+  def sanitize_description(user_input_text)
```
MarkAZhang: `sanitize_project_description`?

PR #2550, opened 2019-09-02
IDSEQ-1152 - Avoid corrupted upload files (idseq-web)

davidrissato:
# Description

Adds stronger validations when decompressing gzip files to avoid undesired behaviors when decompressing corrupted files.

Refer to ticket IDSEQ-1152 in Jira for more deails.

# Tests

Manual tests using regular and corrupted files uploaded from S3 and local (web).

### DIFF
--- a/app/lib/log_util.rb
+++ b/app/lib/log_util.rb
@@ -1,10 +1,16 @@
+# frozen_string_literal: true
+
 class LogUtil
-  def self.log_err_and_airbrake(msg)
-    Rails.logger.error(msg)
+  def self.log_err_and_airbrake(msg, exception = nil)
+    LogUtil.log_backtrace(exception, msg)
     Airbrake.notify(msg)
   end
 
-  def self.log_backtrace(exception)
-    Rails.logger.error("Backtrace:\n\t#{exception.backtrace.join("\n\t")}")
+  def self.log_backtrace(exception, msg = nil)
+    if exception.respond_to? :backtrace
+      Rails.logger.error("#{msg}. Exception: #{exception.try(:message)}. Backtrace:\n\t#{exception.try(:backtrace).try(:join, "\n\t")}")
+    else
+      Rails.logger.error("#{msg}. Exception:\n\t#{exception}")
+    end
   end
 end
--- a/app/lib/log_util.rb
+++ b/app/lib/log_util.rb
@@ -6,11 +6,7 @@ class LogUtil
     Airbrake.notify(msg)
   end
 
-  def self.log_backtrace(exception, msg = nil)
-    if msg.nil?
-      Rails.logger.error("Exception message: #{exception.try(:message)}. Backtrace:\n\t#{exception.try(:backtrace).try(:join, "\n\t")}")
-    else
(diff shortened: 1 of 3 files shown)


app/lib/log_util.rb
```diff
@@ -1,10 +1,16 @@
+# frozen_string_literal: true
+
 class LogUtil
-  def self.log_err_and_airbrake(msg)
-    Rails.logger.error(msg)
+  def self.log_err_and_airbrake(msg, exception = nil)
```
tfrcarvalho: Either exception should not be nil by default or we should check for nil below... 
Weird that it is optional here and required below.

app/lib/log_util.rb
```diff
@@ -1,10 +1,16 @@
...
+    LogUtil.log_backtrace(exception, msg)
     Airbrake.notify(msg)
```
tfrcarvalho: I thought airbrake would automatically send the trace, no?

app/lib/log_util.rb
davidrissato: I agree it looks weird and I felt it too. However, I wanted to keep the previous signatures and just extend their behavior.
These functions have different purposes, which explain the order of these parameters.

`log_err_and_airbrake` sends a *string* message (`msg`) to airbrake. It doesn't accept an exception. For me the name of this operation is wrong, since we are just sending a string to airbrake, not necessarily an error.

So in order to keep it compatible to the current implementation, I added the `exception` field as optional (otherwise it would require a big refactoring to add this second parameter in dozens of different places)

app/lib/log_util.rb
davidrissato: No. These calls might eventually be "injecting" fragments of the exception on that `msg` string, or eventually printing the stack trace in a separate call, which make it very hard to correlate these "airbrake error messages" to their actual stack traces (and there are places where the backtrace and exception message are being completely masked)

app/lib/log_util.rb
tfrcarvalho: As discussed, my preferred approach is to create a new function with the ideal behavior and mark the other one as deprecated.
The new function should be marked has preferred - maybe with some comment to justify it. This will drive adoption of the recommended function as opposed to add some confusion to the current codebase by trying to be backwards compatible. 
We can also refactor `log_err_and_airbrake` to use the new functions...

davidrissato: I believe the right approach would be wrapping the original exception in more specific exception and sending it directly to airbrake. Since this is a huge refactoring, I opted for reverting the method to its original format and improving the `log_backtrace` to support an additional message.

davidrissato: I've completely reverted the error logging to its original form, since this change is completely unrelated to this PR. I missed that extra bit of info while I was making this change, but doing it "right" will require changing multiple places.
We can create a task later on to address a better error handling pattern.

davidrissato:
@tfrcarvalho I'm reviewing the solution here, and I may do some additional changes to this PR. I'll let you know if I need to make any extra changes before merging it (and ask for an additional review).

davidrissato:
@tfrcarvalho I included cut to avoid using additional awk memory to calculate a line length that will be discarded.

PR #2551, opened 2019-09-04
Make project descriptions editable in the DiscoverySidebar

j-x-han:
# Description

Users can now edit the descriptions of projects they have access to by clicking the grey "Edit" button in the DiscoverySidebar after a project has been selected.

image

The component follows the pattern of the editable MetadataSections in the sample details sidebar.

image

Users will not be able to edit the project descriptions of any project (public or private) that they have not been added to. The grey "Edit" button will not appear in these cases

Without access to a public project:

image

Without access to a private project:

image

# Tests

* Verified that users cannot edit project descriptions to be over 700 characters.
* Verified that HTML tags are removed from project descriptions.
* Created non-admin test account and verified that it cannot edit project descriptions without project access
    * On the front-end: the "Edit" button will not be displayed
    * On the back-end: the "update" action will be ignored if the current user does not have update access to the project

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoverySidebar/ProjectDescription.jsx
+++ b/app/assets/src/components/views/discovery/DiscoverySidebar/ProjectDescription.jsx
@@ -88,7 +88,7 @@ class ProjectDescription extends React.Component {
     return (
       <MetadataSection
         editable={this.props.project.editable}
-        onEditToggle={() => this.toggleEditing()}
+        onEditToggle={this.toggleEditing}
         editing={this.state.editing}
         title="Description"
         savePending={this.state.savePending}
@@ -100,7 +100,7 @@ class ProjectDescription extends React.Component {
           <div className={cs.descriptionContainer}>
             <Textarea
               onChange={val => this.handleDescriptionChange(val)}
-              onBlur={() => this.handleDescriptionSave()}
+              onBlur={this.handleDescriptionSave}
               value={description}
               className={cs.textarea}
               maxLength={MAX_DESCRIPTION_LENGTH}
--- a/app/assets/src/components/views/discovery/DiscoverySidebar/ProjectDescription.jsx
+++ b/app/assets/src/components/views/discovery/DiscoverySidebar/ProjectDescription.jsx
@@ -3,8 +3,10 @@ import cx from "classnames";
 import cs from "./project_description.scss";
 
 import PropTypes from "prop-types";
-import { Accordion } from "~/components/layout";
+import Textarea from "~ui/controls/Textarea";
+import MetadataSection from "~/components/common/DetailsSidebar/SampleDetailsMode/MetadataSection";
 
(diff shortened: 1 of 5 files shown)


app/assets/src/components/views/discovery/DiscoverySidebar/ProjectDescription.jsx
```diff
@@ -3,26 +3,81 @@ import cx from "classnames";
...
+  _save = async (projectId, value) => {
+    this.setState({
+      savePending: true,
+    });
+
+    let lastValidDescription = this.state.lastValidDescription;
+
+    const response = await saveProjectDescription(projectId, value);
+
+    // If save fails, revert to last valid description value.
```
gregdingle: why not show an error? and why only handle 422 errors?

app/assets/src/components/views/discovery/DiscoverySidebar/ProjectDescription.jsx
```diff
@@ -31,34 +86,56 @@ class ProjectDescription extends React.Component {
       : false;
 
     return (
-      <Accordion
-        className={cs.descriptionContainer}
-        bottomContentPadding
+      <MetadataSection
+        editable={this.props.project.editable}
+        onEditToggle={() => this.toggleEditing()}
```
gregdingle: shorten to `onEditToggle=[this.toggleEditing}` ?

app/assets/src/components/views/discovery/DiscoverySidebar/ProjectDescription.jsx
```diff
@@ -31,34 +86,56 @@ class ProjectDescription extends React.Component {
...
-            <div
-              className={cx(
-                shouldTruncateDescription && showLess && cs.truncated
-              )}
-            >
-              {description}
+        {this.state.editing ? (
+          <div className={cs.descriptionContainer}>
+            <Textarea
+              onChange={val => this.handleDescriptionChange(val)}
```
gregdingle: shorten these handler expressions as above

app/assets/src/components/views/discovery/DiscoverySidebar/ProjectDescription.jsx
```diff
@@ -3,26 +3,81 @@ import cx from "classnames";
...
+          changed: false,
+        },
+        () => {
+          this._save(this.props.project.id, this.state.description);
+        }
+      );
+    }
+  };
+
+  _save = async (projectId, value) => {
```
gregdingle: I don't think we follow the underscore convention for private JS methods, though I like it. I'd like to hear others's thoughts.

j-x-han: It's used in `SampleDetailsMode.jsx` to save metadata details as well, which is the style I was following!

PR #2553, opened 2019-09-04
Control to sort taxa by genus in heatmap

gregdingle:
# Description

Users have asked for the ability to sort the rows in the heatmap instead of clustering them. This add that ability alongside the new "sort samples" control. 

image

image

image

# Notes

I shrunk the "taxon level" and "taxon categories" controls to make room for the new control. Another solution would be to remove the little used "read specificity" control. 

There are a bunch of weird rows that I did not know how to sort
```
Non-genus-specific reads in family Anelloviridae
All taxa with neither family nor genus classification
uncultured gamma proteobacterium
Vector pAAV-hSyn1-GCaMP6s-P2A-mRuby3
```
I lower cased "all taxa with neither family nor genus classification" so it would be sorted to the bottom like other unclassified rows.

Originally I used the parentId for sorting species but I found an instance where the species parent ID is not the same as the genus:
```
category: "Eukaryota"
id: 340412
index: 23
name: "Aspergillus novofumigatus"
parentId: -1900340412
phage: false
---
category: "Eukaryota"
id: 746128
index: 26
name: "Aspergillus fumigatus"
parentId: -1900746128
phage: false (edited) 
```
I expected the parent IDs to be the same because they are both in the genus "Aspergillus"

# Tests

* Open a variety of heatmaps
* See that the default taxa sort is "cluster"
* Click on "sort taxa: genus"
* See that rows are alphabetical
* Try "genus" taxon level, see alphabetical sorting
* See that "All taxa with neither..." is at the top

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
@@ -35,7 +35,7 @@ const SORT_SAMPLES_OPTIONS = [
   { text: "Alphabetical", value: "alpha" },
   { text: "Cluster", value: "cluster" },
 ];
-const TAXA_SAMPLES_OPTIONS = [
+const SORT_TAXA_OPTIONS = [
   { text: "Genus", value: "genus" },
   { text: "Cluster", value: "cluster" },
 ];
@@ -521,7 +521,7 @@ class SamplesHeatmapView extends React.Component {
     // Client side options
     scales: SCALE_OPTIONS,
     sampleSortTypeOptions: SORT_SAMPLES_OPTIONS,
-    taxaSortTypeOptions: TAXA_SAMPLES_OPTIONS,
+    taxaSortTypeOptions: SORT_TAXA_OPTIONS,
     taxonsPerSample: TAXONS_PER_SAMPLE_RANGE,
     specificityOptions: SPECIFICITY_OPTIONS,
   });
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
@@ -35,6 +35,10 @@ const SORT_SAMPLES_OPTIONS = [
   { text: "Alphabetical", value: "alpha" },
   { text: "Cluster", value: "cluster" },
 ];
+const TAXA_SAMPLES_OPTIONS = [
+  { text: "Genus", value: "genus" },
+  { text: "Cluster", value: "cluster" },
+];
 const TAXONS_PER_SAMPLE_RANGE = {
   min: 0,
   max: 100,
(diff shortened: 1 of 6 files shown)


app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -516,12 +521,13 @@ class SamplesHeatmapView extends React.Component {
     // Client side options
     scales: SCALE_OPTIONS,
     sampleSortTypeOptions: SORT_SAMPLES_OPTIONS,
+    taxaSortTypeOptions: TAXA_SAMPLES_OPTIONS,
```
tfrcarvalho: nit: if we follow the same naming as the sample sort it should be taxon sort.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -126,8 +130,11 @@ class SamplesHeatmapVis extends React.Component {
 
   extractTaxonLabels() {
     return this.props.taxonIds.map(id => {
+      const name = this.props.taxonDetails[id].name;
       return {
-        label: this.props.taxonDetails[id].name,
+        // Moves missing to the top in "genus" sort.
+        // See MISSING_GENUS_ID in TaxonLineage.
+        label: id === -200 ? "* " + name : name,
```
tfrcarvalho: Aren't all negative genus a special case?

tfrcarvalho: Also, I think if treated specially these should be moved to the bottom.

katrinakalantar: agree with Tiago that taxa with missing genus should be moved to the bottom.

happyimadesignr:
We ran into some questions, it's being discussed in Slack here: https://chanzuckerbergteam.slack.com/archives/CM5K7MA03/p1568661697007700

gregdingle:
> The main goal is group by genus or sort alphabetically? We could have done grouping but sort by avg abundance. Seem like a more interesting sort.

Main goal is to group by genus. However, I think it is best to keep things simple in this PR, so I prefer to sort within each genus alphabetically as well. 

> I would have preferred to shorten the read specificity - even if we need to change the name.

@happyimadesignr and I prefer the order as-is, but we decided to shrink the bottom buttons as well, so they appear to have equal importance.

image

> The case you mentioned of the two species with wrong parent id is weird.

Opened a ticket: https://jira.czi.team/browse/IDSEQ-1429

> I would also suggest moving the "unclassified" genus-level calls to the bottom of the heatmap

Done.

app/lib/report_helper.rb
```diff
@@ -503,7 +503,8 @@ def self.cleanup_genus_ids!(taxon_counts_2d)
 
   def self.validate_names!(tax_2d)
     # This converts superkingdom_id to category_name and makes up
-    # suitable names for missing and blacklisted genera and species.
+    # suitable names for missing and blacklisted genera and species. Such
+    # made-up names should be lowercase so they are sorted below proper names.
```
gregdingle: This seems justifiable in general because 

a) it is easy to uppercase downstream
b) it is consistent with missing values defined elsewhere such as "uncultured bacterium"

gregdingle:
> I see you did not comment on my nitpicking comment on the use of plural vs. singular on the variable name. :P

Sorry, that was my mistake! That is a nit worth picking!

PR #2554, opened 2019-09-05
[Data Discovery] Add data layer and generalize async row fetching

tfrcarvalho:
# Description

This PR is the first step to make project and visualizations load data paginated data on-demand without redundant code
* Add a new data layer to hold and manage DD data 
* Move function that loads sample rows to the data layer and generalize

# Notes

* There should be no functionality changes.

# Tests

To test this change, we need to check the table behavior as we scroll or set any conditions that influence the rows to fetch (i.e. filters, search and project selection)/

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoveryDataLayer.js
+++ b/app/assets/src/components/views/discovery/DiscoveryDataLayer.js
@@ -73,7 +73,7 @@ export default class DiscoveryDataLayer {
     }
     if (missingIdxs.length > 0) {
       // currently loads using limit and offset
-      // could eventually lead to redundant fetches if data is not requests in regular continuous chunks
+      // could eventually lead to redundant fetches if data is not requested in regular continuous chunks
       const minNeededIdx = missingIdxs[0];
       const maxNeededIdx = missingIdxs[missingIdxs.length - 1];
       let { fetchedObjects, fetchedObjectIds } = await apiFunction({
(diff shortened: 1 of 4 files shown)


app/assets/src/components/views/discovery/discovery_data_layer.js
```diff
@@ -0,0 +1,104 @@
...
+      : stopIndex;
+    let missingIdxs = range(startIndex, minStopIndex + 1);
+    if (objects.order) {
+      missingIdxs = missingIdxs.filter(
+        idx => !(objects.order[idx] in objects.entries)
+      );
+    }
+    if (missingIdxs.length > 0) {
+      // currently loads using limit and offset
+      // could eventually lead to redundant fetches if data is not requests in regular continuous chunks
```
jshoe: nit: s/requests/requested

app/assets/src/components/views/discovery/DiscoveryDataLayer.js
```diff
@@ -0,0 +1,104 @@
...
+    };
+
+    this.apiFunctions = {
+      samples: this.fetchSamples,
+    };
+  }
+
+  newObjectDb = () => ({
+    entries: {},
+    order: null,
```
jshoe: `order` seems a little vague (at first I was expecting it to be a sorting function) ?

tfrcarvalho: orderedIds?

jshoe: yeah that sounds good

PR #2558, opened 2019-09-08
[Data Discovery] Fix links to samples.

tfrcarvalho:
# Description

Fix sample row links.
Create ObjectCollection class to ease management.

# Tests

* Click sample links and verify it links to the report page.


app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -351,6 +352,7 @@ class SamplesView extends React.Component {
     return (
       <div className={cs.table}>
         <InfiniteTable
+          testId="SamplesTable"
```
jshoe: remove testId?

PR #2560, opened 2019-09-09
Added setup script for linux

morsecodist:
# Description

Created a Linux compatible version of the `bin/setup` script  for Ubuntu/Mint. 

# Tests

* Tested each installation section independently
* Ran the script


morsecodist:
@tfrcarvalho Maybe we could make the bottom one it's own bash script then call it from the separate scripts?

jshoe:
Also feel free to add to https://github.com/chanzuckerberg/idseq-web/wiki/Developer-info

tfrcarvalho:
> @tfrcarvalho Maybe we could make the bottom one it's own bash script then call it from the separate scripts?

Sounds good to me... would you mind doing that on this PR?

morsecodist:
- Created a common multi-platform script so we can reuse the same logic
- Will update the wiki as soon as this is merged in case more changes are needed
- Changed `setup` to `setup-mac` I don't want to inconvenience anyone since everyone else uses mac and likely always will but I think it may be confusing to developers to see `bin/setup` and `bin/setup-ubuntu`. I can change it back if you prefer

bin/shared-setup
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# multi-platform compatible setup script to be shared by platform specific setup scripts
+
+gem install bundler --conservative
+bundle install
```
morsecodist: I noticed  `bundle install` was not in setup but `npm install` is. Is there a reason behind this? If not I think we should add it.

tfrcarvalho:
> * Changed `setup` to `setup-mac`

nit: Maybe `setup-macos` is the best logical match...

bin/shared-setup
tfrcarvalho: IIUC, `bundle install` is ran on the docker image since the gems are stored in non-shared folder.
`npm install` installs packages on `node-modules` folder. 
So, running `bundle install` on your own machine is an optional...

bin/shared-setup
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
tfrcarvalho: nit: Do you mind calling this script `setup-shared`? It will help keep them together when sorting...

bin/shared-setup
morsecodist: This makes sense for the app but you need them for running some of the scripts in `bin`

bin/shared-setup
morsecodist: This makes sense for the app but you need them for running some of the scripts in `bin`

bin/shared-setup
morsecodist: This makes sense for the app but you need them for running some of the scripts in `bin`

tfrcarvalho: I guess we can run it then.
The only thing I do not like is that this can mess up your ruby env, if you are not on a virtual environment.

BTW, by curiosity, which scripts did you notice that depend on it?

morsecodist: That's a good point. The `bin/shell` script depends on some aws-sdk gems which is why I ran into the issue. I can remove that and document it instead to be safe.

tfrcarvalho: Thanks.
Seems overkill to install everything else because of this script... not sure what is the best option long term here (maybe a custom gemfile for setup?).

PR #2565, opened 2019-09-11
Check if proj description is null during ReviewStep

j-x-han:
# Description

* Adding a check to ensure that project descriptions exist before attempting to display the description on the ReviewStep page.
  * During sample upload, ReviewStep page would encounter a null error for projects created prior to descriptions being added.

# Tests

* Verified that the ReviewStep page no longer errors if (locally) uploading a sample to a project created months ago.

### DIFF
--- a/app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx
@@ -131,6 +131,7 @@ class ReviewStep extends React.Component {
   render() {
     const { showUploadModal, showLessDescription } = this.state;
     const shouldTruncateDescription =
+      this.props.project.description &&
       this.countNewLines(this.props.project.description) > 5;
 
     const {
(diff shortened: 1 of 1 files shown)


app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx
```diff
@@ -131,6 +131,7 @@ class ReviewStep extends React.Component {
   render() {
     const { showUploadModal, showLessDescription } = this.state;
     const shouldTruncateDescription =
+      this.props.project.description &&
```
jshoe: nit: it'd be nice to still have the `this.props` consts pulled out so you don't have to repeat `this.props.project.description` throughout

jshoe: I think you should also modify `countNewLines` to just return 0 if it has a null argument. It can pay to be defensive in case that function is used in the future

jshoe:
oh darn I didn't notice you had bot/merge already

PR #2567, opened 2019-09-11
Switch back to shelve/BerkleyDB

morsecodist:
# Description

Change the configs to use the shelve/Berkley DB format instead of the sqlite format. This is to resolve an issue where sqlite3 hangs.

# Tests

- Ran the migration locally
- Will deploy to staging and test the pipeline
- Will monitor the pipeline with cloudwatch


boris-dimitrov:
Pls create JIRA to follow up and confirm the hangs are gone, at which point we can delete all the sqlite code from idseq-dag.

PR #2569, opened 2019-09-12
Added logging for pipeline monitor duration

morsecodist:
# Description

We are highly confident that this process is slow but we don't have specific numbers to quantify how big of an issue it is. This PR will log meaningful data so we can reliably monitor how long this process takes. It also includes information about how many active jobs their are at the time so we can try to see how the process scales. 

# Notes

The logging framework for the tasks doesn't support nested data like this so I just added the JSON myself. I intend to query for these programatically then parse and aggregate them.

# Tests

Test plan: monitor for the appearance of these logs after task deployment.

### DIFF
--- a/lib/tasks/pipeline_monitor.rake
+++ b/lib/tasks/pipeline_monitor.rake
@@ -289,7 +289,7 @@ class CheckPipelineRuns
     max_work_duration = 0
     iter_count = 0
     until @shutdown_requested
-      before_iter_timestamp = Time.now.to_f # unixtime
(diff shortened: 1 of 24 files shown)


lib/tasks/pipeline_monitor.rake
```diff
@@ -288,6 +289,7 @@ class CheckPipelineRuns
     max_work_duration = 0
     iter_count = 0
     until @shutdown_requested
+      before_iter_timestamp = Time.now.to_f # unixtime
```
tfrcarvalho: Better to use `Process.clock_gettime(Process::CLOCK_MONOTONIC)`

PR #2572, opened 2019-09-13
[Data Discovery] Persist DD options into local and session storage

tfrcarvalho:
# Description

* Persist all DD options (e.g. UX config, filters search) into session storage
* Persist UX options (show/hide sidebars, current tab, selected columns) into local storage
* The new stored fields are not sensitive data

# Tests

* Load DD and change filters, search or UX options
* Click my data again or go to the report page and comeback, and see the same configuration
* Close the browser and reopen and see that only UX config changes are still active

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoveryView.jsx
+++ b/app/assets/src/components/views/discovery/DiscoveryView.jsx
@@ -93,6 +93,11 @@ class DiscoveryView extends React.Component {
     });
 
     const urlState = this.urlParser.parse(location.search);
+    let sessionState = this.loadState(sessionStorage, "DiscoveryViewOptions");
+    let localState = this.loadState(localStorage, "DiscoveryViewOptions");
+
+    // values are copied from left to rigth to the last argument (left elements have priority)
+    const savedState = defaults(urlState, sessionState, localState, {});
 
     this.state = defaults(
       {
@@ -125,6 +130,7 @@ class DiscoveryView extends React.Component {
         projectId: projectId,
         projects: [],
         rawMapLocationData: {},
+        sampleActiveColumns: undefined,
         sampleDimensions: [],
         search: null,
         selectedSampleIds: new Set(),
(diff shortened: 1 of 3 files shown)


app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -159,20 +175,30 @@ class DiscoveryView extends React.Component {
...
     const urlState = pick(urlFields, this.state);
+    const localState = pick(localFields, this.state);
 
+    // Saving on URL enables sharing current view with other users
     let urlQuery = this.urlParser.stringify(urlState);
     if (urlQuery) {
       urlQuery = `?${urlQuery}`;
     }
 
+    // History state may include some fields that enable fast loading of previous pages (e.g. project)
```
jshoe: this comment is a little confusing to me. "fast loading" as in direct loading of the previous pages?

tfrcarvalho: Loading when clicking back or the history dropdown on the top-left corner of the browser.

tfrcarvalho: Any suggestions for better wording?

gregdingle:
Just a usability comment: in my experience it is usually a good idea to have a "reset to defaults" button when UI state is persistent.

tfrcarvalho:
> Just a usability comment: in my experience it is usually a good idea to have a "reset to defaults" button when UI state is persistent.

I would agree that could be important for the field selection. But that would be a follow up change. 
Also, we are not sure our current defaults are the thing users want to get back to as default. So, I would first have a way to set the "default" per user.

app/assets/src/components/views/discovery/DiscoveryView.jsx
jshoe: resolved

PR #2573, opened 2019-09-16
Pipeline + Results monitor thread pooling

morsecodist:
# Description

This simplifies the concurrency of the pipeline monitor and I hypothesize that this will also make it more performant. According to the [readme of the parallel gem](https://github.com/grosser/parallel) more threads will increase the performance of IO operations and this library has been designed with `ActiveRecord` in mind. This is also quite a bit simpler to than the sharding logic in my opinion.

# Notes

I tested an unrelated script with `Open3.capture3` to make sure that the approaches were compatible and they are. 

# Tests

- Ran pipeline locally with new changes
- Deploy to staging and monitor with new endpoint


tfrcarvalho:
So, basically, this will push the bottleneck to the database...
Given that is a relatively small change, it seems worth to try and agree that will likely improve performance significantly compared to current. 
Curious to see how the DB will react under this much high concurrency. Would it be worth to set a limit to concurrent queries? Does the gem have some mechanism to do that?

BTW, my opinion, in the medium (or even short) term, we should move to a PUB/SUB strategy (either locally REDIS-based or through AWS step functions, etc.) for both pipeline and result monitor.

morsecodist:
@tfrcarvalho great points.

> Would it be worth to set a limit to concurrent queries?

The library has a maximum thread size which is a limit on concurrent queries because they block the thread still. I have it set to 20 at the moment which I think is pretty safe but we can tweak it.

> So, basically, this will push the bottleneck to the database...

Yes but I think the database can handle what amounts to less than 20 concurrent calls. In general I feel that MySQL is more performant than the app itself.

> BTW, my opinion, in the medium (or even short) term, we should move to a PUB/SUB strategy (either locally REDIS-based or through AWS step functions, etc.) for both pipeline and result monitor.

I totally agree. This just seemed like some low hanging fruit to make things both simpler and faster.

PR #2575, opened 2019-09-17
Change taxon sort name to "alphabetical"

gregdingle:
# Description

Following discussion with @happyimadesignr , we decided to change the name of the recently added feature to be more accurate.

image

# Notes

I also fixed a bug I noticed which caused the save to be inaccurate. 

# Tests

naming
* open a heatmap
* see the change name in the dropdown

saving
* change to alphabetical
* save
* reload
* see alphabetical persist

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
@@ -36,7 +36,11 @@ const SORT_SAMPLES_OPTIONS = [
   { text: "Cluster", value: "cluster" },
 ];
 const SORT_TAXA_OPTIONS = [
-  { text: "Genus", value: "genus" },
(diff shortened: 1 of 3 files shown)


app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -36,7 +36,11 @@ const SORT_SAMPLES_OPTIONS = [
   { text: "Cluster", value: "cluster" },
 ];
 const SORT_TAXA_OPTIONS = [
-  { text: "Genus", value: "genus" },
+  // NOTE: "genus" sort was renamed to "alphabetical" because that is currently
+  // more accurate. To sort by genus, we should sort by taxon parent ID,
+  // which is currently not always accurate, and we should show what the genus
+  // is for viruses, which is not currently shown.
+  { text: "Alphabetical", value: "genus" },
```
tfrcarvalho: Should it be alphabetical in the value too ?
Seems like it can cause confusion.

PR #2578, opened 2019-09-18
Modify "Add Metadata" on heatmap to be more obvious

j-x-han:
# Description

* The `Add Metadata` text is now always showing.
* Plus icon has been made slightly bigger.
* Spacing has been increased.
* Icon, text, and line have a hover state of primary blue.

image

image

# Notes

The plus icon's blue hover state is created by applying a filter to the svg. See:
* https://semisignal.com/using-fecolormatrix-to-dynamically-recolor-icons-part-1-single-color-icons/
* https://alistapart.com/article/finessing-fecolormatrix/

# Tests

* Verified that heatmap behaves as expected when zooming in/out and metadata columns are added or removed.
* Verified that the `Add Metadata` text and icon are not present when the heatmap is downloaded as an svg or png.
image

### DIFF
--- a/app/assets/src/components/visualizations/heatmap/Heatmap.js
+++ b/app/assets/src/components/visualizations/heatmap/Heatmap.js
@@ -11,6 +11,9 @@ import cx from "classnames";
 import symlog from "../../utils/d3/scales/symlog.js";
 import cs from "./heatmap.scss";
 import { CategoricalColormap } from "../../utils/colormaps/CategoricalColormap.js";
+import addSvgColorFilter from "../../utils/d3/svg.js";
+// used for filter to make plus icon blue
+const COLOR_HOVER_LINK = "#3867fa";
 
 // TODO(tcarvalho): temporary hack to send elements to the back.
 // Remove once code is ported to d3 v4, which contains this function.
@@ -55,7 +58,6 @@ export default class Heatmap {
         maxColumnClusterHeight: 100,
         spacing: 10,
         metadataAddLinkHeight: 14,
-        colorHoverLink: "#2b52cd", // used for filter to make icon blue
         transitionDuration: 200,
         nullValue: 0,
         columnMetadata: [],
@@ -269,37 +271,8 @@ export default class Heatmap {
     );
 
     const defs = this.svg.append("defs");
-
-    // Creates a color matrix that can recolor a black icon via a filter. See:
-    // https://semisignal.com/using-fecolormatrix-to-dynamically-recolor-icons-part-1-single-color-icons/
-    const generateColorMatrix = function(rgb) {
-      let rScaled = rgb[0] / 255.0;
-      let gScaled = rgb[1] / 255.0;
-      let bScaled = rgb[2] / 255.0;
-      return `0 0 0 0 ${rScaled}
-              0 0 0 0 ${gScaled}
-              0 0 0 0 ${bScaled}
-              0 0 0 1 0`;
-    };
-
-    // Convert hex string to rgb array
-    const hexToRgb = function(hex) {
-      var r = parseInt(hex.slice(1, 3), 16),
-        g = parseInt(hex.slice(3, 5), 16),
-        b = parseInt(hex.slice(5, 7), 16);
-      return [r, g, b];
-    };
-
     // Create a blue color filter to match $primary-light.
-    defs
-      .append("filter")
-      .attr("id", "blue")
-      .append("feColorMatrix")
-      .attr("type", "matrix")
-      .attr(
-        "values",
-        generateColorMatrix(hexToRgb(this.options.colorHoverLink))
-      );
+    addSvgColorFilter(defs, "blue", COLOR_HOVER_LINK);
 
     this.g = this.svg.append("g");
     this.gRowLabels = this.g.append("g").attr("class", cs.rowLabels);
--- a/app/assets/src/components/visualizations/heatmap/Heatmap.js
+++ b/app/assets/src/components/visualizations/heatmap/Heatmap.js
@@ -13,7 +13,7 @@ import cs from "./heatmap.scss";
 import { CategoricalColormap } from "../../utils/colormaps/CategoricalColormap.js";
 import addSvgColorFilter from "../../utils/d3/svg.js";
 // used for filter to make plus icon blue
-const COLOR_HOVER_LINK = "#3867fa";
+const COLOR_HOVER_LINK = cs.primaryLight;
 
 // TODO(tcarvalho): temporary hack to send elements to the back.
 // Remove once code is ported to d3 v4, which contains this function.
--- a/app/assets/src/components/visualizations/heatmap/Heatmap.js
+++ b/app/assets/src/components/visualizations/heatmap/Heatmap.js
@@ -54,7 +54,8 @@ export default class Heatmap {
         maxRowClusterWidth: 100,
         maxColumnClusterHeight: 100,
         spacing: 10,
-        columnMetadataAddHeight: 14,
+        metadataAddLinkHeight: 14,
+        colorHoverLink: "#2b52cd", // used for filter to make icon blue
         transitionDuration: 200,
         nullValue: 0,
         columnMetadata: [],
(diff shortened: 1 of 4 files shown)


app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -266,6 +267,28 @@ export default class Heatmap {
       `background-color: ${this.options.svgBackgroundColor}`
     );
 
+    const defs = this.svg.append("defs");
+
+    // Creates a color matrix that can recolor a black icon via a filter. See:
+    // https://semisignal.com/using-fecolormatrix-to-dynamically-recolor-icons-part-1-single-color-icons/
```
gregdingle: nice explanation!

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -266,6 +267,28 @@ export default class Heatmap {
...
+                0 0 0 1 0`;
+    };
+
+    // Create a blue color filter to match $primary-light.
+    defs
+      .append("filter")
+      .attr("id", "blue")
+      .append("feColorMatrix")
+      .attr("type", "matrix")
+      .attr("values", generateColorMatrix(0x2b, 0x52, 0xcd));
```
tfrcarvalho: Can we make this color an option?
A string like "2b52cd" that you convert to this format here?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -54,6 +54,7 @@ export default class Heatmap {
         maxRowClusterWidth: 100,
         maxColumnClusterHeight: 100,
         spacing: 10,
+        columnMetadataAddHeight: 14,
```
tfrcarvalho: Maybe `metadataAddLinkHeight`? After all is not a column...

app/assets/src/components/visualizations/heatmap/heatmap.scss
```diff
@@ -99,24 +99,25 @@
...
       &:hover {
         .metadataAddLine {
-          stroke: $medium-grey;
+          stroke: $primary-light;
         }
 
         .metadataAddIcon {
-          opacity: 0.5;
+          // The filter is defined in heatmap.js
+          filter: url("#blue");
```
tfrcarvalho: I guess using filter mean neither `fill` or `stroke` worked for this icon?

tfrcarvalho:
@j-x-han Can you verify that the text and icon are removed once we download as SVG and PNG?

app/assets/src/components/visualizations/heatmap/heatmap.scss
j-x-han: It seems that the plus icon is an svg embedded in the heatmap, so applying `fill` or `stroke` didn't work. @MarkAZhang helped me find the filter workaround!

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -54,6 +54,8 @@ export default class Heatmap {
         maxRowClusterWidth: 100,
         maxColumnClusterHeight: 100,
         spacing: 10,
+        metadataAddLinkHeight: 14,
+        colorHoverLink: "#2b52cd", // used for filter to make icon blue
```
MarkAZhang: This is the value for $primary-medium, not $primary-light. Typo?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -266,6 +268,39 @@ export default class Heatmap {
...
+    // Create a blue color filter to match $primary-light.
+    defs
+      .append("filter")
+      .attr("id", "blue")
+      .append("feColorMatrix")
+      .attr("type", "matrix")
+      .attr(
+        "values",
+        generateColorMatrix(hexToRgb(this.options.colorHoverLink))
+      );
```
MarkAZhang: Could we factor this out into a function like `addSvgColorFilter` and put it in `utils/d3/svg.js`?

You'd then call the function with `addSvgColorFilter(this.svg, "blue", this.options.colorHoverLink)`.

app/assets/src/components/visualizations/heatmap/Heatmap.js
MarkAZhang: Also, I don't think it's necessary to make `colorHoverLink` an option. Since the blue color is hard-coded in CSS for the line and the "ADD METADATA" label, there's not currently a situation where you want to allow the user to pass in a different color (like red) here. You can just make a constant at the top of the file.

MarkAZhang:
A few more comments!

app/assets/src/components/visualizations/heatmap/Heatmap.js
j-x-han: When I input the value for `$primary-light`, the icon actually appeared noticeably lighter than the text and the line. I tried looking into this a little bit but couldn't find a clear answer, so I used `$primary-medium` as a workaround for the time being since I think it matches slightly better. It's likely that I don't fully understand how the filter color matrix works, so I'm still looking into that.

(Pictures below for reference.)

`$primary-light`:
image
`$primary-medium` (current):
image

app/assets/src/components/visualizations/heatmap/Heatmap.js
happyimadesignr: I think the spacing looks good, the text is a bit close, but we tried more space yesterday and don't want the metadata to be too far from the heatmap.

tfrcarvalho: I'm guessing this is a response to the other comment below...
OK, it is your call. I thing it is a but cluttered now that is always visible (it was ok for me when it was on hover).

app/assets/src/components/visualizations/heatmap/Heatmap.js
tfrcarvalho: If you decide to move it to a constant, maybe use the ability of exporting values: https://github.com/css-modules/css-modules/blob/master/docs/values-variables.md
This way we can make sure the values are in sync with css.

j-x-han: Found the answer to my color struggles in a comment [here](https://stackoverflow.com/questions/21977929/match-colors-in-fecolormatrix-filter) btw 😄I had to add an attribute `color-interpolation-filters="sRGB"` to the filter.

The icon color should match now!

image

MarkAZhang: I think Tiago's link only lets you import css variables into other css files. It can be hard to import css into js unless you do some kind of CSS-in-JS solution (which involves using a completely different framework)

MarkAZhang: Does adding color-interpolation-filters="sRGB" to your <feColorMatrix> help? I think it gets the color a lot closer, but still not perfect.

It's a little scary that feColorMatrix doesn't work as advertised...it seems to produce approximate color values for some reason. I looked into the spec and wasn't able to find a reason.

tfrcarvalho: @MarkAZhang No, the `@value` is used to export to js and css.
See, accepted answer here: https://github.com/css-modules/css-modules/issues/106#issuecomment-211607047

jshoe:
so glad to see this T_T!!

app/assets/src/components/visualizations/heatmap/heatmap.scss
```diff
@@ -1,5 +1,7 @@
 @import "~styles/themes/colors";
 @import "~styles/themes/typography";
+@value colors: "../../../styles/themes/_colors.scss";
```
tfrcarvalho: cant use `~styles/`?

j-x-han: Apparently not 😞
image
image

app/assets/src/components/visualizations/heatmap/Heatmap.js
j-x-han: I was able to export the value with `:export` following the example [here](https://github.com/css-modules/css-modules/issues/86#issuecomment-158273243) (Thanks a ton to Tiago for finding this for me!)

MarkAZhang: I'd never heard of this! Very cool.

PR #2579, opened 2019-09-18
Fix S3 upload SIGPIPE issue for large files

davidrissato:
More specific error handling pattern for SIGPIPE

# Description

Fix an issue with large files upload that was generating "[Errno 32] Broken pipe" message

# Tests

* Multiple manual tests

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -298,7 +298,7 @@ class Sample < ApplicationRecord
       while try < max_tries
         # Run the piped commands and save stderr
         err_read, err_write = IO.pipe
-        if fastq =~ /\.gz/
+        if fastq.match?(/\.gz/)
           proc_download, proc_unzip, proc_cut, proc_head, proc_validation, proc_zip, proc_upload = Open3.pipeline(
             ["aws", "s3", "cp", fastq, "-"],
             ["gzip", "-dc"],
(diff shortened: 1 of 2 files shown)


app/models/sample.rb
```diff
@@ -307,25 +309,42 @@ def initiate_s3_cp
...
+          }
         end
         err_write.close
         stderr = err_read.read
 
-        # Ignore proc_download and proc_unzip status because they will always throw exit 1 and
-        # SIGPIPE 13 'pipe broken' unless the entire file was headed. Ignore pipe error in stderr but
-        # we still want to see errs like HeadObject Forbidden.
-        if to_check.all? { |p| p && p.exitstatus && p.exitstatus.zero? } && (stderr.empty? || stderr.include?(InputFile::S3_CP_PIPE_ERROR))
+        # Whenever the file has more lines than the accepted limit,
```
tfrcarvalho: So, just so I understand, it is considered a success but we only upload the rows until we get to our limit?

davidrissato: That's correct, I kept the previous behavior.

gregdingle: And what was the original justification for 

```
    max_lines = 4 * (max_input_fragments || PipelineRun::DEFAULT_MAX_INPUT_FRAGMENTS)
```
and 
```
  DEFAULT_MAX_INPUT_FRAGMENTS = 75_000_000 # max fragments going into the pipeline
```
? 

Doesn't automatic subsampling cover the problem of too much data in the pipeline? It seems problematic to silently truncate large input files in a non-random way that we then refer to as "original fastqs".

tfrcarvalho: * 4 * comes from requiring 4 lines per read (fastq format)
* stopping at 75M is the initial limit. I believe the goal was to prevent users from just uploading gigantic files, but still produce results.

Agree that "original" fastq might be deceiving, but since we should not store it anyways, and we just need them for potential reruns, seems a reasonable decision.

gregdingle: Are reads in a fastq randomly distributed? It looks like they are: https://www.biostars.org/p/372557/ . In that case, I agree truncating is reasonable.

PR #2586, opened 2019-09-19
IDSEQ-1149 Improve Resque action handling

jshoe:
- Please see screenshots and description here and tag me in comments :) https://docs.google.com/document/d/1sVh-IMse6XfAMy2yIsMwfdj9XT3vQDZDGGsf0hprDo0/edit#heading=h.4ozxpxg5oqbp

### DIFF
--- a/app/middleware/resque_middleware.rb
+++ b/app/middleware/resque_middleware.rb
@@ -7,8 +7,6 @@ class ResqueMiddleware
     # Restrict allowed characters going to Resque server params.
     # update_param will modify the original 'env' object.
     if env['PATH_INFO'].start_with?("/resque")
-      puts "ENV 11:48am", env, "END"
-
       req = Rack::Request.new(env)
       req.params.each do |k, v|
         # Don't rely on other sanitize methods that may miss HTML encoding.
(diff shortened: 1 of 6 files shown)


app/middleware/resque_middleware.rb
```diff
@@ -1,19 +1,64 @@
...
+  BASE_CONTROLLER = ActionController::Base.new
+
   def initialize(app)
     @app = app
   end
 
+  # CSRF reference: https://medium.com/rubyinside/a-deep-dive-into-csrf-protection-in-rails-19fa0a42c0ef
   def call(env)
+    # Regular execution for non-Resque paths
+    unless env['PATH_INFO'].start_with?("/resque")
```
tfrcarvalho: What paths are these? I guess there is not way to mount this middleware class only for certain routes?

app/middleware/resque_middleware.rb
```diff
@@ -1,19 +1,64 @@
...
-      req.params.each do |k, v|
-        # Don't rely on other sanitize methods that may miss HTML encoding.
-        req.update_param(k, v.gsub(/[^0-9A-Za-z_]/, ''))
+    rack_request.params.each do |k, v|
+      # Don't rely on other sanitize methods that may miss HTML encoding.
+      req.update_param(k, v.gsub(/[^0-9A-Za-z_]/, '')) unless k == "_csrf"
+    end
+
+    # Verify every request except GET and HEAD
+    unless rack_request.get? || rack_request.head?
```
tfrcarvalho: Is OPTIONS relevant at all?

app/middleware/resque_middleware.rb
```diff
@@ -1,19 +1,64 @@
...
+      end
+    end
+
+    ### OUTPUT ADDITIONS
+
+    # Get regular application response
+    status, headers, response = @app.call(env)
+
+    # Add a _csrf token to all Resque UI forms
+    if response.is_a?(Array) && response.present?
```
tfrcarvalho: Is the response always supposed to be an array?
Never null in case of error or something?

app/middleware/resque_middleware.rb
jshoe: surprisingly I couldn't find a way with Rails 5 https://stackoverflow.com/questions/12679644/trigger-rack-middleware-on-specific-rails-routes

app/middleware/resque_middleware.rb
jshoe: which options actually?

app/middleware/resque_middleware.rb
jshoe: For the responses to modify, I think it would always be array

"The last element is the body of the response. Whatever object you return here must be an enumerable object, such as an Array or an IO object." (https://gist.github.com/markbates/4240848)

PR #2588, opened 2019-09-20
Revoke Basespace access token once all file transfer is complete.

MarkAZhang:
# Description
This is a follow-up to https://github.com/chanzuckerberg/idseq-web/pull/2548

See also https://docs.google.com/document/d/1FKy_bXtDBI-1Uqplv2XHfXfaz2IqYPVxHdpQ1kKaGsM/edit

# Tests

* Verified that multiple samples are uploaded successfully.
* Verified that the token is successfully revoked after multiple samples are uploaded.

### DIFF
--- a/app/helpers/basespace_helper.rb
+++ b/app/helpers/basespace_helper.rb
@@ -17,13 +17,13 @@ module BasespaceHelper
     )
   end
 
-  def verify_access_token_revoked(access_token)
+  def verify_access_token_revoked(access_token, sample_id)
     # Verify that the token was revoked by using it to call an API endpoint.
     # The API endpoint should return a 401.
 
     fetch_from_basespace(BASESPACE_CURRENT_PROJECTS_URL, access_token, {}, true)
 
-    LogUtil.log_err_and_airbrake("BasespaceAccessTokenError Revoke access token check failed")
+    LogUtil.log_err_and_airbrake("BasespaceAccessTokenError: Failed to revoke access token for sample id #{sample_id}")
   rescue
     # The call should fail.
     Rails.logger.info("Revoke access token check succeeded")
--- a/app/jobs/transfer_basespace_files.rb
+++ b/app/jobs/transfer_basespace_files.rb
@@ -11,8 +11,21 @@ class TransferBasespaceFiles
     sample = Sample.find(sample_id)
     sample.transfer_basespace_files(basespace_dataset_id, basespace_access_token)
 
-    # TODO: Revoke the access token, so that it can no longer be used.
-    # Need to do this after ALL samples that use the access token can be uploaded.
+    # If ALL samples relying on this access token are done transferring files, revoke the access token.
+    samples_remaining = Sample.where(basespace_access_token: basespace_access_token, status: Sample::STATUS_CREATED)
+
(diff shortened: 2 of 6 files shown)


app/helpers/basespace_helper.rb
```diff
@@ -17,13 +17,13 @@ def revoke_access_token(access_token)
...
     # Verify that the token was revoked by using it to call an API endpoint.
     # The API endpoint should return a 401.
 
     fetch_from_basespace(BASESPACE_CURRENT_PROJECTS_URL, access_token, {}, true)
 
-    LogUtil.log_err_and_airbrake("BasespaceAccessTokenError Revoke access token check failed")
+    LogUtil.log_err_and_airbrake("BasespaceAccessTokenError: Failed to revoke access token for sample id #{sample_id}")
   rescue
     # The call should fail.
     Rails.logger.info("Revoke access token check succeeded")
```
davidrissato: I know it is not part of this PR, but I feel like this rescue block should be refactored.
First, it is really counterintuitive this logic like "if it fails then it is a success, and if it succeeds then it is an error"
Second, the rescue block should to catching a very specific error, otherwise it will mask other situations. For example: what if it returned a 500? or if access token is null? All these situations would be considered as success, when in fact it should be an error.

app/jobs/transfer_basespace_files.rb
```diff
@@ -11,8 +11,21 @@ def self.perform(sample_id, basespace_dataset_id, basespace_access_token)
...
-    # TODO: Revoke the access token, so that it can no longer be used.
-    # Need to do this after ALL samples that use the access token can be uploaded.
+    # If ALL samples relying on this access token are done transferring files, revoke the access token.
+    samples_remaining = Sample.where(basespace_access_token: basespace_access_token, status: Sample::STATUS_CREATED)
+
+    # If multiple samples finish file transfer at exactly the same time, this block of code may be run multiple times.
+    if samples_remaining.empty?
+      begin
+        BasespaceHelper.revoke_access_token(basespace_access_token)
+      rescue
```
davidrissato: Same thing here, you should catch a specific error. Otherwise, things like "method doesn't exist" will silently succeed. And also, we should log the error. `rescue... nil ... end` is a dangerous construction.

app/helpers/basespace_helper.rb
davidrissato: Ideally, if status_code is important in this case, `fetch_from_basespace` should be able to return that status and not fail with an error in this case. It could be a parameter that could modify its behavior, like `return_failed_status_codes=True` (we can find a better name) and then you can parse this response code.

MarkAZhang:
Hey @davidrissato , I did some research and I would like to switch from using Net::HTTP to http.rb. The API is much better and it will be easier to do the error handling that you mentioned.
https://github.com/httprb/http

I will do this immediately in a follow-up PR.

app/helpers/basespace_helper.rb
MarkAZhang: This is a great point. Thanks for pointing this out.

davidrissato:
Approved under the condition that these issues will be addressed in a follow up PR / ticket.

PR #2595, opened 2019-09-24
Create an "on/off" switch for the automatic pipeline restart

davidrissato:
# Description

Adding a new configuration value `AppConfig::AUTO_RESTART_ALLOWED_STAGES` to enable/disables automatic pipeline restart individually per stage number (ex: you may want to enable it just for stages 3 - Post processing).

# Notes

This PR is also fixing an issue in `PipelineRunStages` factory that had an invalid default value

# Tests

- Automated tests


cdebourcy:
Should we add instructions on how to use this new switch to the On-Call Handbook?

davidrissato:
@cdebourcy That’s a great idea. I’ll do that

PR #2598, opened 2019-09-24
Switch to using http.rb and improve http error handling.

MarkAZhang:
# Description

http.rb has a much simpler API and is a popular library. Switching to it makes our api code simpler.

Improved error handling so only specific errors that we expect are caught, based on feedback from David in https://github.com/chanzuckerberg/idseq-web/pull/2588.

# Tests

* Updated tests. Most existing tests still passed with the refactor, which improves confidence.
* Manually tested, including the edge case where revoke_access_token is called multiple times (see screenshot).

Screen Shot 2019-09-24 at 2 58 20 PM

### DIFF
--- a/app/helpers/basespace_helper.rb
+++ b/app/helpers/basespace_helper.rb
@@ -19,14 +19,18 @@ module BasespaceHelper
 
   def verify_access_token_revoked(access_token, sample_id)
     # Verify that the token was revoked by using it to call an API endpoint.
-    # The API endpoint should return a 401.
 
     fetch_from_basespace(BASESPACE_CURRENT_PROJECTS_URL, access_token, {}, true)
 
+    # If we reach this step, the access token must not have been revoked.
     LogUtil.log_err_and_airbrake("BasespaceAccessTokenError: Failed to revoke access token for sample id #{sample_id}")
-  rescue
-    # The call should fail.
-    Rails.logger.info("Revoke access token check succeeded")
+  rescue HttpHelper::HttpError => e
+    # We expect the API endpoint to return a 401.
+    if e.status_code == 401
+      Rails.logger.info("Revoke access token check succeeded")
+    else
+      raise e
+    end
   end
 
   # In one instance, we send a request expecting it to fail. So we provide a silence_errors option.
--- a/app/jobs/transfer_basespace_files.rb
+++ b/app/jobs/transfer_basespace_files.rb
@@ -18,15 +18,20 @@ class TransferBasespaceFiles
     if samples_remaining.empty?
       begin
         BasespaceHelper.revoke_access_token(basespace_access_token)
-      rescue
+      rescue HttpHelper::HttpError => e
         # If revoke_access_token is called multiple times, it will fail on the subsequent calls. This is okay.
-        nil
+        if e.status_code == 401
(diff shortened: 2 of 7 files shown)


app/helpers/basespace_helper.rb
```diff
@@ -19,14 +19,18 @@ def revoke_access_token(access_token)
 
   def verify_access_token_revoked(access_token, sample_id)
     # Verify that the token was revoked by using it to call an API endpoint.
-    # The API endpoint should return a 401.
 
     fetch_from_basespace(BASESPACE_CURRENT_PROJECTS_URL, access_token, {}, true)
 
+    # If we reach this step, the access token must not have been revoked.
```
MarkAZhang: Hey @davidrissato, I realize you had commented that this function feels counterintuitive. I believe this is a rare usage of the fetch_from_basespace http method, and almost all of the time, we won't expect the request to fail, so I'm okay with the function being a little unusual (I added a little more commenting). I'd rather not modify the http method to make it more complicated at this point, unless these cases become more common.

I did add a custom HttpError type that stores the status code. This ensures that we only catch the one specific error we are expecting and re-raise all other errors.

app/jobs/transfer_basespace_files.rb
```diff
@@ -18,15 +18,17 @@ def self.perform(sample_id, basespace_dataset_id, basespace_access_token)
     if samples_remaining.empty?
       begin
         BasespaceHelper.revoke_access_token(basespace_access_token)
-      rescue
+      rescue HttpHelper::HttpError => e
         # If revoke_access_token is called multiple times, it will fail on the subsequent calls. This is okay.
-        nil
+        if e.status_code == 401
+          Rails.logger.warn("Revoke access token failed. Likely called multiple times by samples sharing the same token.")
```
davidrissato: Should we add the sample id here? (ex: "Revoke access token failed for sample 123. Likely...")

app/jobs/transfer_basespace_files.rb
```diff
@@ -18,15 +18,17 @@ def self.perform(sample_id, basespace_dataset_id, basespace_access_token)
...
+        else
+          raise e
```
davidrissato: I noticed you removed a rescue block in the end of this method that was logging the error and ignoring it (which would eventually make it silently fail).

This new version (where you are re-raising the original error) is better, since it propagates the error condition, but it is important to guarantee that the caller method can actually handle these errors. I don't know where exactly this method is being invoked. Have you checked that?

cdebourcy:
Cool!

app/jobs/transfer_basespace_files.rb
```diff
@@ -18,15 +18,19 @@ def self.perform(sample_id, basespace_dataset_id, basespace_access_token)
...
+        end
       end
 
       # This verification call will work even if called multiple times.
       BasespaceHelper.verify_access_token_revoked(basespace_access_token, sample_id)
     end
   rescue => e
-    Rails.logger.error(e)
+    LogUtil.log_err_and_airbrake("Error transferring basespace files for sample #{sample_id}. Reason: #{e}")
```
davidrissato: I would re-raise the error after logging it to airbrake, so we can eventually restart the failing task using resque (assuming this is being executed by a resque worker)

PR #2599, opened 2019-09-24
Change pathogen info histogram x-axis to log10 scale

j-x-han:
# Description
The rpm counts histogram displayed on a taxon's sidebar previously used a linear scale; changing to a log10 scale to show more detail on the lower distribution.

- Example 1:
**Before (Linear Scale)**
image
Screen Shot 2019-09-24 at 3 29 16 PM
**After (log10 Scale)**
image

- Example 2:
image
**Before (Linear Scale)**
Screen Shot 2019-09-24 at 3 43 07 PM
**After (log10 Scale)**
image

**Info Tooltip Added**
image

# Notes

- Since the log(0) is undefined, all values were incremented by 1 to accommodate using the log scale. The x-axis label has been updated accordingly.
- Some statistics are no longer displayed (lines displaying average and lighter shaded boxes displaying standard deviation) because they were confusing with the log10 scale.
- Updated histogram component so bins have variable width if using the log10 scale.

# Tests

* Checked that histogram now uses a log10 scale for the x-axis and values are modified appropriately to prevent out of range (0) values from being passed in.
* Will need to check CoverageVizBottomSidebar on staging to ensure that the changes made to the Histogram component don't influence the histogram displayed there. Was unable to check on localhost.

### DIFF
--- a/app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
+++ b/app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
@@ -5,6 +5,8 @@ import PropTypes from "prop-types";
 import { getTaxonDescriptions, getTaxonDistributionForBackground } from "~/api";
 import Histogram from "~/components/visualizations/Histogram";
 import { logAnalyticsEvent, withAnalytics } from "~/api/analytics";
+import InfoCircleIcon from "~/components/ui/icons/InfoCircleIcon";
+import ColumnHeaderTooltip from "~/components/ui/containers/ColumnHeaderTooltip";
 
 import cs from "./taxon_details_mode.scss";
 
@@ -184,11 +186,11 @@ export default class TaxonDetailsMode extends React.Component {
       this.state.histogramRpmSeries,
       {
         seriesNames: ["NT", "NR"],
-        labelX: "rpm+1",
-        labelY: "count",
+        labelX: "rPM+1",
+        labelY: "# of Background Samples",
         refValues: [
           {
-            name: "sample",
+            name: "this sample",
             values: [
               this.props.taxonValues.NT.rpm,
               this.props.taxonValues.NR.rpm,
@@ -324,6 +326,17 @@ export default class TaxonDetailsMode extends React.Component {
           <div>
             <div className={cs.subtitle}>
               Reference Background: {this.props.background.name}
+              <ColumnHeaderTooltip
+                trigger={
+                  <span>
+                    <InfoCircleIcon className={cs.infoIcon} />
+                  </span>
+                }
+                content="This chart shows how abundant (rPM) this taxon is in your
+                sample compared to the other samples in the chosen Background Model.
+                The farther your sample line is to the right of the other samples, the
+                higher the likelihood that this taxon is unique to your sample."
+              />
             </div>
             <div
               className={cs.histogram}
--- a/app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
+++ b/app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
@@ -184,7 +184,7 @@ export default class TaxonDetailsMode extends React.Component {
       this.state.histogramRpmSeries,
       {
         seriesNames: ["NT", "NR"],
-        labelX: "rpm",
+        labelX: "rpm+1",
         labelY: "count",
         refValues: [
           {
@@ -195,6 +195,8 @@ export default class TaxonDetailsMode extends React.Component {
             ],
           },
         ],
+        showStatistics: false,
+        xScaleLog: true,
       }
     );
     this.histogram.update();
--- a/app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
+++ b/app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
@@ -184,7 +184,7 @@ export default class TaxonDetailsMode extends React.Component {
       this.state.histogramRpmSeries,
       {
         seriesNames: ["NT", "NR"],
-        labelX: "rpm+1",
+        labelX: "rpm",
         labelY: "count",
         refValues: [
           {
--- a/app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
+++ b/app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
@@ -184,7 +184,7 @@ export default class TaxonDetailsMode extends React.Component {
       this.state.histogramRpmSeries,
       {
         seriesNames: ["NT", "NR"],
-        labelX: "rpm",
+        labelX: "rpm+1",
         labelY: "count",
         refValues: [
           {
--- a/app/assets/src/components/visualizations/Histogram.js
+++ b/app/assets/src/components/visualizations/Histogram.js
@@ -75,6 +75,14 @@ export default class Histogram {
         `translate(0,${this.size.height - this.margins.bottom})`
       )
       .call(axisBottom(x).tickSizeOuter(0));
+
+    // if using log10 scale, set the number of ticks according to the size of
+    // the domain (in powers of 10) to prevent extra labels from showing up
+    if (this.options.xScaleLog) {
+      let nTicks = Math.log10(x.domain()[1]) + 1;
+      g.call(axisBottom(x).ticks(nTicks));
+    }
+
     g
       .append("text")
       .attr("x", this.size.width - this.margins.right)
--- a/app/assets/src/components/visualizations/Histogram.js
+++ b/app/assets/src/components/visualizations/Histogram.js
@@ -163,13 +163,17 @@ export default class Histogram {
     return bins;
   };
 
-  getBarWidth = () => {
+  getBarWidth = (x, bin) => {
     if (this.options.skipBins) {
       return (
         (this.size.width - this.margins.right - this.margins.left) /
         (this.options.numBins ||
           max(map(seriesData => seriesData.length, this.data)))
       );
+    }
+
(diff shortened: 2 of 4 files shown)


app/assets/src/components/visualizations/Histogram.js
```diff
@@ -301,20 +310,38 @@ export default class Histogram {
...
+        ? this.svg
+            .append("g")
+            .attr("class", `bar-${i}`)
+            .attr("fill", colors[i])
+            .selectAll("rect")
+            .data(bins[i])
+            .enter()
+            .append("rect")
+            .attr("class", (_, index) => `rect-${index}`)
+            .attr("x", d => x(d.x0) + i * ((x(d.x1) - x(d.x0)) / 2))
```
j-x-han: Ideally, I would be able to do something like modify `getBarWidth` to calculate this if `this.options.xScaleLog` rather than repeating this giant block of code. @MarkAZhang Do you have any ideas on how to potentially clean this up?

cdebourcy:
This looks SO much better and more useful than before! Nice!

tfrcarvalho:
* The labels seems a bit too close on the second chart. Weren't you able to remove the 2eX labels?

app/assets/src/components/visualizations/Histogram.js
tfrcarvalho: Looking again it seems you should be able to get away with the same code. You would just have to replace `2` by the number of series... But would be good to test with coverage viz.

j-x-han: By number of series, do you mean `this.data.length`?

app/assets/src/components/visualizations/Histogram.js
```diff
@@ -274,10 +276,17 @@ export default class Histogram {
...
+    if (this.options.xScaleLog) {
+      for (let i = 0; i < this.data.length; i++) {
+        this.data[i] = this.data[i].map(d => d + 1);
+      }
+    }
     const domain = this.getDomain();
 
-    let x = scaleLinear()
+    let x = this.options.xScaleLog ? scaleLog() : scaleLinear();
+    x
```
MarkAZhang: Can you use the clamp function as shown here?
https://stackoverflow.com/questions/11322651/how-to-avoid-log-zero-in-graph-using-d3-js
You can clamp the domain to [1, max(maxs)]

This way, you can remove all the special cases around adding +1 for log.

j-x-han:
@tfrcarvalho Just pushed a commit to remove the 2eX labels and updated the picture accordingly. Also added screenshots to show the z-scores!

app/assets/src/components/visualizations/Histogram.js
MarkAZhang: Hey Julie, you can change the "getBarWidth" function to take a bin.

Then you can do

```
getBarWidth = bin => {
  if (this.options.xScaleLog) {
    return (x(bin.x1) - x(bin.x0)) / 2
  }
  
  // previous code here
}
```
 
```
.attr("x", d => x(d.x0) + i * getBarWidth(d))
.attr("width", d => getBarWidth(d))
```

app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
```diff
@@ -184,7 +184,7 @@ export default class TaxonDetailsMode extends React.Component {
       this.state.histogramRpmSeries,
       {
         seriesNames: ["NT", "NR"],
-        labelX: "rpm",
+        labelX: "rpm+1",
```
MarkAZhang: I don't really like `rpm+1`. It adds additional confusion for the user. I would try out the clamp approach and see if that works. The clamp approach will basically treat 0 the same as 1 (so it's impossible to tell them apart in the graph) but that should be fine for our purpose.

tfrcarvalho: @MarkAZhang Agree it is a bit confusing, but it is however, scientifically inaccurate... so I would run it through @katrinakalantar

app/assets/src/components/visualizations/Histogram.js
tfrcarvalho: > By number of series, do you mean `this.data.length`?

Yes

app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
j-x-han: Just tested out reverting to `rpm` rather than `rpm+1` and using clamp, it looks like all the out of range values (<1) get dropped in `getBins()` and don't show up on the graph. Looking into that right now.

(This is the Ralstonia example used above.)
* **rpm+1, no clamp**
image

* **rpm, using clamp**
image

j-x-han: It seems that when generating bins for the histogram, values outside the domain (taken from the x scale) are [ignored](https://github.com/d3/d3-array/blob/v1.2.4/README.md#histogram_domain). Doesn't seem like there's an equivalent clamp function to use for histograms :/

tfrcarvalho: The other question if if we want to get rid of rpm+1 (which is more correct)

cdebourcy: I think it's important to have rpm=0 represented on this graph, i.e. prefer top graph over bottom graph. Often, the whole point will be to see how many samples did not contain the microbe at all, and how unusual it is that the sample does contain it. So dropping the lowest bin as shown in the bottom graph is not an option

cdebourcy: It would be very misleading.

app/assets/src/components/visualizations/Histogram.js
cdebourcy: Doesn't this mean the bin around 0 will be dropped entirely? Usually that bin contains the most information and dropping it would be very misleading

app/assets/src/components/common/DetailsSidebar/TaxonDetailsMode/TaxonDetailsMode.jsx
tfrcarvalho: > dropping the lowest bin as shown in the bottom graph is not an option

I believe the clamp idea is that the 0 rpm would actually return a value of 1 1 (so having 0 and 1 would be represented in the same bar). 
I guess the question is if people will understand the `rpm+1` label, which makes the chart correct.

cdebourcy: I see. I believe rpm+1 is fairly standard to deal with this issue in bioinformatics

app/assets/src/components/visualizations/Histogram.js
j-x-han: Took Mark's suggestion, thanks! I'll be sure to double check the coverage viz once this is on staging as well to make sure this doesn't affect anything there.

happyimadesignr:
Hey @j-x-han and team, I have a few suggestions! What do you all think?
* Change “rpm" to “rPM” - I was asking Katrina about this for the docs and she said we use rPM most consistently.
* Change "count" to "Number of Samples” to be more clear? Either way capitalize it.
* Should we change "sample" to “This sample”? Or "This sample (NT)”? To be more clear?
* Can we center the labels?
* Should we put a tool tip on the "rPM +1”? Unsure what that means initially.
* Can you match the key in the phylogenetic tree? The circle, size, and style to be consistent.
* Can we add an info icon and tooltip about this chart? Suggested copy with input from Tiago and Katrina: “This chart shows how abundant (rPM) this taxon is in your sample compared to the other samples in the chosen Background Model. The farther your sample line is to the right of the other samples, the higher the likelihood that this taxon is unique to your sample.”

cdebourcy:
> Hey @j-x-han and team, I have a few suggestions! What do you all think?
Thanks for the input!

> * Change “rpm" to “rPM” - I was asking Katrina about this for the docs and she said we use rPM most consistently.
Agree

> * Change "count" to "Number of Samples” to be more clear? Either way capitalize it.
I suggest "Number of Background Samples" to make it even clearer what this means! That will be a huge improvement. 

> * Should we change "sample" to “This sample”? Or "This sample (NT)”? To be more clear?
"This Sample" would be great.

> * Can we center the labels?
Could do that if it's not too much of a time sink for Julie :)

> * Should we put a tool tip on the "rPM +1”? Unsure what that means initially.
Let's hold off on that since its not defined yet, so that Julie can get her change in. In my experience as a developer, it is extremely frustrating having to wait for additional requirements to come in. If it is really necessary, it can be a follow-up PR. In any case, `rPM+1` doesn't leave any room for ambiguity, it's a precise definition.

> * Can you match the key in the phylogenetic tree? The circle, size, and style to be consistent.
Isn't the point that the nodes on the tree are circles and the bars here are rectangles though? So the key matches what's in the chart? To me, the current style makes more sense and I think we should keep it.

> * Can we add an info icon and tooltip about this chart? Suggested copy with input from Tiago and Katrina: “This chart shows how abundant (rPM) this taxon is in your sample compared to the other samples in the chosen Background Model. The farther your sample line is to the right of the other samples, the higher the likelihood that this taxon is unique to your sample.”
I like it, perhaps once we add this people will actually know what a background model is :)

happyimadesignr:
I'm so glad we can make the wording more clear and give descriptions!! =D

- Regarding the key, interesting point about matching the shape. We have circles for the key in a few places like with the heatmap key (though I see now it reverted) and phylogenetic tree key, I think they should be consistent. I can look into how other people do this.
- Let's change the key box sizing thought to 14px width and height and make the labels 600 weight for consistency.

j-x-han:
Style and label changes have been made, as well as adding the tooltip explaining the chart. Screenshots have been updated to show the changes!

PR #2600, opened 2019-09-25
Pipeline monitor thread pooling + log improvements

morsecodist:
# Description

This implements the thread pooling earlier but fixes the error it caused. The error it caused was caused by writing too long a string to the DB. Switching to the pooled connection means this string was no longer auto truncated and instead it threw an error. The logging here also didn't log the error messages so I added that as well.

# Tests

* Ensured the benchmark worked locally
* Ran locally for many concurrent jobs


lib/tasks/pipeline_monitor.rake
```diff
@@ -3,6 +3,17 @@ require 'English'
 require 'json'
 require 'thread/pool'
 
+Rails.application.config.after_initialize do
+  ActiveRecord::Base.connection_pool.disconnect!
+
+  ActiveSupport.on_load(:active_record) do
+    config = ActiveRecord::Base.configurations[Rails.env] ||
+              Rails.application.config.database_configuration[Rails.env]
+    config['pool'] = 20
```
tfrcarvalho: Is this to match the THREAD_COUNT? Should we use the same constant?

lib/tasks/pipeline_monitor.rake
```diff
@@ -294,16 +296,19 @@ class CheckPipelineRuns
...
-      shard_id = 0
-      while shard_id < num_shards
-        pid = Process.fork { update_jobs(num_shards, shard_id, pr_ids, pt_ids) }
-        fork_pids << pid
-        shard_id += 1
+      pool = Thread.pool(THREAD_COUNT)
+      pr_ids.each do |prid|
+        pool.process do
+          update_pr(prid)
+          puts ActiveRecord::Base.connection_pool.stat()
```
tfrcarvalho: Forgot to remove?

lib/tasks/pipeline_monitor.rake
```diff
@@ -259,7 +261,7 @@ class CheckPipelineRuns
       begin
         create_sample_for_benchmark(s3path, pipeline_commit, web_commit, bm_pipeline_branch, bm_user, bm_proj, bm_host, bm_comment, t_now)
       rescue => exception
-        LogUtil.log_err_and_airbrake("Failed to create sample for benchmark #{s3path}")
+        LogUtil.log_err_and_airbrake("Creating sample for benchmark failed with exception: #{exception.message}")
```
tfrcarvalho: Can you leave the s3path on the message anyway?

lib/tasks/pipeline_monitor.rake
morsecodist: Good catch thanks

PR #2601, opened 2019-09-25
Improve lineagedb rake task

gregdingle:
# Description

This improves the usability of this rake task. In following https://github.com/chanzuckerberg/idseq-web/wiki/%5BDev%5D-Index-Update-Instructions, I wanted to understand better the lineage update step. As I played with the code, I made some improvements.

Improvements:
* make version to default to plus one
* make script fail on failed inline shell command `set -e`
* make script print all inline shell commands `set -x`
* add `dryrun` mode
* fix db creds to work in local development (not sure when this broke)
* refactored into two methods
* added some print statements
* added simpler `NCBI_DATE` task param
* added rake task description

# Tests

* run with and without `NCBI_DATE`: `NCBI_DATE=2019-09-17 rails update_lineage_db`
* run with and without dryrun: `NCBI_DATE=2019-09-17 rails update_lineage_db[dryrun]`
* check version increments with dryrun
* force bad database creds and observe early exit
* run to completion once on actual data in local env


lib/tasks/update_lineagedb.rake
```diff
@@ -1,30 +1,67 @@
...
+                      end
+
+  puts "\n\nStarting import of #{reference_s3_path} ...\n\n"
+  import_lineage_database!(reference_s3_path, current_date) unless args.dryrun
+  puts "\n\nDone import of #{reference_s3_path}."
 
-  reference_s3_path = ENV['REFERENCE_S3_FOLDER'].gsub(%r{([/]*$)}, '') # trim any trailing '/'
   current_lineage_version = ENV['LINEAGE_VERSION'].to_i
+  if current_lineage_version.zero?
+    current_lineage_version = AlignmentConfig.last.lineage_version + 1
```
cdebourcy: We should use `AlignmentConfig.max("lineage_version")` here rather than `AlignmentConfig.last.lineage_version` to guarantee it's an increment. 

`AlignmentConfig.last` could be a revert of a previously generated AlignmentConfig with higher lineage_version, for example. We shouldn't introduce a requirement that the last AlignmentConfig is the one currently in use or the one with the most recent NCBI data

gregdingle: Good point. I was just following what was in the guide.

lib/tasks/update_lineagedb.rake
```diff
@@ -1,30 +1,67 @@
...
    ## Set work directory
    mkdir -p #{local_taxonomy_path};
    cd #{local_taxonomy_path};
 
+   ## Check database connection
+   mysql -h #{host} #{lp} -e "SELECT 1"
+
    ## Get old lineage file
-   mysql -h #{host} -u $DB_USERNAME --password=$DB_PASSWORD -e "SELECT #{column_names},started_at FROM idseq_#{Rails.env}.taxon_lineages WHERE ended_at = (SELECT MAX(ended_at) FROM idseq_#{Rails.env}.taxon_lineages);" | tr "\t" "," | tail -n +2 > old_taxon_lineages_with_started_at.csv
+   mysql -h #{host} #{lp} -e "SELECT #{column_names},started_at FROM idseq_#{Rails.env}.taxon_lineages WHERE ended_at = (SELECT MAX(ended_at) FROM idseq_#{Rails.env}.taxon_lineages);" | tr "\t" "," | tail -n +2 > old_taxon_lineages_with_started_at.csv
```
jshoe: what does lp stand for?

PR #2602, opened 2019-09-25
Rake task to create new alignment config

gregdingle:
# Description

This is a necessary step for using newly generated indexes in IDseq. The rake task simplifies the step and checks that the files exist. 

# Tests

* run on existing files `NCBI_DATE=2019-09-17 rails create_alignment_config`, see save message
* run on non-existing files `NCBI_DATE=2019-09-18 rails create_alignment_config`, see error message


lib/tasks/create_alignment_config.rake
```diff
@@ -0,0 +1,48 @@
+desc 'Creates a new AlignmentConfig for a set of indexes'
+
+task create_alignment_config: :environment do
+  name = ENV['NCBI_DATE'] # YYYY-MM-DD
+  raise "Must have a $NCBI_DATE" unless name
+
+  # lineage_version should be in sync with that used in lib/tasks/update_lineagedb.rake
+  lineage_version = ENV['LINEAGE_VERSION'] || AlignmentConfig.last.lineage_version + 1
```
cdebourcy: Should we throw an error if user doesn't supply LINEAGE_VERSION so we're sure the user is conscious of the choice? As you say, it needs to be in sync with that used in update_lineagedb.rake. If we must have a default, should we favor AlignmentConfig.max("lineage_version") over AlignmentConfig.last.lineage_version? I can imagine a scenario in which the last AlignmentConfig is a revert of a previous one with higher version number, for example.

gregdingle: That's a good point. What do you think of using this as the default?

```
TaxonLineage.maximum('version_start')
```
That way you will at least get a version that has data.

cdebourcy: This won't achieve the desired goal because in principle you cannot assume that the latest lineage_version is represented as a `version_start`. Only new TaxonLineage records get a `version_start` -- we actually expect that the bulk of the TaxonLineage table will remain untouched. I.e. version_start of taxid 562 still has version_start == 1 because it wasn't touched during the updates to versions 2 and 3. (Maybe it helps if I clarify how the versions are used: the app check whether alignment_config.lineage_version is between taxon_lineage.version_start and taxon_lineage.version_end to decide whether it should use taxon_lineage -- it's range-based.)

I think we should force the operator to think about the version rather than selecting one silently. Thoughts on how to select one automatically should go into the notes on how to automate the whole process. Part of the challenge with the index update automation will be to figure out how to coordinate all these loose ends

gregdingle: I updated it to throw an error. 

And thanks for the explaining how the versioning is supposed to work. I couldn't find any docs on it anywhere.

gregdingle: By the way, I started a doc here: https://czi.quip.com/QwMWAe4JXwu2/2019-09-25-Plan-to-automate-IDseq-index-generation .

PR #2604, opened 2019-09-26
IDSEQ-1151 - Start switching dag templates from erb to jbuilder

jshoe:
### Description
- JIRA: https://jira.czi.team/browse/IDSEQ-1151
- We decided to switch the dag templates from ERB to Jbuilder, which is the default Rails tool for constructing JSON responses. This makes constructing the dags cleaner and less fault-prone.
- This is the first PR to do setup work + convert the host_filter dag to start.

### Notes
- The code will work with both .jbuilder and .erb now, although after rollout we can discard the old code.
- For operational safety, I decided to try and flag it with `AppConfig::USE_JBUILDER_TEMPLATES`. After I add all the next templates, I'll start to switch it on to roll out.
- To verify the template, you should read side-by-side with the original file: https://github.com/chanzuckerberg/idseq-web/blob/master/app/lib/dags/host_filter.json.erb

### Tests
- Verified a variety of samples in local development and compared the outputs (ex: `Sample.find(i).pipeline_runs[0].pipeline_run_stages[0].host_filtering_command`). Outputs were identical with tools like jsondiff.com.
- Submitted multiple local samples and had successful host filter runs: https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logEventViewer:group=/aws/batch/job;stream=aegea_batch_job_57d721c3/default/f463c0da-dace-4c99-a4f9-95998f8cadb9

### DIFF
--- a/app/lib/dag_generator.rb
+++ b/app/lib/dag_generator.rb
@@ -23,7 +23,7 @@ class DagGenerator
     if File.extname(@template_file) == ".jbuilder"
       rendered = ActionController::Base.new.render_to_string(
         file: @template_file,
-        :locals => {:attr => @attribute_dict},
+        locals: { attr: @attribute_dict }
       )
       # Couldn't find a more direct way to prettify from template
       json_object = JSON.parse(rendered)
--- a/app/lib/dag_generator.rb
+++ b/app/lib/dag_generator.rb
@@ -21,10 +21,13 @@ class DagGenerator
 
   def render
     if File.extname(@template_file) == ".jbuilder"
-      ActionController::Base.new.render_to_string(
+      rendered = ActionController::Base.new.render_to_string(
         file: @template_file,
         :locals => {:attr => @attribute_dict},
       )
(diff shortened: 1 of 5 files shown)


jshoe:
> This is awesome! Could you add comments in dag generator to point developers very explicitly to the corresponding jbuilder file? In the past, I have found the magic around where the .jbuilder file lives and when it gets used to be incredibly confusing for newcomers

added 👍

app/lib/dag_generator.rb
```diff
@@ -7,13 +7,32 @@ def initialize(template_file, project_id, sample_id, host_genome, attribute_dict
...
+    attribute_dict[:sample_id] = sample_id
+
     @attribute_dict = attribute_dict
     @dag_vars_dict = dag_vars_dict
   end
 
+  # See our dag templates in app/lib/dags.
   def render
-    ERB.new(@template).result(binding)
+    # Interoperable between .jbuilder and .erb templates
```
davidrissato: Nice

PR #2605, opened 2019-09-26
[Data Discovery] Return list of all project ids to enable pagination

tfrcarvalho:
### Description

* Returns list of project ids
* Turns out that to return the list in an efficient way, it was necessary to refactor projects a bit. Moving all the filtering above and expanding data for projects below. Also trying to reduce number of queries necessary to compute the aggregations.

### Tests

* All previous unit and integrations tests pass
* Added two integration tests to test presence/absence of list all ids
* Checked upload page and projects pages that use this endpoint

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -90,7 +90,6 @@ class ProjectsController < ApplicationController
           projects = projects.where(samples: { id: filter_samples(current_power.samples, params) })
         end
         projects = projects.order(Hash[order_by => order_dir])
-
         limited_projects = limit ? projects.offset(offset).limit(limit) : projects
 
         if basic
@@ -99,7 +98,7 @@ class ProjectsController < ApplicationController
           ]
           names = attrs.map { |attr| attr.split(' AS ').last }
           render json: {
-            projects: limited_projects.pluck(*attrs).map { |p| names.zip(p).to_h },
+            projects: limited_projects.group(:id).pluck(*attrs).map { |p| names.zip(p).to_h },
           }
         else
           limited_projects = limited_projects
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -94,7 +94,6 @@ class ProjectsController < ApplicationController
         limited_projects = limit ? projects.offset(offset).limit(limit) : projects
 
         if basic
-          limited_projects = limited_projects
           attrs = [
             'id', 'name', 'description', 'created_at', 'public_access', 'COUNT(DISTINCT samples.id) AS number_of_samples',
           ]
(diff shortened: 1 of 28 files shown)


app/controllers/projects_controller.rb
```diff
@@ -62,94 +62,78 @@ def index
...
+          render json: {
+            projects: limited_projects.pluck(*attrs).map { |p| names.zip(p).to_h },
+          }
+        else
+          limited_projects = limited_projects
+                             .includes(samples: [:host_genome, :user, { metadata: [:metadata_field, :location] }])
+                             .group(:id)
+                             .references(:samples)
+
+          group_concat_host = "GROUP_CONCAT(DISTINCT host_genomes.name SEPARATOR '::') AS hosts"
```
jshoe: maybe a comment on what GROUP_CONCAT does?

app/controllers/projects_controller.rb
```diff
@@ -62,94 +62,78 @@ def index
...
-            number_of_samples: sample_count_by_project_id[project.id] || 0,
-            hosts: host_genome_names_by_project_id[project.id] || [],
-            tissues: tissues_by_project_id[project.id] || [],
-            owner: owner_by_project_id[project.id],
-            locations: locations_by_project_id[project.id] || [],
-            editable: updatable_projects.include?(project.id),
-            users: updatable_projects.include?(project.id) ? project.users.map { |user| { name: user[:name], email: user[:email] } } : []
-          )
+        if basic
+          limited_projects = limited_projects
```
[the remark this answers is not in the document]
tfrcarvalho: You never know... :) Removed.

app/controllers/projects_controller.rb
tfrcarvalho: Added

app/controllers/projects_controller.rb
```diff
@@ -62,94 +62,76 @@ def index
...
+          }
+        else
+          limited_projects = limited_projects
+                             .includes(samples: [:host_genome, :user, { metadata: [:metadata_field, :location] }])
+                             .group(:id)
+                             .references(:samples)
+          # get aggregated lists of association values in string by using MySQL's GROUP_CONCAT (should update to JSON_ARRAYAGG when possible)
+          group_concat_host = "GROUP_CONCAT(DISTINCT host_genomes.name SEPARATOR '::') AS hosts"
+          group_concat_sample_type = "GROUP_CONCAT(DISTINCT CASE WHEN metadata_fields.name = 'sample_type' THEN metadata.string_validated_value ELSE NULL END SEPARATOR '::') AS tissues"
+          group_concat_location = "GROUP_CONCAT(DISTINCT CASE WHEN metadata_fields.name = 'collection_location' THEN IFNULL(locations.name, metadata.string_validated_value) ELSE NULL END SEPARATOR '::') AS locations"
```
davidrissato: It might be too late for this review (looks like it is already merged), but since this field is based on user input, we cannot guarantee it won’t have “::” sequence, unless we sanitize that input. Same for fields sample type, emails and eventually user name.

tfrcarvalho: This was discussed, but facing the lack of a better alternative we moved forward with this. I believe we validate at least user name and email, but not sample_type. 
As the comment suggests, I would like to use `JSON_ARRAYAGG` as soon as we make it available.

The worst case that can happen with this is some data being cut short or split too much. If we have `::` on our data, then that is an issue in itself and we should clean anyway.

I plan to revisit this in a future PR (next week), but needed to push the functionality. One possibility is actually using MySQL's `REPLACE()` to sub `::` by something else

PR #2607, opened 2019-09-26
Switch back to processes

morsecodist:
# Description

The error with this happened because rake tasks don't eagerly load for performance reasons in prod/staging. This causes issues with autoloading when you multi-thread. I switched back to processes. I think we will still get performance gains from increasing the connection pool size to the db and from the smarter assignment of tasks to processes by the library.

# Tests

*  Ran monitor locally

### DIFF
--- a/lib/tasks/pipeline_monitor.rake
+++ b/lib/tasks/pipeline_monitor.rake
@@ -1,9 +1,9 @@
 # Jos to check the status of pipeline runs
 require 'English'
 require 'json'
-require 'thread/pool'
+require 'parallel'
 
-THREAD_COUNT = 20
+PROCESS_COUNT = 20
 
 Rails.application.config.after_initialize do
   ActiveRecord::Base.connection_pool.disconnect!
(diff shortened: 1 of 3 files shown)


gregdingle:
LGTM but I will leave someone familiar with the previous work to approve.

lib/tasks/pipeline_monitor.rake
```diff
@@ -296,18 +296,15 @@ class CheckPipelineRuns
       t_iter_start = t_now
       pr_ids = PipelineRun.in_progress.pluck(:id)
       pt_ids = PhyloTree.in_progress.pluck(:id)
-      pool = Thread.pool(THREAD_COUNT)
-      pr_ids.each do |prid|
-        pool.process do
-          update_pr(prid)
-        end
+      Parallel.each(pr_ids, in_processes: PROCESS_COUNT) do |prid|
+        @reconnected ||= ActiveRecord::Base.connection.reconnect! || true
```
tfrcarvalho: Sorry, what is the `|| true` supposed to do?

morsecodist: This makes it only attempt reconnect once if it fails I believe it's from the docs. I will switch back to the old way which just has the reconnect, it worked before.

PR #2608, opened 2019-09-26
[Bug:JIRA-1468] Refresh project page changes selected tab

tfrcarvalho:
### Description

This bug was created due to a previous change that did not save the last active tab to local storage if in the project page. This aims to guarantee that when we click a different domain (my data, public) we do remain in the samples page.

This fix maintains that behavior but saves the active tab to url.

# Tests

* Click a project
* Refresh and check that it remains in the samples tab and shows list of samples
* Click some other domain. Check that you it will display the projects tab for that domain.


gregdingle:
Will this also cover url sharing between users?

tfrcarvalho:
> Will this also cover url sharing between users?

Yes, everything that gets saved on url fields it saved on the shared link.

PR #2615, opened 2019-09-30
Add read counts table to sample details sidebar

j-x-han:
# Description

Add a table to the Pipeline tab in the Sample Details Sidebar displaying the number of reads remaining after each step in the pipeline run.
Screen Shot 2019-10-01 at 4 35 14 PM

The `Host Filtering Step` label can be hovered over to display a tooltip that will link to [here](https://github.com/chanzuckerberg/idseq-dag/wiki/IDseq-Pipeline-Stage-%231:-Host-Filtering-and-QC) to provide more information about host filtering and the steps involved.
Screen Shot 2019-10-01 at 4 35 28 PM

The name of each step can be hovered over to display a tooltip briefly describing the step.
Screen Shot 2019-09-30 at 4 09 12 PM

# Notes

If the sample failed or the read counts were otherwise unable to be retrieved, `No data` will be displayed instead.
Screen Shot 2019-10-01 at 4 34 50 PM

# Tests

* Verified that the table correctly displays the number of reads remaining after each step in the pipeline.
* Verified that the table displays placeholder value if the read counts were unable to be fetched.


app/assets/src/components/common/DetailsSidebar/SampleDetailsMode/constants.js
```diff
@@ -51,3 +51,42 @@ export const PIPELINE_INFO_FIELDS = [
     key: "lastProcessedAt",
   },
 ];
+
+export const PIPELINE_RUN_STEPS = [
```
cdebourcy: Is there a way to avoid hardcoding these? If we change the steps in the pipeline in the future, they could go out of sync with this constant

cdebourcy: @tfrcarvalho How did Emily handle this for pipeline viz again? Did she just parse the underscores and re-capitalize to turn "validate_input_out" into "Validate Input"? That would seem like a good solution

tfrcarvalho: Basically yes: https://github.com/chanzuckerberg/idseq-web/blob/master/app/services/retrieve_pipeline_viz_graph_data_service.rb#L238

app/assets/src/components/common/DetailsSidebar/SampleDetailsMode/sample_details_mode.scss
```diff
@@ -37,6 +37,31 @@
...
+
+  .narrowMetadataValueContainer {
+    width: 100px;
+    margin-right: 15px;
+    display: inline-flex;
+    text-align: right;
+    justify-content: flex-end;
+
+    .labelText {
+      font-weight: 600;
```
tfrcarvalho: Should this have a typography define font?

app/assets/src/components/common/DetailsSidebar/SampleDetailsMode/sample_details_mode.scss
```diff
@@ -66,6 +91,7 @@
 
   .emptyValue {
     color: $light-grey;
+    font-weight: 400;
```
tfrcarvalho: Same as before... Even if not a full font, you should use `$font-weight-regular`

app/assets/src/components/common/DetailsSidebar/SampleDetailsMode/PipelineTab.jsx
```diff
@@ -158,10 +159,13 @@ class PipelineTab extends React.Component {
...
                     position="top left"
                     trigger={
-                      <div className={cs.labelText}>Host Filtering Step</div>
+                      <div className={cx(cs.labelText, cs.header)}>
+                        Host Filtering Step
+                      </div>
                     }
                     content={this.state.pipelineStepDict["stage_description"]}
                     title="Host Filtering"
+                    link="https://github.com/chanzuckerberg/idseq-dag/wiki/IDseq-Pipeline-Stage-%231:-Host-Filtering-and-QC"
```
tfrcarvalho: Can we put this is a constant somewhere?

PR #2617, opened 2019-10-01
[WIP] Update docker-compose default alignment config

gregdingle:
# Description

This updates the default when there is no env var, following the updates of NCBI indexes.

# Notes

* This should wait until we have fully verified that 2019-09-17 is good in staging and prod.

* I'm also wondering how we should update our dev environments. Maybe we should be using `db/seeds.rb`?

# Tests

None


gregdingle:
Please review now @cdebourcy .

PR #2618, opened 2019-10-01
Add detailed timing instrumentation and apply to dimensions endpoints

tfrcarvalho:
# Description

* Adds a timer util class that helps tracking time over multiple substeps.
* Timer uses common util functions to publish metrics
* Enable timer for projects and samples controller actions and detailed monitoring to dimensions endpoints

Screen Shot 2019-10-01 at 12 39 45 PM

# Tests

* Run get requests to dimensions endpoints. Check that metrics are correctly published to Datadog (requires api key).
* Unit tests


app/controllers/application_controller.rb
```diff
@@ -125,4 +125,10 @@ def set_current_context_for_logging!
   rescue => e
     Rails.logger.error(e)
   end
+
+  def instrument_with_timer
+    @timer = Timer.new("#{params[:controller]}.#{params[:action]}")
```
gregdingle: any chance that `@timer` could be created twice, overwriting the first instance?

tfrcarvalho: I believe each action is called only once. Are you thinking of any specific scenario?
We use this pattern to fill in other parameters like `@sample`...

gregdingle: No, it just requires the programmer to understand that. Maybe add a comment "This should only be called once per request" or put 

`raise "must be called only once" unless @timer.nil?`

tfrcarvalho:
> You can also replace the timer info I put in the app/lib/report_helper.rb

Yeah, I will probably do that on a follow up PR. I want to make a parallel between those measurements and the new report sample service I was working on.

app/controllers/projects_controller.rb
```diff
@@ -39,6 +39,8 @@ class ProjectsController < ApplicationController
   before_action :check_access
   before_action :no_demo_user, only: [:create, :new]
 
+  around_action :instrument_with_timer
```
jshoe: 👍

app/lib/timer.rb
```diff
@@ -0,0 +1,41 @@
...
+  def add_tags(tags)
+    @tags += tags
+  end
+
+  def split(name)
+    timestamp = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    @splits << [name, timestamp]
+
+    if @splits.length > 1
+      elapsed = timestamp - (@splits.length > 1 ? @splits[-2][1] : @start_timestamp)
```
jshoe: `if @splits.length > 1` and then checking again `@splits.length > 1 ?`  (?)

app/lib/timer.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def publish
+    Rails.logger.debug("Timer for #{@prefix}")
+    end_timestamp = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    Rails.logger.debug("Timer:Total: #{end_timestamp - @start_timestamp}")
+    MetricUtil.put_metric_now("#{@prefix}.timer", end_timestamp - @start_timestamp, tags: @tags)
+
+    previous_timestamp = @start_timestamp
+    @splits.each do |name, split_timestamp|
+      Rails.logger.debug("Timer:CP[#{@prefix}.#{name}]: #{split_timestamp - previous_timestamp}")
```
jshoe: does CP mean clock prefix?

tfrcarvalho: Good catch... this meant checkpoint, but since then I changed it to split. Will fix.

app/lib/timer.rb
tfrcarvalho: Good catch, this was just for debug since I log everything in the end.

app/controllers/application_controller.rb
tfrcarvalho: It is not "must" but a "should"... thus added logging.
Do not want instrumentation to be a blocker and thus prefer not raising exception, since the application will not get into an error state.

PR #2620, opened 2019-10-02
IDSEQ-1442 Fix buggy Basespace upload counter.

MarkAZhang:
Also fix header text alignment on upload table.

# Description

Before:
Screen Shot 2019-10-01 at 6 21 23 PM

After:
Screen Shot 2019-10-01 at 6 19 57 PM

# Notes

This was caused by buggy logic around handling the selected sample ids when merging in new samples. This PR fixes the bad logic.

I also added comments and tried to rename the variables a bit to make it more clear.

A recent change also caused the table header to be center aligned instead of left aligned. Fixed that as well.

# Tests
* Verified that clicking "Connect to Project" multiple times doesn't increase the upload counter incorrectly.
* Verified that if Sample 1 is selected, Sample 2 is not selected, and I try to add Sample 1, Sample 2, and Sample 3 (via Connect to Project), the counter shows "3 samples" and Sample 2 is still not selected (as expected)
* Verified that table headers are correctly aligned for all 3 sample upload tables (local, S3, basespace)

### DIFF
--- a/app/assets/src/components/views/SampleUploadFlow/UploadSampleStep.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/UploadSampleStep.jsx
@@ -26,6 +26,7 @@ import {
   difference,
   union,
   uniqueId,
+  intersection,
 } from "lodash/fp";
 
 import PropTypes from "~/components/utils/propTypes";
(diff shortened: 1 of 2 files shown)


app/assets/src/components/views/SampleUploadFlow/UploadSampleStep.jsx
```diff
@@ -531,35 +532,42 @@ class UploadSampleStep extends React.Component {
...
 
-    const newSelectedSampleIds = new Set(
-      union(
-        Array.from(this.state[selectedSampleIdsKey]),
-        map(SELECT_ID_KEY, validatedSamples)
-      )
+    const mergedSampleIds = map(SELECT_ID_KEY, mergedSamples);
+
+    // Get the selectIds of all newly added samples that weren't already added.
+    const dedupedNewSampleIds = intersection(
```
tfrcarvalho: Why call this deduped? according to the description its just `newlyAddedSampleIds`.

MarkAZhang: Makes sense! Will fix.

PR #2621, opened 2019-10-02
IDSEQ-1151 - Finish changing dag template configs to jbuilder templates

jshoe:
### Description
- This adds jbuilder templates for the alignment, postprocess, and experimental stages, as well as the phylo tree job command.
- JIRA: https://jira.czi.team/browse/IDSEQ-1151
- Continuation to: https://github.com/chanzuckerberg/idseq-web/pull/2604

### Notes
- To verify the template, you can read side-by-side with the original erb files: https://github.com/chanzuckerberg/idseq-web/blob/master/app/lib/dags

### Tests
- Verified a variety of samples in local development and compared the erb vs. jbuilder outputs in JSON differ. Ex:
```
Sample.find(i).pipeline_runs[0].pipeline_run_stages[0].host_filtering_command
Sample.find(i).pipeline_runs[0].pipeline_run_stages[1].alignment_command
Sample.find(i).pipeline_runs[0].pipeline_run_stages[2].postprocess_command
Sample.find(i).pipeline_runs[0].pipeline_run_stages[3].experimental_command
PhyloTree.find(i).send(:job_command)
```
- Currently running local samples and planning to run additional samples in staging.


app/models/pipeline_run_stage.rb
```diff
@@ -216,7 +216,7 @@ def prepare_dag(attribute_dict, key_s3_params = nil)
     attribute_dict[:bucket] = SAMPLES_BUCKET_NAME
 
     # Temp flag for rolling out jbuilder templates
-    dag_ext = if AppConfigHelper.get_app_config(AppConfig::USE_JBUILDER_TEMPLATES) == "1" && step_number == 1
+    dag_ext = if AppConfigHelper.get_app_config(AppConfig::USE_JBUILDER_TEMPLATES) == "1"
```
davidrissato: Would it make sense to completely remove the `.erb` files and this temp flag since you are now moving all these files to jbuilder?

jshoe: I put it there for operational safety to avoid a hotfix if we discover a problem.. :)

jshoe: but will do a follow-up to remove them maybe 1 week after it all goes to prod

app/lib/dags/experimental.json.jbuilder
```diff
@@ -0,0 +1,164 @@
...
+    json.s3_dir "s3://#{attr[:bucket]}/samples/#{attr[:project_id]}/#{attr[:sample_id]}/postprocess/#{attr[:pipeline_version]}/assembly"
+  end
+
+  json.fastqs do
+    json.s3_dir "s3://#{attr[:bucket]}/samples/#{attr[:project_id]}/#{attr[:sample_id]}/fastqs"
+  end
+
+  json.nonhost_fasta do
+    json.s3_dir "s3://#{attr[:bucket]}/samples/#{attr[:project_id]}/#{attr[:sample_id]}/postprocess/#{attr[:pipeline_version]}/assembly"
+  end
```
MarkAZhang: I find this a little harder to read than
```
nonhost_fasta: {
  s3_dir: "s3://...",
}
```
The jbuilder files do feel a little harder to read than the erb ones. It's harder to see where the object keys are at a glance and it feels a little more verbose. Anything you can do to organize the file better would be appreciated, but it's not urgent.

jshoe: Unfortunately it won't let me change the nested hash to just:
```
json.nonhost_fasta { s3_dir: "s3://..." }
```
but I find this syntax about the same after getting used to it.

On balance I think the jbuilder construction is cleaner and avoids constructs like:
```
      "in": [
        <% @attribute_dict[:taxon_byteranges].keys.each_with_index do |pipeline_run_id, idx| %>
          "prepare_taxon_fasta_<%= pipeline_run_id %>_out"
          <% unless idx == @attribute_dict[:taxon_byteranges].size - 1 %>
            ,
          <% end %>
        <% end %>
      ],
```
```
    "host_filter_out": [
      "gsnap_filter_1.fa"
      <% if @attribute_dict[:input_file_count] > 1 %>
        , "gsnap_filter_2.fa"
        , "gsnap_filter_merged.fa"
      <% end %>
    ],
```
and inconsistent formatting in the generated files (as well as the main benefits of string handling). For overall verbosity I think jbuilder could be even more DRY in re-using variables, but I'm hearing that people want the "at a glance" template style as you do.

PR #2622, opened 2019-10-02
Add info_db to alignment config.

MarkAZhang:
# Description

Previously, we were hard-coding the info_db s3 path when passing it to the coverage viz step. Now we pass it dynamically. We also add a default if no info_db value can be found.

Also change the create alignment config rake task to populate info_db.

# Notes

If we want to change the info_db from sqlite3 to BDB in the future, we can modify the coverage viz step to accept both file formats temporarily. This doesn't preclude that change from happening in the future.

If @jshoe's PR 2621 gets in before mine, will need to update the jbuilder as well.

# Tests

* Verified that rake task works.
* Verified that the default is used when the current alignment config has no info_db value.

Need to test with additional samples in staging, including the samples that were failing before.

### DIFF
--- a/app/lib/dags/experimental.json.erb
+++ b/app/lib/dags/experimental.json.erb
@@ -113,7 +113,7 @@
       "class": "PipelineStepGenerateCoverageViz",
       "module": "idseq_dag.steps.generate_coverage_viz",
       "additional_files": {
-        "info_db": "s3://idseq-database/alignment_data/2018-12-01/nt_info.sqlite3"
+        "info_db": "<%= @attribute_dict[:nt_info_db] %>"
       },
       "additional_attributes": {}
     },
--- a/db/migrate/20191002195108_add_nt_info_to_alignment_config.rb
+++ b/db/migrate/20191002195108_add_nt_info_to_alignment_config.rb
@@ -1,0 +1,5 @@
+class AddNtInfoToAlignmentConfig < ActiveRecord::Migration[5.1]
+  def change
+    add_column :alignment_configs, :s3_nt_info_db_path, :text
+  end
+end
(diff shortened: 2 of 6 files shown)


app/lib/dags/experimental.json.erb
```diff
@@ -113,7 +113,7 @@
       "class": "PipelineStepGenerateCoverageViz",
       "module": "idseq_dag.steps.generate_coverage_viz",
       "additional_files": {
-        "info_db": "s3://idseq-database/alignment_data/2018-12-01/nt_info.sqlite3"
+        "info_db": "<%= @attribute_dict[:nt_info_db] %>"
```
jshoe: just merged mine, so in the jbuilder you can just change to `info_db: attr[:nt_loc_db]` as well

db/migrate/20191002195108_add_nt_info_to_alignment_config.rb
```diff
@@ -0,0 +1,5 @@
+class AddNtInfoToAlignmentConfig < ActiveRecord::Migration[5.1]
+  def change
+    add_column :alignment_configs, :s3_nt_info_db_path, :text
```
gregdingle: please add a comment such as ```

add_column :alignment_configs, :s3_nt_info_db_path, :text, comment: "Used for Reference Length and Reference Name  in the coverage viz. Column added 2019-10-03."
```

File: lib/tasks/create_alignment_config.rake [hunk 3]
```diff
@@ -21,6 +21,9 @@ task create_alignment_config: :environment do
     s3_nt_loc_db_path: "#{bucket}/alignment_data/#{name}/nt_loc.#{db_file_ext}",
     s3_nr_db_path: "#{bucket}/alignment_data/#{name}/nr",
     s3_nr_loc_db_path: "#{bucket}/alignment_data/#{name}/nr_loc.#{db_file_ext}",
+    # TODO(mark): Convert nt_info to use BDB.
```
gregdingle: It would be nice to start referencing jira tickets in TODOs (https://jira.czi.team/browse/IDSEQ-1536) though that crosses the open and closed source boundary.

db/migrate/20191002195108_add_nt_info_to_alignment_config.rb
MarkAZhang: Hey Greg, sorry I missed this comment. Will remember to add comments going forward.

PR #2624, opened 2019-10-03
Cleanup redundant :authenticate_user! invocations from controllers

davidrissato:
# Description

Cleanup redundant :authenticate_user! invocations from controllers. Adding these extra calls to `before_action authenticate_user!` changes the order of execution of that method, which can cause some issues with `current_power do ... end` blocks, specially when you have authentication using tokens.

# Notes

This cleanup is required as part of implementing auth0.

# Tests

* Automated tests passed
* Manual tests (including CLI)

### DIFF
--- a/app/controllers/metadata_controller.rb
+++ b/app/controllers/metadata_controller.rb
@@ -3,6 +3,7 @@ class MetadataController < ApplicationController
 
   # Token auth needed for CLI uploads
   TOKEN_AUTH_METHODS = [:validate_csv_for_new_samples].freeze
+  skip_before_action :verify_authenticity_token, only: TOKEN_AUTH_METHODS
   prepend_before_action :authenticate_user_from_token!, only: TOKEN_AUTH_METHODS
 
   def dictionary
--- a/app/controllers/metadata_controller.rb
+++ b/app/controllers/metadata_controller.rb
@@ -2,13 +2,8 @@ class MetadataController < ApplicationController
   include MetadataHelper
 
   # Token auth needed for CLI uploads
-  skip_before_action :verify_authenticity_token, only: [:validate_csv_for_new_samples]
-  before_action :authenticate_user!, except: [:validate_csv_for_new_samples]
-  before_action :authenticate_user_from_token!, only: [:validate_csv_for_new_samples]
-
-  current_power do # Needed for CLI
-    Power.new(current_user)
-  end
+  TOKEN_AUTH_METHODS = [:validate_csv_for_new_samples].freeze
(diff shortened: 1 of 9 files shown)


app/controllers/samples_controller.rb
```diff
@@ -31,11 +31,10 @@ class SamplesController < ApplicationController
                    :dimensions, :all, :show_sample_names, :cli_user_instructions, :metadata_fields, :samples_going_public,
                    :search_suggestions, :stats, :upload, :validate_sample_files,].freeze
   OWNER_ACTIONS = [:raw_results_folder].freeze
+  TOKEN_AUTH_ACTIONS = [:create, :update, :bulk_upload, :bulk_upload_with_metadata].freeze
```
tfrcarvalho: I saw that you did not remove the `verify_authenticity_token` on this controller.

app/controllers/metadata_controller.rb
```diff
@@ -2,13 +2,8 @@ class MetadataController < ApplicationController
   include MetadataHelper
 
   # Token auth needed for CLI uploads
-  skip_before_action :verify_authenticity_token, only: [:validate_csv_for_new_samples]
```
tfrcarvalho: What was the goal of this call? Why would it skip it? Is it done anywhere else?

davidrissato: Thank you for catching this. I cherry picked that commit from a different branch and it was a test that I wasn't supposed to commit. (we have an issue in our git pre-commit script that commit all the lines even when they are not added to the commit, but this is a different discussion).

It looks like it is done in different places. If I understood it correct, this is because CLI doesn't send CSRF tokens, and therefore all endpoints that are invoked by CLI need to disable some of these constraints.

Actually I really don't like the idea we have these endpoints shared between CLI and WEB, because it compromises the security. We should either have separate endpoints, or have better support to these bearer tokens (that's something that I'm planning to do during auth0 migration), or maybe both.

This can be even a larger discussion. For me, ideally, we would only use bearer tokens, and would have a single page app, but this is not rails works by default (it uses cookies to store authentication tokens and renders pages as you click them, and for that reason it requires that CSRF tokens approach, which are usually limited to post operations). However, such change would be huge, not sure if we can fit all this extra work in our auth0 tasks. Would love to discuss more about that.

davidrissato:
Testing new CLI, looks like I may need to fix something else.

app/controllers/samples_controller.rb
davidrissato: Actually, I shouldn't have removed the other ones.

davidrissato:
Tested CLI with success

PR #2630, opened 2019-10-03
Update links to Help Center

j-x-han:
# Description

Add/update all the links to point to the new help center!
* Username dropdown menu has additional option to the help center (below Create User, above Wiki)
* The `Learn More` links in the tooltips of the column headers for the samples table have been updated.
* The `Learn More` links in the tooltips of the column headers for the sample report page have been updated.
* The link to the CLI instructions in the sample upload flow has been updated.

# Notes

Links were updated according to https://czi.quip.com/0y0xAXGutQfR/Tooltips-and-Documentation-in-IDseq


app/assets/src/components/common/Header.jsx
```diff
@@ -172,7 +188,7 @@ const UserMenuDropDown = ({
         }
       />,
       <BareDropdown.Item
-        key="5"
+        key="6"
```
jshoe: this was there before and you don't need to change it, but for your future reference keys should be more unique https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

app/assets/src/components/views/SampleUploadFlow/SampleUploadFlowHeader.jsx
```diff
@@ -47,7 +47,7 @@ class SampleUploadFlowHeader extends React.Component {
                 <div className={cs.subtitle}>
                   Rather use our command-line interface?
                   <a
-                    href="/cli_user_instructions"
+                    href="https://help.idseq.net/hc/en-us/articles/360034790414-Upload-with-Command-Line"
```
jshoe: this should also change to use `<ExternalLink>` ;) I think you saw ESLint flag it `51:21  error  Using target="_blank" without rel="noopener noreferrer" is a security risk: see https://mathiasbynens.github.io/rel-noopener  react/jsx-no-target-blank`

PR #2632, opened 2019-10-04
[Release fix] Data Discovery bug fixes

tfrcarvalho:
# Description

* Fixes a crash when clicking on a project on the search box results
* Fixes a bug when selecting samples and changing to the map view on samples views
* Mask an issue where location hierarchy is inconsistent (not easy to reproduce in development)

# Tests

* Start searching for a project that will not show up on the first 50 results of the project page. Click on the result below and verify it does not crash.
* Select two or more samples and switch to the map tab on the Samples view. Verify that it does not crash.

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoveryView.jsx
+++ b/app/assets/src/components/views/discovery/DiscoveryView.jsx
@@ -859,10 +859,6 @@ class DiscoveryView extends React.Component {
   handleMapLevelChange = mapLevel => {
     const { rawMapLocationData, currentTab } = this.state;
 
-    if (!rawMapLocationData) {
-      return;
-    }
-
     const ids = currentTab === "samples" ? "sample_ids" : "project_ids";
     const clusteredData = {};
 
@@ -876,7 +872,7 @@ class DiscoveryView extends React.Component {
 
     const addToAncestor = (entry, ancestorLevel) => {
       const ancestorId = entry[`${ancestorLevel}_id`];
-      if (ancestorId) {
+      if (ancestorId && rawMapLocationData[ancestorId]) {
         if (!clusteredData[ancestorId]) {
           clusteredData[ancestorId] = copyLocation(
             rawMapLocationData[ancestorId]
--- a/app/assets/src/components/views/discovery/DiscoveryView.jsx
+++ b/app/assets/src/components/views/discovery/DiscoveryView.jsx
@@ -547,7 +547,10 @@ class DiscoveryView extends React.Component {
         break;
       }
       case "project": {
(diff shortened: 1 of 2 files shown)


app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -547,7 +547,9 @@ class DiscoveryView extends React.Component {
         break;
       }
       case "project": {
-        this.handleProjectSelected({ project: this.projects.get(value) });
+        this.handleProjectSelected({
```
MarkAZhang: I don't understand this part of the code, but this is confusing in isolation.

Here, it's okay to just set a "placeholder" project without any fields besides id? That doesn't lead to any errors where another component is expecting project to have more fields?

Maybe add a comment explaining why it's okay...

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -252,17 +252,9 @@ class SamplesView extends React.Component {
...
-      samples,
-      selectedSampleIds,
-    } = this.props;
+    const { samples, selectedSampleIds } = this.props;
 
-    // NOTE(jsheu): For mapSidebar sample names to appear in CollectionModal,
-    // they need to be presently loaded/fetched. Otherwise the ids work but says "and more..." for un-fetched samples.
-    const targetSamples =
-      currentDisplay === "map" ? mapPreviewedSamples : samples.loaded;
+    const targetSamples = samples.loaded;
```
MarkAZhang: This isn't removing a feature where the user selects some samples on the map and then goes to the sample list? Just making sure. Don't know the code.

MarkAZhang:
Could you add a comment on the line that's masking the inconsistent location hierarchy? (so we know that's still an outstanding issue.

app/assets/src/components/views/discovery/DiscoveryView.jsx
tfrcarvalho: Yes, it tries to load the whole project if possible to avoid requesting from the server and accelerate rendering.
If not send the id only and the project will be loaded from server.

app/assets/src/components/views/samples/SamplesView.jsx
tfrcarvalho: No, this samples for the sidebar were loaded independently from the samples for the main table. The last changes changed that, and all the data comes from the same collection now.

tfrcarvalho:
> Could you add a comment on the line that's masking the inconsistent location hierarchy

There will be a Jira ticket for that.

PR #2634, opened 2019-10-04
Set Batch job names

kislyuk:
# Description

Set job names for Batch jobs, like "idseq-staging-12345-align", so they can be easier to identify and track in the Batch/ECS API.

### DIFF
--- a/app/models/pipeline_run_stage.rb
+++ b/app/models/pipeline_run_stage.rb
@@ -262,7 +262,7 @@ class PipelineRunStage < ApplicationRecord
     batch_command = [install_pipeline(pipeline_run.pipeline_commit), upload_version(pipeline_run.pipeline_version_file), dag_commands].join("; ")
 
     # Dispatch job. Use the himem settings for host filtering.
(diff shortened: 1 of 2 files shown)


app/models/pipeline_run_stage.rb
```diff
@@ -262,7 +262,7 @@ def host_filtering_command
     batch_command = [install_pipeline(pipeline_run.pipeline_commit), upload_version(pipeline_run.pipeline_version_file), dag_commands].join("; ")
 
     # Dispatch job. Use the himem settings for host filtering.
-    aegea_batch_submit_command(batch_command, vcpus: Sample::DEFAULT_VCPUS_HIMEM, job_queue: Sample::DEFAULT_QUEUE_HIMEM, memory: Sample::HIMEM_IN_MB)
+    aegea_batch_submit_command(batch_command, vcpus: Sample::DEFAULT_VCPUS_HIMEM, job_queue: Sample::DEFAULT_QUEUE_HIMEM, memory: Sample::HIMEM_IN_MB, stage_name: "hostfilter", sample_id: sample.id)
```
jshoe: nit: could we change these to DAG_NAME_HOST_FILTER, DAG_NAME_ALIGNMENT, etc (also in this file)?

kislyuk: Fixed, thanks.

jshoe:
btw you can also test this locally in Rails console with commands like `Sample.last.pipeline_runs[0].pipeline_run_stages[0].host_filtering_command` `Sample.last.pipeline_runs[0].pipeline_run_stages[1].alignment_command`

PR #2640, opened 2019-10-07
[Release fix] Fix bugs in sidebar read counts table

j-x-han:
# Description

* Fix `NoMethodError` for `Waiting` samples caused by attempting to fetch the pipeline results file when a sample hasn't had a pipeline run yet.
* Fix bug for `Failed` samples caused by attempting to convert a `null` value (nonexistent read counts) into a string.

# Tests

* Open sample with `Waiting` status and open sample details sidebar to the pipeline tab, verify that error doesn't occur.
* Open sample with `Failed` status and open sample details sidebar to the pipeline tab, verify that page does not crash.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -981,7 +981,7 @@ class SamplesController < ApplicationController
   def results_folder
     can_see_stage1_results = (current_user.id == @sample.user_id)
     @exposed_raw_results_url = can_see_stage1_results ? raw_results_folder_sample_url(@sample) : nil
-    @file_list = @sample.first_pipeline_run ? @sample.first_pipeline_run.outputs_by_step(can_see_stage1_results) : nil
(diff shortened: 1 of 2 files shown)


app/controllers/samples_controller.rb
```diff
@@ -981,7 +981,7 @@ def raw_results_folder
   def results_folder
     can_see_stage1_results = (current_user.id == @sample.user_id)
     @exposed_raw_results_url = can_see_stage1_results ? raw_results_folder_sample_url(@sample) : nil
-    @file_list = @sample.first_pipeline_run.outputs_by_step(can_see_stage1_results)
+    @file_list = @sample.first_pipeline_run ? @sample.first_pipeline_run.outputs_by_step(can_see_stage1_results) : nil
```
tfrcarvalho: Should you return an empty list instead?

PR #2643, opened 2019-10-07
Adapt triggering of pipeline benchmark in pipeline monitor to new benchmarks

tfrcarvalho:
# Description

* Triggers the new format benchmarks.
* Maintains backward compatibility with previous benchmark format (to be removed once we stabilize the feature).
* We will keep to configuration in S3 until we deploy to production. We can delete the old configuration and the deprecated example then.

# Tests

* Ran pipeline monitor locally from a local configuration.
* Ran pipeline monitor from new S3 configuration and checked that both an old and a new benchmark was triggered.


config/idseq-bench-config.v2.json
```diff
@@ -0,0 +1,54 @@
...
+        " You can test and deploy changes to it by running                               ",
+        "                                                                                ",
+        "    aws s3 cp idseq-bench-config.v2.json s3://idseq-bench/config.v2.json        ",
+        "                                                                                ",
+        " This process allows changing webapp configuration without the cost             ",
+        " and disruption of a production push.                                           ",
+        "                                                                                ",
+        " The recommended process for deploying a new benchmark is as follows:           ",
+        "                                                                                ",
+        "    1. Instanll idseq-bench, create new benchmarks through configuration        ",
```
jshoe: sp: Instanll

lib/tasks/pipeline_monitor.rake
```diff
@@ -118,41 +119,59 @@ class CheckPipelineRuns
...
   end
 
-  def self.create_sample_for_benchmark(s3path, pipeline_commit, web_commit, bm_pipeline_branch, bm_user, bm_project, bm_host, bm_comment, t_now)
+  def self.benchmark_has_git_commit(metadata)
+    return metadata['idseq_bench_reproducible'] && metadata['idseq_bench_git_hash'].length == 40
+  end
+
+  def self.benchmark_has_configuration(metadata, s3_bucket, s3_key)
+    if metadata['iss_version'] && metadata['idseq_bench_version'] && metadata['prefix']
+      Rails.logger.debug("bucket: #{s3_bucket}, key: #{s3_key}/#{metadata['prefix']}.yaml")
```
jshoe: did you still want this debug line?

tfrcarvalho: no, thanks!

config/idseq-bench-config.v2.json
tfrcarvalho: Fixed

PR #2644, opened 2019-10-07
Refactor ToolbarIcons and improve hover styling.

MarkAZhang:
Pull popup text logic of various icons into ToolbarIcons.
Add circular hover shadow to toolbar icons and also popup arrows.

Screen Shot 2019-10-07 at 2 27 35 PM

# Notes

* Moved the `<Popup>` wrappers from various icons into `<ToolbarIcon>`, so that the icons can just be the `<svg>` element.
* `<ToolbarIcon>` has non-basic popups, and a circular hover shadow.
* All styling of icon dimensions moved to `<ToolbarIcon>`. For now, ToolbarIcon is intended to come in one size only (additional sizes can be added via a size prop). 

# Tests

* Verified that styling and behavior is correct, both when icon is enabled and disabled (when less than two samples are selected).


app/assets/src/components/views/samples/samples_view.scss
```diff
@@ -63,7 +63,7 @@
   .separator {
     flex: 0 0 auto;
     height: 27px;
-    margin: 0 20px;
+    margin: 0 10px 0 20px;
```
tfrcarvalho: Why do you need to reduce this?

app/assets/src/components/views/samples/samples_view.scss
```diff
@@ -73,31 +73,7 @@
...
-        }
-
-        &.heatmap {
-          margin-left: 4px;
-        }
-
-        &.download {
-          width: 28px;
-        }
+        margin-left: 2px;
```
tfrcarvalho: Why the reduction?

app/assets/src/components/views/samples/ToolbarIcon.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+import BasicPopup from "~/components/BasicPopup";
+import DownloadIcon from "~ui/icons/DownloadIcon";
+import HeatmapIcon from "~ui/icons/HeatmapIcon";
+import PhyloTreeIcon from "~ui/icons/PhyloTreeIcon";
+import SaveIcon from "~ui/icons/SaveIcon";
+
+import cs from "./toolbar_icon.scss";
+
+const ICON_COMPONENTS = {
```
tfrcarvalho: Why cant we receive an icon directly? Is it because they have different shapes and need adjustments? 
If it is the second, we should probably fix the icons.

app/assets/src/components/views/samples/samples_view.scss
MarkAZhang: This is because the size of the whole icon is much bigger now. Up from 21px to 42px. To include the gray hover circle. So the margin needed to be reduced.

app/assets/src/components/views/samples/samples_view.scss
MarkAZhang: After increasing the size of the icon to include the gray hover circle, the distance between the leftmost icon and the separator became too big.

app/assets/src/components/views/samples/ToolbarIcon.jsx
MarkAZhang: I did it mainly to keep it simple when it's being used.

```
<ToolbarIcon icon="save" />
```


instead of 
```
const saveIcon = <SaveIcon />
<ToolbarIcon icon={saveIcon} />
```

The customizability of letting you pass in whatever icon you want didn't seem that important, since we're trying to standardize styles across IDseq, which means restricting the number of options. We should encourage developers to use the pre-styled version.

tfrcarvalho: I actually disagree a bit. But it depends on how its going to be used. If this is the only way to use these icons, than it is good design, but if we start adding other "wrappers" to icons depending on location, it becomes hard to maintain... i.e. everytime you add a new icon you need to change multiple places.

MarkAZhang: That makes sense about having multiple wrappers, and referencing the icon component in each of them. I will make the change.

happyimadesignr:
Looks awesome!!! Thanks for tagging!

PR #2647, opened 2019-10-08
IDSEQ-1542 - Export legacy user data to auth0

davidrissato:
# Description

This rake test will export users from idseq mysql database to a legacy auth0 database.

This script is idempotent, and executing it multiple times will update the legacy user data and hashed password for user profiles that haven't been migrated yet.

We expect to execute this script only once, right before applying the definitive auth0 changes to idseq-web.

# Notes

Users will be automatically migrated by auth0 on their first login.

# Tests

* Lots of manual tests


lib/tasks/export_users_auth0.rake
```diff
@@ -0,0 +1,164 @@
...
+      'updated' => response["created_at"].present?,
+      'error' => response["error"].present?,
+      'response' => response.to_h,
+    }
+  end
+
+  private def format_legacy_user_record(user)
+    {
+      'connection' => AUTH0_DB_CONNECTION_NAME,
+      'email' => "legacy_idseq|#{user.email}",
```
tfrcarvalho: What is the goal of these prefixes i.e. `legacy_idseq`, `auth0`?
Can you add some comments to the PR? OR point to a doc?

davidrissato: I believe I may have to create a doc explaining this migration strategy.
We are using auth0 database user store (https://auth0.com/docs/connections/database), and that's why it is prefixed with `auth0`.
The `legacy_idseq` is just a constant defined to help identifying these profiles during automatic migration (https://auth0.com/docs/users/guides/configure-automatic-migration)

lib/tasks/export_users_auth0.rake
```diff
@@ -0,0 +1,164 @@
+# frozen_string_literal: true
+
+desc 'Export legacy users to Auth0'
+
+task 'export_users_to_auth0' => :environment do |_t, _args|
+  ActiveRecord::Base.logger.level = :info
+
+  options = { emails: [], export_all: false }
+
+  op = OptionParser.new do |o|
```
gregdingle: nice to see this class... I haven't seen us use it in any other rake tasks 😄

davidrissato: Yes, it is nice. The only issue with rake tasks and parameters is that you need to add two extra dashes (`--`) before the actual parameters, like in `rake my_task -- --my-parameter`, otherwise `rake` will think that is another task to be executed.

lib/tasks/export_users_auth0.rake
davidrissato: @tfrcarvalho I managed to make it work without prefixing the email.

PR #2649, opened 2019-10-08
[IDSEQ-1578] Fix causes of potential duplicate collection location entries

jshoe:
### Description
- This fixes bugs that led to multiple entries occurring for the same locations, e.g. multiple "USAs".

### Notes
Two main bugs:
(1) We made a previous insufficiently-informed assumption that place_ids from the LocationIQ provider would be stable over time. This turned out not to be true, and we found some locations to change once a month or so. Ex: "USA" would get a new higher ID. So we have to fallback to string comparisons on, say, country/state/subdivision/city names.
(2) The location provider data has name changes that may not be internally consistent. For example, the entry for "Nevada" was updated to say its country is "United States of America" (instead of just "USA" before). But the entry for the country only has the name "USA". Therefore, we can resolve this by explicitly normalizing "United States of America" => "USA" for ourselves. Unfortunately we don't have real pointers to say "Nevada is in USA" besides the string fields, although we could also use the country_code in just this case.
- Docs: https://locationiq.com/docs-html/index.html#forward_response

Other fixes:
- Added some fixes for hardening against page crashes when I had inconsistent data in my local dev db.
- Catch invalid NaN coordinates.
- Fix for trying to get MetadataFields with no corresponding Metadatum entry.

### Tests
- See code tests.
- Debug test cases: Pick an entry you have in local db, e.g. "Nevada, USA". Update its locationiq_id to 0. Go through the upload flow and create a new sample with "Nevada, USA". See no duplicate entries for "Nevada" and "USA" were created.

### DIFF
--- a/app/helpers/location_helper.rb
+++ b/app/helpers/location_helper.rb
@@ -21,12 +21,12 @@ module LocationHelper
                 end
 
     loc = {
-      name: body["display_name"],
+      name: normalize_name_aliases(body["display_name"]),
       geo_level: geo_level,
-      country_name: address[country_key] || "",
-      state_name: address[state_key] || "",
-      subdivision_name: address[subdivision_key] || "",
-      city_name: address[city_key] || "",
+      country_name: normalize_name_aliases(address[country_key] || ""),
+      state_name: normalize_name_aliases(address[state_key] || ""),
+      subdivision_name: normalize_name_aliases(address[subdivision_key] || ""),
+      city_name: normalize_name_aliases(address[city_key] || ""),
       # Round coordinates to enhance privacy
       lat: body["lat"] ? body["lat"].to_f.round(2) : nil,
       # LocationIQ uses 'lon'
@@ -121,7 +121,15 @@ module LocationHelper
     end
   end
 
+  # Normalize some location names from the provider for matching consistency.
+  # Ex: Treat "United States of America" and "USA" as the same in cases where
+  # the provider is not internally consistent.
   def self.normalize_name_aliases(name)
-
+    # See config/initializers/location_name_aliases.rb.
+    if LOCATION_NAME_ALIASES && LOCATION_NAME_ALIASES.key?(name)
+      LOCATION_NAME_ALIASES[name]
+    else
+      name
+    end
   end
 end
--- a/app/helpers/location_helper.rb
+++ b/app/helpers/location_helper.rb
@@ -20,13 +20,19 @@ module LocationHelper
                   ""
                 end
 
+    name = normalize_name_aliases(body["display_name"], geo_level)
+    country_name = normalize_name_aliases(address[country_key] || "", Location::COUNTRY_LEVEL)
+    state_name = normalize_name_aliases(address[state_key] || "", Location::STATE_LEVEL)
+    subdivision_name = normalize_name_aliases(address[subdivision_key] || "", Location::SUBDIVISION_LEVEL)
+    city_name = normalize_name_aliases(address[city_key] || "", Location::CITY_LEVEL)
+
     loc = {
-      name: normalize_name_aliases(body["display_name"]),
+      name: name,
       geo_level: geo_level,
-      country_name: normalize_name_aliases(address[country_key] || ""),
-      state_name: normalize_name_aliases(address[state_key] || ""),
-      subdivision_name: normalize_name_aliases(address[subdivision_key] || ""),
-      city_name: normalize_name_aliases(address[city_key] || ""),
+      country_name: country_name,
+      state_name: state_name,
+      subdivision_name: subdivision_name,
+      city_name: city_name,
       # Round coordinates to enhance privacy
       lat: body["lat"] ? body["lat"].to_f.round(2) : nil,
       # LocationIQ uses 'lon'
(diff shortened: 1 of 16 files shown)


app/helpers/location_helper.rb
```diff
@@ -120,4 +120,18 @@ def self.filter_by_name(samples_with_metadata, query)
       samples.where("`metadata`.`string_validated_value` IN (?)", query)
     end
   end
+
+  # Normalize some location names from the provider for matching consistency.
```
gregdingle: it is surprising that we need to do this ourselves

config/initializers/location_name_aliases.rb
```diff
@@ -0,0 +1,7 @@
+begin
+  path = "app/lib/location_name_aliases.json"
+  LOCATION_NAME_ALIASES = JSON.parse(File.read(path))
+rescue SystemCallError, JSON::ParserError => err
+  Rails.logger.error("Couldn't load location name aliases. #{err.message}")
```
gregdingle: wouldn't we want to crash here?

app/helpers/location_helper.rb
```diff
@@ -21,12 +21,12 @@ def self.adapt_location_iq_response(body)
...
+      name: normalize_name_aliases(body["display_name"]),
       geo_level: geo_level,
-      country_name: address[country_key] || "",
-      state_name: address[state_key] || "",
-      subdivision_name: address[subdivision_key] || "",
-      city_name: address[city_key] || "",
+      country_name: normalize_name_aliases(address[country_key] || ""),
+      state_name: normalize_name_aliases(address[state_key] || ""),
+      subdivision_name: normalize_name_aliases(address[subdivision_key] || ""),
+      city_name: normalize_name_aliases(address[city_key] || ""),
```
gregdingle: did you consider running a migration to normalize the data that's already in prod?

app/helpers/samples_helper.rb
```diff
@@ -587,9 +587,16 @@ def self.samples_by_metadata_field(sample_ids, field_name)
...
-        .includes(:metadata_field)
-        .group(Metadatum.where(key: field_name).first.validated_field)
+      # TODO(jsheu): Modify to get validated field from MetadataField.
+      first_datum = Metadatum.where(key: field_name).first
+      if first_datum
+        query
+          .includes(:metadata_field)
+          .group(first_datum.validated_field)
+      else
+        # No data for that field. Return empty grouped <ActiveRecord::Relation []> to be safe.
```
gregdingle: 👍

app/helpers/location_helper.rb
```diff
@@ -120,4 +120,18 @@ def self.filter_by_name(samples_with_metadata, query)
...
   end
+
+  # Normalize some location names from the provider for matching consistency.
+  # Ex: Treat "United States of America" and "USA" as the same in cases where
+  # the provider is not internally consistent.
+  #
+  # See config/initializers/location_name_aliases.rb and add important known
+  # aliases there.
+  def self.normalize_name_aliases(name)
+    if LOCATION_NAME_ALIASES && LOCATION_NAME_ALIASES.key?(name)
```
gregdingle: there is a chance that we could want two same-named entities to have different aliases. for example, washington the state and washington the city.

jshoe:
> It's pretty crazy that's not true. Do they not provide _some_ id that is stable?
> And why do you think their location data is changing? If they are fixing bugs in their data we should try to benefit.

Unfortunately no. The core reason I've found is that LocationIQ (and many other similar providers) are heavily based on the OpenStreetMap project data, which can get 5 million changes a day. The OSM project has never had stable IDs and the same IDs can refer to very different locations over time. I suspect that LocationIQ does imports from OSM and other sources periodically and then just increments all their place_ids.

Related links:
- https://wiki.openstreetmap.org/wiki/Permanent_ID
- https://blog.emacsen.net/blog/2018/02/16/osm-is-in-trouble/ "Lack of Permanent IDs" section
- https://engineering.fb.com/ml-applications/mars/ OSM change management background

test/models/location_test.rb
```diff
@@ -96,12 +96,13 @@ class LocationTest < ActiveSupport::TestCase
     assert mock.verify
   end
 
-  test "should find existing Location by API ID" do
+  test "should find existing Location by matching fields" do
+    location_info = { name: "Nevada, USA", geo_level: "state", country_name: "USA", state_name: "Nevada", subdivision_name: "", city_name: "" }
```
tfrcarvalho: Should we write a test for not yet normalized fields?

app/helpers/location_helper.rb
jshoe: I've corrected the existing ones already but I do plan to do it again manually 👍

config/initializers/location_name_aliases.rb
jshoe: Sure, changed it to re-raise and crash

app/helpers/location_helper.rb
jshoe: That's a great point. Adapted to split by geo_levels 👍

app/models/location.rb
```diff
@@ -86,27 +86,39 @@ def self.geosearch_by_osm_id(osm_id, osm_type)
...
     else
       # If osm_id and osm_type are missing, just use the original params.
       # 'New' without saving so make sure caller saves.
-      new_from_params(loc)
+      new_from_params(loc_info)
     end
   end
 
+  # Consider it a match if all fields to refer to the same place.
+  def self.find_with_fields(loc_info)
```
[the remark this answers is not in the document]
jshoe: We had:
`t.index ["country_name", "state_name", "subdivision_name", "city_name"]`
`t.index ["geo_level"]`
`t.index ["name"]`

Theoretically MySQL could use an index merge with them, but I just added another composite with these 6 to optimize.

app/helpers/samples_helper.rb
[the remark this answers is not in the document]
jshoe: 👍

test/models/location_test.rb
jshoe: Discussed offline: the existing fields would still need manual or migration correction unless we also wanted to add the normalization to `find_with_fields`

PR #2651, opened 2019-10-09
[IDSEQ-1441] Move required MetadataFields to front of CSV template + Fix location values in template

jshoe:
### Description
- JIRA: https://jira.czi.team/browse/IDSEQ-1441
- This is about the metadata CSV template that users can download when uploading samples or when editing on a project level later.

- Two changes in this ticket:
(1) Move is_required fields to the leftmost of the CSV but otherwise keep the order. Also hide the legacy location field.
(2) Get location names to show up in the template CSVs. Before they were showing as blank.

### Tests
- Go to URLs like `http://localhost:3000/metadata/metadata_template_csv?project_id=i` to download the sample templates. See collection_location in one of the first columns and see existing location names in the CSV as well.

Screen Shot 2019-10-08 at 4 51 56 PM

### DIFF
--- a/app/helpers/metadata_helper.rb
+++ b/app/helpers/metadata_helper.rb
@@ -122,8 +122,13 @@ module MetadataHelper
             elsif project.nil?
               generate_metadata_default_value(field, sample[:host_genome_name])
             else
-              puts "FIELD INFO: 3:25pm", sample[:metadata][field.name]
-              sample[:metadata][field.name] ? sample[:metadata][field.name].raw_value : nil
+              datum = sample[:metadata][field.name]
+              if datum
+                val = datum.validated_value
+                # Special-case for object values (e.g. Location types)
+                # rubocop:disable Metrics/BlockNesting
+                val.is_a?(Hash) ? val[:name] : val
+              end
             end
           end
         end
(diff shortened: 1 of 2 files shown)


app/helpers/metadata_helper.rb
```diff
@@ -115,7 +122,19 @@ def metadata_template_csv_helper(project_id, new_sample_names)
...
+              if datum
+                base_type = datum.metadata_field.base_type
+
+                # rubocop:disable Metrics/BlockNesting
+                if base_type == Metadatum::LOCATION_TYPE
+                  # Special-case for Location types
+                  val = datum.validated_value
+                  val.is_a?(Hash) ? val[:name] : val
+                else
+                  datum.raw_value
```
gregdingle: I feel like this code block above would better as a method of `Metadatum`, possibly overriding `raw_value`. 

Is the special case relevant to CSV only or other potential call sites?

jshoe: That's a good suggestion 👍moved it to `Metadatum#csv_template_value`. It is mostly custom for CSV to have a CSV-friendly string value instead of formatting a date or a hash of location info

PR #2652, opened 2019-10-09
Add cropped versions of typography mixins.

MarkAZhang:
Also add typography playground page.

Screen Shot 2019-10-08 at 5 34 25 PM

(click screenshot for full image)
Screen Shot 2019-10-09 at 12 23 59 PM

**Notes**

We can move over to the cropped versions gradually.
This is a result of a conversation with Jenn.


happyimadesignr:
Thanks Mark, this looks great! Only nitpick is it looks like it was done to the lowercase letters rather than the uppercase. The uppercase is a bit lower than the ascenders of the lowercase letters. It's probably fine though. =D

PR #2653, opened 2019-10-09
Add BulkDownload model.

MarkAZhang:
# Tests

* Verified in Rails console that BulkDownload object works as expected.

### DIFF
--- a/db/migrate/20191009004932_create_bulk_downloads_table.rb
+++ b/db/migrate/20191009004932_create_bulk_downloads_table.rb
@@ -1,0 +1,18 @@
+class CreateBulkDownloadsTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :bulk_downloads do |t|
+      t.text "params_json", comment: "JSON of the params for this bulk download"
+      t.string "download_type", comment: "The type of bulk download", null: false
+      t.string "status", comment: "The current status of the download, e.g. waiting, running, error, success", null: false
+      t.string "error_message", comment: "An error message to display to the user."
+    end
+
+    create_join_table :pipeline_runs, :bulk_downloads do |t|
+      t.index [:pipeline_run_id]
+      t.index [:bulk_download_id]
+    end
+
+    add_foreign_key "bulk_downloads_pipeline_runs", :bulk_downloads, name: "bulk_downloads_pipeline_runs_bulk_download_id_fk"
+    add_foreign_key "bulk_downloads_pipeline_runs", :pipeline_runs, name: "bulk_downloads_pipeline_runs_pipeline_run_id_fk"
+  end
+end
--- a/app/models/bulk_download.rb
+++ b/app/models/bulk_download.rb
@@ -1,0 +1,3 @@
+class BulkDownload < ApplicationRecord
+  has_and_belongs_to_many :pipeline_runs
+end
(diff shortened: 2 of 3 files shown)


MarkAZhang:
Requesting @gregdingle's review he worked on foreign keys.

db/migrate/20191009004932_create_bulk_downloads_table.rb
```diff
@@ -0,0 +1,18 @@
+class CreateBulkDownloadsTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :bulk_downloads do |t|
+      t.text "params_json", comment: "JSON of the params for this bulk download"
```
gregdingle: love the comments 😉

db/migrate/20191009004932_create_bulk_downloads_table.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.string "download_type", comment: "The type of bulk download"
```
gregdingle: if we want to want to get really fancy, we could use rails enum types here, and for `status`. disclaimer: I have not used them before.

db/migrate/20191009004932_create_bulk_downloads_table.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.string "status", comment: "The current status of the download, e.g. waiting, running, error, success"
+      t.string "error_message", comment: "An error message to display to the user."
+    end
+
+    create_join_table :pipeline_runs, :bulk_downloads do |t|
+      t.index [:pipeline_run_id]
+      t.index [:bulk_download_id]
+    end
+
+    add_foreign_key "bulk_downloads_pipeline_runs", :bulk_downloads, name: "bulk_downloads_pipeline_runs_bulk_download_id_fk"
```
gregdingle: 👍

db/migrate/20191009004932_create_bulk_downloads_table.rb
```diff
@@ -0,0 +1,18 @@
+class CreateBulkDownloadsTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :bulk_downloads do |t|
+      t.text "params_json", comment: "JSON of the params for this bulk download"
+      t.string "download_type", comment: "The type of bulk download"
+      t.string "status", comment: "The current status of the download, e.g. waiting, running, error, success"
+      t.string "error_message", comment: "An error message to display to the user."
```
gregdingle: it seems like every column except this one should be not null. if so, please add `null: false` to the definitions.

db/migrate/20191009004932_create_bulk_downloads_table.rb
MarkAZhang: My main concern, apart from us having never used it before, is that enums get stored as numbers in the database, which makes using the Rails console much more annoying. (you have to remember what "0" and "1" mean) Will defer for now.

db/migrate/20191009004932_create_bulk_downloads_table.rb
gregdingle: so `params_json` may be null?

MarkAZhang: Yeah. For some types of downloads there are no params. We could require everything to have an empty json `{}`, or just interpret `null` as `{}`.

MarkAZhang:
Okay, I will look into the ActiveRecord way of defining enum as I flesh out the model more. Thanks!

PR #2654, opened 2019-10-09
Udpate to non-sharded mosquito index genome

morsecodist:
# Description

This updates the configuration to use the new non-sharded mosquito indexes.

# Tests

I manually added these indexes to staging with a different name and ran several different runs against them. They all worked out.

### DIFF
--- a/db/migrate/20191009171351_mosquito_genome_update_2019_10_02.rb
+++ b/db/migrate/20191009171351_mosquito_genome_update_2019_10_02.rb
@@ -1,23 +1,9 @@
 class MosquitoGenomeUpdate20191002 < ActiveRecord::Migration[5.1]
-  def up
-    # The previous mosquito genome was not added via a migration so we need to destroy it
-    old_hg = HostGenome.find_by(name: "Mosquito").destroy
-    old_hg.destroy if old_hg
-
-    hg.name = "Mosquito"
+  def change
+    hg = HostGenome.find_by(name: "Mosquito")
+    return unless hg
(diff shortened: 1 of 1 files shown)


db/migrate/20191009171351_mosquito_genome_update_2019_10_02.rb
```diff
@@ -0,0 +1,23 @@
+class MosquitoGenomeUpdate20191002 < ActiveRecord::Migration[5.1]
+  def up
+    # The previous mosquito genome was not added via a migration so we need to destroy it
+    old_hg = HostGenome.find_by(name: "Mosquito").destroy
```
gregdingle: Won't this break references between all current samples and their host genomes? (foreign key constraints should prevent this) 

Why not create a Mosquito v2?

db/migrate/20191009171351_mosquito_genome_update_2019_10_02.rb
```diff
@@ -0,0 +1,23 @@
...
+    old_hg.destroy if old_hg
```
gregdingle: it looks like `.destroy` is called twice

db/migrate/20191009171351_mosquito_genome_update_2019_10_02.rb
```diff
@@ -0,0 +1,9 @@
+class MosquitoGenomeUpdate20191002 < ActiveRecord::Migration[5.1]
+  def change
```
gregdingle: is this actually reversible? rails must be really magic then!

if not then change the method to `up`

morsecodist: Nope, thanks for clearing that up.

PR #2655, opened 2019-10-09
Add ReviewStep to BulkDownloadModal.

MarkAZhang:
Rename downloadType to downloadTypeMetadata.
Rename download "options" to download "fields".

# Description

Also small changes like:
* Disable download button if no samples selected.
* Pass selected sampleIds into bulk download modal and display in review step.
* When the user re-opens the modal, everything is reset.

Screen Shot 2019-10-09 at 2 24 40 PM

Screen Shot 2019-10-09 at 2 24 47 PM

# Tests
* Download button is only enabled if a sample is chosen.
* Correct information is displayed in review step.
* Modal is reset if modal is closed.

### DIFF
--- a/app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
+++ b/app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
@@ -1,19 +1,34 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { set } from "lodash/fp";
+import { find, get, set } from "lodash/fp";
+import memoize from "memoize-one";
 
 import { getBulkDownloadTypes } from "~/api";
 import Modal from "~ui/containers/Modal";
 
 import ChooseStep from "./ChooseStep";
+import ReviewStep from "./ReviewStep";
+
+const assembleSelectedDownload = memoize(
+  (selectedDownloadTypeName, allSelectedFields, sampleIds) => {
+    const fields = get(selectedDownloadTypeName, allSelectedFields);
+
+    return {
+      type: selectedDownloadTypeName,
+      fields,
+      sampleIds: Array.from(sampleIds),
+    };
+  }
+);
 
 class BulkDownloadModal extends React.Component {
   state = {
     bulkDownloadTypes: null,
-    // We save the options for ALL download types.
+    // We save the fields for ALL download types.
     // If the user clicks between different download types, all their selections are saved.
-    selectedOptions: {},
-    selectedDownloadType: null,
+    selectedFields: {},
+    selectedDownloadTypeName: null,
+    currentStep: "choose",
   };
 
   async componentDidMount() {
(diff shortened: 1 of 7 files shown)


app/assets/src/api/index.js
```diff
@@ -341,7 +341,7 @@ const getSamplesLocations = ({ domain, filters, projectId, search }) =>
 const getSamplePipelineResults = id =>
   get(`/samples/${id}/results_folder.json`);
 
-const getBulkDownloadTypes = () => get("/bulk_downloads/types");
+const getBulkDownloadTypeMetadata = () => get("/bulk_downloads/type_metadata");
```
MarkAZhang: Renamed this "type" to "type metadata" to distinguish it from the "type" field of a bulk download object.
"type metadata" is an object with fields like "description", "display name" for that particular download type.
"type" on a bulk download object is a string such as "sample_taxons_summary" which specifies the download type.

app/helpers/bulk_download_types_helper.rb
```diff
@@ -17,8 +17,9 @@ module BulkDownloadTypesHelper
       display_name: "Combined Sample Taxon Results",
       description: "The value of a particular metric (e.g. total reads, rPM) for all taxons in all selected samples, in a single data table",
       category: "report",
-      options: [
+      fields: [
```
MarkAZhang: Renamed from options to fields, since option is already used to describe the different choices for a field (e.g. ".csv", ".biom") below. `option.options` is super confusing.

app/assets/src/api/index.js
tfrcarvalho: But what does this return? Isn't it all the types + their metadata?

app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
```diff
@@ -1,62 +1,128 @@
...
+import memoize from "memoize-one";
 
-import { getBulkDownloadTypes } from "~/api";
+import { getBulkDownloadTypeMetadata } from "~/api";
 import Modal from "~ui/containers/Modal";
 
 import ChooseStep from "./ChooseStep";
+import ReviewStep from "./ReviewStep";
+
+const assembleSelectedDownload = memoize(
```
tfrcarvalho: Why do you need to memoize this?
Not sure if I understand whats happening but wouldnt storing the selected on a state variable a simpler alternative?

app/controllers/bulk_downloads_controller.rb
```diff
@@ -2,8 +2,8 @@
 class BulkDownloadsController < ApplicationController
   include BulkDownloadTypesHelper
 
-  # GET /bulk_downloads/types
-  def types
-    render json: BulkDownloadTypesHelper::BULK_DOWNLOAD_TYPES
+  # GET /bulk_downloads/type_metadata
+  def type_metadata
```
tfrcarvalho: Seem to me that this is returning a list of download types, just like `projects`, `samples` etc. They all return metadata but we do not call them explicitly say its metadata.

app/assets/src/components/views/bulk_download/ReviewStep.jsx
```diff
@@ -0,0 +1,81 @@
...
+        <div className={cs.value}>{value}</div>
+      </div>
+    );
+  };
+
+  render() {
+    const { selectedDownload, downloadTypeMetadata, onBackClick } = this.props;
+
+    return (
+      <div className={cs.reviewStep}>
```
tfrcarvalho: I am curious if there aren't any small element that are reusable on this component? 
This seems like yet another modal that will have slightly different styles.
Do you see any opportunities to consolidate?

app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
MarkAZhang: This is IMO a cleaner way to do it.

If we store the selected download in the state, then we would need to remember to update the state value every time certain props changed. Here, the `memoize` only recalculates the value if any of the input parameters changed. So it takes care of all that logic for us.

Granted, the "assembleSelectedDownload" function isn't very expensive, so the memoize is more of a best practice here. Even if we ran it every render, it probably wouldn't be a problem.

app/controllers/bulk_downloads_controller.rb
MarkAZhang: So do you think I should keep the original naming? I could do either. The reason I changed it was that I was running into a bit of confusion with "selectedDownloadType", which could either mean a string like "samples_taxon_summary" or the metadata object. I ran into a situation where I was using both in the same function and couldn't figure out how to name differently.

app/assets/src/components/views/bulk_download/ReviewStep.jsx
MarkAZhang: There are very small opportunities. I could factor out the <Footer>. But the overlap is quite small overall. It doesn't seem worth it to me right now, if the overlap became more substantial I would consolidate then.

app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
tfrcarvalho: Ok, I did not know that this library just memoizes the last one (hence the name... ) so seems fine. I actually find it more confusing as you got to know that. But leave it up to you...

app/controllers/bulk_downloads_controller.rb
tfrcarvalho: I believe the most consistent would be to name as `selectedDownloadTypeName`. It is long but more accurate. And then you can have `downloadsTypes[selectedDownloadTypeName].<metadata_property>` or event `selectedDownloadType.<metadata_property>`

MarkAZhang: Okay that makes sense. I will change to "type name" where there's a conflict..

app/assets/src/components/views/bulk_download/ReviewStep.jsx
tfrcarvalho: This was more of a philosophical/long term strategy question. Maybe part of a larger consolidation effort with @happyimadesignr.

MarkAZhang:
Fixed the naming of "type metadata". Changed back to "type".

PR #2658, opened 2019-10-10
Clean-up: finish moving type fields from Metadatum -> MetadataField

jshoe:
### Description
- This is some cleanup to finish moving type-related calls onto the MetadataField model. Separated from another PR.
- We started off with `base_type` on `Metadatum` but it evolved to move onto `MetadataField`. So logically function calls like `validated_field` and `convert_type_to_string` should also stay within `MetadataField`.

### Notes
- It's a little unusual to also edit the old migration files, but I think it's correct because new builds run them all + they won't run again on dbs that already ran them.
- Moving the constants should be quite safe because it is just a find and replace to another file as opposed to actually moving data.

### Tests
- Unit tests.
- Checked metadata sample upload flow, project-level metadata, report sidebar metadata.


app/helpers/samples_helper.rb
```diff
@@ -587,16 +589,9 @@ def self.samples_by_metadata_field(sample_ids, field_name)
           ] + Location::GEO_LEVELS.map { |l| "#{l.pluralize}_locations.name" }
         )
     else
-      # TODO(jsheu): Modify to get validated field from MetadataField.
-      first_datum = Metadatum.where(key: field_name).first
-      if first_datum
-        query
-          .includes(:metadata_field)
-          .group(first_datum.validated_field)
```
jshoe: ^ This was one recent hacky example. Shouldn't need to fetch the Metadatum just to call `validated_field`

PR #2659, opened 2019-10-10
Add creation logic to bulk download modal.

MarkAZhang:
# Description

Mostly updates to the model and controller.
Also add error and success messages to the bulk download modal.

Screen Shot 2019-10-10 at 12 58 48 PM

Screen Shot 2019-10-10 at 12 59 04 PM

# Tests
* Wrote tests for controller/model changes.
* Verify that bulk download objects are correctly created via the modal.
* Verify that users without the flag cannot access controller methods.


app/assets/src/components/views/bulk_download/ReviewStep.jsx
```diff
@@ -1,32 +1,123 @@
...
+    }
+
+    if (createStatus === "success") {
+      return (
+        <React.Fragment>
+          <div className={cs.successMessage}>
+            <CircleCheckmarkIcon className={cs.icon} />
+            <div className={cs.text}>
+              Your download was successfully submitted. We&apos;ll send you an
+              e-mail when your files are ready.
```
jshoe: nit nit: looks like we use `email` usually

app/controllers/bulk_downloads_controller.rb
```diff
@@ -2,8 +2,55 @@
...
   def types
     render json: BulkDownloadTypesHelper::BULK_DOWNLOAD_TYPES
   end
+
+  # POST /bulk_downloads
+  def create
+    bulk_download_params = params.permit(:download_type, sample_ids: [], params: {})
+
+    # Convert sample ids to pipeline run ids.
+    pipeline_run_ids = Sample.where(id: bulk_download_params[:sample_ids]).includes(:pipeline_runs)
```
jshoe: This `Sample.where` definitely needs access control!! 😬

app/controllers/bulk_downloads_controller.rb
```diff
@@ -2,8 +2,55 @@
...
+                             .select do |sample|
+                         # Check that the most recent pipeline run succeeded.
+                         pr = sample.pipeline_runs.first
+                         !pr.nil? && pr.finalized && pr.job_status == PipelineRun::STATUS_CHECKED
```
jshoe: nit: did an autoformatter do this? The indentation is hard to read

app/controllers/bulk_downloads_controller.rb
```diff
@@ -2,8 +2,55 @@
...
+      render json: {
+        error: "Some samples have either failed or not finished running yet. Please contact us for help.",
+      },
+             status: :unprocessable_entity
+      return
+    end
+
+    # Convert the sample ids to pipeline run ids.
+    bulk_download_params[:pipeline_run_ids] = pipeline_run_ids
+    bulk_download_params.delete(:sample_ids)
```
tfrcarvalho: Not great to change the input params. I would create a copy or pass the pipeline runs separately.

app/controllers/bulk_downloads_controller.rb
tfrcarvalho: You'll want to read from power here.

app/controllers/application_controller.rb
```diff
@@ -27,6 +27,10 @@ def admin_required
     redirect_to root_path unless current_user && current_user.admin?
   end
 
+  # To use in before_action with parameters, do
+  # before_action do
+  #   allowed_feature_required(...)
+  # end
   def allowed_feature_required(allowed_feature, allow_admin = false)
     redirect_to root_path unless current_user && (
```
tfrcarvalho: This is old but... should we redirect to root here? Why not just an 404 or something like that or maybe add a parameter for that?

app/controllers/bulk_downloads_controller.rb
tfrcarvalho: Also, we should probably create a method `first_pipeline_runs` or something to do this in PipelineRun model that will do basically something like:
```
def first_pipeline_runs(samples)
	PipelineRun.where(samples: samples).group(:sample_id).having('MAX(`pipeline_runs`.`created_at`) >= `pipeline_runs`.`created_at`')
end
```
This is a very common use case... 
We can also add your conditions in there.

app/controllers/bulk_downloads_controller.rb
MarkAZhang: Yeah, autoformatter. Tried to make it better.

app/controllers/bulk_downloads_controller.rb
MarkAZhang: Looks like the function already exists. Will use. Thanks!

app/controllers/application_controller.rb
MarkAZhang: I believe we discussed offline that the downside of returning 403 is that it would indicate that there is something hidden at that URL. Right now, if I type in a random url that doesn't match anything, it redirects to home for some reason. Will defer addressing for now.

MarkAZhang:
Rewrote the `get_pipeline_run_ids_for_samples` method to return more specific errors. Planning to use this method for the validate bulk download step as well.

app/controllers/bulk_downloads_controller.rb
```diff
@@ -1,9 +1,44 @@
...
+
+  # POST /bulk_downloads
+  def create
+    # Convert sample ids to pipeline run ids.
+    begin
+      pipeline_run_ids = get_pipeline_run_ids_for_samples(bulk_download_params[:sample_ids])
+    rescue => e
+      # Throw an error if any sample doesn't have a valid pipeline run.
+      # The user should never see this error, because the validation step should catch any issues.
+      render json: { error: e }, status: :unprocessable_entity
```
jshoe: could you use `log_err_and_airbrake` here if it's never supposed to be happen?

app/assets/src/components/views/bulk_download/review_step.scss
```diff
@@ -61,5 +65,50 @@
...
+    }
+
+    .errorMessage {
+      display: flex;
+
+      .icon {
+        fill: $error;
+        width: 18px;
+        height: 18px;
+        margin-right: 6px;
```
jshoe: Jenn would probably recommend using `$space-xxs` for margin-right ;)

app/controllers/bulk_downloads_controller.rb
```diff
@@ -1,9 +1,44 @@
...
   # GET /bulk_downloads/types
   def types
     render json: BulkDownloadTypesHelper::BULK_DOWNLOAD_TYPES
   end
+
+  # POST /bulk_downloads
+  def create
+    # Convert sample ids to pipeline run ids.
+    begin
+      pipeline_run_ids = get_pipeline_run_ids_for_samples(bulk_download_params[:sample_ids])
```
jshoe: maybe a comment to say the access control is checked in `get_pipeline_run_ids_for_samples` :)

app/assets/src/components/views/bulk_download/review_step.scss
MarkAZhang: I'm planning on going through and re-doing the spacing values in a separate PR. Jenn sent me some new mocks that use the new values.

MarkAZhang:
Decided to change to just a one-to-many relationship between bulk downloads and users. The MVP doesn't have plans for sharing between users. If we need sharing later, we can add an additional `has_many :through` relationship through a "BulkDownloadCollaborator" table.

app/helpers/bulk_downloads_helper.rb
```diff
@@ -0,0 +1,27 @@
+module BulkDownloadsHelper
+  SAMPLE_NO_PERMISSION_ERROR = "You do not have permission to access all of the selected samples. Please contact us for help.".freeze
```
tfrcarvalho: Why would they even see it in this case?
If this a potential impossible state, maybe we should be alerted for it.

MarkAZhang: Great point about alerting. I added an alert to the controller create method.

app/helpers/bulk_downloads_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+  SAMPLE_STILL_RUNNING_ERROR = "Some selected samples have not finished running yet. Please remove these samples and try again.".freeze
+  SAMPLE_FAILED_ERROR = "Some selected samples failed their most recent run. Please remove these samples and try again.".freeze
+
+  def get_pipeline_run_ids_for_samples(sample_ids)
```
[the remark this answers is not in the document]
MarkAZhang: Thanks for raising this. I didn't consider whether it should be in the pipeline run model.

After thinking about it, I do think this is related to bulk downloads. This gets valid pipeline runs, where "valid" is in the context of whether the pipeline run is ready to be part of a bulk download. I can rename the function and add a comment to make this more clear. The errors that are returned are also written for display in the bulk download modal.

You could argue that we should make this function more generic, into something like "get_successful_pipeline_run_ids_for_samples" in pipeline_runs model and then have this function call that one. But in bulk downloads we've chosen to fail if even one sample is invalid, and a generic function might want to just return all the valid pipeline run ids and not error out. We might relax the invalid constraint in bulk downloads in the future, but right now I think this function has enough unique requirements that it's fine to put in the bulk download helper.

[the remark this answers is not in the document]
tfrcarvalho: I think that should be a `strict mode` option which can raise or return the errors. There are a bunch of examples where this would be useful, and knowing about running or error samples is important even if treatment is different: heatmap, phylo tree creation, etc. I just fear that is yet another logic duplication.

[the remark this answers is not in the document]
tfrcarvalho: I also changed my comment from `related` -> `specific`.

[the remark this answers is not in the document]
MarkAZhang: Great point about heatmap and other use cases. Okay, I'll make the change.

app/helpers/pipeline_runs_helper.rb
```diff
@@ -221,4 +224,27 @@ def check_for_user_error(failed_stage)
...
+
+  # Return all pipeline runs that have succeeded for given samples
+  # Only check the first pipeline run.
+  # samples should be an ActiveRecord relation
+  # If strict mode is turned on, error out even if one pipeline run did not succeed.
+  def get_succeeded_pipeline_runs_for_samples(samples, strict = false)
+    # Include the pipeline run for efficient queries.
+    samples = samples.includes(:pipeline_runs)
+
+    first_pipeline_runs = samples.map(&:first_pipeline_run)
```
tfrcarvalho: This is very slow... one query per sample.

MarkAZhang: Hmm...I was hoping the `includes` would take care of this, but it looks like it doesn't. Any ideas how to make this faster?

tfrcarvalho: I was looking... i know it MySQL: `SELECT pr.id, pr.sample_id FROM (SELECT sample_id, MAX(created_at) AS max_created_at FROM pipeline_runs pr GROUP BY sample_id) AS x INNER JOIN pipeline_runs AS pr ON pr.sample_id=x.sample_id AND pr.created_at=x.max_created_at;`
Still looking for the best way to do it in rails...

MarkAZhang: The only thing I can think of is to manually sort the pipeline runs in Ruby and then pick the correct one. So rewriting the `order_by` logic in Ruby. I'm assuming that if you call order_by, it will make a Rails query no matter what?

app/helpers/pipeline_runs_helper.rb
```diff
@@ -221,4 +224,30 @@ def check_for_user_error(failed_stage)
...
+  # Only check the first pipeline run.
+  # samples should be an ActiveRecord relation
+  # If strict mode is turned on, error out even if one pipeline run did not succeed.
+  # Note: Does NOT do access control checks.
+  def get_succeeded_pipeline_runs_for_samples(samples, strict = false)
+    # Gets the first pipeline runs for multiple samples in an efficient way.
+    created_dates = PipelineRun.select("sample_id, MAX(created_at) as created_at").where(sample_id: samples.pluck(:id)).group(:sample_id)
+    first_pipeline_runs = PipelineRun
+                          .select(:finalized, :id, :job_status)
+                          .where("(sample_id, created_at) IN (?)", created_dates)
```
tfrcarvalho: * You can remove `pipeline_runs` prefix actually! even more dry...
* Just add `.where(finalized: 1)` and you can remove the line below because pr is never going to be nil.

PR #2660, opened 2019-10-10
Unstick heatmap filters and make row/column labels sticky

j-x-han:
# Description

The heatmap filters are no longer sticky; users can scroll the page so the heatmap is the only thing shown on the page.

Screen Shot 2019-10-10 at 4 42 25 PM
Screen Shot 2019-10-10 at 4 42 42 PM

The sample titles, taxon titles, and metadata fields are now sticky on the heatmap.

**Scrolled to the bottom of the heatmap.**
Screen Shot 2019-10-10 at 4 42 53 PM
**Zoomed in, scrolled to the middle and far right of the heatmap.**
Screen Shot 2019-10-10 at 4 44 16 PM

# Notes

* To prevent double scrolling (both the page and the heatmap scrolling), the page will not scroll if the mouse is hovering over the heatmap unless the heatmap has already been scrolled to the top/bottom.
* Since the metadata fields (including their labels) are grouped together, the metadata labels and `Add metadata` button aren't sticky.

# Tests

* Verified that scrolling works as expected for different heatmaps.
* Verified that the heatmap scrolling is bounded (i.e. the heatmap can't be scrolled off the page).
* Verified that zooming still works as expected.


cdebourcy:
This is awesome!

Quick question looking at the screenshot: when scrolling horizontally, doesn't the current behavior mean that cells will line up with the wrong metadata?

tfrcarvalho:
> doesn't the current behavior mean that cells will line up with the wrong metadata

I think they are aligned... What was confusing is that the labels and the cells that are now out of the heatmap did not disappear. We should either place a white rectangle in top left corner or apply clipping (https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Clipping_and_masking) to the g elements that hold the row  labels and metadata cells. Does the same happen with column labels?

MarkAZhang:
Hey Julie! This looks great!

My one big concern is that with this implementation, it looks like the only way to scroll horizontally is with the mouse wheel. There's no horizontal scrollbar. For users without a mouse that supports horizontal scrolling, they might not be able to scroll horizontally at all. Maybe @happyimadesignr can weigh in on whether that would be an issue.

MarkAZhang:
Right now, it's only possible to scroll over the heatmap cells, not the heatmap labels. Would it be hard to enable scrolling on the labels as well?

For the pathogen name labels, only vertical scrolling would need to be enabled. And for sample name, only horizontal scrolling.

MarkAZhang:
Screen Shot 2019-10-21 at 3 20 00 PM

I see a little visual artifact (the gray line) above "Collection Location". Maybe one of the background rects is a couple pixels off?

MarkAZhang:
Could we change the cursor to the "move" cursor when dragging is happening? (see https://developer.mozilla.org/en-US/docs/Web/CSS/cursor)

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -594,43 +595,54 @@ class SamplesHeatmapView extends React.Component {
...
+                options={this.getControlOptions()}
+                selectedOptions={this.state.selectedOptions}
+                onSelectedOptionsChange={this.handleSelectedOptionsChange}
+                loading={this.state.loading}
+                data={this.state.data}
+              />
+            </NarrowContainer>
+          </div>
+        )}
+        {this.state.hideFilters && <Divider style="shadow" />}
```
MarkAZhang: I wouldn't use the `<Divider>` component in this case. It doesn't feel like this is a "divider". It's not visually separating two sections of content. Instead, it's hinting that a panel can be opened. We generally try not to add new types or "styles" of a component without good reason.

For now, I would just make this a div. You might nest it under the `filterToggleContainer` div since it's technically part of that component (you would then have to move the onClick to a child)

app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
```diff
@@ -31,6 +34,23 @@
     }
   }
 
+  .filterToggleContainer {
+    margin: auto;
```
MarkAZhang: It doesn't seem like you need `margin: auto, width: 47px or display: block`.

app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
```diff
@@ -31,6 +34,23 @@
...
+    width: 47px;
+    display: block;
+    text-align: center;
+    position: relative;
+
+    .arrowIcon {
+      color: $dark-grey;
+      background-color: $white;
+      font-size: $font-size-h4;
+      padding: 0 18px;
```
MarkAZhang: Do any of the spacing variables in `elements.scss` work instead?

app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
```diff
@@ -31,6 +34,23 @@
...
+      border-radius: 0 0 4px 4px;
+      box-shadow: 0 2px 2px 0 rgba(34, 36, 38, 0.2);
```
MarkAZhang: Could you use the box-shadows in colors.scss/elements.scss or add an additional constant? (but make sure the ones that are already there aren't good enough)

app/assets/src/components/views/compare/samples_heatmap_vis.scss
```diff
@@ -61,4 +64,46 @@
...
+    .bannerText {
+      @include font-body-xxs;
+      color: $medium-grey;
+      background: $white;
+      border-radius: 4px;
+      box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.15);
+      cursor: default;
+      display: inline-block;
+      margin: 8px auto auto;
+      padding: 6px 8px;
```
MarkAZhang: Use spacing variables.

app/assets/src/components/views/compare/samples_heatmap_vis.scss
```diff
@@ -61,4 +64,46 @@
...
+    &.hide {
+      opacity: 0;
+    }
+
+    &:hover {
+      opacity: 1;
+    }
+
+    .bannerText {
+      @include font-body-xxs;
```
MarkAZhang: We're switching to the `crop` version of these typography mixins. Can you use font-body-xxs-crop?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -400,6 +439,165 @@ export default class Heatmap {
...
+    d3.event.stopPropagation();
+  }
+
+  pan(deltaX, deltaY) {
+    // Define the scrolling boundaries for the svg.
+    // Upper limits are determined by the difference between the container and svg sizes,
+    // scaled by the zoom factor.
+    let containerWidth = this.container[0][0].offsetWidth,
+      containerHeight = this.container[0][0].offsetHeight;
+    let xScrollMax =
```
MarkAZhang: Nit: I think this is actually the xScrollMin since it's usually a negative number.

MarkAZhang: Also, I think this should take the `marginLeft` into account as well. Maybe it's `containerWidth - this.svg.atr("width") - this.options.marginLeft - this.options.marginRight`? Otherwise, as you zoom in more and more, I think there will be a bigger and bigger gap at the bottom and right when you try to scroll all the way.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -400,6 +439,165 @@ export default class Heatmap {
...
+        this.rowLabelsWidth +
+          this.columnLabels.length * this.cell.width -
+          labelsDx
+      );
+
+    // Translating the column labels in the opposite y direction of the svg.
+    let columnLabelsCurrent = d3.transform(this.gColumnLabels.attr("transform"))
+      .translate;
+    let labelsDy = Math.max(
+      this.columnLabelsHeight,
```
MarkAZhang: Could you use `clamp` from `lodash/fp` instead of `Math.max(Math.min`? I find it's a little easier to tell what the upper and lower bound are with the `clamp` function, so it makes it more readable. Especially since you use min/max quite a bit here.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -400,6 +439,165 @@ export default class Heatmap {
...
+      this.mouseY = d3.event.clientY;
+    }
+  }
+
+  scroll() {
+    this.pan(d3.event.wheelDeltaX, d3.event.wheelDeltaY);
+    d3.event.stopPropagation();
+  }
+
+  pan(deltaX, deltaY) {
```
MarkAZhang: This function has a lot of duplicated code with placeContainers. The only difference is that placeContainers uses `0` while this method uses `labelsDx` and `labelsDy` when placing various containers. We discussed in person factoring out the place code for the two sets of d3 elements that "move" when the heatmap is shifted, which will help de-dupe the code.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -935,14 +1133,30 @@ export default class Heatmap {
     columnMetadataLabelEnter
       .append("rect")
       .attr("class", cs.hoverTarget)
-      .attr("width", this.rowLabelsWidth)
+      .attr("x", -this.options.marginLeft)
+      .attr("width", d => {
+        const xOffset =
```
MarkAZhang: Maybe we can factor this logic out into a function, since it is being called in several different situations.

MarkAZhang:
While I had a few comments, this PR looks really good overall! Thanks for taking the time to go through all the different iterations. Your final approach was easy to understand, which is a great accomplishment given that the Heatmap code has been historically tough to work with! Awesome job.

app/assets/images/icons/close.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tfrcarvalho: Do you still need this svg?

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -405,6 +444,20 @@ class SamplesHeatmapVis extends React.Component {
...
         )}
+        <div
+          className={cx(
+            cs.bannerContainer,
+            this.state.displayControlsBanner ? cs.show : cs.hide
+          )}
+        >
+          <div className={cs.bannerText}>
+            Hold SHIFT to scroll horizontally and SPACE BAR to pan.
+            <CloseIcon
```
tfrcarvalho: We have a "remove" icon that is very similar (except for the circle). We cannot reuse that one? 
@happyimadesignr do we have a reasoning to use on or the other?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -594,43 +595,54 @@ class SamplesHeatmapView extends React.Component {
...
+        <div
+          className={cs.filterToggleContainer}
+          onClick={this.toggleDisplayFilters}
```
gregdingle: I forgot to say that this action should log to analytics so we can see how many users click on it over time.

app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
```diff
@@ -1,6 +1,9 @@
 @import "~styles/themes/colors";
+@import "~styles/themes/typography";
+@import "~styles/themes/elements";
 
 .heatmap {
+  margin-top: -10px;
```
MarkAZhang: Can you use -$header-margin-bottom here instead?

app/assets/src/components/views/compare/samples_heatmap_vis.scss
```diff
@@ -18,13 +19,15 @@
...
-    // of the viewport instead of the bottom of the page. 334px is the height
-    // of the all the elements above the heatmapContainer. It *will* need to
-    // change if the height of the elements above ever change.
-    // TODO(mark): Use flexbox instead of hard-coding the offset.
-    height: calc(100vh - 334px);
+    overflow: hidden;
+    // Set the height so the heatmap takes up the rest of the screen.
+    // 358px is the height of the elements above the heatmapContainer.
+    // It *will* need to change if the height of the elements above ever change.
+    height: calc(100vh - 358px);
```
MarkAZhang: Could you use $header-height in this expression? 

```
  height: calc(100vh - #{$header-height} - 308px);
```

app/assets/src/components/views/compare/samples_heatmap_vis.scss
```diff
@@ -18,13 +19,15 @@
...
+
+    &.fullScreen {
+      height: calc(100vh - 85px);
```
MarkAZhang: I notice a tiny gap at the bottom of the screen. Is that intentional? If not, could you increase the height to the largest value before a scrollbar appears?

app/assets/src/components/views/compare/samples_heatmap_vis.scss
```diff
@@ -18,13 +19,15 @@
...
+    }
```
MarkAZhang: For the `height`, could you also take into account whether any threshold filters are applied? When threshold filters are not applied, there's currently a large gap under the heatmap. You should be able to detect whether any threshold filters are passed into SamplesHeatmapVis and apply another css class if there are threshold filters and it's not full-screen.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -400,6 +423,140 @@ export default class Heatmap {
...
+
+    // Translating the column and metadata labels in the opposite y direction of the svg.
+    let columnLabelsCurrent = d3.transform(this.gColumnLabels.attr("transform"))
+      .translate;
+    let labelsDy = clamp(
+      this.columnLabelsHeight,
+      this.columnLabelsHeight + this.options.marginTop - yScrollMax,
+      columnLabelsCurrent[1] - deltaY
+    );
+    this.placeColumnAndMetadataLabels(labelsDy);
```
MarkAZhang: Can we rename this to `placeColumnLabelAndMetadataContainers`? Otherwise, it's confusing why we're calling `renderColumnMetadataLabels(labelsDx)` above and `placeColumnAndMetadataLabels(labelsDy)` here. You would expect both dx and dy to be passed into the "place" method and nothing to be passed to the "render" method based on the naming. The reason is that the dy translation is applied to the container, while dx is applied to each individual metadata label, and the renaming makes that a little more clear. (also, `renderColumnMetadataLabels` both renders and places)

Also, rename `placeRowLabels` to `placeRowLabelContainers` for consistency.

Finally, I think a further improvement would be to split up the renderColumnMetadataLabel and placeColumnMetadataLabel logic into separate function and only call the place function here. So that you only execute the exact logic that you need. However, we can defer that to a later time, to avoid endlessly refactoring code.

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -407,6 +407,9 @@ class SamplesHeatmapVis extends React.Component {
         <div
           className={cx(
             cs.heatmapContainer,
+            (!isEmpty(this.props.thresholdFilters) ||
```
MarkAZhang: Great catch on the taxon categories!!

PR #2661, opened 2019-10-11
Rewrite update lineage db script

gregdingle:
# Description

This is a complete rewrite of the lineage update script. The script used to do most its work in inline bash and the update delta was computed with unix tools. It was very hard to tell exactly what the unix commands were doing, and there was an issue where the lineage version of rows was set to NULL. 

Now the script is in pure ruby, except mysqlimport, and the update delta is computed in SQL. The intermediate state is all stored as SQL tables for easy inspection. The final irreversible queries are executed only after a user prompt and they are wrapped in SQL transaction.

# Notes

* Even after our security talk, I didn't use SQL placeholders, because the values are hardcoded in the file, but I could refactor to use placeholders
* See the log output below to see the sequence of events.
* See the test data to understand the inputs. 
* I did not compare output of the old script to the new, but I could do so. 
* The current version is now defined as TaxonLineage.maximum('version_end'), and the next, +1. So the version must only increase in steps of 1. 
* After this goes live, it would be good to set the version columns to be NOT NULL 
* The `started_at` and `ended_at` date format is now the same YYYY-MM-DD as the NCBI date instead of the `Time.now`. I feel this represents better the intent of the script and I can't think of any downside.
* The script is 4x faster to run locally than before, taking 15 min instead of an hour

# Tests

I extracted some test data for this script then added some rows to force inserts and updates. By inspecting database, I saw each type of row was handled correctly. Comparing the 11 test rows to the existing rows from which they were randomly sampled:

```
Retire records...
1920336 records

Insert records...
1 records

Update records...
1 records

Keep same records...
9 records
```

```
TEST RUN

Starting import of s3://idseq-database/taxonomy/2019-09-17 ...



--------------
DROP TABLE IF EXISTS _new_names
--------------

--------------
CREATE TABLE _new_names(taxid text, name text, common_name text)
--------------

Connecting to db
Selecting database idseq_development
Loading data from LOCAL file: /app/tmp/taxonomy/2019-09-17/_new_names.csv into _new_names
idseq_development._new_names: Records: 50  Deleted: 0  Skipped: 0  Warnings: 0
Disconnecting from db
--------------
ALTER TABLE _new_names ADD UNIQUE idx(taxid(255))
--------------

--------------
DROP TABLE IF EXISTS _new_taxid_lineages
--------------

--------------
CREATE TABLE _new_taxid_lineages(taxid text, superkingdom_taxid text, kingdom_taxid text, phylum_taxid text, class_taxid text, order_taxid text, family_taxid text, genus_taxid text, species_taxid text)
--------------

Connecting to db
Selecting database idseq_development
Loading data from LOCAL file: /app/tmp/taxonomy/2019-09-17/_new_taxid_lineages.csv into _new_taxid_lineages
idseq_development._new_taxid_lineages: Records: 11  Deleted: 0  Skipped: 0  Warnings: 0
Disconnecting from db
--------------
DROP TABLE IF EXISTS _new_taxon_lineages
--------------

--------------
CREATE TABLE _new_taxon_lineages(taxid text, started_at text, version_start text, version_end text, superkingdom_taxid text, phylum_taxid text, class_taxid text, order_taxid text, family_taxid text, genus_taxid text, species_taxid text, superkingdom_name text, phylum_name text, class_name text, order_name text, family_name text, genus_name text, species_name text, superkingdom_common_name text, phylum_common_name text, class_common_name text, order_common_name text, family_common_name text, genus_common_name text, species_common_name text, kingdom_taxid text, kingdom_name text, kingdom_common_name text)
--------------

--------------
ALTER TABLE _new_taxon_lineages ADD UNIQUE idx(taxid(255))
--------------

--------------
INSERT INTO _new_taxon_lineages
      SELECT l.taxid,
      2019-09-17,
      5,
      5,
      `superkingdom`.taxid, `phylum`.taxid, `class`.taxid, `order`.taxid, `family`.taxid, `genus`.taxid, `species`.taxid, `superkingdom`.name, `phylum`.name, `class`.name, `order`.name, `family`.name, `genus`.name, `species`.name, `superkingdom`.common_name, `phylum`.common_name, `class`.common_name, `order`.common_name, `family`.common_name, `genus`.common_name, `species`.common_name, `kingdom`.taxid, `kingdom`.name, `kingdom`.common_name
      FROM _new_taxid_lineages l
      JOIN _new_names superkingdom ON l.superkingdom_taxid = superkingdom.taxid
      JOIN _new_names kingdom ON l.kingdom_taxid = kingdom.taxid
      JOIN _new_names phylum ON l.phylum_taxid = phylum.taxid
      JOIN _new_names class ON l.class_taxid = class.taxid
      JOIN _new_names `order` ON l.order_taxid = `order`.taxid
      JOIN _new_names family ON l.family_taxid = family.taxid
      JOIN _new_names genus ON l.genus_taxid = genus.taxid
      JOIN _new_names species ON l.species_taxid = species.taxid
--------------

Retire records...
1920336 records

Insert records...
1 records

Update records...
1 records

Keep same records...
9 records

Do you wish to execute the above irreversible changes? [y/N]

```


lib/tasks/update_lineagedb.rake
```diff
@@ -4,14 +4,30 @@ require 'English'
 
 desc 'Imports NCBI lineage data into IDseq'
 
-task 'update_lineage_db', [:dryrun] => :environment do |_t, args|
+# Lineage records have a [version_start, version_end] inclusive range. Each
+# pipline run has an AlignmentConfig that has a lineage_version. The pipeline
```
cdebourcy: To be slightly more precise, I would say something like: "The version of the NCBI taxonomy that specifies the lineages is kept in sync with the version of NCBI NT/NR used during alignment by comparing alignment_config.lineage_version to taxon_lineages.version_start / taxon_lineages.version_end. When serving reports, the correct lineage should be selected using the following condition: ..."
(you'll definitely word it better than me)

gregdingle: I added some more docs but I'm still not sure if I captured what you wanted. 

> When serving reports, the correct lineage should be selected 

From my reading of the code, the correct lineage is determined at write-time in `generate_aggregate_counts`. The report then merely selects the records associated with the current pipeline run.

cdebourcy: It's found dynamically when showing the taxon tree: https://github.com/chanzuckerberg/idseq-web/blob/e114826b65f9029af2e60257017dbbbb54f90c32/app/models/taxon_lineage.rb#L196

gregdingle:
> If you don't mind adding one more thing, I think it would be great to dump the old state of the table into a file and archive it somewhere in S3. That way if something goes wrong, we can just restore the previous state. It seems like an important addition for peace of mind.

I'll make a copy in mysql itself, then delete it in a month if there have been no issues.

cdebourcy:
> > If you don't mind adding one more thing, I think it would be great to dump the old state of the table into a file and archive it somewhere in S3. That way if something goes wrong, we can just restore the previous state. It seems like an important addition for peace of mind.
> 
> I'll make a copy in mysql itself, then delete it in a month if there have been no issues.

Ok that's fine. I meant in the update script itself in case someone else operates it in the future :)

PR #2664, opened 2019-10-12
Fix hover code in coverage viz for overlapping hits.

MarkAZhang:
# Description

Also separate the contig and read vizs and add the number of reads/contigs in the label (as requested by Jenn)

Previously, the hover code assumed that hits in the coverage viz would never overlap. Unfortunately, they occasionally do.

The previous code tried to use some optimized binary search fanciness to find the hit that was being hovered (and assumed that hits would never overlap)

The new hover code uses a naive approach. It just highlights the smallest overlapping hit. If there is no overlapping hit, it checks if there is a hit within a small buffer (which makes it easier to select tiny hits).

Furthermore, when a hit is highlighted, we ensure that it always appear on top of all other hits. (by re-drawing it in a container that is positioned above other hits)

New algorithm performs well when there are many hits. The coverage viz has a maximum of 500 hits, so the naive approach shouldn't have any issue (the gif below shows an accession with 471 hits).

Oct-14-2019 14-40-27

When there are overlapping hits, the appearance is still a little confusing. (the lighter highlighted flashes in the gif are when we hover over a read that appear underneath the contigs) But it is an improvement from before, when sometimes you would hover over a hit and nothing would appear at all.

Oct-14-2019 14-40-22

# Tests

* Verified that hover code still works properly when there are no overlapping hits.
* Verified that hover code behaves as expected when there are overlapping hits.
* Verified that the read viz doesn't show up when there are no loose reads.
* Verified that bands that represent aggregated reads and contigs show up in both vizs, and are colored differently.
* Will want to double-check on staging that there are no performance issues.

### DIFF
--- a/app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
+++ b/app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { find, get } from "lodash/fp";
+import { sum, find, get, isEmpty } from "lodash/fp";
 import cx from "classnames";
 import ReactDOM from "react-dom";
 
@@ -25,6 +25,8 @@ import {
   getHistogramTooltipData,
   generateCoverageVizData,
   getSortedAccessionSummaries,
+  selectContigsFromHitGroups,
+  selectReadsFromHitGroups,
 } from "./utils";
 import cs from "./coverage_viz_bottom_sidebar.scss";
 
@@ -237,7 +239,7 @@ export default class CoverageVizBottomSidebar extends React.Component {
       [[0, data.total_length, 0]],
       {
         domain: [0, data.total_length],
-        colors: [REF_ACC_COLOR],
+        color: REF_ACC_COLOR,
       }
     );
     this.refAccesssionViz.update();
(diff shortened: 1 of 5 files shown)


davidrissato:
Would it make sense to use transparency (fill-oppacity) on these blue bars to visually accommodate overlapping?

davidrissato:
Another thing that I noticed is that we are setting the height and top values for each `rect` element. Also noticed the current version defines a different class name for each of them as well.
Is there a special reason for that? I bet it might be faster using a CSS style, or if it is not possible, maybe we shouldn't create a class name for each entry, because that will allocate a lot of unnecessary CSS entries.

MarkAZhang:
Hey David, thanks for the feedback!

I removed the custom classname for each rect (rect-0, rect-1) as you suggested and used a filter instead.

It's hard to pass the `this.props.height` and `this.props.color` properties to CSS to implement your proposed solution. We would have to hard-code them.
I don't see a huge drawback to keeping the height and top and fill properties as is, as we are not drawing a ton of shapes. If we run into performance issues, we can revisit.

For the fill opacity, I think that should be a separate discussion. It can be tricky to find the right opacity value. (if you make it too low, then when nothing overlaps the shape will be too light. If you make it too high, then it's hard to tell between 2 overlaps and 5 overlaps.) It's also not that clear that the number of overlaps is that important to the user.

happyimadesignr:
This is awesome!! It seems to work a lot better! 

Can you make the space in between the bars smaller from 10px to 6px (space-xs)? They feel a bit far away and we can minimize the height a bit.

PR #2667, opened 2019-10-15
skip CDHITDUP if sample name contains PLEASE_SKIP_CDHITDUP

boris-dimitrov:
# Description

Allow users to request skipping CDHITDUP but adding text PLEASE_SKIP_CDHITDUP to the sample name

# Notes

This sets an attribute in the DAG json, which is then noticed by the pipeline, and has the effect of replacing cdhitdup with the identity function, copying inputs unchanged to outputs.

# Tests

Should confirm that in staging the automatically submitted benchmark sample does run with cdhitdup, and that submitting a new sample with appropriate name runs without it.

### DIFF
--- a/app/lib/dags/host_filter.json.erb
+++ b/app/lib/dags/host_filter.json.erb
@@ -123,7 +123,7 @@
       "class": "PipelineStepRunCDHitDup",
       "module": "idseq_dag.steps.run_cdhitdup",
       "additional_files": {},
-      "additional_attributes": {}
+      "additional_attributes": { "skip_cdhitdup": <%= @attribute_dict[:skip_cdhitdup] %> }
     },
     {
       "in": ["cdhitdup_out"],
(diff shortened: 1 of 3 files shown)


boris-dimitrov:
It's a little silly and goes against all principles.

But isn't it better than running custom branches?    Custom branches are a pain when stuff gets accidentally rerun in future.

At the very least this might be a foundation for someone to complete it into a real feature.

boris-dimitrov:
There is also a thing called `dag_vars` that's even persisted in the sample schema and pipeline_run schema.   I am not sure if that's what should be used instead.

cdebourcy:
> There is also a thing called `dag_vars` that's even persisted in the sample schema and pipeline_run schema. I am not sure if that's what should be used instead.

Yes, that's the intended purpose of dag_vars :)

tfrcarvalho:
Lets please not add this type of hacks to our codebase. If the current workflow (custom branches + `dag_vars`) is not enough for some reason or is not working as expected, we can review it and fix it.

PR #2669, opened 2019-10-16
[IDSEQ-1529] [Geosearch] Add autocompletion API calls for better results

jshoe:
### Description
- This adds new results from an autocompletion API from our location provider + other adjustments to make the results more relevant for our use cases. So now we call both `search` + `autocomplete` and zip the results (b/c combined results seemed better than either endpoint alone).
- See more explanation about 'search' vs 'autocomplete' in the comments.
- JIRA: https://jira.czi.team/browse/IDSEQ-1529

### Notes
- I paired with Jenn to figure out how to combine the geosearch vs. autocomplete results. We settled on zipping both lists, with the #1 from `autocomplete` coming first.
- This was somewhat subjective naturally, and there wasn't a clear winner even with a lot of sample queries. We also looked at the most common queries in Periscope. Happy to demo some comparisons in-person. Some examples: "San Fran", "San F", "USA", "United St", "Cam", "Cambodia", "Bangladesh", etc.
- `#external_search_action` was mostly extracted from `LocationsController#external_search`.

### Tests
- Manually test links such as: `http://localhost:3000/locations/external_search?query=USA` and check out the results.
- See unit tests.


spec/controllers/locations_controller_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe LocationsController, type: :controller do
+  describe "#external_search_action" do
+    # Extracted from LocationsController#external_search.
+    # See tests for external_search path in test/controllers/locations_controller_test.rb.
```
tfrcarvalho: Any plans to migrate these?

app/helpers/location_helper.rb
```diff
@@ -2,11 +2,13 @@ module LocationHelper
   # Adapter function to munge responses from Location IQ API to our format
   def self.adapt_location_iq_response(body)
     address = body["address"]
+    # Note(jsheu): 'type' field may contain a helpful exact name match.
```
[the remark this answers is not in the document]
jshoe: clarified 👍

app/helpers/location_helper.rb
```diff
@@ -134,10 +136,47 @@ def self.filter_by_name(samples_with_metadata, query)
   # See config/initializers/location_name_aliases.rb and add important known
   # aliases there.
   def self.normalize_name_aliases(name, geo_level)
+    # NOTE(jsheu): Provider v1/autocomplete endpoint often has repetitive
+    # country names such as "Cambodia, Cambodia". Dedupe them.
```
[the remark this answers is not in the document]
jshoe: I'm guessing whatever system generates the results for that endpoint considers it as "a place Cambodia in the country of Cambodia" and it happens for every single country. I probably wouldn't report it but it is one of the many idiosyncrasies of the systems..

app/helpers/location_helper.rb
```diff
@@ -134,10 +136,47 @@ def self.filter_by_name(samples_with_metadata, query)
...
+
+  # Combine results from provider autocomplete and geosearch endpoints for
+  # better results.
+  def self.handle_external_search_results(results)
+    autocomplete_results = results[Location::GEOSEARCH_ACTIONS[0]] || []
+    search_results = results[Location::GEOSEARCH_ACTIONS[1]] || []
+
+    # Zip/interpolate both lists (#1 from autocomplete, #1 from geosearch, #2
+    # from autocomplete, #2 from geosearch, etc).
+    if search_results.size > autocomplete_results.size
```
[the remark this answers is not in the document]
jshoe: yes, added a comment 👍

app/helpers/location_helper.rb
```diff
@@ -134,10 +136,47 @@ def self.filter_by_name(samples_with_metadata, query)
...
     if LOCATION_NAME_ALIASES.dig(geo_level, name)
       LOCATION_NAME_ALIASES[geo_level][name]
     else
       name
     end
   end
+
+  # Combine results from provider autocomplete and geosearch endpoints for
+  # better results.
+  def self.handle_external_search_results(results)
```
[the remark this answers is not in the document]
jshoe: That's a good idea. We could potentially tag each entry with that info and report it up in the analytics. I could also spot check when we get new data and see what list it likely came from for their queries.. will add a ticket for this as extra credit

spec/controllers/locations_controller_spec.rb
jshoe: moved two to get the ball rolling 👍 will move the others in the file gradually

app/controllers/locations_controller.rb
```diff
@@ -14,23 +14,23 @@ def external_search
...
-        msg = GEOSEARCH_RATE_LIMIT_ERR
-        msg += ": #{resp}" if resp
-        LogUtil.log_err_and_airbrake(msg)
-        raise msg
+      responses = {}
+      threads = []
+      Location::GEOSEARCH_ACTIONS.each do |action|
+        t = Thread.new { external_search_action(action, query, limit, responses) }
+        # Suppress verbose backtrace on handled exceptions
+        t.report_on_exception = false
```
MarkAZhang: Nit: Should we suppress this? Since you don't have a `rescue` block around the entire `external_search_action`, if an unexpected error happens, I think it will fail silently?

Would it be better to not suppress this, and also not re-raise exceptions that are expected (such as the rate limit one) if we are logging them?

I notice this is the first time we're using multi-threading in the code base, so paying a bit more attention to conventions.

app/helpers/location_helper.rb
```diff
@@ -2,11 +2,14 @@ module LocationHelper
   # Adapter function to munge responses from Location IQ API to our format
   def self.adapt_location_iq_response(body)
     address = body["address"]
+    # Note(jsheu): 'type' field may contain a helpful exact name match, but not
+    # necessarily. Ex: city, state, administrative, river, university, station..
+    category = body["type"]
```
MarkAZhang: Add an rspec test for this to `adapt_location_iq_response`?

app/helpers/location_helper.rb
```diff
@@ -134,10 +137,50 @@ def self.filter_by_name(samples_with_metadata, query)
   # See config/initializers/location_name_aliases.rb and add important known
   # aliases there.
   def self.normalize_name_aliases(name, geo_level)
```
MarkAZhang: Would it make sense to rename this to `clean_location_name`? `normalize_name_aliases` is kind of confusing, and you're now doing an additional kind of cleaning.

app/controllers/locations_controller.rb
jshoe: Actually the errors inside will still be raised when `join` is called (and the unit tests include this). The difference is that if this is `true` it would exit immediately and you will see a verbose stdout when running tests even if it was expected and handled. Will clarify this code comment.

`If it is guaranteed the Thread will be joined with #join or #value, then it is safe to disable this report with Thread.current.report_on_exception = false when starting the Thread` https://ruby-doc.org/core-2.5.0/Thread.html#method-c-report_on_exception

It would look like this when unsuppressed but successful ;)
Screen Shot 2019-10-16 at 1 55 42 PM

app/helpers/location_helper.rb
```diff
@@ -134,10 +137,50 @@ def self.filter_by_name(samples_with_metadata, query)
...
+      .map { |r| adapt_location_iq_response(r) }
+      .select { |r| Location::OSM_SEARCH_TYPES_TO_USE.include?(r[:osm_type]) }
+      .uniq { |r| [r[:name], r[:geo_level]] }
+      .uniq { |r| r[:osm_id] }
+  end
+
+  # Zip/interpolate two arrays. #1 from A, #1 from B, #2 from A, #2 from B...
+  # If one list is longer than the other, extra entries will be appended to the
+  # end of the combined list.
+  def self.zip_arrays(list_a, list_b)
```
MarkAZhang: Thoughts on moving this to `array_helper.rb`? Also renaming it "merge_arrays" since zip already means something?

app/controllers/locations_controller.rb
MarkAZhang: Ah I see. So `report_on_exception` will do an additional log.

MarkAZhang: Thanks for clarifying!

app/helpers/location_helper.rb
jshoe: added 👍

app/helpers/location_helper.rb
jshoe: good idea, changed 👍

app/helpers/location_helper.rb
jshoe: makes sense, extracted 👍

spec/lib/array_util_spec.rb
```diff
@@ -0,0 +1,53 @@
+require "rails_helper"
+
+RSpec.describe MetricUtil do
```
MarkAZhang: ArrayUtil, not MetricUtil

spec/lib/array_util_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  describe "#merge_arrays" do
+    it "zips List A and List B of equal length" do
+      a = [1, 3, 5]
+      b = [2, 4, 6]
+      actual = LocationHelper.zip_arrays(a, b)
```
MarkAZhang: Think you'll need to update this.

spec/lib/array_util_spec.rb
jshoe: thnx!

spec/lib/array_util_spec.rb
jshoe: thnx!

app/helpers/location_helper.rb
tfrcarvalho: Humm, I would not rename it... Lets not forget what is the real purpose of this function: It is to get a `normalized` name for each location to make sure that we do not have duplicates. 
It is not just cleaning...
If you do rename, please add a comment enforcing that goal.

jshoe: Hah, I like the `normalize` part, so how about just `normalize_location_name`

tfrcarvalho: Sure

jshoe:
✅ Jenn and Olivia

jshoe:
^ I decided to remove the `Thread.new` because I found odd bugs with threads hanging after trying to `raise` within them (and the exception handling is trickier than I thought). It seems just as fast this way..

PR #2670, opened 2019-10-16
Add BulkDownloadList page

MarkAZhang:
Also add common components BlankScreenMessage, LoadingMessage.

Add timestamp to BulkDownloads model

BulkDownloadModal now links to BulkDownloadList page.

Also add the Download link to the User Dropdown for users with the feature flag only.

Screen Shot 2019-10-15 at 6 22 13 PM

Screen Shot 2019-10-15 at 6 23 30 PM

Screen Shot 2019-10-15 at 6 24 07 PM

# Tests

* Test that bulk download modal still creates downloads correctly, and automatically takes user to the list page.
* Test that list page displays downloads correctly.
* Test that user without feature flag cannot access new controller endpoints or go to the bulk download list page or see the dropdown link in the user menu.
* Verify that components that now use BlankScreenMenu and LoadingMessage display properly.
* Wrote unit tests for the new controller methods.


app/assets/src/components/views/bulk_download/BulkDownloadDetailsModal.jsx
```diff
@@ -0,0 +1,91 @@
...
+    downloadType: null,
+  };
+
+  async componentDidUpdate(prevProps) {
+    if (
+      this.props.bulkDownload &&
+      prevProps.bulkDownload !== this.props.bulkDownload
+    ) {
+      this.setState({
+        bulkDownloadDetails: null,
```
tfrcarvalho: You do not need to clear download type too (since you set it below)?

app/assets/src/api/bulk_downloads.js
```diff
@@ -8,3 +8,8 @@ export const createBulkDownload = bulkDownload =>
     sample_ids: bulkDownload.sampleIds,
     params: bulkDownload.fields,
   });
+
+export const getBulkDownloads = () => get("/bulk_downloads.json");
+
+export const getBulkDownloadDetails = bulkDownloadId =>
```
tfrcarvalho: Could call it just `getBulkDownload`... more in like with the REST action `show`.

app/assets/src/components/views/bulk_download/BulkDownloadDetailsModal.jsx
```diff
@@ -0,0 +1,91 @@
...
+          ))}
+        </div>
+      </div>
+    );
+  };
+
+  render() {
+    const { open, onClose } = this.props;
+    return (
+      <Modal narrow open={open} tall onClose={onClose}>
```
tfrcarvalho: I guess this is an issue we have with plenty of Modals... The whole details are rendered independently of the modal being open or not ? Could we constrain the below elements to being open?

app/assets/src/components/views/bulk_download/BulkDownloadTableRenderers.jsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+
+import DownloadIcon from "~ui/icons/DownloadIcon";
+import LoadingIcon from "~ui/icons/LoadingIcon";
+
+import cs from "./bulk_download_table_renderers.scss";
+
+export default class BulkDownloadTableRenderers extends React.Component {
+  static renderDownload = ({ rowData }) => {
```
tfrcarvalho: I guess this will never be an asynchronous table, right?
Just good practice for renderers to have a behavior when rowData is null, but that will probably never happen with this one.

app/views/layouts/_page_header.html.erb
```diff
@@ -9,6 +9,7 @@
         signOutEndpoint: '<%=destroy_user_session_path%>',
         userName: '<%= current_user.name %>',
         userSignedIn: true,
+        allowedFeatures: JSON.parse('<%= raw escape_json(@current_user.allowed_feature_list) %>'),
         disableNavigation: <%= @disable_header_navigation %>,
       }, 'page_header', JSON.parse('<%= raw escape_json(request_context)%>'))
```
tfrcarvalho: Note the request context being passed here... so you should not need to pass allowedFeatures.

db/migrate/20191015190650_add_timestamp_to_bulk_downloads.rb
```diff
@@ -0,0 +1,6 @@
+class AddTimestampToBulkDownloads < ActiveRecord::Migration[5.1]
+  def change
+    add_column :bulk_downloads, :created_at, :datetime, null: false # rubocop:disable NotNullColumn
```
tfrcarvalho: These fields should not be overwritten.
We made the error of using these before. If you need explicit date fields you should create IN ADDITION to these control fields and let rails handle these...

app/assets/src/components/common/Header.jsx
```diff
@@ -123,15 +123,18 @@ Header.propTypes = {
   userSignedIn: PropTypes.bool,
   disableNavigation: PropTypes.bool,
   showBlank: PropTypes.bool,
+  allowedFeatures: PropTypes.arrayOf(PropTypes.string),
 };
 
+// TODO(mark): Separate user dropdown into sections with dividers, to make it more readable.
```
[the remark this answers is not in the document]
MarkAZhang: Created https://jira.czi.team/browse/IDSEQ-1641 and removed the TODO.

app/assets/src/components/views/bulk_download/BulkDownloadTableRenderers.jsx
MarkAZhang: I added a
```
if (!rowData) return null
```

Let me know if returning `null` is not correct here.

db/migrate/20191015190650_add_timestamp_to_bulk_downloads.rb
MarkAZhang: I forgot to add `t.timestamp` when I first created the table. It looks like Rails doesn't do this automatically unless you add that line to the table creation. If I add the timestamp later, it uses this syntax, but I'm actually intending for Rails to auto-populate these timestamps. (and Rails does handle it properly)

tfrcarvalho: Ah, cool. Thanks for clarifying

app/assets/src/components/views/bulk_download/BulkDownloadTableRenderers.jsx
tfrcarvalho: That works.

PR #2671, opened 2019-10-16
[JIRA-325] Upgrade to React 16.10.2

tfrcarvalho:
# Description
JIRA-325: https://jira.czi.team/browse/IDSEQ-325

* Upgrades React to last stable version (16.10.2)
* Upgrades a few other dependencies that were still using deprecated lifecycle methods:
	* rc-slider
	* react-moment
	* semantic-ui-calendar
	* semantic-ui-react
* Migrates one of our components (`ScatterPlot`) to use the `UNSAFE_` lifecycle methods. Eventually those should to be refactored before migrating further.

# Notes

* With this version (>16.5) we now have access to the profiler tool from react dev tools.

# Tests

* Hard to fully test the changes. Navigated through the website and have not found any issues.


MarkAZhang:
It looks like `AccessionViz` has a `componentWillUpdate` method. Append `UNSAFE` before that as well?

Should we make a JIRA task to convert all the UNSAFE methods? (luckily there are only a couple)

tfrcarvalho:
> It looks like `AccessionViz` has a `componentWillUpdate` method. Append `UNSAFE` before that as well?
> 
> Should we make a JIRA task to convert all the UNSAFE methods? (luckily there are only a couple)

Thank you the catch. Fixed and https://jira.czi.team/browse/IDSEQ-1639

PR #2674, opened 2019-10-17
[IDSEQ-1017] Use display names in ReviewStep in the upload flow

jshoe:
### Description
- JIRA: https://jira.czi.team/browse/IDSEQ-1017
- Before, the review step was not using the display names for the metadata column headers (and didn't have that info loaded in the component).

### Notes
- This adds an extra call to `getProjectMetadataFields`, which seemed cheap enough and cleaner to me than the alternative of moving that state onto the lowest common parent `SampleUploadFlow` and managing the state like project changes. This is parallel to `MetadataUpload` which also calls `getProjectMetadataFields` and can be used independently.
- The component hierarchy is: SampleUploadFlow => [UploadMetadataStep => MetadataUpload (calls getProjectMetadataFields), ReviewStep].

### Tests
- Go to /samples/upload, go through the steps, get to the review step, see display names like "Sample Type" instead of "sample_type".

#### Before:
Screen Shot 2019-10-16 at 6 14 58 PM

#### After:
Screen Shot 2019-10-16 at 6 02 17 PM

### DIFF
--- a/app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx
@@ -59,22 +59,18 @@ class ReviewStep extends React.Component {
   };
 
   getDataHeaders = () => {
-    const { uploadType } = this.props;
+    const { uploadType, metadata } = this.props;
     const { projectMetadataFields } = this.state;
 
     // Omit sample name, which is the first header.
     let metadataHeaders = without(
       ["Sample Name", "sample_name"],
-      this.props.metadata.headers
+      metadata.headers.map(
+        // Convert to use friendly name.
+        h => projectMetadataFields[h] && projectMetadataFields[h].name
+      )
     );
-    if (projectMetadataFields) {
-      metadataHeaders = metadataHeaders.map(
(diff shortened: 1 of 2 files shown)


app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx
```diff
@@ -23,10 +33,19 @@ const processMetadataRows = metadataRows =>
 class ReviewStep extends React.Component {
   state = {
     consentChecked: false,
-    showUploadModal: false,
+    projectMetadataFields: null,
     showLessDescription: true,
+    showUploadModal: false,
   };
 
+  async componentDidMount() {
```
tfrcarvalho: Lets keep lifecycle methods lean and not blocking.
You should move the fetch call to server to its own function and have `componentDidMount` return immediately.
Even if it works, since this is not a function you yourself call, it makes no sense to change its behavior.
Also, because you are not calling it directly, it make no sense to guarantee that you return a Promise.

jshoe: extracted 👍

PR #2675, opened 2019-10-18
Heatmap sort taxa

gregdingle:
# Description

Users have asked for the ability to sort the rows in the heatmap instead of clustering them. This add that ability alongside the new "sort samples" control.

image

The rows of the same genus are highlighted on hover.

image

# Notes

* See also https://github.com/chanzuckerberg/idseq-web/pull/2553
* And https://github.com/chanzuckerberg/idseq-web/pull/2676
* I also fixed a small CSS bug where pointer was showing for a non-clickable area

# Tests

Open a variety of heatmaps
See that the default taxa sort is "cluster"
Click on "sort taxa: genus"
See that rows are alphabetical
Try "genus" taxon level, see sorting
See that "All taxa with neither..." is at the bottom

Hover over row, See highlight
Hover off, See highlight disappear
Change to genus level, see no highlights on hover

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
@@ -218,6 +218,31 @@ export default class SamplesHeatmapControls extends React.Component {
     );
   }
 
+  onSortTaxaChange = order => {
+    if (order === this.props.selectedOptions.taxaSortType) {
+      return;
+    }
+
+    this.props.onSelectedOptionsChange({ taxaSortType: order });
+    logAnalyticsEvent("SamplesHeatmapControls_sort-taxa-select_changed", {
+      taxaSortType: order,
+    });
+  };
+
+  renderSortTaxaSelect() {
+    return (
+      <Dropdown
+        fluid
+        rounded
+        options={this.props.options.taxaSortTypeOptions}
+        value={this.props.selectedOptions.taxaSortType}
+        label="Sort Taxa"
+        onChange={this.onSortTaxaChange}
+        disabled={!this.props.data}
+      />
+    );
+  }
+
   onDataScaleChange = scaleIdx => {
     if (scaleIdx == this.props.selectedOptions.dataScaleIdx) {
       return;
@@ -397,6 +422,7 @@ export default class SamplesHeatmapControls extends React.Component {
           <div className="col s3">{this.renderTaxonLevelSelect()}</div>
           <div className="col s3">{this.renderCategoryFilter()}</div>
           <div className="col s2">{this.renderSortSamplesSelect()}</div>
+          <div className="col s2">{this.renderSortTaxaSelect()}</div>
           <div className="col s2">{this.renderMetricSelect()}</div>
           <div className="col s2">{this.renderBackgroundSelect()}</div>
         </div>
@@ -450,6 +476,12 @@ SamplesHeatmapControls.propTypes = {
         value: PropTypes.string,
       })
     ),
+    taxaSortTypeOptions: PropTypes.arrayOf(
+      PropTypes.shape({
+        text: PropTypes.string,
+        value: PropTypes.string,
+      })
+    ),
     sortTaxaOptions: PropTypes.arrayOf(
       PropTypes.shape({
         text: PropTypes.string,
@@ -483,6 +515,7 @@ SamplesHeatmapControls.propTypes = {
     ),
     readSpecificity: PropTypes.number,
     sampleSortType: PropTypes.string,
+    taxaSortType: PropTypes.string,
     dataScaleIdx: PropTypes.number,
     taxonsPerSample: PropTypes.number,
   }),
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
@@ -419,10 +419,10 @@ export default class SamplesHeatmapControls extends React.Component {
       <div className={cs.menu}>
         <Divider />
         <div className={`${cs.filterRow} row`}>
-          <div className="col s3">{this.renderTaxonLevelSelect()}</div>
-          <div className="col s3">{this.renderCategoryFilter()}</div>
-          <div className="col s2">{this.renderSortSamplesSelect()}</div>
+          <div className="col s2">{this.renderTaxonLevelSelect()}</div>
+          <div className="col s2">{this.renderCategoryFilter()}</div>
           <div className="col s2">{this.renderSortTaxaSelect()}</div>
+          <div className="col s2">{this.renderSortSamplesSelect()}</div>
           <div className="col s2">{this.renderMetricSelect()}</div>
           <div className="col s2">{this.renderBackgroundSelect()}</div>
         </div>
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
@@ -427,8 +427,8 @@ export default class SamplesHeatmapControls extends React.Component {
           <div className="col s2">{this.renderBackgroundSelect()}</div>
         </div>
         <div className={`${cs.filterRow} row`}>
-          <div className="col s3">{this.renderThresholdFilterSelect()}</div>
-          <div className="col s3">{this.renderSpecificityFilter()}</div>
+          <div className="col s2">{this.renderThresholdFilterSelect()}</div>
+          <div className="col s2">{this.renderSpecificityFilter()}</div>
           <div className="col s2">{this.renderScaleSelect()}</div>
           <div className="col s2">{this.renderTaxonsPerSampleSlider()}</div>
           <div className="col s2">{this.renderLegend()}</div>
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
@@ -193,14 +193,14 @@ export default class SamplesHeatmapControls extends React.Component {
     );
   }
 
-  onSortSamplesChange = order => {
-    if (order === this.props.selectedOptions.sampleSortType) {
+  onSortSamplesChange = selectedSortType => {
+    if (selectedSortType === this.props.selectedOptions.sampleSortType) {
       return;
     }
 
-    this.props.onSelectedOptionsChange({ sampleSortType: order });
+    this.props.onSelectedOptionsChange({ sampleSortType: selectedSortType });
     logAnalyticsEvent("SamplesHeatmapControls_sort-samples-select_changed", {
-      sampleSortType: order,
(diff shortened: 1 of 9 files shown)


app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -591,13 +603,38 @@ export default class Heatmap {
...
+      this.rowLabels,
+      label => label.sortKey || label.label,
+      direction
+    );
+    this.rowLabels.forEach((label, idx) => {
+      label.pos = idx;
+    });
+
+    // Need to sort this array as well. It is created from rowLabels earlier.
+    this.filteredRowLabels = orderBy(
```
tfrcarvalho: Isn't this the only thing that is shown and thus the only one you actually need to sort?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -836,6 +877,23 @@ export default class Heatmap {
...
+      .append("line")
+      .attr("x1", 0)
+      .attr("x2", this.rowLabelsWidth)
+      .attr("y1", this.cell.height)
+      .attr("y2", this.cell.height)
+      .attr("class", cs.genusBorder)
+      .classed(cs.hideGenusBorder, (label, i, nodes) => {
+        const nextLabel = this.filteredRowLabels[i + 1];
+        if (nextLabel) {
+          return label.label.split(" ")[0] === nextLabel.label.split(" ")[0];
```
tfrcarvalho: nit: you might want to change the name of the first variable... `label.label` fills very confusing.

app/assets/src/components/visualizations/heatmap/Heatmap.js
gregdingle: That's correct, but I thought it was important to also sort the `rowLabels` because that variable is used elsewhere and it would be surprising if it was in a different order.

gregdingle: Alternatively, naming it `filteredAndSortedRowLabels` would make the intentions clearer.

app/assets/src/components/visualizations/heatmap/Heatmap.js
gregdingle: Sorry this line was out of date. 

Anyway, the `label.label` name predates this change and is also in use by the AMR heatmap.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -52,7 +52,8 @@ export default class Heatmap {
         zoom: null, // multiplier for zooming in and out
         minHeight: 500,
         clustering: true,
-        shouldSortColumns: true,
+        shouldSortColumns: false,
```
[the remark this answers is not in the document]
tfrcarvalho: Since it is an option that you set out side makes more sense to me to use the imperative... But up to you

app/assets/src/components/visualizations/heatmap/Heatmap.js
tfrcarvalho: Yeah, if rowLabels is used somewhere else it would be important. I guess I find it weird if it is, since they do not match... Do you know what it is used for?

gregdingle:
See also https://github.com/chanzuckerberg/idseq-web/pull/2676.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -836,11 +896,28 @@ export default class Heatmap {
...
+          return true;
+        }
+      });
+
     rowLabelEnter
       .append("text")
       .attr("class", cs.removeIcon)
       .text("X")
-      .attr("transform", `translate(0, ${this.cell.height / 2})`)
+      .attr("transform", `translate(2, ${this.cell.height / 2})`)
```
gregdingle: I changed this to give the X some padding inside the gray background highlighting

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -809,8 +869,8 @@ export default class Heatmap {
       .enter()
       .append("g")
       .attr("class", cs.rowLabel)
-      .on("mousein", this.options.onRowLabelMouseIn)
-      .on("mouseout", this.options.onRowLabelMouseOut);
+      .on("mouseover", this.handleRowLabelMouseOver)
```
gregdingle: `this.options.onRowLabelMouseIn` and `Out` were not used anywhere. 

I made `handleRowLabelMouseOver` a method of `Heatmap` because the behavior is internal to the `Heatmap` and it is enabled by `shouldSortRows` option.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -591,13 +603,38 @@ export default class Heatmap {
 
   sortColumns(direction) {
     this.columnClustering = null;
+    this.gColumnDendogram.classed(cs.shouldSortColumns, true);
```
tfrcarvalho: Is the dendogram still present if we are sorting?
This functions should operate on data structures only.. .. the render functons would change anything related to display.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -591,13 +603,38 @@ export default class Heatmap {
...
         label.pos = idx;
       }
     );
   }
 
+  // Re-sorts the rows. The rendered order of rows is determined solely by
+  // the `pos` property of each rowLabel and filteredRowLabel.
+  sortRows(direction) {
+    this.rowClustering = null;
+    this.gRowDendogram.classed(cs.shouldSortRows, true);
```
tfrcarvalho: See comment above about this line.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -591,13 +603,38 @@ export default class Heatmap {
...
+    orderBy(
```
tfrcarvalho: Why do you need to sort this? Wouldnt it suffice to sort the filtered?

app/assets/src/components/visualizations/heatmap/Heatmap.js
tfrcarvalho: Can you put this as an option above?
Even if not changed often I prefer having this magical numbers defined above (a constant as an alternative)

app/assets/src/components/visualizations/heatmap/Heatmap.js
tfrcarvalho: This would be used if the client wants to add behavior. For instance, you would like to add a tooltip for this event, right? That could be done on the client.

app/assets/src/components/visualizations/heatmap/Heatmap.js
gregdingle: 👍

app/assets/src/components/visualizations/heatmap/Heatmap.js
gregdingle: Yes, no clients were adding behavior with those options. That's why I removed them, to avoid confusion with the new internal mouseover behavior. 

For the tooltip that was in @happyimadesignr 's mock, I'm planning now on adding that now in a follow-up PR.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -218,6 +218,31 @@ export default class SamplesHeatmapControls extends React.Component {
     );
   }
 
+  onSortTaxaChange = order => {
+    if (order === this.props.selectedOptions.taxaSortType) {
```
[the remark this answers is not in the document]
gregdingle: Hmmm....  it's named that way to go with the UI label "Sort taxa by". `currentSortType` wouldn't work because there is also a `sampleSortType` in `selectedOptions`.

app/assets/src/components/visualizations/heatmap/Heatmap.js
gregdingle: The dendrogram is hidden by CSS when sorting. I will add some comments and move those lines to the render functions.

tfrcarvalho: It's probably better to just return after `this.gColumnDendogram.select("g").remove();` in the render function. Why draw it?

app/assets/src/components/visualizations/heatmap/Heatmap.js
gregdingle: I'll remove that sort.

gregdingle: Will remove .

app/assets/src/components/visualizations/heatmap/Heatmap.js
gregdingle: No clients are using those options, and so I removed them to avoid confusion with `handleRowLabelMouseOver`.

For the tooltip as in @happyimadesignr 's mock, I'm planning on adding that in a follow-up PR.

app/assets/src/components/visualizations/heatmap/Heatmap.js
gregdingle: done

app/assets/src/components/visualizations/heatmap/Heatmap.js
gregdingle: I didn't want to touch the existing rendering because I didn't understand it. (First time working with SVG or D3). So taking your suggestion I will remove the CSS hiding and make the rendering conditional on `this.rowClustering`.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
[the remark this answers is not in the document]
tfrcarvalho: Sorry, I meant rename `order` -> `currentSortType`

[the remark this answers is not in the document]
gregdingle: ok good point

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -809,8 +862,8 @@ export default class Heatmap {
       .enter()
       .append("g")
       .attr("class", cs.rowLabel)
-      .on("mousein", this.options.onRowLabelMouseIn)
-      .on("mouseout", this.options.onRowLabelMouseOut);
+      .on("mouseover", this.handleRowLabelMouseOver)
```
tfrcarvalho: You want `mousein` here instead of `mouseover - or in the next pr - so it does not continuously trigger.

gregdingle: Good point. I switched to mouseover from mousein because it did not bubble up. When I hovered over the text inside the box the tooltip disappeared. Do you have any solutions for that?

tfrcarvalho: Was it triggering a mouse out as you were moving? 

Typically, on `mousein` you should create the tooltip and save the position and clear it on `mouseout`. Mouse over should change anything...

gregdingle: The problem is that the mouseout is firing when hovering over the text label inside the g element.

PR #2676, opened 2019-10-18
Fix genus assignment in heatmap

gregdingle:
# Description

This fixes a bug in heatmaps that was revealed by sorting by genus ID. See https://github.com/chanzuckerberg/idseq-web/pull/2675 . `cleanup_missing_genus_counts!` assumes that "there should be a genus_pair for every species" in the result set, which is true for the report page but not the heatmap. 

The fix is simply to put `cleanup_missing_genus_counts` calls in the report page code path only. 

# Notes

* It looks like the genus ID was being incorrectly overridden for all heatmap taxa for the past two years. 
* When I first started working with the heatmap, I noticed the copious warnings in the logs, I asked if anyone knew what it meant, and then I learned to ignore it. 
* It also looks like the heatmap SQL is unnecessarily fetching genus counts then discarding them. I will remove this in a follow up PR.
* There is still a bunch of copied and pasted code between the report and heatmap

# Tests
Before: endpoint json contains all negative genus IDs
After: endpoint json contains correct genus IDs

Before
image

After
image

Report page
* unchanged
* correct IDs in json endpoint


app/lib/report_helper.rb
```diff
@@ -438,6 +438,9 @@ def self.fake_genus!(tax_info)
     # This is a workaround for a bug we are fixing soon... but leaving
     # in the code lest that bug recurs.  A warning will be logged.
     #
+    # Two years ago, the above comment was written. It is not clear whether the
```
jshoe: do you want to date these comments? 😅

tfrcarvalho: Should probably generate an alert for this... Make the bug more visible.

gregdingle: I thought git blame was enough for dating the comments. 

As for the alert, I think it is better to ask @boris-dimitrov or whoever would know and try to fix to root cause. There was a log warning already but we ignored it.

tfrcarvalho: Exactly because it was ignored, we have to raise its visibility to an error+alert. Until we fix it. It was ignore because it was a warning...

tfrcarvalho: OK, maybe we need to clarify this logic before raising a more drastic alert...

gregdingle: @jshoe said it was related to an old "hit calling" feature which affected samples for a few weeks. So it should still work correctly as a workaround for the report page. No need to raise an alert.

PR #2678, opened 2019-10-21
Revert monitor parallelism change

morsecodist:
# Description

Reverts the changes to parallelism that I made. 
- Removes the `parallel` gem (though it is still a dependency of rubocop)
- Returns the sharded approach to parallelism

I am keeping some of the error message handling and logging I added.

# Tests

To test this I started the application locally and ensured that the benchmarks were being properly run.


gregdingle:
What's the hold-up here?

PR #2680, opened 2019-10-22
Make taxon lineage versions not nullable

gregdingle:
# Description

This modifies the SQL schema of `taxon_lineages` to avoid data corruption. 

# Notes

* Today I deleted all existing rows in `taxon_lineages` that contained `NULL` in the two columns
* This change will take a few minutes to execute during deployment but it is push-safe because the old and new versions of the schema work with prod code. 

# Tests

`rails db:migrate`


gregdingle:
> Any estimate of how long the deploy will take in prod?

Each `alter table modify` took 3 min locally, and prod has the same data as local, so I estimate it will add 6 min to the deployment. 

> Looks like Travis still has an error

Should be fixed now.

jshoe:
> Each `alter table modify` took 3 min locally, and prod has the same data as local, so I estimate it will add 6 min to the deployment.

Ok, heads up @davidrissato / @j-x-han for when this goes out

PR #2683, opened 2019-10-23
[IDSEQ-1331] Disable redundant use of UglifyJS to speed up deploys

jshoe:
### Description
- It looks like we have an extra invocation of UglifyJS that we don't need and slows down our containers on deployment (specifically on the first request post-boot up).
- There may be more deploy-related PRs to follow this.

### Notes
- In our Dockerfile we have `npm run build-img`, which calls `webpack --config webpack.config.prod.js` which includes `minimizer` `UglifyJsPlugin` and already generates packed versions of `app/assets/src` files. As discovered before, this uses a lot of memory.
Screen Shot 2019-10-23 at 3 18 19 PM

- In staging.rb and prod.rb, we also have `config.assets.js_compressor = Uglifier.new(harmony: true)`, which uses the Uglifier gem (https://github.com/lautis/uglifier) which also uses UglifyJS. This is almost entirely redundant except for trivial files that are in `app/assets` but not in `app/assets/src` (where we do want to keep all the JS code).

- In my testing, I found that `js_compressor` line to be the likely culprit behind very slow first requests. In the asset pipeline docs, there are modes that reference "on the first request the assets are compiled" (https://guides.rubyonrails.org/v4.0.0/asset_pipeline.html) and that seems to be how our Rails app was behaving. Unfortunately I couldn't get the compressor to generate a nice verbose output, which would've been additional evidence.

Screen Shot 2019-10-23 at 2 18 29 PM

- Indeed, trying to run the compilation locally and manually was slow and I ran out of memory.

Screen Shot 2019-10-23 at 2 45 17 PM

### Testing
- As in the notes, I'm confident that nothing of value will be lost, if anything.
- Pushed a testing branch to staging with these settings. You can verify that the site still behaves as expected, and the assets served as still very ugly.
Screen Shot 2019-10-23 at 4 23 13 PM
Screen Shot 2019-10-23 at 4 06 08 PM

- Looked through files in Sources and Network tabs in Chrome in staging vs. prod and found that they had the same formatting.
Screen Shot 2019-10-23 at 3 48 26 PM

- I pushed my test branch multiple times and ran `curl -i https://staging.idseq.net` on a loop. Machines came up much faster and no longer saw the lag and timeout. In the ECS logs, service idseq-staging-web deploy logs were as expected.

### DIFF
--- a/config/environments/prod.rb
+++ b/config/environments/prod.rb
@@ -28,9 +28,8 @@ Rails.application.configure do
   # Apache or NGINX already handles this.
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
-  # Compress JavaScripts and CSS.
-  config.assets.js_compressor = Uglifier.new(harmony: true)
-  # config.assets.css_compressor = :sass
+  # Don't add an asset compressor here because we already minimize with webpack.
+  # Check out webpack.config.prod.js.
 
   config.assets.debug = true
   # Suppress logger output for asset requests.
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -28,9 +28,8 @@ Rails.application.configure do
   # Apache or NGINX already handles this.
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
-  # Compress JavaScripts and CSS.
-  config.assets.js_compressor = Uglifier.new(harmony: true)
-  # config.assets.css_compressor = :sass
+  # Don't add an asset compressor here because we already minimize with webpack.
+  # Check out webpack.config.prod.js.
 
   config.assets.debug = true
   # Suppress logger output for asset requests.


cdebourcy:
Amazing detective work, Jonathan! 

>  and the assets served as still very ugly.
ahaha

Out of curiosity, any number we can put on the speedup for posterity?

jshoe:
> Out of curiosity, any number we can put on the speedup for posterity?

The time from `Booting Puma` to the first `Completed 200 OK` should really be no more than 15 seconds (assuming you switched the logging to a more verbose mode like in dev). Versus the 2-3 min times in the before case.

config/environments/prod.rb
```diff
@@ -28,9 +28,8 @@
   # Apache or NGINX already handles this.
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
-  # Compress JavaScripts and CSS.
-  config.assets.js_compressor = Uglifier.new(harmony: true)
-  # config.assets.css_compressor = :sass
+  # Don't add an asset compressor here because we already minimize with webpack.
```
gregdingle: ❤️

config/environments/staging.rb
```diff
@@ -28,9 +28,8 @@
   # Apache or NGINX already handles this.
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
-  # Compress JavaScripts and CSS.
-  config.assets.js_compressor = Uglifier.new(harmony: true)
-  # config.assets.css_compressor = :sass
+  # Don't add an asset compressor here because we already minimize with webpack.
```
gregdingle: Out of curiosity, is there a way to share rails environments so we don't need to copy code like this?

jshoe:
> Question: when did `js_compressor` start slowing down deploys? I seem to remember it wasn't an issue when I first joined idseq.

I actually don't know, but I suspect it's been sneaking up on us and we've just been approaching breaking points where we didn't notice it as much before.. according to this post I read: "if you feed the minifier double the JavaScript, it will take longer than double the time to minify it" (https://medium.com/@adamhooper/make-uglifyjs-way-faster-by-using-it-sooner-c2c39a9ad27e) <-- Not sure if that's always true but we have added a lot of JS in 2019.

config/environments/staging.rb
jshoe: looks like yes 👍 https://stackoverflow.com/questions/14947833/remove-duplication-between-staging-rb-and-production-rb Would be a good clean-up task.

MarkAZhang:
Super cool!! Crazy that uglification uses that much RAM.

PR #2684, opened 2019-10-24
Add explanation to tooltips for grayed out Save and Heatmap buttons

j-x-han:
# Description
Add an explanation of the 2 sample minimum in order to Save a Collection or create a Heatmap. 
(The `Select at least 2 samples` line will not appear if 2+ samples have been selected.)

image
image

# Notes

* The tooltip text for ToolbarIcon has also been updated to have 0.3px letter-spacing as requested by Jenn (applied through `font-body-xxs-crop`).

### DIFF
--- a/app/assets/src/components/views/samples/SamplesView.jsx
+++ b/app/assets/src/components/views/samples/SamplesView.jsx
@@ -199,6 +199,7 @@ class SamplesView extends React.Component {
         disabled
         icon={heatmapIcon}
         popupText="Heatmap"
+        popupSubtitle="Select at least 2 samples"
       />
     ) : (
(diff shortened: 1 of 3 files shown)


app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -199,6 +199,7 @@ class SamplesView extends React.Component {
         disabled
         icon={heatmapIcon}
         popupText="Heatmap"
```
gregdingle: by the way, it would be nice to be consistent in how we phrase pop text as either verbs "Save a collection" or nouns "Heatmap", @happyimadesignr

PR #2686, opened 2019-10-25
Show genus tooltip on hover in heatmap

gregdingle:
# Description

This adds a tooltip on hover of the new genus groups in the heatmap according to https://chanzuckerberg.invisionapp.com/d/main#/console/18553478/385796250/inspect. See https://github.com/chanzuckerberg/idseq-web/pull/2675. 

image

# Notes

* This is most useful for viruses which don't have their genus in their species name
* The tooltip is fixed width to make it possible to center it
* The width appears wider than in the design because the width of the column is of the longest row labels
* It might be nice to add a delay effect to only show after 1 second
* I didn't use the built in BasicPopup because it doesn't play nice with SVG
* This took a lot of tweaking to get right!

# Tests

* Load heatmap, hover over labels
* See correct genus

* Try unknown genus

* Try zooming in and out and the same thing

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
@@ -202,7 +202,10 @@ class SamplesHeatmapVis extends React.Component {
     this.setState({
       rowGroupLegend: {
         label: `Genus: ${rowGroup.genusName || "Unknown"}`,
-        tooltipLocation: { left: rect.left, top: rect.top },
+        tooltipLocation: {
+          left: rect.left + rect.width / 2, // center
+          top: rect.top - 25, // depends on heights defined in CSS
+        },
       },
     });
     logAnalyticsEvent("SamplesHeatmapVis_row-group_hovered", {
@@ -428,14 +431,6 @@ class SamplesHeatmapVis extends React.Component {
             />
           )}
         {rowGroupLegend && <RowGroupLegend {...rowGroupLegend} />}
-        {
-          <RowGroupLegend
-            {...{
-              label: `Genus: Unknown`,
-              tooltipLocation: { left: 100, top: 400 },
-            }}
-          />
-        }
         {addMetadataTrigger && (
           <MetadataSelector
             addMetadataTrigger={addMetadataTrigger}
--- a/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
@@ -428,6 +428,14 @@ class SamplesHeatmapVis extends React.Component {
             />
           )}
         {rowGroupLegend && <RowGroupLegend {...rowGroupLegend} />}
+        {
+          <RowGroupLegend
+            {...{
+              label: `Genus: Unknown`,
+              tooltipLocation: { left: 100, top: 400 },
+            }}
+          />
+        }
         {addMetadataTrigger && (
           <MetadataSelector
             addMetadataTrigger={addMetadataTrigger}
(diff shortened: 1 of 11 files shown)


app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -733,6 +747,10 @@ export default class Heatmap {
     return Math.round(scale(value));
   }
 
+  verticalPosition(d) {
```
gregdingle: I added this just for clarity

gregdingle:
@tfrcarvalho  Thanks for the detailed review. I made most of the suggested changes. I changed the behavior of the tooltip to show at the top of the row labels if it would be too high instead of hiding it. 

image

I'm still struggling to make `mousein` work as desired. I'd appreciate a few min to discuss in person.

gregdingle:
@tfrcarvalho I corrected the mouse handling events with help from @MarkAZhang . They are now firing only when needed.

tfrcarvalho:
> @tfrcarvalho I corrected the mouse handling events with help from @MarkAZhang . They are now firing only when needed.

nit: Would you mind changing the naming of the function that handle the events to match the actual events that trigger them as per comment on the review?

gregdingle:
> nit: Would you mind changing the naming of the function that handle the events to match the actual events that trigger them as per comment on the review?

Yes, I did that as well. Sorry I forgot to push the changes.

MarkAZhang:
Can you check on the styling of the tooltip with @happyimadesignr? (should it be right-aligned, or is center-aligned fine?) Otherwise LGTM.

gregdingle:
I talked with @happyimadesignr today which led to the CSS adjustments in https://github.com/chanzuckerberg/idseq-web/pull/2686/commits/8a48b2c842b13b4541374eea50107dfe12d7bf96 .

app/assets/src/components/views/compare/SamplesHeatmapVis.jsx
```diff
@@ -211,6 +216,24 @@ class SamplesHeatmapVis extends React.Component {
...
 
+  handleRowGroupEnter = (rowGroup, rect, minTop) => {
+    const tooltipWidth = Math.max(MAX_ROW_GROUP_TOOLTIP_WIDTH, rect.width / 2);
+    this.setState({
+      rowGroupLegend: {
+        label: `Genus: ${rowGroup.genusName || "Unknown"}`,
+        tooltipLocation: {
+          left: rect.left + rect.width / 2 - tooltipWidth / 2, // center
+          top: Math.max(minTop, rect.top),
+          width: tooltipWidth,
```
gregdingle: @MarkAZhang and @tfrcarvalho: @happyimadesignr  and I would like to know whether you can think of any way to center the tooltip in the row label column while the tooltip's size is dynamic based on the text? 

All I can think of would be to render all the tooltips, keep them hidden with CSS, and show them only on hover. Thoughts?

tfrcarvalho: You can render an invisible placeholder that has the width of the label column (width of rect) and left 0 from the left of the enclosing rectangle. And then align the tooltip inside the placeholder.

MarkAZhang: You can do, in `RowGroupLegend`
```
{{
  transform: "translate(-50%)",
  left: tooltipLocation.left + 4, // offset because of right-alignment of labels
  top: tooltipLocation.top - 44, // depends on heights defined in CSS
}}
```

And then here, just do `left: rect.left + rect.width / 2`.

`translate(-50%)` translate by 50% of the dynamic width of the tooltip, which is what you want.

Percentage when using `translate` refers to the current element's width, whereas most CSS rules (like left, top, margin) refer to the parent element's width. The difference is a little confusing to remember, but helpful for cases like this.

gregdingle: Translate  / transform works perfectly! Bravo @MarkAZhang . You are making my CSS knowledge feel out of date!

PR #2689, opened 2019-10-25
Use instance storage on himem Batch CE

kislyuk:
# Description

This switches the Batch job scratch space manager from self-managed EBS to instance ephemeral storage.

# Notes

This is necessary to avoid orphaning volumes when Batch jobs get erroneously terminated before they have a chance to clean up their EBS volumes.

# Tests

Tested in isolation by running `aegea batch submit` with relevant options.

### DIFF
--- a/app/helpers/pipeline_runs_helper.rb
+++ b/app/helpers/pipeline_runs_helper.rb
@@ -66,7 +66,12 @@ module PipelineRunsHelper
                                  stage_name: "misc")
     command = "aegea batch submit --command=#{Shellwords.escape(base_command)} "
     command += " --name=idseq-#{Rails.env}-#{sample_id}-#{stage_name} "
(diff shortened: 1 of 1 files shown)


app/helpers/pipeline_runs_helper.rb
```diff
@@ -129,7 +134,7 @@ def upload_dag_json_and_return_job_command(dag_json, dag_s3, dag_name, key_s3_pa
 
   def install_pipeline(commit_or_branch)
     "pip install --upgrade git+git://github.com/chanzuckerberg/s3mi.git; " \
-    "cd /mnt; " \
+    "cd /mnt; rm -rf *; df -h; " \
```
davidrissato: I guess this probably is not an issue, but `rm -rf *` won't remove hidden files at root level, correct?

kislyuk: Correct. But we don't care because the only files of significant size would not be hidden.

PR #2692, opened 2019-10-28
Remove slow query

gregdingle:
# Description

This query was logically unnecessary but it was fast locally so I left it in. It actually appears to be slow in prod, with some very slow outliers (>10s). See https://chanzuckerberg.airbrake.io/projects/158403/performance/2439523645692588671?dur=90d%2F1d&last_n=90&sortColumn=stat.p95&sortDesc=true&page=1 . 

# Notes

The original active record query is compiled to: 
```
SELECT DISTINCT genus_taxid,
                family_taxid
FROM `taxon_counts`
INNER JOIN `pipeline_runs` ON `pipeline_runs`.`id` = `taxon_counts`.`pipeline_run_id`
WHERE `pipeline_runs`.`sample_id` IN
    (SELECT `samples`.`id`
     FROM `samples`
     INNER JOIN `projects` ON `projects`.`id` = `samples`.`project_id`
     WHERE ((project_id IN (?)
             OR projects.public_access = ?
             OR DATE_ADD(samples.created_at, INTERVAL projects.days_to_keep_sample_private DAY) < ?))
       AND `samples`.`id` IN (?))
  AND `taxon_counts`.`tax_id` IN (?)

```

* IMPORTANT: I took the opportunity to remove the `family_taxid` from `parent_ids`. I have no idea why this was originally included and it seems like those counts should always be filtered out downstream in `only_species_or_genus_counts!`.

# Tests

* `rails t app/helpers/heatmap_helper_test.rb`
* compare heatmap before and after visually
* compare set of parent ids before and after


app/helpers/heatmap_helper.rb
```diff
@@ -84,7 +84,7 @@ def self.sample_taxons_dict(params, samples, background_id)
 
     unless taxon_ids.empty?
       # Refetch at genus level using species level
-      parent_ids = species_selected ? [] : HeatmapHelper.fetch_parent_ids(taxon_ids, samples)
```
gregdingle: notice how `taxon_ids` was plucked from `details` then used to fetch `parents_ids`. instead we can include the parent ids in `details`.

cdebourcy:
FYI, I think report lines like this are the reason we fetch the family as well:
image

For non-genus-specific rows (negative genus taxids), we want to say which family they fall into. Looking at the code, it seems like the ability to represent those rows was dropped from the heatmap at one point? Should we then remove the "specific only" filter on the heatmap?

gregdingle:
> Looking at the code, it seems like the ability to represent those rows was dropped from the heatmap at one point

I've never seen any reference to family names in the code. However, the "specific only" feature still works, though I would support it being removed from the heatmap or at least renamed to something clearer.

jshoe:
^ I'd be supportive of removing the `specific` and family rows if it's so rarely relevant to simplify things. Thoughts from @katrinakalantar ?

gregdingle:
Could I get a review here? To be clear, the change should not change the `specific` feature at all in the heatmap. At most, it reveals that the feature was broken.

PR #2693, opened 2019-10-29
[Bug] Disable browser autocomplete of location fields

jshoe:
### Description
- Quick fix to a bug introduced last sprint.
- We changed the placeholder text of the location box to say "Enter a city, region, or country". Words like 'city' cause Chrome to try to show a suggestions box.
- We don't want this because it could cause user confusion vs. our intended suggestions box, so this disables the browser autocomplete.
- Also useful in other contexts like not showing email suggestions on a different flow (came up before).

### Notes
Before:
Oct-28-2019 17-31-46

After:
Oct-28-2019 17-32-28

### Tests
- See screenshots / try out the location box in the sample upload flow and other flows.

### DIFF
--- a/app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
+++ b/app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
@@ -292,6 +292,7 @@ BareDropdown.propTypes = forbidExtraProps({
   open: PropTypes.bool,
   selectOnBlur: PropTypes.bool,
   trigger: PropTypes.node.isRequired,
+  disableAutocomplete: PropTypes.bool,
 });
 
 BareDropdown.defaultProps = {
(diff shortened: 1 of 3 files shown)


app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
```diff
@@ -227,6 +229,7 @@ class BareDropdown extends React.Component {
         {...baseDropdownProps}
         className={dropdownClassName}
         onBlur={e => e.stopPropagation()}
+        disableAutocomplete={disableAutocomplete}
```
MarkAZhang: I don't think semantic-ui's Dropdown accepts disableAutocomplete. If not, maybe remove this?

jshoe: Thanks! not sure why I added it there..

PR #2694, opened 2019-10-29
[Bulk Download] Add aegea script and download status update endpoints.

MarkAZhang:
# Description

* Added aegea kickoff script to Rails back-end, which accepts the correct values for source files for the "Original Input File" bulk download type.
* Added *public* controller endpoints that accept a per-download access token, allowing the ecs task to ping the Rails server with success, error, and progress updates.
* Added UI for various statuses in bulk download list.
Screen Shot 2019-10-28 at 6 14 33 PM
* Factored out a couple of components (LoadingBar, StatusLabel)

# Notes
* How the bulk download works is:
  * When a bulk download is created, we kick off an ECS task with aegea.
  * We give the ECS task a "success-url", an "error-url" and a "progress-url", which it can ping with updates. Each of these URLs has a form like "/bulk_downloads/5/success/fbb546f2930fff6", where "fbb546f2930fff6" is an access token assigned to that particular download.
  * The per-download access token removes the need for any kind of user authentication.
  * After the ECS task completes, it will either ping the success-url or the error-url and the BulkDownload status field will update.
* This all assumes that the ECS task completes without failing in the middle. The ECS task (see idseq-services) DOES have an error block that surrounds the entire task so it should catch all normal errors. But we will probably also need an occasional lambda function that checks if any BulkDownloads have been "running" for a long time, in case tasks hang. (Thanks for David for suggesting)
* In order to test BulkDownloads locally, you will need to use something like `ngrok`. After running `ngrok http 3000`, you can restart your web service with `DEVELOPMENT_TUNNEL_URL=https://eb9417b1.ngrok.io docker-compose up web`. This allows the ECS task to ping your local web server. Make sure to bring down your tunnel when you are done.

# Tests
* Manually tested the bulk downloads locally, and verified that the UI works properly.
* Wrote rspec tests for all model and controller functions.


app/models/bulk_download.rb
```diff
@@ -25,4 +30,161 @@ def convert_params_to_json
...
+    ecr_image: "idseq-s3-tar-writer:latest",
+    fargate_cpu: 4096,
+    fargate_memory: 8192
+  )
+    command = "aegea ecs run --command=#{Shellwords.escape(shell_command)} "
+    command += "--task-role #{Shellwords.escape(task_role)} "
+    command += "--ecr-image #{Shellwords.escape(ecr_image)} "
+    command += "--fargate-cpu #{fargate_cpu} "
+    command += "--fargate-memory #{fargate_memory}"
+    command
```
davidrissato: I can see you using `Shellwords.escape` here, but it is even safer to pass parameters individually to open3.

Ex:
```ruby
def aegea_ecs_submit_command(...)
    ["aegea", "ecs", "run",
     "--command", shell_command,
     "--ecr-image", ecr_image,
     "--fargate-cpu", fargate_cpu,
     "--fargate-memory", fargate_memory ]
end

...

  def kickoff_ecs_task(shell_command)
    aegea_command = aegea_ecs_submit_command(shell_command)
    # notice that I'm using the splat operator (asterisk) here, because capture 3 wants separated arguments instead of an array
    command_stdout, command_stderr, status = Open3.capture3(*aegea_command)
```

app/models/bulk_download.rb
```diff
@@ -25,4 +30,161 @@ def convert_params_to_json
...
+      download_src_urls = samples.map do |sample|
+        sample.input_files.map { |input_file| "s3://#{ENV['SAMPLES_BUCKET_NAME']}/#{input_file.file_path}" }
+      end.flatten
+
+      # We use the sample name in the input name because it's what is visible to the user.
+      # Also, there might be duplicates in the original file name.
+      download_tar_names = samples.map do |sample|
+        # We assume that the first input file is R1 and the second input file is R2.
+        sample.input_files.map.with_index do |input_file, input_file_index|
+          "#{input_file.sample.name}__original_R#{input_file_index + 1}.#{input_file.file_type}"
```
davidrissato: Is there any special reason why not `input_file.name` here instead of trying to guess the name and order?
I guess this is because we can rename samples, but should we rename the "original" files?

I have a suggestion, instead of adding the word `_original` to the file name, what if we store it in a subfolder containing the sample name, like `samples/CURRENT_SAMPLE_NAME/fastqs/ORIGINAL_SAMPLE_NAME_R1.fastq.gz`.

app/controllers/bulk_downloads_controller.rb
```diff
@@ -75,6 +93,69 @@ def show
     render json: { error: BulkDownloadsHelper::BULK_DOWNLOAD_NOT_FOUND }, status: :not_found
   end
 
+  # GET /bulk_downloads/:id/presigned_output_url
+  def presigned_output_url
+    bulk_download_id = params[:id]
+
+    bulk_download = current_power.viewable_bulk_downloads.find(bulk_download_id)
```
jshoe: could you use `.find_by_id` here instead of `.find`? I'm wondering if someone could put strings in here like case statements

app/controllers/bulk_downloads_controller.rb
```diff
@@ -75,6 +93,69 @@ def show
...
+
+  # POST /bulk_downloads/:id/progress/:access_token
+  def progress_with_token
+    # set bulk download and validate access token in before_action
+    @bulk_download.update(progress: params[:progress].to_f)
+    render json: { status: "success" }
+  end
+
+  def set_bulk_download_from_id
+    @bulk_download = BulkDownload.find(params[:id])
```
jshoe: this should really use `current_power.viewable_bulk_downloads.find`, right? because some other methods assume that @bulk_download is something you are authorized to use

app/helpers/bulk_downloads_helper.rb
```diff
@@ -4,7 +4,15 @@ module BulkDownloadsHelper
   SAMPLE_NO_PERMISSION_ERROR = "You do not have permission to access all of the selected samples. Please contact us for help.".freeze
   SAMPLE_STILL_RUNNING_ERROR = "Some selected samples have not finished running yet. Please remove these samples and try again.".freeze
   SAMPLE_FAILED_ERROR = "Some selected samples failed their most recent run. Please remove these samples and try again.".freeze
-  BULK_DOWNLOAD_NOT_FOUND = "No bulk download was found with this id".freeze
+  BULK_DOWNLOAD_NOT_FOUND = "No bulk download was found with this id.".freeze
+  OUTPUT_FILE_NOT_SUCCESSFUL = "Bulk download has not succeeded. Can't get output file url.".freeze
+  INVALID_ACCESS_TOKEN = "The access token was invalid for this bulk download.".freeze
+  KICKOFF_FAILURE = "Unexpected error kicking off bulk download.".freeze
+  KICKOFF_FAILURE_HUMAN_READABLE = "Could not kick off bulk download. Please contact us for help.".freeze
+  PRESIGNED_URL_GENERATION_ERROR = "Could not generated a presigned url.".freeze
```
jshoe: nit: "Could not generated"

app/models/bulk_download.rb
```diff
@@ -25,4 +30,161 @@ def convert_params_to_json
...
+                                        expires_in: OUTPUT_DOWNLOAD_EXPIRATION).to_s
+    rescue StandardError => e
+      LogUtil.log_err_and_airbrake("BulkDownloadPresignError: #{e.inspect}")
+    end
+    nil
+  end
+
+  def server_host
+    if Rails.env == "prod"
+      "https://idseq.net"
```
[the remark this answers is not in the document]
jshoe: +1

app/assets/src/components/views/bulk_download/BulkDownloadList.jsx
```diff
@@ -33,11 +35,47 @@ const TABLE_COLUMNS = [
...
     dataKey: "status",
     label: "",
-    width: 240,
+    width: 190,
     cellRenderer: BulkDownloadTableRenderers.renderStatus,
   },
 ];
 
+const STATUS_TYPES = {
+  waiting: null,
```
jshoe: are these supposed to be `null`?

app/models/bulk_download.rb
```diff
@@ -25,4 +30,161 @@ def convert_params_to_json
   def self.viewable(user)
     user.bulk_downloads
   end
+
+  def validate_access_token(access_token)
+    return self.access_token == access_token
```
jshoe: could you also do `self.access_token.present?` here? I'm wondering if we'll ever get into a situation where access_token is "" or accidentally not set, and then anyone can access it

db/migrate/20191023235303_add_access_token_to_bulk_download.rb
```diff
@@ -0,0 +1,7 @@
+class AddAccessTokenToBulkDownload < ActiveRecord::Migration[5.1]
+  def change
+    add_column :bulk_downloads, :access_token, :string, null: false, default: ""
+    add_column :bulk_downloads, :progress, :float, limit: 24
```
jshoe: haha do we need the `limit: 24`?

app/models/bulk_download.rb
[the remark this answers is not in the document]
MarkAZhang: Done!

app/assets/src/components/views/bulk_download/BulkDownloadList.jsx
MarkAZhang: Yes, they are supposed to be the default (which renders as a gray label). I can remove them though, so it'll just resolve to undefined.

app/assets/src/components/views/bulk_download/BulkDownloadList.jsx
```diff
@@ -33,11 +35,47 @@ const TABLE_COLUMNS = [
...
+
+const getStatusDisplay = bulkDownload => {
+  if (bulkDownload.status === "success" && bulkDownload.error_message) {
+    return "complete with issue";
+  }
+  return STATUS_DISPLAY[bulkDownload.status];
+};
+
+const getTooltipText = bulkDownload => {
+  if (bulkDownload.status === "success" && bulkDownload.error_message) {
```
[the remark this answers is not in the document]
MarkAZhang: Yeah exactly. I'll add a comment.

app/controllers/bulk_downloads_controller.rb
MarkAZhang: Discussed offline.

app/controllers/bulk_downloads_controller.rb
MarkAZhang: I combined this method with the one below to make it less ambiguous. This one is for endpoints using the access token authentication, so there isn't a current user in this case.

app/assets/src/components/views/bulk_download/BulkDownloadList.jsx
```diff
@@ -71,6 +115,28 @@ class BulkDownloadList extends React.Component {
...
+      const outputFilePresignedUrl = await getPresignedOutputUrl(
+        bulkDownload.id
+      );
+      openUrl(outputFilePresignedUrl);
+    }
+  };
+
+  onCopyUrlClick = async bulkDownload => {
+    // This should only be clickable when the bulk download has succeeded
+    // TODO(mark): Handle error case. Want to change popup text, which involves a forceUpdate on the table.
```
tfrcarvalho: Tip: If you do not want to forceUpdate, you can always change the data value of the table. Should be seamless.

app/assets/src/components/views/bulk_download/BulkDownloadTableRenderers.jsx
```diff
@@ -24,13 +36,46 @@ export default class BulkDownloadTableRenderers extends React.Component {
...
+              Download File
+            </div>
+            <div className={cs.separator}>|</div>
+            <BasicPopup
+              trigger={
+                <div className={cs.link} onClick={rowData.onCopyUrlClick}>
+                  Copy Url
+                </div>
+              }
+              content="Copy a signed URL for your file which expires in 24 hours. For advanced users."
```
tfrcarvalho: Do we need the `For advanced users`? Seems bit condescending...

app/models/bulk_download.rb
MarkAZhang: That's a great point regarding truncation! Will make a note to add that to the docs...although most users will not be using our S3 upload (that's just for biohub)

Regarding the input-file name, input files can be named the same thing. Multiple samples could have input files named "sample_R1.fastq". Whereas I believe sample names must be unique. Also, it's more intuitive for the user when we name it based on the name of the current sample. They might not remember what the sample was originally called, or what the file was originally named when they uploaded it, so giving them that additional information doesn't seem to add much.

I tried to avoid using subfolders in the tarfile because I worried a more complex directory structure could be more confusing for users to deal with. Tried to keep it simple.

app/assets/src/components/views/bulk_download/BulkDownloadList.jsx
MarkAZhang: Thanks for the tip!

app/assets/src/components/views/bulk_download/BulkDownloadTableRenderers.jsx
```diff
@@ -1,10 +1,14 @@
...
 
 import DownloadIcon from "~ui/icons/DownloadIcon";
-import LoadingIcon from "~ui/icons/LoadingIcon";
+import StatusLabel from "~ui/labels/StatusLabel";
+import BasicPopup from "~/components/BasicPopup";
+import LoadingBar from "~ui/controls/LoadingBar";
 
 import cs from "./bulk_download_table_renderers.scss";
 
+const PROGRESS_BAR_MINIMUM = 0.04;
```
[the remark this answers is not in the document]
MarkAZhang: Good point. I added a prop to `<LoadingBar>` to show a hint and removed this magic variable.

app/models/bulk_download.rb
jshoe: sample names are unique per project, so is it possible to select from multiple projects?

app/models/bulk_download.rb
```diff
@@ -25,4 +31,159 @@ def convert_params_to_json
...
+    unless config_ecr_image.nil?
+      ecr_image = config_ecr_image
+    end
+    shell_command_escaped = shell_command.map { |s| Shellwords.escape(s) }.join(" ")
+    command = ["aegea", "ecs", "run"]
+    command += ["--command=#{shell_command_escaped}"]
+    command += ["--task-role", task_role]
+    command += ["--ecr-image", ecr_image]
+    command += ["--fargate-cpu", fargate_cpu]
+    command += ["--fargate-memory", fargate_memory]
```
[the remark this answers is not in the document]
davidrissato: Just a few style suggestions:
```
        ["aegea", "ecs", "run",
         "--command=#{shell_command_escaped}",
         "--task_role", task_role...
```
(https://github.com/rubocop-hq/ruby-style-guide#align-multi-line-arrays)
(note: I'm not suggesting using %w, just suggesting how to align multiline arrays)

and also about using return: (https://github.com/rubocop-hq/ruby-style-guide#no-explicit-return)

(actually you don't even need the command variable, if the array is the last thing on that method)

app/models/bulk_download.rb
```diff
@@ -25,4 +31,159 @@ def convert_params_to_json
...
+    success_url: self.success_url,
+    error_url: self.error_url,
+    progress_url: self.progress_url
+  )
+    command = ["python", "s3_tar_writer.py"]
+    command += ["--src-urls"]
+    command += src_urls
+    command += ["--tar-names"]
+    command += tar_names
+    command += ["--dest-url", dest_url]
```
davidrissato: In this case, there is an easy way to do that without using array concatenation:
```ruby
command = ["python", "s3_tar_writer.py",
                      "--src-urls", *src_urls,
                      "--tar-names", *tar_names,
                      "--dest-url", *dest_url]
```

app/models/bulk_download.rb
davidrissato: I personally think folders are more organized than have long names that doesn't reflect the original one. Specially considering what @jshoe just said.
Should we have something like "#{project_name}/#{sample_name}/#{original_file_name}"?
Opinions?
(Note that we may need to escape the project name, since it doesn't have the same constraints as input files have)

MarkAZhang: Let me check with @happyimadesignr. I'm okay with it.

MarkAZhang: I think this decision rests on how users will use our files.

One reason why it might be good for all the files to be in one directory is that it's easier for users to select multiple files to upload to an online tool (like NCBI).

I could see having a bunch of directories being annoying if you want to select multiple files.

Even programmatically, instead of saying "all the files in this directory", you have to do a little extra work to traverse the directories.

davidrissato: I can see this is a subjective topic.

The alternative for unique names without folders would probably be something like `<project_name>__<inferred_sample_name>.<file_type>`.

Checked production database and collected some stats:
```
Project.maximum("length(name)")
=> 51
Sample.maximum("length(name)")
=> 127
```

So potentially it could be something like:

`MY_PROJECT_NAME_WITH_51_CHARACTERSXXXXXXXXXXXXXXXX__MY_SAMPLE_NAME_WITH_121_CHARACTERSXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.fastq.gz`

I also collected the average name length, so we can have an informed decision:
```
Project.average("length(name)").to_i
=> 15
Sample.average("length(name)").to_i
=> 28
```
(average may not be the best metric, but just to have an idea)

So it will probably looks like this:
`MY_PROJECT_NAME__MY_SAMPLE_NAME_WITH_28_CHA.fastq.gz`

davidrissato: Additionally, it is worth pointing out that we may need to deduplicate project names (doesn't matter if it a folder or part of the file name), because there may have a collision after escaping the file name:

```
PROJECT 1
PROJECT_1
PROJECT*1
```
they all would need to be normalized to something like `PROJECT_1` (we don't want special characters in the file name)

MarkAZhang: Talked to Jenn and Katrina. It seems like scientists are used to long names, and would appreciate having everything in the root directory.

We can do `{sample_name}_{project_name}__original_r1.fastq.gz`. Sample name first so they can sort by file name.

Regarding the fact that we infer which file is "R1" and which one is "R2" based on index in `sample.input_files`, it looks like that's how we handle it when passing files to the pipeline as well. So at least it's consistent.

`pipeline_run_stage.rb`:
```
attribute_dict = {
      fastq1: sample.input_files[0].name,
}
attribute_dict[:fastq2] = sample.input_files[1].name if sample.input_files[1]  
```

MarkAZhang: Just saw your comments above. Thanks for doing those queries!

MarkAZhang: What's the best way to escape the project name? (is there a standard function I can use, or will have to replace a regex with `_`) I think we can do `{sample_name}__{escaped_project_name}__{project_id}__original.fastq.gz`. The project id will ensure uniqueness.

davidrissato: I would suggest using something similar to The sample name (https://github.com/chanzuckerberg/idseq-web/blob/8ed4fd810ec1cf66e7c9c1993ad875a5639b1a4c/app/models/input_file.rb#L9   )

Basically anything that is not a letter, number, underscore, minus sign or a dot, should be replaced by an underscore.

We also don’t want to allow the first letter of the name to be a dot or a minus sign because these things have a special meaning in that position (dot is a hidden file, and a minus sign could be interpreted as a parameter by some commands)

MarkAZhang:
@davidrissato I'm going to go ahead and submit this PR, and work on a follow-up PR today that handles the file naming. Since the file naming seems a little tricky, and the rest of this PR is already reviewed. None of this is visible to users yet, so we can continue to iterate.

app/models/bulk_download.rb
```diff
@@ -25,4 +31,169 @@ def convert_params_to_json
...
+  )
+    aegea_command = aegea_ecs_submit_command(shell_command)
+
+    command_stdout, command_stderr, status = Open3.capture3(*aegea_command)
+    if status.exitstatus.zero?
+      output = JSON.parse(command_stdout)
+      # Store the taskArn for debugging purposes.
+      self.ecs_task_arn = output['taskArn']
+      self.status = STATUS_RUNNING
+      save
```
davidrissato: I would use `save!` here, since we are not handling this return value.

app/models/bulk_download.rb
```diff
@@ -25,4 +31,169 @@ def convert_params_to_json
...
+    else
+      self.status = STATUS_ERROR
+      self.error_message = BulkDownloadsHelper::KICKOFF_FAILURE
+      save
```
davidrissato: Same here (`save!`)

PR #2695, opened 2019-10-29
Upgrade airbrake

gregdingle:
# Description

Airbrake released some new perf monitoring features which they've told me would be better with the upgrade. We've also fallen a full version behind because the last upgrade caused perf monitoring to break. I'm hoping this upgrade will not have that problem.

# Tests

docker-compose web build
load idseq in localhost, click around see no errors in console or rails logs


jshoe:
I think travis errored b/c you still need to add the Gemfile.lock

PR #2697, opened 2019-10-29
Implement initial version of new report service

j-x-han:
# Description

Refactor of the report service prototype implemented on [this branch](https://github.com/chanzuckerberg/idseq-web/commit/196dbd699bc66443d2f86286a35b081bf3ed2f12).

# Notes

* Includes rudimentary version of the new frontend from Tiago's branch.
* No tests for correctness of the results generated by the new service yet, those are to come in future PRs.
* Protected by `report_v2` feature flag, so users without access to that feature will be redirected to the home page instead.

# Tests

* Added tests to ensure that only users with the access `report_v2` feature can view the results generated from the new report.


app/controllers/samples_controller.rb
```diff
@@ -738,6 +743,12 @@ def samples_going_public
     render json: samples.to_json(include: [{ project: { only: [:id, :name] } }])
   end
 
+  def report_v2
+    @pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
```
tfrcarvalho: We can use a local variable here.

app/services/pipeline_report_service.rb
```diff
@@ -0,0 +1,279 @@
...
+    @timer = Timer.new("pipeline_report_service")
+    report = generate
+    @timer.publish
+    return report
+  end
+
+  def generate
+    pipeline_run = PipelineRun.find(@pipeline_run_id)
+    adjusted_total_reads = (pipeline_run.total_reads - pipeline_run.total_ercc_reads.to_i) * pipeline_run.subsample_fraction
+    @timer.split("initialize and adjust reads")
```
tfrcarvalho: Since we are now sending this to datadog we should not use spaces.

PR #2698, opened 2019-10-30
Improve location dropdown behavior

gregdingle:
# Description

This addresses the first five usability issues of IDSEQ-1512. Essentially, it changes the search input component to open up the dropdown whenever focus is on the input and there are results. 

I also adjusted `delayTriggerSearch` to make the search feel more responsive. 

# Notes

* I hacked a dom event ordering issue. See https://stackoverflow.com/questions/17769005/onclick-and-onblur-ordering-issue . Other solutions welcome. 

# Tests

* Start upload of a sample locally 
* On metadata page...
* click into blank location field, see no menu
* type 3 chars, see menu
* click out of menu, see menu disappear
* click on item in menu, see menu disappear
* tab into menu, see menu appear
* tab out of menu, see menu disappear
* on change of chars, see loading icon appear in 100ms
* enter "asdfasdf", see no plain text (no match) result

### DIFF
--- a/app/assets/src/components/ui/controls/LiveSearchPopBox.jsx
+++ b/app/assets/src/components/ui/controls/LiveSearchPopBox.jsx
@@ -123,7 +123,7 @@ class LiveSearchPopBox extends React.Component {
   handleBlur = _ => {
     // Give a chance for handleResultSelect to happen. Otherwise, the dropdown
     // will disappear before the item onClick fires.
-    setTimeout(() => this.setState({ focus: false }), 10);
+    setTimeout(() => this.setState({ focus: false }), 100);
   };
 
   buildItem = (categoryKey, result, index) => (
--- a/app/assets/src/components/ui/controls/LiveSearchPopBox.jsx
+++ b/app/assets/src/components/ui/controls/LiveSearchPopBox.jsx
@@ -43,16 +43,11 @@ class LiveSearchPopBox extends React.Component {
     }
   };
 
-  resetComponent = () => {
-    this.setState({
-      isLoading: false,
-      results: [],
-    });
-  };
-
   handleResultSelect = ({ currentEvent, result }) => {
     const { onResultSelect } = this.props;
-    this.resetComponent();
+    this.setState({
+      isLoading: false,
+    });
     onResultSelect && onResultSelect({ currentEvent, result });
   };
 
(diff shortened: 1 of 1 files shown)


app/assets/src/components/ui/controls/LiveSearchPopBox.jsx
```diff
@@ -99,25 +94,38 @@ class LiveSearchPopBox extends React.Component {
...
-        )}
-        icon="search"
-        loading={isLoading}
-        placeholder={placeholder}
-        onChange={this.handleSearchChange}
-        onKeyPress={this.handleKeyDown}
-        value={value}
-        disableAutocomplete={true}
-      />
+      <div onFocus={this.handleFocus} onBlur={this.handleBlur}>
```
gregdingle: This wrapper is necessary because semantic-ui input has its own onfocus behavior defined.

app/assets/src/components/ui/controls/LiveSearchPopBox.jsx
```diff
@@ -99,25 +94,38 @@ class LiveSearchPopBox extends React.Component {
...
   };
 
+  handleFocus = _ => {
+    this.setState({ focus: true });
+  };
+
+  handleBlur = _ => {
+    // Give a chance for handleResultSelect to happen. Otherwise, the dropdown
+    // will disappear before the item onClick fires.
+    setTimeout(() => this.setState({ focus: false }), 100);
```
gregdingle: another solution would be to use `mousedown` event instead of `onclick`, but `mousedown` is not supported by the item component

jshoe:
Thanks for doing this, seems like a huge improvement trying it locally!

For #3 and #4:
- #3: "If still processing, display “Loading” in the drop down menu and show loading indicator instead of magnifying glass icon." -> I think this was Olivia's idea and the idea is to have the box pop up and actually have the text "Loading" so that it's more prominent than just the magnifying glass.
- #4: "If there are no results but the Plain Text option, still display dropdown menu with “No Results” under “Location Results”, as well as the Plain Text option" -> This would be still having the sub-divider for "Location Results" but showing "No Results" under it so it's more prominent.

These can be separate from this PR though so I'm going to approve.

cc Olivia although I can't tag her here

gregdingle:
Thanks for bringing up those points @jshoe .

> "If still processing, display “Loading” in the drop down menu and show loading indicator

I think this is addressed by reducing the delay from 1s to 200ms. I think showing "loading" in the drop down would be too much. 

> If there are no results but the Plain Text option, still display dropdown menu with “No Results” under “Location Results”, 

Ok I was confused by this because the current plain text message already says "no results". I will take a closer look when adding the divider in the next PR. 

Thanks for testing locally. I agree it is much better now!

PR #2701, opened 2019-10-31
Add separator between location results and plain text option

gregdingle:
# Description

As title. 

image

image

# Notes

* I matched the CSS style I've saw elsewhere and I increased the margin to make the visual separation obvious. 
* I also changed the message when there are no results, to address https://jira.czi.team/browse/IDSEQ-1516 . It was going to be more work than I had time to add a new type of dropdown item that was non-selectable and different style than the `category` type. 
image
* I also took the opportunity to improve a workaround from my last PR. Now we use onMouseDown to get around the onBlur onClick race condition. 

# Tests

* open menu, see line
* type "asdf", see "no results" message

### DIFF
--- a/app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
+++ b/app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
@@ -94,6 +94,13 @@ class GeoSearchInputBox extends React.Component {
       });
     }
 
+    // TODO (gdingle): Horizontal Line Between Location Results and Plain Text Sections
+
+    // TODO (gdingle):
+    // If there are no results but the Plain Text option, still display dropdown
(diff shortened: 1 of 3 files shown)


jshoe:
can we reduce the spacing on the top? it looks off in the regular case

Screen Shot 2019-10-31 at 9 51 55 AM

oliviabholmes:
Can we also increase the vertical spacing between the horizontal rule and the "Use Plain Text" title?
67910697-f9446a80-fb40-11e9-84d0-bbce731a705e

gregdingle:
@jshoe @oliviabholmes I updated the spacing. See the screens in the PR description.

oliviabholmes:
Looks great!

PR #2706, opened 2019-10-31
Paginate s3 requests during upload

j-x-han:
# Description

`list_object_v2` only returns up to 1000 objects in a bucket, so if a user has >1000 miscellaneous files in their bucket, we wouldn't detect any valid sample files.

This change uses the [built-in enumerator](
https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/paging-responses.html) for a response object to paginate the responses and the first 10k objects in a bucket.

# Notes

10k limit imposed to avoid potential timeouts and memory issues (@davidrissato ).

I briefly looked into filtering the objects we scan so we only look at valid sample files, but it doesn't seem like AWS has a way to filter by suffix/extension, [only by prefix](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_ResponseSyntax).

# Tests

* Connected to `idseq-samples-test` bucket during the sample upload flow and used Rails logger to confirm that all 5000+ objects were scanned rather than just the first 1000.


app/helpers/samples_helper.rb
```diff
@@ -165,7 +165,11 @@ def parsed_samples_for_s3_path(s3_path, project_id, host_genome_id)
     s3_prefix = parsed_uri.path.sub(%r{^/(.*?)/?$}, '\1/')
 
     begin
-      entries = S3_CLIENT_LOCAL.list_objects_v2(bucket: s3_bucket_name, prefix: s3_prefix).contents.map(&:key)
+      entries = []
+      S3_CLIENT_LOCAL.list_objects_v2(bucket: s3_bucket_name, prefix: s3_prefix).each do |response|
+        entries += response.contents.map(&:key)
+        break if entries.length >= 10_000
```
jshoe: I think we don't bubble these errors back up, but for on call debugging assistance could you do something like `if entries.length >= 10_000` then `Rails.logger.info("User tried to list more than 10,000 objects in #{s3_path}")`

jshoe:
Thanks for finding this!!

app/helpers/samples_helper.rb
```diff
@@ -165,7 +165,14 @@ def parsed_samples_for_s3_path(s3_path, project_id, host_genome_id)
     s3_prefix = parsed_uri.path.sub(%r{^/(.*?)/?$}, '\1/')
 
     begin
-      entries = S3_CLIENT_LOCAL.list_objects_v2(bucket: s3_bucket_name, prefix: s3_prefix).contents.map(&:key)
+      entries = []
+      S3_CLIENT_LOCAL.list_objects_v2(bucket: s3_bucket_name, prefix: s3_prefix).each do |response|
+        entries += response.contents.map(&:key)
+        if entries.length >= 10_000
```
cdebourcy: Nice! I wonder, should the 10,000 limit be an app config or a named constant rather than a number hardcoded here?

PR #2707, opened 2019-10-31
Metadata "apply all" only when multiple rows

gregdingle:
# Description

As title. See IDSEQ-1528

# Tests

* Start an upload, go to metadata page
* See "apply to all" when more than one sample
* See none when one sample

* Do same for metadata upload

### DIFF
--- a/app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
+++ b/app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
@@ -156,7 +156,9 @@ class MetadataCSVLocationsMenu extends React.Component {
             isHuman={isRowHuman(row)}
             warning={CSVLocationWarnings[sampleName]}
           />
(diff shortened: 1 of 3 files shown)


app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
```diff
@@ -156,7 +156,9 @@ class MetadataCSVLocationsMenu extends React.Component {
             isHuman={isRowHuman(row)}
             warning={CSVLocationWarnings[sampleName]}
           />
-          {applyToAllSample === sampleName && this.renderApplyToAll()}
+          {applyToAllSample === sampleName &&
+            this.props.metadata.rows > 1 &&
```
jshoe: this should be `this.props.metadata.rows.length`, right?

gregdingle: good catch! I hadn't tested that yet 😛

PR #2708, opened 2019-10-31
Location match warning improvements

gregdingle:
# Description

See IDSEQ-1521.

* Show the warning only when there is a nonzero length value
* Show the warning on blur if applicable 

image

# Notes

* The other request in the IDSEQ-1521, the vertical alignment, turned out to be quite tricky. I'm not sure if it is worth it. 

# Tests

* Select plain text, see warning
* Remove text, see warning disappear
* Tab out of input, see warning disappear

### DIFF
--- a/app/assets/src/components/common/MetadataInput.jsx
+++ b/app/assets/src/components/common/MetadataInput.jsx
@@ -77,7 +77,7 @@ class MetadataInput extends React.Component {
           <GeoSearchInputBox
             className={className}
             // .warning reference in old .idseq-ui.input file
-            inputClassName={cx(warning && "warning")}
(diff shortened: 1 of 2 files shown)


app/assets/src/components/common/MetadataInput.jsx
```diff
@@ -89,14 +89,15 @@ class MetadataInput extends React.Component {
             }}
             value={value}
           />
-          {warning && (
-            <div className={cs.warning}>
-              <div className={cs.icon}>
-                <AlertIcon />
+          {warning &&
+            value && (
```
gregdingle: not sure why eslint is formatting on two lines :/

PR #2711, opened 2019-11-01
Fix metadata input vert align

gregdingle:
# Description

See https://jira.czi.team/browse/IDSEQ-1524. 

image

# Notes

* It was hard to tell the origin of the `middle` alignment because of CSS bundling. How can I verify that it is from semantic UI? 

* I had to remove some previous special case CSS from @jshoe 

# Tests

* Open metadata upload
* Check "apply to all" and location warning does not affect vert align

### DIFF
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -294,10 +294,6 @@ class MetadataManualInput extends React.Component {
 
           const inputClasses = cx(
             cs.input,
-            // Add extra bottom padding to get inputs to align with the input that has Apply to All under it.
-            column !== this.state.applyToAllCell.column &&
-              sample.name === this.state.applyToAllCell.sampleName &&
-              cs.extraPadding,
             // Add extra width to location inputs for long names.
             column.startsWith("collection_location") && cs.extraWidth
           );
--- a/app/assets/src/components/common/metadata_manual_input.scss
+++ b/app/assets/src/components/common/metadata_manual_input.scss
@@ -32,7 +32,12 @@
   .inputTable {
     margin-bottom: 20px;
 
+    td {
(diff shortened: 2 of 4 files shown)


oliviabholmes:
One other test case that I noticed things getting misaligned was 1) make location warning appear 2) have location warning disappear 3) confirm that vertical alignment is 👌

jshoe:
- Not sure about the bundle.min.. sometimes I see it show up in elements and say materialize.css or semantic.

- I noticed another bug when clicking around with this although it's probably from the previous PR. Getting a page crash when the field is empty and blurs.

Screen Shot 2019-11-01 at 9 44 07 AM
Nov-01-2019 09-43-44

MarkAZhang:
Hey Greg, for some reason it doesn't look like CSS sourcemaps are working locally for me, so everything appears as `bundle.min.css`. However, the `bundle.min.css` is annotated (certain lines will have a comment before them them that says Semantic UI), so you can click on the line number and see which minified block it belongs to.

Screen Shot 2019-11-01 at 9 53 33 AM

Screen Shot 2019-11-01 at 9 55 08 AM

Hopefully that helps while we figure out the css sourcemaps.

jshoe:
Two other bugs I noticed:

It won't close the other dropdowns now without selecting anything:
Nov-01-2019 09-55-46

The "non-specific" message is no longer showing up (ignore the missing county info which is probably a different issue):
Nov-01-2019 09-59-46

should be like this
Screen Shot 2019-11-01 at 10 02 29 AM

jshoe:
But feel free to merge this PR in since it's just CSS..

gregdingle:
@jshoe I fixed both bugs you reported in the last two commits. Good catches!

jshoe:
Sweet, thanks!

PR #2713, opened 2019-11-01
Added insert size metrics option

morsecodist:
# Description

This modifies the host filtering DAG JSON generator to generate insert size metrics. It adds the insert size metrics step for Paired-End reads. It adds an alignments bam file output to STAR fir Paired-End reads.

# Notes

See this PR https://github.com/chanzuckerberg/idseq-dag/pulls/215.

# Tests

Tested pipeline for Single-End and Paired-End reads as well as testing the assertions.


app/lib/dags/host_filter.json.jbuilder
```diff
@@ -73,6 +74,16 @@ json.steps do
     additional_files: { star_genome: attr[:star_genome] },
     additional_attributes: { output_gene_file: 'reads_per_gene.star.tab' },
   }
+  if attr[:fastq2]
+    steps << {
+      in: ['star_out'],
+      out: 'insert_size_metrics_out',
```
gregdingle: see comments in https://github.com/chanzuckerberg/idseq-dag/pull/215 about naming

app/lib/dags/host_filter.json.jbuilder
```diff
@@ -2,6 +2,10 @@ json.name attr[:dag_name]
 
 json.output_dir_s3 "s3://#{attr[:bucket]}/samples/#{attr[:project_id]}/#{attr[:sample_id]}/results"
 
+compute_insert_size_metrics = attr[:fastq2] and
+  attr[:host_genome] == 'human' and
+  attr[:sample_template] == 'DNA'
```
cdebourcy: All `sample_*` columns are deprecated. You have to go via Metadatum / MetadataField

app/lib/dags/host_filter.json.jbuilder
cdebourcy: How about just naming it verbatim after the tool that was used (CollectInsertSizeMetrics)? This has been the convention with STAR, bowtie2, etc for exmaple. That way it's clear where the results come from. Greg would you be comfortable with that?

cdebourcy:
I will be comfortable approving after the the nucleotide type is determined from the correct place (metadata table, not sample_template column). Would you mind also adding comments that all `sample_*` columns are deprecated in the relevant places where you got misled? We should really drop those columns, that migration was never finished up

cdebourcy:
> I will be comfortable approving after the the nucleotide type is determined from the correct place (metadata table, not sample_template column). Would you mind also adding comments that all `sample_*` columns are deprecated in the relevant places where you got misled? We should really drop those columns, that migration was never finished up

except `sample_notes`. It would be a nice courtesy for the team if you could help us bring that over the finish line or at least prevent future developers from getting misled like you were :)

morsecodist:
In light of some changes recently discussed I will rework this

PR #2714, opened 2019-11-01
Fix issue where pressing spacebar on search dropdown closes dropdown.

MarkAZhang:
# Description

This fixes the immediate issue (which seems quite serious) but introduces other annoying issues. Specifically, now when you click on the trigger or any items, the dropdown no longer closes at all. You have to click outside the dropdown.

I think the underlying issue is that we've implemented a lot of custom behavior around the semantic ui component, which semantic ui doesn't know about. The long-term solution might be to either write our own component, or adopt semantic-ui's behavior instead of our custom behavior.

### DIFF
--- a/app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
+++ b/app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
@@ -12,7 +12,7 @@
 // redundant with props.items.customNode and props.children.
 import React from "react";
 import cx from "classnames";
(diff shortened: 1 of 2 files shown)


MarkAZhang:
Fixes this issue: https://jira.czi.team/projects/IDSEQ/issues/IDSEQ-1705

Not super happy with the fix because it creates other UX issues, but it does solve the serious issue of the dropdown unexpectedly closing when spacebar is pressed in the search bar.

PR #2716, opened 2019-11-01
Add sample taxon report bulk download type.

MarkAZhang:
# Description
* This is the first resque-style bulk download.

* Expose background index endpoint to non-admin so we can show the background options in the download modal.
Screen Shot 2019-11-01 at 4 48 09 PM

* If backgrounds are loading, dropdown shows a "Loading" message.

Screen Shot 2019-11-01 at 2 11 38 PM

* Add "execution_type" field to download types to indicate how to execute the download.
* Refactor BulkDownloadTypeHelper a bit to make it easier to test.
* Add s3_tar_writer class that accepts data and streams it to an s3 tar file, similar to the s3_tar_writer service in idseq-services.

# Tests

* Added model and controller tests.
* Manually verified that previous download types still work, and sample taxon report generates the correct csv (taking the background into account)

### DIFF
--- a/app/assets/src/api/index.js
+++ b/app/assets/src/api/index.js
@@ -308,6 +308,11 @@ const createBackground = ({ description, name, sampleIds }) =>
     sample_ids: sampleIds,
   });
 
+const getBackgrounds = async () => {
+  const response = await get("/backgrounds.json");
+  return response.backgrounds;
+};
+
 const getCoverageVizSummary = sampleId =>
   get(`/samples/${sampleId}/coverage_viz_summary`);
 
@@ -384,4 +389,5 @@ export {
   validatePhyloTreeName,
   validateSampleFiles,
   validateSampleNames,
+  getBackgrounds,
 };
--- a/app/assets/src/components/views/bulk_download/ChooseStep.jsx
+++ b/app/assets/src/components/views/bulk_download/ChooseStep.jsx
@@ -52,7 +52,8 @@ class ChooseStep extends React.Component {
   renderOption = (downloadType, field) => {
     const { selectedFields, onFieldSelect, fieldOptions } = this.props;
     const selectedField = get([downloadType.type, field.type], selectedFields);
-    let dropdownOptions, loadingOptions;
+    let dropdownOptions = [];
+    let loadingOptions = false;
 
     switch (field.type) {
       case "file_format":
--- a/app/assets/src/components/views/bulk_download/ChooseStep.jsx
+++ b/app/assets/src/components/views/bulk_download/ChooseStep.jsx
@@ -50,9 +50,9 @@ class ChooseStep extends React.Component {
   };
 
   renderOption = (downloadType, field) => {
-    const { selectedFields, onFieldSelect } = this.props;
+    const { selectedFields, onFieldSelect, fieldOptions } = this.props;
     const selectedField = get([downloadType.type, field.type], selectedFields);
-    let dropdownOptions;
+    let dropdownOptions, loadingOptions;
 
     switch (field.type) {
       case "file_format":
@@ -63,7 +63,7 @@ class ChooseStep extends React.Component {
 
         return (
           <div className={cs.field} key={field.type}>
-            <div className={cs.label}>File Format:</div>
+            <div className={cs.label}>{field.display_name}:</div>
             <Dropdown
               fluid
               options={dropdownOptions}
@@ -85,10 +85,36 @@ class ChooseStep extends React.Component {
 
         return (
           <div className={cs.field} key={field.type}>
-            <div className={cs.label}>Taxon:</div>
+            <div className={cs.label}>{field.display_name}:</div>
             <Dropdown
               fluid
               options={dropdownOptions}
+              onChange={(value, displayName) =>
+                onFieldSelect(downloadType.type, field.type, value, displayName)
+              }
+              value={selectedField}
+            />
+          </div>
+        );
+      // TODO(mark): Consolidate this with the above fields once the set of fields is stable.
+      case "background":
+        if (fieldOptions.backgrounds) {
+          dropdownOptions = fieldOptions.backgrounds.map(background => ({
+            text: background.name,
+            value: background.id,
+          }));
+        } else {
+          loadingOptions = true;
+        }
+
+        return (
+          <div className={cs.field} key={field.type}>
+            <div className={cs.label}>{field.display_name}:</div>
+            <Dropdown
+              fluid
+              disabled={loadingOptions}
+              placeholder={loadingOptions && "Loading..."}
+              options={dropdownOptions || []}
               onChange={(value, displayName) =>
                 onFieldSelect(downloadType.type, field.type, value, displayName)
               }
@@ -185,6 +211,7 @@ ChooseStep.propTypes = {
   selectedFields: PropTypes.objectOf(PropTypes.objectOf(PropTypes.string)),
   onFieldSelect: PropTypes.func.isRequired,
   onContinue: PropTypes.func.isRequired,
+  fieldOptions: PropTypes.objectOf(PropTypes.arrayOf(PropTypes.any)),
 };
 
 export default ChooseStep;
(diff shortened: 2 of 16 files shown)


app/jobs/generate_bulk_download.rb
```diff
@@ -0,0 +1,12 @@
+# Job to generate a bulk download
+class GenerateBulkDownload
+  @queue = :generate_bulk_download
+  def self.perform(bulk_download_id)
+    Rails.logger.info("Start generating bulk download #{bulk_download_id}")
+    BulkDownload.find(bulk_download_id).generate_download_file
+  rescue => e
+    failure_message = "Bulk download generation failed for id #{bulk_download_id}"
+    Rails.logger.error(e)
+    LogUtil.log_err_and_airbrake(failure_message)
```
jshoe: maybe include the `e` in the failure_message so it's exposed in the Airbrake?

app/lib/s3_tar_writer.rb
```diff
@@ -0,0 +1,32 @@
...
+  def start_streaming
+    @stdin, _stdout, @wait_thr = Open3.popen2("aws", "s3", "cp", "-", @s3_dest_url)
+    # Set stdin to binary mode, since GzipWriter outputs binary.
+    @stdin.binmode
+    @gz = Zlib::GzipWriter.new(@stdin)
+    @tar = Gem::Package::TarWriter.new(@gz)
+  end
+
+  def add_file_with_data(file_path, data)
+    @tar.add_file_simple(file_path, 0o444, data.length) do |io|
```
jshoe: could you add a comment for the `0o444`

app/models/bulk_download.rb
```diff
@@ -258,10 +258,82 @@ def bulk_download_ecs_task_command
...
+            "#{get_output_file_prefix(pipeline_run.sample, cleaned_project_names)}" \
+              "taxon_report.csv",
+            report_csv
+          )
+        rescue
+          failed_sample_ids << pipeline_run.sample.id
+        end
+      end
+
+      s3_tar_writer.close
```
jshoe: do you think it's possible that the `close` actions will hang? / should we have a timeout for this?

app/helpers/bulk_download_types_helper.rb
```diff
@@ -15,12 +15,20 @@ module BulkDownloadTypesHelper
       display_name: "Sample Overviews",
       description: "Sample metadata and QC metrics",
       category: "report",
+      execution_type: "resque",
```
jshoe: maybe also put "resque" / "ecs" / "variable" in constants?

app/models/bulk_download.rb
MarkAZhang: That's a good point.  I think it's more likely that the `process_status` method will hang because that's the one that waits for the `aws s3 cp` thread to complete. Any suggestions for the best way to implement a timeout?

jshoe: I think this looks like what we want? https://flushentitypacket.github.io/ruby/2015/02/21/ruby-timeout-how-does-it-even-work.html

MarkAZhang: There's a link at the bottom of that that says ruby Timeout is "dangerous". I'm also not sure what a good timeout to set is. Maybe a monitoring solution like we have for our long-running samples would be best.

Would you be okay with us deferring this until later? I can add some log statements to this Resque task which will hopefully make it more obvious if something has hung.

app/lib/s3_tar_writer.rb
MarkAZhang: Done, also changed to 644 so the file owner can write to it.

app/models/bulk_download.rb
jshoe: yep sounds good 👍

app/assets/src/api/index.js
```diff
@@ -308,6 +308,11 @@ const createBackground = ({ description, name, sampleIds }) =>
     sample_ids: sampleIds,
   });
 
+const getBackgrounds = async () => {
+  const response = await get("/backgrounds.json");
```
tfrcarvalho: Be careful because this endpoint does not do access control.
We were supposed to add it as part of backgrounds/collections work but we have not started that project yet.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -50,9 +50,9 @@ class ChooseStep extends React.Component {
   };
 
   renderOption = (downloadType, field) => {
-    const { selectedFields, onFieldSelect } = this.props;
+    const { selectedFields, onFieldSelect, fieldOptions } = this.props;
     const selectedField = get([downloadType.type, field.type], selectedFields);
-    let dropdownOptions;
+    let dropdownOptions, loadingOptions;
```
tfrcarvalho: nit: initializing `loadingOptions` to false hints about the type of variable. Good for readability.

app/controllers/backgrounds_controller.rb
```diff
@@ -2,13 +2,18 @@ class BackgroundsController < ApplicationController
   include BackgroundsHelper
   before_action :login_required
   before_action :no_demo_user
-  before_action :admin_required, except: [:create, :show_taxon_dist]
+  before_action :admin_required, except: [:create, :show_taxon_dist, :index]
```
tfrcarvalho: Why is admin required for index? 
This should not be the case since, and we will need to use it in future projects.

app/controllers/bulk_downloads_controller.rb
```diff
@@ -54,6 +54,7 @@ def create
       rescue => e
         # If the kickoff failed, set to error.
         bulk_download.update(status: BulkDownload::STATUS_ERROR)
+        Rails.logger.error(e)
```
tfrcarvalho: `log_err_and_airbrake` already does this. you do not need it. 
On the other hand we might not have the full stack trace.

app/controllers/bulk_downloads_controller.rb
```diff
@@ -33,7 +33,7 @@ def create
     rescue => e
       # Throw an error if any sample doesn't have a valid pipeline run.
       # The user should never see this error, because the validation step should catch any issues.
-      LogUtil.log_backtrace(e)
+      Rails.logger.error(e)
```
tfrcarvalho: Same as below.

app/jobs/generate_bulk_download.rb
```diff
@@ -0,0 +1,12 @@
+# Job to generate a bulk download
+class GenerateBulkDownload
+  @queue = :generate_bulk_download
+  def self.perform(bulk_download_id)
+    Rails.logger.info("Start generating bulk download #{bulk_download_id}")
+    BulkDownload.find(bulk_download_id).generate_download_file
+  rescue => e
+    failure_message = "Bulk download generation failed for id #{bulk_download_id}"
+    Rails.logger.error(e)
```
tfrcarvalho: Same as above, if you include e on the error message.

app/lib/s3_tar_writer.rb
tfrcarvalho: Actually isn't 644 too open?
Who needs read access?

app/models/bulk_download.rb
```diff
@@ -128,7 +128,7 @@ def s3_tar_writer_command(
   end
 
   def download_output_key
-    "downloads/#{id}/#{BulkDownloadTypesHelper::BULK_DOWNLOAD_TYPE_TO_DISPLAY_NAME[download_type]}.tar.gz"
+    "downloads/#{id}/#{BulkDownloadTypesHelper.bulk_download_type_display_name(download_type)}.tar.gz"
```
tfrcarvalho: This `id` is the id of the bulk download, right? So it is guaranteed to be unique?

MarkAZhang: That's correct.

app/controllers/backgrounds_controller.rb
MarkAZhang: This excludes index from being admin_required, if I understand your comment correctly.

app/lib/s3_tar_writer.rb
MarkAZhang: Okay, changed to 600 so only the file owner has access.

app/controllers/backgrounds_controller.rb
tfrcarvalho: Oh yeah, sorry. Read it wrong.

PR #2719, opened 2019-11-02
[Release Fix] Fix some issues when search is enabled on Dropdown.

MarkAZhang:
# Description

This is a follow-up to https://github.com/chanzuckerberg/idseq-web/pull/2714.

To fix the previous bug, we passed the "search" prop to semantic-ui so it wouldn't close the dropdown when Spacebar is pressed.
This caused other issues, which are fixed in this PR.

# Notes

* These solutions are kind of hacky, in that it is manually overriding behavior in semantic-ui. If we continue to run into more issues like this, we should think about a better long-term solution.

# Tests

* Manually tested that the ProjectSelect now works.
* Manually verify that ThresholdDropdown and MinContigSizeFilter don't close when you click on them. (since they use closeOnClick)

### DIFF
--- a/app/assets/src/components/ui/controls/dropdowns/bare_dropdown.scss
+++ b/app/assets/src/components/ui/controls/dropdowns/bare_dropdown.scss
@@ -66,6 +66,15 @@
     display: none;
   }
 
+  // semantic-ui applies max-height when search is true.
+  // We have our max-height on .innerMenu, so override this max-height
+  // to prevent double scrollbars.
+  &:global(.search) {
+    .menu {
+      max-height: none;
+    }
+  }
+
   .item {
     cursor: pointer;
 
--- a/lib/tasks/export_users_auth0.rake
+++ b/lib/tasks/export_users_auth0.rake
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'aws-sdk-ssm'
+
 desc 'Export legacy users to Auth0'
 
 task 'export_users_to_auth0' => :environment do |_t, _args|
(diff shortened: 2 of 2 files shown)


app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
```diff
@@ -141,17 +158,41 @@ class BareDropdown extends React.Component {
...
+    // for some reason. Manually close it in this case.
+    if (this.props.search) {
+      // When search is true, inject an onClick handler onto the trigger element.
+      wrappedTrigger = React.cloneElement(trigger, {
+        onClick: () => {
+          // Manually close the base dropdown if it's open.
+          if (
+            this.props.search &&
+            this.baseDropdownRef.current &&
+            this.baseDropdownRef.current.state.open
```
MarkAZhang: Accessing the state of a child component is generally bad practice. Doing it here only because it's a 3rd party component that doesn't support the behavior we want.

MarkAZhang: Adding this as a comment in the code directly.

tfrcarvalho: Isn't this a symptom that we are using the wrong component? Maybe we are officially using dropdown for too much?
It looks to me that we should plan for a long term solution for this...

MarkAZhang: Totally agree. We'd have to do a spike to see how much work replacing the Dropdown component would be. Does this feel high-priority or medium-priority to you?

tfrcarvalho: Maybe create a Jira ticket with medium priority for now... and lets chat next week.

app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
```diff
@@ -111,6 +113,20 @@ class BareDropdown extends React.Component {
     }
   };
 
+  handleMenuClick = e => {
+    const { closeOnClick, search } = this.props;
+
+    // When search is true, semantic-ui doesn't close the dropdown when the menu is clicked
+    // for some reason. Manually close it in this case.
+    if (search && this.baseDropdownRef.current) {
```
tfrcarvalho: Should we also check for `closeOnClick` on this if?
Not sure I fully understand it... but what would happen on a multiple selection with search, for instance? Like the metadata selector? Does a click in the option go through here?

MarkAZhang:
Closing this PR and separating into two PRs, one with less risky changes as a Release Fix.

PR #2727, opened 2019-11-06
Reduce complexity in LiveSearchPopBox.

MarkAZhang:
# Description

This is a refactor to make some of the logic in LiveSearchPopBox less confusing.

The primary source of confusion was that multiple variables were storing the same thing. `currentResult` and `value` were both storing what the user was typing, but being used in different ways. `currentResult` was also being modified in unintuitive ways, and was being used to overwrite the value on blur, which resulted in some special-case code that was hard to understand.

The component is also tricky because there are multiple ways to interact with it. The user can press enter, blur, or select an element. Furthermore, the user can click "Apply to All" to change the value of the element without interacting with it at all.

This refactor elements `currentResult` and `value` and replaces it with a single variable `inputValue` which stores the string that is currently being displayed in the input (a variable that is easy to understand). When the input is blurred, the `inputValue` is compared with `this.props.value`. If they are different, this means the user has typed something and then blurred without selecting, so we call `onResultSelect`.

# Notes

* Also removed some dead code.
* Also removed some instances where `value` was being stored in the state unnecessarily. Changed it to just use `props.value`.
* This does NOT solve the "warning" issue, where if you click Apply to All on a value, the warnings don't update. That will be in a follow-up PR.

# Tests
Tested the following:
* When the user blurs without clicking, it sets the currently typed text as plain text and shows a warning.
* If user presses enter while typing, it does the same.
* If user selects a location option, it should not show a warning.
* If user clicks Apply to All, it should apply the value to all.
* Focusing and blurring without typing should never change the value. (even after Apply to All)
* If the user deletes the text in the input, and then blurs (or clicks Continue directly), the UI should correctly register that the input is empty.
* If the user submits samples after Apply to All, pressing enter, etc, the samples are uploaded with the correct location metadata (whether it is plain-text or location object), and display properly in the map.

### DIFF
--- a/app/assets/src/components/ui/controls/LiveSearchPopBox.jsx
+++ b/app/assets/src/components/ui/controls/LiveSearchPopBox.jsx
@@ -2,7 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import Input from "~ui/controls/Input";
 import { BareDropdown } from "~ui/controls/dropdowns";
-import { forEach, sumBy, values, clone } from "lodash/fp";
+import { forEach, sumBy, values } from "lodash/fp";
 import cx from "classnames";
 import cs from "./live_search_pop_box.scss";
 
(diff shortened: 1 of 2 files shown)


app/assets/src/components/ui/controls/LiveSearchPopBox.jsx
```diff
@@ -13,55 +13,59 @@ class LiveSearchPopBox extends React.Component {
     this.state = {
       isLoading: false,
       results: [],
-      value: this.props.initialValue,
-      selectedResult: null,
-      currentResult: "", // either string or object
+      // the current value of the search input
+      inputValue: this.props.initialValue,
```
gregdingle: much better 😃

PR #2730, opened 2019-11-06
IDSEQ-1552 - Implement Auth0 Alpha Login Strategy

davidrissato:
# Description

Add auth0 login strategy in parallel to Devise.

We are currently using devise to handle authentication and authorization. As an intermediary migration step, add support to auth0 login strategy without removing Devise legacy authentication.

# Tests

* Manual tests using auth0
* Manual tests using previous authentication method


davidrissato:
Looks like I still need to fix something here

davidrissato:
Let me close this PR while I fix the issue. I'll reopen it soon.

app/assets/src/components/utils/links.js
```diff
@@ -43,10 +43,31 @@ const openUrlInPopupWindow = (url, windowName, windowWidth, windowHeight) => {
   );
 };
 
+const postToUrlWithCSRF = (url, params) => {
```
jshoe: this function might be better in `src/api/core.js`

app/controllers/application_controller.rb
```diff
@@ -27,6 +24,34 @@ def admin_required
     redirect_to root_path unless current_user && current_user.admin?
   end
 
+  # This method is used to verify if the user is already logged in,
+  # and will redirect it to the homepage in case it is not.
```
jshoe: nit: "in case they are not"

app/controllers/application_controller.rb
```diff
@@ -27,6 +24,34 @@ def admin_required
...
+  # This method overrides a default behavior from Devise to allow auth0
+  # authentication working simultaneously with legacy devise database mode,
```
jshoe: nit: "to work simultaneously"

app/controllers/application_controller.rb
```diff
@@ -27,6 +24,34 @@ def admin_required
...
+    if resp.nil?
+      # invoke devise if auth0 token is not present
+      super
+    else
+      if resp[:authenticated]
+        auth_payload = resp[:auth_payload]
+        return if current_user&.email == auth_payload["email"]
+        auth_user = User.find_by(email: auth_payload["email"])
+        if auth_user.present?
+          sign_in auth_user, store: false
```
jshoe: what does `store: false` do?

app/controllers/application_controller.rb
```diff
@@ -81,26 +106,28 @@ def get_background_id(sample)
     sample.default_background_id
   end
 
-  private
-
-  # Modified from https://gist.github.com/josevalim/fb706b1e933ef01e4fb6
-  def authenticate_user_from_token!
+  # This method is only used to loging users using tokens
```
jshoe: sp: s/loging/login

app/controllers/application_controller.rb
```diff
@@ -81,26 +106,28 @@ def get_background_id(sample)
...
+  # It is not intended to replace the authenticate_user! method, which
+  # must be invoked regardless.
+  # authenticate_user! is used for verifying if the user is already logged in
+  # and redirecting to the homepage in case it is not.
```
jshoe: nit: "in case they are not"

app/controllers/auth0_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  include Auth0Helper
+
+  def auth0_callback
+    # Store the user token that came from Auth0 and the IdP
+    # Auth0Helper#auth0_session
+    self.auth0_session = request.env['omniauth.auth'].credentials.id_token
+
+    # Redirect to the URL you want after successful auth
+    redirect_to "/home"
```
jshoe: can you do `redirect_to root_path` instead

app/helpers/auth0_helper.rb
```diff
@@ -0,0 +1,62 @@
...
+  delegate :present?, to: :auth0_session, prefix: true
+
+  def auth0_session=(value)
+    session[:jwt_id_token] = value
+  end
+
+  protected
+
+  def auth0_signout_url
+    domain = Rails.application.secrets.auth0_domain
```
jshoe: should we also get these from `ENV["AUTH0_DOMAIN"]` to be consistent?

app/lib/json_web_token.rb
```diff
@@ -0,0 +1,43 @@
...
+      iss: JWT_TOKEN_ISS,
+      verify_iss: true,
+      aud: Rails.application.secrets.auth0_api_audience,
+      verify_aud: true
+    ) do |header|
+      jwks_hash[header['kid']]
+    end
+  end
+
+  def self.jwks_hash
```
jshoe: could you add a comment or two to these methods about their function for future readers?

app/views/home/landing.html.erb
```diff
@@ -7,7 +7,8 @@
       {
         contactEmail: "accounts@idseq.net",
         showBulletin: <%= @show_bulletin %>,
-        browserInfo: <%= @browser_info.to_json.html_safe %>
+        browserInfo: <%= @browser_info.to_json.html_safe %>,
+        signInStrategy: location.href.match(/auth0/) ? 'AUTH0' : 'DEVISE',
```
jshoe: nit: could you move this line into the `HomeController#landing` method like `@sign_in_strategy`? Just want to keep these erb files very slim on logic

bin/entrypoint.sh
```diff
@@ -1,9 +1,16 @@
...
+fi
+echo ENVIRONMENT is set to \"$ENVIRONMENT\"
+
+if [ "$OFFLINE" = "1" ]
+then
+    echo Offline mode detected
     exec bundle exec "$@"
 else
     # Use Chamber to inject secrets via environment variables.
+    echo Chamber will reading entries from AWS SSM /idseq-$ENVIRONMENT-web/
```
jshoe: nit: s/reading/read

app/assets/src/components/utils/links.js
davidrissato: I took a look on `core.js`, and it seems the functions over there are all used to invoke json endpoints, whereas `links.js` has functions that replace the current location (which is the case here, with the difference I'm using a post operation to do that).
That said, it looked to me to be more logical to put this method on `links.js`. What do you think?

app/controllers/auth0_controller.rb
davidrissato: I noticed `root_path == "/"`. I wanted to redirect to home, since user is already authenticated. Let me see if there is a more elegant way to do that.

app/helpers/auth0_helper.rb
davidrissato: Actually I should be invoking this method instead. I started moving it here but it seems I ended up not completely doing that. Thank you for catching this.

app/controllers/auth0_controller.rb
davidrissato: Ok, looks like `home_path` is what I was looking for

app/assets/src/components/utils/links.js
jshoe: I see, `links.js` sounds good then

app/controllers/application_controller.rb
davidrissato: Devise documentation doesn't have info about that, but from what I could check, it prevents warden from creating a session, which is exactly what I wanted since I'm handling the session in a different place.
I could find some extra info about this behavior [here](https://github.com/plataformatec/devise/issues/3981)

davidrissato:
Found an issue in this PR. I'm closing it while I fix it, and I'll reopen it later on.

davidrissato:
Reopening this PR since. There was only a small issue related to not verifying the audience.
I also added comments to `json_web_token.rb` and added a caching mechanism for `jwks_hash`

app/lib/json_web_token.rb
davidrissato: Done

PR #2734, opened 2019-11-06
Add taxon with reads dropdown for bulk downloads.

MarkAZhang:
# Description

This adds a `taxon_with_reads_suggestions` endpoint that uses the existing elasticsearch-backed `taxon_search` and also adds a front-end dropdown that allows users to search for taxons that have reads (in the bulk download modal).

Screen Shot 2019-11-06 at 2 07 53 PM

Refactored the bulk download render options method to remove duplicate code.

# Notes
Added some additional search-related props to dropdown.

One prop adds the header directly above the search results (Taxa/Samples)
Another prop shows a "no results" message if we aren't still loading results and there are no results.

Note that "no results found" only appears after the search response comes back. We could also show a loading indicator, but that seems out of scope of this PR.
Nov-06-2019 14-07-58

The search querying is debounced (will only fire after the user hasn't typed for 200ms), and handles race conditions from response ordering. 

This doesn't cover the back-end implementation for this bulk download type. (dividing into smaller PRs)

# Tests

* Added controller tests.
* Verified that data discovery taxon search is unaffected by changes.
* Verified that component behaves as expected.
* Verified that other dropdowns in bulk download modal still work as expected.

### DIFF
--- a/app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
+++ b/app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
@@ -144,6 +144,9 @@ class BareDropdown extends React.Component {
       menuClassName,
       disableAutocomplete,
       trigger,
+      optionsHeader,
+      showNoResultsMessage,
+      loadingSearchOptions,
       ...otherProps
     } = this.props;
 
@@ -240,9 +243,17 @@ class BareDropdown extends React.Component {
             />
           </div>
         )}
+        {optionsHeader && (
+          <div className={cs.optionsHeader}>{optionsHeader}</div>
+        )}
         <BaseDropdown.Menu scrolling className={cs.innerMenu}>
           {filteredItems}
         </BaseDropdown.Menu>
+        {filteredItems.length === 0 &&
+          showNoResultsMessage &&
+          !loadingSearchOptions && (
+            <div className={cs.noResultsMessage}>No results found.</div>
+          )}
       </BaseDropdown.Menu>
     );
 
@@ -294,6 +305,8 @@ BareDropdown.propTypes = forbidExtraProps({
   hideArrow: PropTypes.bool,
   smallArrow: PropTypes.bool,
   menuLabel: PropTypes.string,
+  // Optional header that displays between the search box and the options.
+  optionsHeader: PropTypes.node,
   // whether the dropdown should close when you click on the menu. Useful for custom dropdown menus.
   closeOnClick: PropTypes.bool,
   search: PropTypes.bool,
@@ -303,6 +316,11 @@ BareDropdown.propTypes = forbidExtraProps({
   // If search is true, but you want to customize behavior of search function, e.g. async search,
   // you should provide your own handler
   onFilterChange: PropTypes.func,
+  showNoResultsMessage: PropTypes.bool,
+  // Don't show the no results message if search options are still loading.
+  // TODO(mark): Visually indicate that search options are loading even if
+  // there are old search results to display.
+  loadingSearchOptions: PropTypes.bool,
 
   // Custom props for rendering options
   options: PropTypes.arrayOf(
--- a/app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
+++ b/app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
@@ -146,7 +146,7 @@ class BareDropdown extends React.Component {
       trigger,
       optionsHeader,
       showNoResultsMessage,
-      loadingSearchOptions,
+      isLoadingSearchOptions,
       ...otherProps
     } = this.props;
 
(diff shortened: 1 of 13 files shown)


app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
```diff
@@ -303,6 +316,11 @@ BareDropdown.propTypes = forbidExtraProps({
   // If search is true, but you want to customize behavior of search function, e.g. async search,
   // you should provide your own handler
   onFilterChange: PropTypes.func,
+  showNoResultsMessage: PropTypes.bool,
+  // Don't show the no results message if search options are still loading.
+  // TODO(mark): Visually indicate that search options are loading even if
+  // there are old search results to display.
+  loadingSearchOptions: PropTypes.bool,
```
jshoe: I'm a little confused by `loadingSearchOptions` (maybe the naming or the comment). If true, then it doesn't display "no results" while loading?

app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -1,16 +1,55 @@
...
 
 import cs from "./choose_step.scss";
 
+const AUTOCOMPLETE_DEBOUNCE_DELAY = 200;
+
 class ChooseStep extends React.Component {
+  state = {
+    backgroundOptions: null,
+    taxonWithReadsOptions: null,
+    loadingTaxonWithReadsOptions: false,
```
jshoe: similar for `loadingTaxonWithReadsOptions`.. if it's called `Options` I'm not usually expecting a bool

app/controllers/samples_controller.rb
```diff
@@ -1258,6 +1259,29 @@ def coverage_viz_data
     }
   end
 
+  # GET /samples/taxon_with_reads_suggestions
+  # Get taxon search suggestions, where taxon are restricted to the provided sample ids.
```
jshoe: nit: "where taxa are restricted"

app/controllers/samples_controller.rb
```diff
@@ -1258,6 +1259,29 @@ def coverage_viz_data
...
+  # Also include, for each taxon, the number of samples that contain the taxon.
+  def taxon_with_reads_suggestions
+    sample_ids = (params[:sampleIds] || []).map(&:to_i)
+    query = params[:query]
+    samples = current_power.viewable_samples.where(id: sample_ids)
+
+    # User should not be querying for unviewable samples.
+    if samples.length != sample_ids.length
+      LogUtil.log_err_and_airbrake("Get taxons with reads error: Unauthorized access of samples")
```
jshoe: nit: s/taxons/taxa

app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -49,83 +88,154 @@ class ChooseStep extends React.Component {
...
+    }
+
+    if (!dropdownOptions) {
+      return null;
     }
+
+    return (
+      <div className={cs.field} key={field.type}>
+        <div className={cs.label}>{field.display_name}:</div>
+        <Dropdown
```
tfrcarvalho: It just does not feel like this should be using the Dropdown directly... So none of the other higher level components makes sense here?
It seems like setting up this dropdown was too complex...

app/controllers/samples_controller.rb
MarkAZhang: OMG this whole time I thought taxon was the plural of taxa. I've been doing this all wrong.

app/assets/src/components/ui/controls/dropdowns/BareDropdown.jsx
MarkAZhang: Would `isLoadingSearchOptions` be better? This indicates to the dropdown that the parent component is currently waiting for a response from the server. This would actually change between true and false on the same component depending on whether we're currently waiting for a response.

jshoe: yeah that makes more sense to me

MarkAZhang:
That's a great point, especially if the user has a lot of samples selected. Will keep an eye on it.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
MarkAZhang: Good point. `<LiveSearchPopBox>` is the one with the closest set of features, in that it also fetches data with a debounce. But that one is also quite different, because it allows users to pick an option OR type whatever they want, whereas in my case the user is required to pick an option. There's no clear alternative dropdown to use.

We should definitely think about refactoring these dropdowns soon.

MarkAZhang: One concrete thing we can do is refactor the ability to load options asynchronously into a wrapper component or something like that.

tfrcarvalho: Yeah, i guess i left this comment more as an extra note for future discussion.

PR #2735, opened 2019-11-06
Add Sample Overviews download type.

MarkAZhang:
# Description

Use existing csv generation code.

Refactor `generate_download_file` slightly as we start adding more download types.

# Notes

# Tests

* Added test for samples overview download type.
* Verified that CSV matches what is produced in the current download button.
* Verified that previous resque-style download (sample taxon reports) still works.

### DIFF
--- a/app/models/bulk_download.rb
+++ b/app/models/bulk_download.rb
@@ -2,6 +2,7 @@ require 'shellwords'
 
 class BulkDownload < ApplicationRecord
   include AppConfigHelper
+  include SamplesHelper
   include Rails.application.routes.url_helpers
   has_and_belongs_to_many :pipeline_runs
   belongs_to :user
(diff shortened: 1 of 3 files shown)


app/models/bulk_download.rb
```diff
@@ -307,23 +301,45 @@ def generate_download_file
...
+    Rails.logger.info("Waiting for streaming to complete...")
+    unless s3_tar_writer.process_status.success?
+      raise BulkDownloadsHelper::BULK_DOWNLOAD_GENERATION_FAILED
+    end
+
+    Rails.logger.info("Success!")
+    Rails.logger.info(format("Tarfile of size %s written successfully in %3.1f seconds", StringUtil.human_readable_file_size(s3_tar_writer.total_size_processed), Time.now.to_f - start_time))
+    update(status: STATUS_SUCCESS)
   rescue
     update(status: STATUS_ERROR)
```
tfrcarvalho: No logging?

MarkAZhang: This function is called by the Resque task (`app/jobs/generate_bulk_download.rb`), and the logging is done in a resque block there. Should I log here as well?

tfrcarvalho: I guess since it raises immediately we can keep the old behavior.

PR #2741, opened 2019-11-12
[Report Page] Initial table view

tfrcarvalho:
# Description

* Implements a table view based on react virtualized.
* Currently only for development purposes and guarded by flag.
* Current status: report page only, no corner cases handled (genus shows up blank)

# Tests

* Open show_v2 route on a samples /samples/<smaple_id>/show_v2
* See report table
* Test sorting of all columns (asc and desc)
* Test collapsing genus and species


tfrcarvalho:
> Would it worth exploring Netflix `fast_jsonapi`? ([Netflix/fast_jsonapi](https://github.com/Netflix/fast_jsonapi)).
> 
> If I understood it right, Netflix is claiming it is faster than other methods (including OJ). ([Netflix/fast_jsonapi:performance_methodology.md@`master`#benchmark-results](https://github.com/Netflix/fast_jsonapi/blob/master/performance_methodology.md#benchmark-results) [](/Netflix/fast_jsonapi/blob/HEAD@{2019-10-29T23:44:00Z}/performance_methodology.md#benchmark-results))

As discussed offline fast_jsonapi seems to focus on higher level of preparing active record objects to json but uses default JSON library or OJ to do the actually dumping into string. So, we should explore it in the future, but not for this particular service.

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
```diff
@@ -0,0 +1,404 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Table } from "~/components/visualizations/table";
```
MarkAZhang: Nit: Can we separate these imports into 3 groups: 3rd party imports ("react", "classnames"), imports with "~" or "../..", and imports from the same directory ("./")? It's a convention we've been using in other files.

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
```diff
@@ -0,0 +1,404 @@
...
+        dataKey: "ntnrSelector",
+        disableSort: true,
+        headerClassName: cs.ntnrSelectorHeader,
+        headerRenderer: this.renderNtNrSelector,
+        width: 40,
+      },
+    ];
+  }
+
+  nestedSortFunction = ({ data, path, sortDirection, nullValue, limits }) => {
```
MarkAZhang: This seems to be the most complex function in this class. Could you add some more comments?

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
```diff
@@ -0,0 +1,404 @@
...
+    // Uses lodash's orderBy function.
+    // It uses a sorting key that enables nested sorting of genus and species, while guaranteeing that
+    // genus is always on top of its children species
+    return orderBy(
+      [
+        rowData =>
+          rowData.genus
+            ? getOr(nullValue, ["genus"].concat(path), rowData)
+            : getOr(nullValue, path, rowData),
+        // this value guarantees that we keep species within their genus, even if the first value is duplicated
```
MarkAZhang: Not sure what this comment means. "the first value is duplicated" is vague.

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
```diff
@@ -0,0 +1,404 @@
...
+  }
+
+  nestedSortFunction = ({ data, path, sortDirection, nullValue, limits }) => {
+    // Uses lodash's orderBy function.
+    // It uses a sorting key that enables nested sorting of genus and species, while guaranteeing that
+    // genus is always on top of its children species
+    return orderBy(
+      [
+        rowData =>
+          rowData.genus
```
MarkAZhang: It looks like if `rowData.genus` exists, then that row is actually a species row? And if it doesn't exist, then it's a genus row. Maybe a comment explaining that. If rowData is for a species row and the genus is open, then it's always the case that `rowData.genus` is set? Even for species with non-unidentified genuses? (no worries if you haven't dealt with this case yet)

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
```diff
@@ -0,0 +1,404 @@
...
+            ? getOr(nullValue, ["genus"].concat(path), rowData)
+            : getOr(nullValue, path, rowData),
+        // this value guarantees that we keep species within their genus, even if the first value is duplicated
+        rowData => (rowData.genus ? rowData.genus.taxId : rowData.taxId),
+        rowData =>
+          rowData.genus
+            ? getOr(nullValue, path, rowData)
+            : sortDirection === "asc"
```
MarkAZhang: I'm a little confused why you need this `limits[0] : limits[1]` here. If `rowData.genus` doesn't exist, then it's a genus row, so it should already be sorted by the first key, right? It seems like it would be okay to have it always be the same number since it doesn't matter, instead of two numbers (limit[0] and limit[1]).

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
```diff
@@ -0,0 +1,404 @@
...
+    return this.renderNtNrValues({
+      cellData: cellData.map(val =>
+        TableRenderers.formatNumberWithCommas(
+          Number(val).toFixed(decimalPlaces || 0)
+        )
+      ),
+    });
+  };
+
+  handleNtNrChange = selectedDbType => {
```
MarkAZhang: Nit: Could we group all the render functions at the bottom and group similar functions together, instead of interspersing render functions with event handlers?

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
```diff
@@ -0,0 +1,404 @@
...
+        cellRenderer: ({ cellData }) =>
+          this.renderNtNrDecimalValues({ cellData, decimalPlaces: 1 }),
+        dataKey: "percent_identity",
+        label: "%id",
+        sortFunction: ({ data, sortDirection }) =>
+          this.nestedNtNrSortFunction({
+            data,
+            sortDirection,
+            path: ["percent_identity"],
+            nullValue: 0,
```
MarkAZhang: I'm not sure about this `nullValue` implementation. This means that if you sort `ascending`, then all the rows with null value would be the top, since null is interpreted at 0. Is that what you want, or do you always want the rows with null value to be at the bottom of the table?

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
```diff
@@ -0,0 +1,404 @@
...
+            path: ["contig_reads"],
+            nullValue: 0,
+            limits: NUMBER_NULL_VALUES,
+          }),
+        width: 75,
+      },
+      {
+        cellDataGetter: ({ rowData }) => {
+          return [
+            (rowData.nt || {}).percent_identity || 0,
```
MarkAZhang: Could we abstract calls like

```
[
  (rowData.nt || {}).contig_reads || 0,
  (rowData.nr || {}).contig_reads || 0,
];
```

into a getter function?

app/models/ui_config.rb
```diff
@@ -1,2 +1,11 @@
 class UiConfig < ApplicationRecord
+  after_initialize :set_default_values
```
MarkAZhang: Maybe move to a future PR? Not sure how this fits in with this PR.

app/assets/src/components/visualizations/table/Table.jsx
```diff
@@ -49,21 +49,24 @@ class Table extends React.Component {
...
+          sortDirection: sortDirectionStr,
+        });
+      } else {
+        sortedData = orderBy(
+          sortColumn.sortKey || sortBy,
+          [sortDirectionStr],
+          data
+        );
+      }
+    }
```
MarkAZhang: Looks great!

MarkAZhang: Maybe add some documentation around the parameters and outputs of `sortColumn.sortFunction` to the `propTypes` below.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -1,16 +1,68 @@
...
+          taxId: genusTaxId,
+          taxLevel: "genus",
+          species: speciesData,
+        })
+      );
+    });
+    this.setState({
+      data,
+    });
+    return data;
```
MarkAZhang: Nit: Return data not necessary?

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
MarkAZhang: Maybe my confusion can suggest an appropriate comment to put here.

tfrcarvalho: It matters because it is nested. No matter asc or desc you want the genus on top of their species.

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
tfrcarvalho: Commented.

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
tfrcarvalho: Added a comment.
rowData.genus is always set for a species and points to its genus row.
Non-unindentified genus have their own row with a particular negative id. They just have the name empty at this point (would add that later).

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
tfrcarvalho: Added some.

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
tfrcarvalho: Sure.

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
tfrcarvalho: Thats what i want. You should always be able to find the values even if null. So if you pushed them to the bottom when sorting ascending, they should be on top once you invert.
In some cases (with negative values) we might want to put them in the middle.

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
```diff
@@ -355,6 +288,67 @@ class ReportTableV2 extends React.Component {
...
+  }
+
+  nestedSortFunction = ({ data, path, sortDirection, nullValue, limits }) => {
+    // Uses lodash's orderBy function.
+    // It uses a triple sorting key that enables nested sorting of genus and species, while guaranteeing that
+    // genus is always on top of its children species
+    return orderBy(
+      [
+        // 1st value: value defined by path for the genus (guarantees all genus together)
+        // note: a species row as a point .genus to their genus
```
MarkAZhang: Nit: a species row has a field ".genus" that points to their genus.

app/assets/src/components/views/SampleViewV2/ReportTableV2.jsx
```diff
@@ -355,6 +288,67 @@ class ReportTableV2 extends React.Component {
...
+        rowData =>
+          rowData.genus
+            ? getOr(nullValue, ["genus"].concat(path), rowData)
+            : getOr(nullValue, path, rowData),
+        // 2nd value: the genus tax id
+        // this value guarantees that we keep species within their genus, even if the first value is duplicated
+        // e.g. if two genus have 2 reads they would be together on top and they species after them,
+        // adding tax id guarantees all the species are below their respective genus
```
MarkAZhang: This comment was super helpful. Thanks for that. I definitely didn't understand all the nuances before.

tfrcarvalho:
> I wish there was a way to generalize this to a "nested table" component that understood the nesting and automatically handled the two layer sorting.

I think it is totally doable. But given that this is our only nested table, i felt that this flattened sorting was worth it. The sorting function, after you understand the logic, is short, not that complex and can easily be extended to more levels.
If we have to support more nested tables, I think it would be worth doing that extra NestedTabled component that abstracts that logic.

PR #2745, opened 2019-11-13
[IDSEQ-1719] Add Auth0 functionality to "new user new project" flow

jshoe:
### Description
- JIRA: https://jira.czi.team/browse/IDSEQ-1719
- This allows auto-creation of new users via being added in project settings. The flow is as identical as possible to the existing one.

### Notes
- This updates the endpoint for the current `password reset request` form in IDseq. Unfortunately I couldn't find a way to link directly to the Auth0-hosted `forgot password` page so this is second best.
- The flows are flagged by `AppConfig::USE_AUTH0_FOR_NEW_USERS`. After we migrate all the users, we'll flip this switch.
- The `new_auth0_user_new_project` email has to be sent from our end because Auth0 doesn't allow that level of customization on their emails. You'd have to use the "password reset" email template to get the token included and you can't create new template types.
- See below for details.

### Tests

Full flow:
(1) Existing user adds new user from project settings.

(2) IDseq sends the user this email (same as current design).

Screen Shot 2019-11-12 at 5 11 37 PM

(3) User clicks "Activate your account" which takes them to this page:

Screen Shot 2019-11-12 at 5 11 39 PM

(4) User sets their password.

Screen Shot 2019-11-13 at 2 11 25 PM

- Tested that old Devise-based flows still work for `new user new project` and password resets.


app/controllers/projects_controller.rb
```diff
@@ -545,17 +545,39 @@ def validate_sample_names
   def create_new_user_random_password(name, email)
     Rails.logger.info("Going to create new user via project sharing: #{email}")
     user_params_with_password = { email: email, name: name }
-    random_password = SecureRandom.hex(10)
+    random_password = SecureRandom.base64(10)
```
jshoe: This is to satisfy a stricter password complexity policy

app/views/devise/passwords/new.html.erb
```diff
@@ -2,7 +2,7 @@
   <%= javascript_tag do %>
     react_component('ForgotPassword', {
       errors: `<%= raw devise_error_messages! %>`,
-      endpoint: '<%=  password_path(resource) %>',
+      endpoint: '/users/auth0_request_password_reset',
```
jshoe: new endpoint works with both Devise and Auth0

app/views/user_mailer/new_auth0_user_new_project.html.erb
```diff
@@ -0,0 +1,8 @@
+<p><b>Join IDseq!</b></p>
```
jshoe: This is copied with minor variable name changes from `app/views/devise/mailer/new_user_new_project.html.erb`, which should be deleted later

app/controllers/projects_controller.rb
davidrissato: I would say a random base64 is not a guarantee that will get a combination of all required characters (which I believe is at least one of each: uppercase/lowercase/digit/special_character).
This random base64 could eventually only contains number and letters and fail to pass. I believe we should either use a library to generate these bogus initial passwords (they will never be used, correct?) or append at least one character of each class to this password (something like `SecureRandom.base64(10) + "aA1!"`)

jshoe: We have it as "3 out of 4 of lowercase/uppercase/digits/special characters" (and that's the highest level they have). But you're right that it could still fail. Will update 👍

app/models/user.rb
```diff
@@ -193,4 +224,15 @@ def generate_authentication_token
...
     end
   end
+
+  # Set up Auth0 management client for actions like adding users.
+  # See: https://github.com/auth0/ruby-auth0/blob/master/README.md#management-api-v2
+  private_class_method def self.auth0_management_client
+    @auth0_management_client ||= Auth0Client.new(
+      client_id: ENV["AUTH0_MANAGEMENT_CLIENT_ID"],
+      client_secret: ENV["AUTH0_MANAGEMENT_CLIENT_SECRET"],
+      domain: ENV["AUTH0_DOMAIN"],
```
davidrissato: It would worth verifying if using custom domains here (like `login.idseq.net`) is fine, but I guess you may need to use tenant url here (ex: `czi-idseq-dev.auth0.com`). So you may need to set a `ENV["AUTH0_MANAGEMENT_DOMAIN"]`.

jshoe: will update 👍

jshoe: added `ENV["AUTH0_MANAGEMENT_DOMAIN"]` 👍

app/controllers/projects_controller.rb
jshoe: updated 👍

app/helpers/auth0_helper.rb
```diff
@@ -43,6 +43,23 @@ def auth0_signout_url
...
 
+  # See: https://auth0.com/docs/api/authentication#authorization-code-flow
+  def auth0_login_url
+    client_id = ENV["AUTH0_CLIENT_ID"]
+    connection = ENV["AUTH0_CONNECTION"]
+    domain = ENV["AUTH0_DOMAIN"]
+    request_params = {
+      client_id: client_id,
+      connection: connection,
+      redirect_uri: root_url + 'auth/auth0/callback',
```
davidrissato: NIT: `URI::join(root_url, 'auth/auth0/callback').to_s`

PR #2746, opened 2019-11-14
Adapt old sample report tests to new report service

j-x-han:
# Description

Converts [this test](https://github.com/chanzuckerberg/idseq-web/blob/master/test/controllers/samples_controller_test.rb#L110) into rspec and tests the new report service.

The test is split up into two:

1. species taxid 573 with genus taxid 570

- species NT and NR are both present in both the sample and the background model

2. species taxid 1313 and species taxid 28037, both from genus taxid 1310

- species 1313 NT and NR are present in the background model, but the sample only has NR
- species 28037 NT and NR are present in the background model, but the sample only has NT

# Notes

**Bugs addressed** (more details [here](https://czi.quip.com/u5ajAc1o5Chq/Report-Page-Optimization#bCHACAP4x4C)):
* Return the correct z-score when the taxon is missing in the sample (-100 if it's present in the background model, 0 if it's also missing in the background model.)
* If the species has NT/NR values but the genus is missing them, then use 100 as a placeholder for the genus z-score when calculating aggregate score.
  * The genus z-score will be displayed as 0 on the frontend.
  * [Logging](https://github.com/chanzuckerberg/idseq-web/pull/2746/files#diff-fe537067be5407d2c7c9e3a14329237cR263) has been added for these occurrences.

The new report service rounds values differently, so some of the expected values are not directly carried over from the old tests. The modified values in `pipeline_report_service_spec` are commented with what the old test previously expected.

The TaxonCounts factory was changed to pass `tax_id` as an input to TaxonLineage rather than always having the value be sequentially generated. `projects_controller_spec` had one test case slightly modified to account for this change.

### DIFF
--- a/app/services/pipeline_report_service.rb
+++ b/app/services/pipeline_report_service.rb
@@ -257,9 +257,15 @@ class PipelineReportService
   def compute_aggregate_scores(species_counts, genus_counts)
     species_counts.each do |tax_id, species|
       genus = genus_counts[species[:genus_tax_id]]
-      # Workaround placeholder for bad data (species NR is present in TaxonSummary but genus NR isn't)
+      # Workaround placeholder for bad data (e.g. species counts present in TaxonSummary but genus counts aren't)
       genus_nt_zscore = genus[:nt].present? ? genus[:nt][:z_score] : 100
       genus_nr_zscore = genus[:nr].present? ? genus[:nr][:z_score] : 100
+      if species[:nt].present? && genus[:nt].blank?
+        Rails.logger.warn("NT data present for species #{tax_id} but missing for genus #{species[:genus_tax_id]}.")
+      end
+      if species[:nr].present? && genus[:nr].blank?
+        Rails.logger.warn("NR data present for species #{tax_id} but missing for genus #{species[:genus_tax_id]}.")
+      end
 
       species[:agg_score] = (species[:nt].present? ? genus_nt_zscore.abs * species[:nt][:z_score] * species[:nt][:rpm] : 0) \
         + (species[:nr].present? ? genus_nr_zscore.abs * species[:nr][:z_score] * species[:nr][:rpm] : 0)
--- a/app/services/pipeline_report_service.rb
+++ b/app/services/pipeline_report_service.rb
@@ -121,7 +121,7 @@ class PipelineReportService
         sortedGenus: sorted_genus_tax_ids,
         highlighted_tax_ids: highlighted_tax_ids
       )
-    @timer.split("convert_to_json_with_sJSON")
+    @timer.split("convert_to_json_with_JSON")
 
     return json_dump
   end
@@ -147,7 +147,11 @@ class PipelineReportService
   end
 
   def zero_metrics(tax_id, tax_level, count_type, mean, stdev)
-    # fill in default zero values for FIELDS_TO_PLUCK
+    # Fill in default zero values for FIELDS_TO_PLUCK.
+    # Necessary for taxons absent from sample to match the taxon_counts_and_summaries structure,
+    # since they're fetched from TaxonSummary, which doesn't have some columns listed in FIELDS_TO_PLUCK.
+    # The two nil values (genus_taxid and name) will be filled in when counts_by_tax_level is
+    # transformed to be indexed by taxid and count type.
     [
       tax_id,
       nil,
@@ -253,8 +257,12 @@ class PipelineReportService
   def compute_aggregate_scores(species_counts, genus_counts)
     species_counts.each do |tax_id, species|
       genus = genus_counts[species[:genus_tax_id]]
-      species[:agg_score] = (species[:nt].present? && genus[:nt].present? ? genus[:nt][:z_score].abs * species[:nt][:z_score] * species[:nt][:rpm] : 0) \
-        + (species[:nr].present? && genus[:nr].present? ? genus[:nr][:z_score].abs * species[:nr][:z_score] * species[:nr][:rpm] : 0)
+      # Workaround placeholder for bad data (species NR is present in TaxonSummary but genus NR isn't)
+      genus_nt_zscore = genus[:nt].present? ? genus[:nt][:z_score] : 100
+      genus_nr_zscore = genus[:nr].present? ? genus[:nr][:z_score] : 100
+
+      species[:agg_score] = (species[:nt].present? ? genus_nt_zscore.abs * species[:nt][:z_score] * species[:nt][:rpm] : 0) \
+        + (species[:nr].present? ? genus_nr_zscore.abs * species[:nr][:z_score] * species[:nr][:rpm] : 0)
       genus[:agg_score] = species[:agg_score] if genus[:agg_score].nil? || genus[:agg_score] < species[:agg_score]
       # TODO : more this to a more logical place
       if !genus[:children]
--- a/app/services/pipeline_report_service.rb
+++ b/app/services/pipeline_report_service.rb
@@ -20,6 +20,7 @@ class PipelineReportService
   Z_SCORE_MIN = -99
   Z_SCORE_MAX =  99
   Z_SCORE_WHEN_ABSENT_FROM_BACKGROUND = 100
+  Z_SCORE_WHEN_ABSENT_FROM_SAMPLE = -100
 
   DEFAULT_SORT_PARAM = :agg_score
   MIN_CONTIG_SIZE = 0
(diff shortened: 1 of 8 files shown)


spec/services/pipeline_report_service_spec.rb
```diff
@@ -0,0 +1,261 @@
...
+          "z_score" => 2.208123824886411, # previously rounded to 2.2081236
+          "e_value" => 81.478,
+        },
+        "nr" => {
+          "count" => 2.0,
+          "rpm" => 1782.5311942959001, # previously rounded to 1782.531
+          "z_score" => 1.6768346926439197, # previously rounded to 1.6768345
+          "e_value" => 9.3,
+        },
+        "agg_score" => 73_603.777,
```
j-x-han: The [test](https://github.com/chanzuckerberg/idseq-web/blob/master/test/controllers/samples_controller_test.rb#L176) expects a genus aggregate score of 73603.777, but the report returns 12583.634591815486 (same as the species aggregate score).

spec/services/pipeline_report_service_spec.rb
```diff
@@ -0,0 +1,261 @@
...
+                           },])
+
+      @report = PipelineReportService.call(@pipeline_run.id, @background.id)
+    end
+
+    it "should get correct species values" do
+      species_result = {
+        "genus_tax_id" => 1301,
+        "name" => "Streptococcus pneumoniae",
+        "nt" => {
```
j-x-han: The returned report is currently missing nt.

spec/services/pipeline_report_service_spec.rb
cdebourcy: Thanks for flagging this! @tfrcarvalho do you know why?

tfrcarvalho: This a bug in the new report service - we are still investigating.
This a draft PR to facilitate communication between Julie and I.

j-x-han: Caught the bug, there was a typo in the test and the background model was missing some taxon summary data.

app/services/pipeline_report_service.rb
```diff
@@ -137,6 +146,48 @@ def fetch_taxon_counts(_pipeline_run_id, _background_id)
...
+    taxons_absent_from_sample = TaxonSummary
+                                .joins(
+                                  " LEFT JOIN taxon_counts ON ("\
+                                    " taxon_counts.count_type = taxon_summaries.count_type"\
+                                    " AND taxon_counts.tax_level = taxon_summaries.tax_level"\
+                                    " AND taxon_counts.tax_id = taxon_summaries.tax_id"\
+                                    " AND taxon_summaries.background_id = #{@background_id}"\
+                                    " AND taxon_counts.pipeline_run_id = #{@pipeline_run_id}"\
+                                  " )"\
+                                " WHERE"\
```
tfrcarvalho: Can we convert this to ActiveRecord where clauses?

app/services/pipeline_report_service.rb
```diff
@@ -137,6 +146,48 @@ def fetch_taxon_counts(_pipeline_run_id, _background_id)
...
+                                  " taxon_summaries.background_id = #{@background_id}"\
+                                  " AND taxon_counts.count IS NULL"\
+                                  " AND taxon_summaries.tax_id IN"\
+                                    " (SELECT DISTINCT tax_id FROM taxon_counts"\
```
tfrcarvalho: You should also be able to write this one as ActiveRecord query.

app/services/pipeline_report_service.rb
```diff
@@ -137,6 +146,48 @@ def fetch_taxon_counts(_pipeline_run_id, _background_id)
     return taxon_counts_and_summaries_query.pluck(*FIELDS_TO_PLUCK)
   end
 
+  def zero_metrics(tax_id, tax_level, count_type, mean, stdev)
+    # Fill in default zero values for FIELDS_TO_PLUCK.
+    # Necessary for taxons absent from sample to match the taxon_counts_and_summaries structure,
+    # since they're fetched from TaxonSummary, which doesn't have some columns listed in FIELDS_TO_PLUCK.
+    # The two nil values (genus_taxid and name) will be filled in when counts_by_tax_level is
+    # transformed to be indexed by taxid and count type.
+    [
```
tfrcarvalho: This format is hard to maintain and understand. If you had fields to pluck it is hard to keep track of the values here.
Have you consider adding a FIELDS_DEFAULTS to include defaults for plucked fields e.g.
```
FIELDS_TO_PLUCK = {
    genus_taxid => nil,
    count => 0,
    percent_identity => 0,
	...
]
```
Then you can build this dynamically using this hash the FIELDS_TO_PLUCK.
End goal is to make it more obvious when adding fields and make it more readable.

app/services/pipeline_report_service.rb
```diff
@@ -197,15 +248,27 @@ def compute_z_scores(taxa_counts, adjusted_total_reads)
...
       taxon_counts[:max_z_score] = nr_z_score.nil? || (nt_z_score && nt_z_score > nr_z_score) ? nt_z_score : nr_z_score
     end
   end
 
   def compute_aggregate_scores(species_counts, genus_counts)
     species_counts.each do |tax_id, species|
       genus = genus_counts[species[:genus_tax_id]]
-      species[:agg_score] = (species[:nt].present? ? genus[:nt][:z_score].abs * species[:nt][:z_score] * species[:nt][:rpm] : 0) \
-        + (species[:nr].present? ? genus[:nr][:z_score].abs * species[:nr][:z_score] * species[:nr][:rpm] : 0)
+      # Workaround placeholder for bad data (e.g. species counts present in TaxonSummary but genus counts aren't)
```
tfrcarvalho: Maybe add a TODO here to fix with a link to a JIRA ticket to fix the data issue?

spec/controllers/projects_controller_spec.rb
```diff
@@ -574,11 +574,11 @@
...
-            create(:project, users: [@user], samples_data: [{ pipeline_runs_data: [{ taxon_counts_data: [{ taxon_name: "klebsiella", nt: 0 }], job_status: "CHECKED" }] }])
-            expected_projects << create(:project, users: [@user], samples_data: [{ pipeline_runs_data: [{ taxon_counts_data: [{ taxon_name: "klebsiella", nt: 10 }], job_status: "CHECKED" }] }])
+            create(:project, users: [@user], samples_data: [{ pipeline_runs_data: [{ taxon_counts_data: [{ taxon_name: "klebsormidium", nt: 10, tax_id: 1 }], job_status: "CHECKED" }] }])
+            create(:project, users: [@user], samples_data: [{ pipeline_runs_data: [{ taxon_counts_data: [{ taxon_name: "klebsiella", nt: 0, tax_id: 2 }], job_status: "CHECKED" }] }])
+            expected_projects << create(:project, users: [@user], samples_data: [{ pipeline_runs_data: [{ taxon_counts_data: [{ taxon_name: "klebsiella", nt: 10, tax_id: 2 }], job_status: "CHECKED" }] }])
 
-            get :index, params: { format: "json", domain: domain, taxon: TaxonLineage.find_by(tax_name: "klebsiella").id }
+            get :index, params: { format: "json", domain: domain, taxon: TaxonLineage.find_by(tax_name: "klebsiella").taxid }
```
tfrcarvalho: Can't you just use `2` here? Saves an extra DB call.
Probably - since its used multiple times - the best is to define a constant like KLEBSIELLA_TAX_ID = 2 and reuse that.

spec/services/pipeline_report_service_spec.rb
```diff
@@ -0,0 +1,313 @@
+require 'rails_helper'
+require 'json'
+require 'pp'
```
tfrcarvalho: Not need anymore?

app/services/pipeline_report_service.rb
```diff
@@ -146,46 +160,32 @@ def fetch_taxon_counts(_pipeline_run_id, _background_id)
...
-      nil,
-    ]
+    FIELDS_INDEX.each do |field, index|
+      taxon[index] = FIELDS_DEFAULTS[field] unless taxon[index]
+    end
+    return taxon
   end
 
   def fetch_taxons_absent_from_sample(_pipeline_run_id, _background_id)
+    tax_ids = TaxonCount.select(:tax_id).where(pipeline_run_id: @pipeline_run_id).distinct.pluck(:tax_id)
```
tfrcarvalho: Just remove the pluck here.
You should not pluck here. You can inject he query directly and then rails is smart enough to not do 2 queries.

app/services/pipeline_report_service.rb
```diff
@@ -258,6 +258,8 @@ def compute_aggregate_scores(species_counts, genus_counts)
     species_counts.each do |tax_id, species|
       genus = genus_counts[species[:genus_tax_id]]
       # Workaround placeholder for bad data (e.g. species counts present in TaxonSummary but genus counts aren't)
+      # TODO: investigate why a count type appearing in species is missing in its genus. It's possible the data is
+      # only missing in local/staging and is present on prod.
```
tfrcarvalho: Please remove the last sentence... It is also missing in prod.
Did you add a Jira ticket?

spec/controllers/projects_controller_spec.rb
```diff
@@ -574,11 +574,12 @@
             expected_projects = []
             create(:project, users: [@user])
             create(:project, :with_sample, users: [@user])
+            klebsiella_tax_id = 2
```
tfrcarvalho: Constants should be all caps.

PR #2748, opened 2019-11-14
Fix issue where phylo tree modal was closing when mouse left browser.

MarkAZhang:
# Description

PhyloTreeCreationModal was originally inside `<HoverActions>` so it would unmount whenever hoveractions unmounted (which happened on onMouseOut, which would trigger when the user moved their mouse out of the browser).

Moved the PhyloTreeCreationModal higher up in the component tree so this doesn't happen.

# Tests

* Verified that the bug no longer occurs.
* Verified that modal still behaves correctly.


MarkAZhang:
Thoughts on cherry-picking this to staging? Some of our users had noticed the issue.

cdebourcy:
> Thoughts on cherry-picking this to staging? Some of our users had noticed the issue.

Sounds like a good idea

PR #2751, opened 2019-11-15
[WIP] Switch nt_info to Berkeley DB

morsecodist:
**Do not merge** until I update the file in s3 and the index generation instructions.

# Description

This updates `nt_info.sqlite3` to the `.db` extension for use with the Berkeley DB format. I had to update it in a few places. It seems that staging does not have this field set in the database. This config seems to not have been updated by ActiveRecord so staging and production are not in sync. I created a migration that will update both staging and production to the appropriate value. 

There is also a hard-coded default that is used currently for staging. Since we are deprecating sqlite3 support for key-value stores I changed this default as well.

I also updated the create alignment config rake task.

# Tests

To test this I will update the files in s3 and the index generation instructions then run a sample with this configuration set on a machine. Then we can test a sample in staging after deployment.

### DIFF
--- a/db/migrate/20191115200413_update_nt_info_to_berkeley_db.rb
+++ b/db/migrate/20191115200413_update_nt_info_to_berkeley_db.rb
@@ -1,0 +1,7 @@
+class UpdateNtInfoToBerkeleyDb < ActiveRecord::Migration[5.1]
+  def change
+    AlignmentConfig.where(name: "2019-09-17").update_all(
+      s3_nt_info_db_path: "s3://idseq-database/alignment_data/2019-09-17/nt_info.db",
+    )
+  end
+end
--- a/app/models/pipeline_run_stage.rb
+++ b/app/models/pipeline_run_stage.rb
@@ -25,7 +25,7 @@ class PipelineRunStage < ApplicationRecord
   DAG_NAME_EXPERIMENTAL = "experimental".freeze
 
   # Older alignment configs might not have an s3_nt_info_db_path field, so use a reasonable default in this case.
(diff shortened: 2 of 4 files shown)


db/migrate/20191115200413_update_nt_info_to_berkeley_db.rb
```diff
@@ -0,0 +1,7 @@
+class UpdateNtInfoToBerkeleyDb < ActiveRecord::Migration[5.1]
+  def change
+    AlignmentConfig.where(name: "2019-09-17").update_all(
+      s3_nt_info_db_path: "s3://idseq-database/alignment_data/2019-09-17/nt_info.db",
```
cdebourcy: Conceptually, isn't that a new alignment config? I worry that by updating it (rather than creating a new one), we could lose documentation of how a sample was run. Although the reference files used are also reflected in the DAG JSON that gets produced, so maybe we just need to be aware that the DAG is the source of truth for how a sample was run, not the alignment config the pipeline_run is pointed to.

morsecodist: These configs should both be identical key value stores that are stored in a different format so the internal data should be the same. I know it has a different path so there may be some confusion, but I just wanted to make this clear.

We want to eliminate the sqlite3 dependency at runtime which means we no longer want to keep supporting this as per @boris-dimitrov 's cleanup initiatives. I also agree this is a good thing. The level of complexity we have introduced to do something conceptually simple is quite high. This will mean running with these older configs will break after this switch happens which is why I want to do this migration.

Also the dag JSON still will reference the file extension used so no information is lost.

jshoe:
@morsecodist is this one still in progress / should still be switched?

PR #2753, opened 2019-11-16
Fix issue where pipeline tab doesn't respect pipeline version.

MarkAZhang:
# Description

Fixes some issue where the pipeline tab doesn't respond to pipeline version.

The red boxes were regions that previously didn't respond to pipeline version.
Screen Shot 2019-11-15 at 4 12 09 PM

Also on the result folder and raw result folder pages:
Screen Shot 2019-11-15 at 4 13 27 PM

Screen Shot 2019-11-15 at 4 13 49 PM

# Tests
* Verified that all affected pages/endpoints now respect pipeline version, and the pipeline_version is propagated properly into various URLs.
* Wrote controller rspec test.

### DIFF
--- a/spec/controllers/samples_controller_spec.rb
+++ b/spec/controllers/samples_controller_spec.rb
@@ -12,15 +12,45 @@ RSpec.describe SamplesController, type: :controller do
     describe "GET raw_results_folder (nonadmin)" do
       it "can see raw results on the user's own samples" do
         project = create(:project, users: [@joe])
-        sample = create(:sample, project: project)
+        sample = create(:sample, project: project, user: @joe)
         get :raw_results_folder, params: { id: sample.id }
-        expect(response).to have_http_status :unauthorized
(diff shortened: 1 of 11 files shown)


spec/controllers/samples_controller_spec.rb
```diff
@@ -11,16 +11,46 @@
...
       end
+    end
 
-      it "cannot see raw results on another user's sample" do
+    describe "GET results_folder (nonadmin)" do
+      it "can see raw results on the user's own samples" do
         project = create(:project, users: [@joe])
         sample = create(:sample, project: project, user: @joe)
-        get :raw_results_folder, params: { id: sample.id }
+        get :results_folder, params: { id: sample.id }
```
cdebourcy: It seems like there is a mismatch here between "can see raw results" and the endpoint that's actually tested (results_folder, not raw_results_folder)

PR #2758, opened 2019-11-19
Fix issue with pipeline_version url param.

MarkAZhang:
Also update query-string package.

# Description

This fixes an issue where if you are looking at a report with pipeline_version=3.10, on refresh the pipeline_version becomes 3.1, because the value was being parsed as an integer.

To do this, we had to update the query-string package because the "parseNumbers" option wasn't added until a later version.

# Notes

Also updated the UrlQueryParser to not parse booleans and numbers by default. In the previous implementation, numbers and booleans were automatically parsed.

# Tests

* Verified that sample report url params still work.
* Verified that data discovery url params still work.

### DIFF
--- a/app/assets/src/helpers/url.js
+++ b/app/assets/src/helpers/url.js
@@ -26,10 +26,16 @@ export const parseUrlParams = () => {
     arrayFormat: "bracket",
   });
   for (var key in urlParams) {
-    try {
-      urlParams[key] = JSON.parse(urlParams[key]);
-    } catch (e) {
-      // pass
+    // Don't parse pipeline_version even though it looks like a number.
+    // 3.10 is different from 3.1
+    // TODO(mark): Switch to using UrlQueryParser instead.
+    if (key !== "pipeline_version") {
+      try {
+        // This parses booleans and numbers.
+        urlParams[key] = JSON.parse(urlParams[key]);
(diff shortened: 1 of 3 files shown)


MarkAZhang:
Held off on refactoring `helpers/url` to use `UrlQueryParser` (which would require refactoring the fillUrlParam code in PipelineSampleReport) because presumably the report page refactor would remove that code. If this isn't the case, happy to do the refactor.

app/assets/src/components/utils/UrlQueryParser.js
```diff
@@ -10,7 +10,10 @@ class UrlQueryParser {
   }
 
   parse(query) {
-    let params = QueryString.parse(query);
+    let params = QueryString.parse(query, {
+      parseNumbers: false,
```
tfrcarvalho: According to the github repo, these values are actually the default (https://github.com/sindresorhus/query-string#parsenumbers)... why did you need this?

tfrcarvalho: Also, UrlQueryParser is not used below, right?

MarkAZhang: Thanks for the catch regarding the defaults. Don't need to make a change in this file then.

PR #2759, opened 2019-11-19
Change default number of taxa in heatmap

gregdingle:
# Description

Following team discussion, we decided to lower this value. 

image

# Tests

Open a new heatmap, see slider at value 10 instead of 30


app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -40,7 +40,7 @@ const SORT_TAXA_OPTIONS = [
   { text: "Cluster", value: "cluster" },
 ];
 const TAXONS_PER_SAMPLE_RANGE = {
-  min: 0,
+  min: 1,
```
gregdingle: I also changed this but I don't see how zero can ever make sense.

PR #2760, opened 2019-11-20
Add reads non-host bulk download for single taxa.

MarkAZhang:
# Description

When the user selects a single taxa, the file format field changes.

Screen Shot 2019-11-20 at 11 00 40 AM

On the review screen, the file format is omitted.

Screen Shot 2019-11-20 at 11 00 48 AM

Compare this to the All Taxon download, where the file format is shown.

Screen Shot 2019-11-20 at 11 43 25 AM

Screen Shot 2019-11-20 at 11 00 29 AM

# Notes

* Also fixed a bug where when you re-opened the taxa select, it would show an empty search filter string, but the results would still be filtered. We now save the search filter string.

# Tests

* Verified that existing single taxa fasta download still works on sample report page, and produces the same data as before.
* Verified that reads non-host bulk download produces identical data to sample report page download.
* Wrote model rspec tests. (and refactored tests to reduce redundancy)

### DIFF
--- a/app/assets/src/components/views/bulk_download/ChooseStep.jsx
+++ b/app/assets/src/components/views/bulk_download/ChooseStep.jsx
@@ -9,6 +9,8 @@ import {
   map,
   isUndefined,
   orderBy,
+  isNumber,
+  reject,
 } from "lodash/fp";
 import cx from "classnames";
 import memoize from "memoize-one";
@@ -27,6 +29,8 @@ import PrimaryButton from "~/components/ui/controls/buttons/PrimaryButton";
 import cs from "./choose_step.scss";
 
 const AUTOCOMPLETE_DEBOUNCE_DELAY = 200;
+// Reads non-host download type has some special cases.
+const READS_NON_HOST_DOWNLOAD_TYPE = "reads_non_host";
 
 class ChooseStep extends React.Component {
   state = {
@@ -67,34 +71,55 @@ class ChooseStep extends React.Component {
     });
   }
 
-  isDownloadValid = () => {
-    const {
-      selectedDownloadTypeName,
-      downloadTypes,
-      selectedFields,
-    } = this.props;
+  getSelectedDownloadType = () => {
+    const { selectedDownloadTypeName, downloadTypes } = this.props;
 
     if (!selectedDownloadTypeName) {
-      return false;
+      return null;
     }
 
-    const downloadType = find(
-      ["type", selectedDownloadTypeName],
-      downloadTypes
-    );
+    return find(["type", selectedDownloadTypeName], downloadTypes);
+  };
+
+  // Get all the fields we need to validate for the selected download type.
+  getRequiredFieldsForSelectedType = () => {
+    const { selectedFields } = this.props;
+    const downloadType = this.getSelectedDownloadType();
+
+    if (!downloadType) return null;
+
+    let requiredFields = downloadType.fields;
+
+    // Don't require file_format field if single taxa selected for reads non-host download type.
+    if (
+      downloadType.type === READS_NON_HOST_DOWNLOAD_TYPE &&
+      isNumber(get([downloadType.type, "taxa_with_reads"], selectedFields))
+    ) {
+      requiredFields = reject(["type", "file_format"], requiredFields);
+    }
+
+    return requiredFields;
+  };
+
+  isDownloadValid = () => {
+    const { selectedFields } = this.props;
+
+    const downloadType = this.getSelectedDownloadType();
 
     if (!downloadType) {
       return false;
     }
 
-    if (downloadType.fields) {
+    const requiredFields = this.getRequiredFieldsForSelectedType();
+
+    if (requiredFields) {
       if (
         some(
           Boolean,
           map(
             field =>
               isUndefined(get([downloadType.type, field.type], selectedFields)),
-            downloadType.fields
+            requiredFields
           )
         )
       ) {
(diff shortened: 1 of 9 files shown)


app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -178,6 +203,20 @@ class ChooseStep extends React.Component {
     // Set different props for the dropdown depending on the field type.
     switch (field.type) {
       case "file_format":
+        if (
+          downloadType.type === READS_NON_HOST_DOWNLOAD_TYPE &&
```
gregdingle: extract this pattern to a methods such as `isTaxonWithReads`

app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -237,9 +276,23 @@ class ChooseStep extends React.Component {
...
           options={dropdownOptions}
-          onChange={(value, displayName) =>
-            onFieldSelect(downloadType.type, field.type, value, displayName)
-          }
+          onChange={(value, displayName) => {
+            onFieldSelect(downloadType.type, field.type, value, displayName);
+
+            // If the user has selected a single taxa, reset the file format field.
+            if (
+              downloadType.type === READS_NON_HOST_DOWNLOAD_TYPE &&
```
gregdingle: ideally extract this with above

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -140,6 +140,18 @@ def taxon_name(taxid, tax_level)
     taxon["#{tax_level}_name"] if taxon
   end
 
+  def get_taxid_fasta_from_pipeline_run_combined_nt_nr(pipeline_run, taxid, tax_level)
```
gregdingle: `get_taxid_fasta` sounds off to me... isn't the fasta a property of the taxon, not the taxid?

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -140,6 +140,18 @@ def taxon_name(taxid, tax_level)
...
+    nt_array = get_taxid_fasta_from_pipeline_run(pipeline_run, taxid, tax_level, 'NT').split(">")
+    nr_array = get_taxid_fasta_from_pipeline_run(pipeline_run, taxid, tax_level, 'NR').split(">")
+    combined_array = ((nt_array | nr_array) - [''])
```
gregdingle: so are users expecting that we combine nt and nr fastas? do we document this anywhere? it seems like a detail I would want to know as a user

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -140,6 +140,18 @@ def taxon_name(taxid, tax_level)
...
+
+    if combined_array.empty?
+      return nil
```
gregdingle: why would a taxon we know to have hits in a sample not have a fasta? please add a comment in the code

app/models/bulk_download.rb
```diff
@@ -158,16 +159,24 @@ def kickoff_ecs_task(
     end
   end
 
-  def get_param_value(key)
+  def get_param_field(key, field)
     # If params is nil, try to set it from params_json
     if params.nil? && params_json.present?
       self.params = JSON.parse(params_json)
```
gregdingle: what's the point of `params`? is it only to avoid the cost of parsing the JSON on each get?

app/models/bulk_download.rb
```diff
@@ -293,10 +302,31 @@ def write_output_files_to_s3_tar_writer(s3_tar_writer)
...
+                                    # We avoid pre-fetching "taxon_byteranges" here, even though it has to be fetched once per pipeline_run in the code below,
+                                    # because of the potentially large number of taxon_byteranges per pipeline_run.
+                                    pipeline_runs.includes(:sample)
                                   else
                                     pipeline_runs
                                   end
 
+    # Get the tax_level for the selected taxid by looking at the first available TaxonCount entry.
+    # Since TaxonCount isn't normalized, any TaxonCount entry would provide the same information.
+    if download_type == BulkDownloadTypesHelper::READS_NON_HOST_BULK_DOWNLOAD_TYPE
```
gregdingle: for readability, I'd love to see all these boolean expressions converted to instance methods such as 

`isNonHostReadsType`

app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -178,6 +203,20 @@ class ChooseStep extends React.Component {
...
+        if (
+          downloadType.type === READS_NON_HOST_DOWNLOAD_TYPE &&
+          isNumber(get([downloadType.type, "taxa_with_reads"], selectedFields))
+        ) {
+          return (
+            <div className={cs.field} key={field.type}>
+              <div className={cs.label}>{field.display_name}:</div>
+              <div className={cs.forcedOption}>.fasta</div>
+              <div className={cs.info}>
+                Note: Only .fasta is available when selecting one taxa.
```
gregdingle: my latin is rusty, but shouldn't `one taxa` be `one taxon`?

app/models/bulk_download.rb
MarkAZhang: Yeah that's right.

app/helpers/pipeline_outputs_helper.rb
MarkAZhang: Made a note to add in the docs. Thanks! In the future, we may add an option to let them pick nt or nr.

app/helpers/pipeline_outputs_helper.rb
MarkAZhang: Makes sense. Thanks.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
MarkAZhang: `isReadsNonHostDownloadType(downloadType)` barely feels shorter. This particular case doesn't feel worth it to me.

app/helpers/pipeline_outputs_helper.rb
MarkAZhang: Added comments to make it more clear. We don't actually have that guarantee about the taxon for this method (it IS guaranteed for how the bulk download uses the function, but it's possible that someone calls this method later with taxa that might not have reads).

app/models/bulk_download.rb
MarkAZhang: I removed the `BulkDownloadTypesHelper::` by including that helper file in the model, which removes some of the wordiness. Instance methods for each download type feel a little overkill here.

PR #2761, opened 2019-11-20
IDSEQ-1785 - Implement background sliding session refresh

davidrissato:
# Description

Auth0 sliding sessions feature require you to fetch the authorization endpoint to refresh a bearer token and keep a user session alive.

We want to silently refresh bearer tokens in the background to keep the user session alive, specially during file uploads.

# Notes

There is an extra bit of complexity involved in this PR to prevent multiple tabs from refreshing the token at once.

# Tests

* Manual tests using both strategies.

### DIFF
--- a/app/assets/src/components/common/Header.jsx
+++ b/app/assets/src/components/common/Header.jsx
@@ -118,6 +118,7 @@ class Header extends React.Component {
             // Initialize the toast container - can be done anywhere (has absolute positioning)
           }
           <ToastContainer />
+          <iframe style={{ display: "none" }} src="/auth0/background_refresh" />
         </div>
       )
     );
--- a/app/assets/src/components/common/Header.jsx
+++ b/app/assets/src/components/common/Header.jsx
@@ -118,7 +118,10 @@ class Header extends React.Component {
             // Initialize the toast container - can be done anywhere (has absolute positioning)
           }
           <ToastContainer />
-          <iframe style={{ display: "none" }} src="/auth0/background_refresh" />
+          <iframe
+            className={cs.backgroundRefreshFrame}
+            src="/auth0/background_refresh"
+          />
         </div>
       )
     );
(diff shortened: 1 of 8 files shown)


app/assets/src/components/common/Header.jsx
```diff
@@ -118,6 +118,7 @@ class Header extends React.Component {
             // Initialize the toast container - can be done anywhere (has absolute positioning)
           }
           <ToastContainer />
+          <iframe style={{ display: "none" }} src="/auth0/background_refresh" />
```
jshoe: nit: ideally this style would be in `header.scss`

app/controllers/auth0_controller.rb
```diff
@@ -41,4 +72,39 @@ def request_password_reset
...
+
+  def filter_value(value, set_of_values)
+    value if set_of_values.include?(value)
+  end
+
+  def background_refresh_values
+    @auth0_token = auth0_decode_auth_token
+    if @auth0_token && @auth0_token[:auth_payload]
+      exp = @auth0_token[:auth_payload]["exp"]
+      iat = @auth0_token[:auth_payload]["iat"] || (exp - 15.minutes.in_seconds)
```
jshoe: could you add a comment in this block (e.g. the 15 minutes and the lifespan*0.2) ?

app/controllers/auth0_controller.rb
```diff
@@ -41,4 +72,39 @@ def request_password_reset
...
+      active: auth0_session.present?,
+      exp: exp || 0,
+      iat: iat || 0,
+      lifespan: lifespan,
+      expires_in: expires_in || 0,
+      should_refresh_in: should_refresh_in || 0,
+      should_refresh: should_refresh,
+      expired: expired,
+      refresh_endpoint: "/auth0/refresh_token?mode=background_refresh",
+      reload_wait_seconds: [lifespan / 2, (15.minutes / 1.second)].min,
```
jshoe: same for the "half of lifespan or 15 min" ?

app/controllers/auth0_controller.rb
davidrissato: That's a good observation. I'm adding comments and improving this logic.

PR #2762, opened 2019-11-21
IDSEQ-1784 - Login counters being incremented on every request when using auth0

davidrissato:
# Description

This PR makes devise store the authentication state in the user's session to prevent incrementing login counters on every request.

# Tests

* Manual tests using both strategies

### DIFF
--- a/app/controllers/auth0_controller.rb
+++ b/app/controllers/auth0_controller.rb
@@ -52,11 +52,6 @@ class Auth0Controller < ApplicationController
     end
   end
 
-  def remove_auth0_session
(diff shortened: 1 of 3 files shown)


app/controllers/auth0_controller.rb
```diff
@@ -52,11 +52,6 @@ def callback
     end
   end
 
-  def remove_auth0_session
```
jshoe: no longer needed?

davidrissato: No, it wasn't being invoked at all.

PR #2763, opened 2019-11-21
make admin projects private forever by default

cdebourcy:
# Description
If an admin creates a project, samples in that project will by default "never" go public to avoid clutter.


app/controllers/projects_controller.rb
```diff
@@ -42,6 +42,7 @@ class ProjectsController < ApplicationController
   respond_to :json
 
   MAX_BINS = 34
+  FAR_FUTURE_DAYS = 100_000
```
tfrcarvalho: Ideally we would make this explicitly private rather than this fake future date.
The general problem with these approaches is if an attacker is able to compromise system time, for instance. That would allow them to access private data.

cdebourcy: Makes sense, thanks for the review. Do you have a suggestion for implementing that in this case?

tfrcarvalho: Actually, nm. You are not introducing any new vulnerabilities here since this affects all projects. So good to go. :)

PR #2764, opened 2019-11-21
IDSEQ-1795 - Synchronize auth0 cookies in the homepage

davidrissato:
# Description

Forces an auth0 user logout in the home page when users click login button

# Tests

* Manual tests using both strategies

### DIFF
--- a/app/assets/src/components/views/Landing.jsx
+++ b/app/assets/src/components/views/Landing.jsx
@@ -19,7 +19,6 @@ import DiscoverIcon from "../ui/icons/DiscoverIcon";
 import DetectIcon from "../ui/icons/DetectIcon";
 import DecipherIcon from "../ui/icons/DecipherIcon";
 import LogoIcon from "../ui/icons/LogoIcon";
-import { postToUrlWithCSRF } from "~utils/links";
 
 class Landing extends React.Component {
   constructor(props) {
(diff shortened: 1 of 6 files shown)


app/controllers/auth0_controller.rb
```diff
@@ -20,6 +20,10 @@ def refresh_token
     render :refresh_token, layout: false
   end
 
+  def login
+    redirect_to auth0_signout_url(url_for(action: :refresh_token, only_path: false, params: { mode: "login" }))
```
jshoe: nit: line 24 is a little hard to read. Might be cleaner on two lines?

app/assets/src/components/views/Landing.jsx
```diff
@@ -78,7 +78,7 @@ class Landing extends React.Component {
   render() {
     const signInLink = () => {
       if (this.props.signInStrategy == "AUTH0") {
-        postToUrlWithCSRF("/auth/auth0");
+        location.href = "/auth0/login";
```
jshoe: nice this will be much easier to use

app/controllers/auth0_controller.rb
jshoe: Also maybe a comment on the mode 'login'? It seems unintuitive at first that the login action appears to go to `auth0_signout_url`

davidrissato: I just figured out a better approach to this problem. I don't need to invoke the sign_out URL anymore, I need to send a parameter `prompt=login` AND `connection=...` to the authorization endpoint.
It is worth pointing out that I had tried to send `prompt=login` before without any success, but it seems that `prompt=login` only works as expected if you send the connection name as well.
I'm making changes to this PR now.

PR #2765, opened 2019-11-21
Migrate cleanup functions to the new report service

j-x-han:
# Description

Migrate the [cleanup functions](https://github.com/chanzuckerberg/idseq-web/blob/master/app/lib/report_helper.rb#L507) from the old ReportHelper to use in the new report service.
1) `validate_names`  fills in names for species/genera that are either blacklisted or missing names for various reasons
2) `cleanup_missing_genus_counts` creates a fake genus if a species is lacking one

# Notes

It's unclear if `cleanup_missing_genus_counts` is necessary to migrate over; some more investigation will be necessary to determine if/when a species is missing a genus, and if that can be resolved on the database side.

Logging has been added to `validate_names` to see how frequently each case of a missing name occurs.

# Tests

* Verified that species/genus names are no longer empty in the report.

### DIFF
--- a/app/helpers/reports_helper.rb
+++ b/app/helpers/reports_helper.rb
@@ -33,15 +33,15 @@ module ReportsHelper
               parent_name = "taxon #{parent_tax_id}"
             end
             tax_info[:name] = "non-#{level_str}-specific reads in #{parent_level} #{parent_name}"
-            Rails.logger.debug("Pipeline run #{pipeline_run_id} contains non-#{level_str}-specific reads in #{parent_level} #{parent_name}")
+            Rails.logger.info("Pipeline run #{pipeline_run_id} contains non-#{level_str}-specific reads in #{parent_level} #{parent_name}")
 
           elsif tax_id == TaxonLineage::BLACKLIST_GENUS_ID
             tax_info[:name] = "all artificial constructs"
-            Rails.logger.debug("Blacklisted genus id appeared in report for pipeline run #{pipeline_run_id}.")
+            Rails.logger.info("Blacklisted genus id appeared in report for pipeline run #{pipeline_run_id}.")
 
           elsif !(TaxonLineage::MISSING_LINEAGE_ID.values.include? tax_id) && tax_id != TaxonLineage::MISSING_SPECIES_ID_ALT
             tax_info[:name] += " #{tax_id}"
-            Rails.logger.debug("Pipeline run #{pipeline_run_id} missing lineage for #{tax_id}.")
+            Rails.logger.info("Pipeline run #{pipeline_run_id} missing lineage for #{tax_id}.")
           end
 
         elsif !tax_info[:name]
--- a/app/helpers/reports_helper.rb
+++ b/app/helpers/reports_helper.rb
@@ -4,7 +4,7 @@ module ReportsHelper
   # the fake genus IDs start here:
   FAKE_GENUS_BASE = -1_900_000_000
 
-  def self.validate_names!(counts, lineage_by_tax_id)
+  def self.validate_names(counts, lineage_by_tax_id, pipeline_run_id)
     # This makes up suitable names for missing and blacklisted genera and species. Such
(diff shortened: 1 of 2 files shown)


app/helpers/reports_helper.rb
```diff
@@ -1,2 +1,109 @@
...
+        level_str = TaxonLineage.level_name(tax_level)
+
+        if tax_id < 0
+          # Usually -1 means accession number did not resolve to species.
+          # TODO: Can we keep the accession numbers to show in these cases?
+          # NOTE: important to be lowercase for sorting below uppercase valid genuses
+          tax_info[:name] = "all taxa with neither family nor genus classification"
+
+          if tax_id < TaxonLineage::INVALID_CALL_BASE_ID
+            parent_str = tax_level == TaxonCount::TAX_LEVEL_SPECIES ? genus_str : family_str
```
tfrcarvalho: This is a level right? If so, should call it parent_level.

app/services/pipeline_report_service.rb
```diff
@@ -103,14 +112,27 @@ def generate
     ]
 
     tax_ids = counts_by_tax_level[TaxonCount::TAX_LEVEL_GENUS].keys
+    # If a species has an undefined genus (id < 0), the TaxonLineage id is based off the
+    # species id rather than genus id, so select those species ids as well.
+    species_with_missing_genus = []
```
tfrcarvalho: Did you notice the impact of this?

tfrcarvalho: Let's add another todo here to clear this up once we clean up the data.

app/helpers/reports_helper.rb
```diff
@@ -1,2 +1,109 @@
...
+            parent_tax_id = tax_level == TaxonCount::TAX_LEVEL_SPECIES ? tax_info[genus_tax_id] : convert_neg_taxid(tax_id)
+            parent_name = fetch_parent_name(tax_level, tax_id, tax_info, lineage_by_tax_id)
+            # If no parent (genus/family) name was found, then use the parent id as the name.
+            unless parent_name
+              missing_parents[parent_tax_id] = tax_id
+              parent_name = "taxon #{parent_tax_id}"
+            end
+            tax_info[:name] = "non-#{level_str}-specific reads in #{parent_str} #{parent_name}"
+            MetricUtil.log_analytics_event("non-#{level_str}-specific reads in #{parent_str} #{parent_name}", nil)
```
tfrcarvalho: As I think about lets just move Metric to logs. Metrics need to follow a particular naming structure. 
Can we also complement all logs with pipeline_run_id? Just so we can find it later.

app/services/pipeline_report_service.rb
j-x-han: This definitely slows things down. I eyeballed the times for the `fetch_taxon_lineage` for a sample, and it goes from roughly 0.4 to 0.6.

app/helpers/reports_helper.rb
```diff
@@ -25,36 +24,35 @@ def self.validate_names!(counts, lineage_by_tax_id)
...
             parent_name = fetch_parent_name(tax_level, tax_id, tax_info, lineage_by_tax_id)
             # If no parent (genus/family) name was found, then use the parent id as the name.
             unless parent_name
               missing_parents[parent_tax_id] = tax_id
               parent_name = "taxon #{parent_tax_id}"
             end
-            tax_info[:name] = "non-#{level_str}-specific reads in #{parent_str} #{parent_name}"
-            MetricUtil.log_analytics_event("non-#{level_str}-specific reads in #{parent_str} #{parent_name}", nil)
+            tax_info[:name] = "non-#{level_str}-specific reads in #{parent_level} #{parent_name}"
+            Rails.logger.debug("Pipeline run #{pipeline_run_id} contains non-#{level_str}-specific reads in #{parent_level} #{parent_name}")
```
tfrcarvalho: Lets change this to `warn`.

PR #2766, opened 2019-11-21
Add test for report when taxon is missing from background model

j-x-han:
# Description

Add a test for the new report service checking that the correct z-scores are returned when a taxon has a count type missing from the background model. 

If the count type is present in the sample, the z-score should be 100

If the count type is also missing in the sample, then the report will not contain that count type data for the taxon, and the 0 values should be filled in on the frontend instead.

### DIFF
--- a/spec/services/pipeline_report_service_spec.rb
+++ b/spec/services/pipeline_report_service_spec.rb
@@ -309,4 +309,56 @@ RSpec.describe PipelineReportService, type: :service do
       expect(JSON.parse(@report)["counts"]["2"]["1301"]).to include_json(genus_result)
     end
   end
+
+  context "taxon missing from background" do
+    before do
+      ResqueSpec.reset!
+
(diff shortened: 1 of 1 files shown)


spec/services/pipeline_report_service_spec.rb
```diff
@@ -309,4 +309,56 @@
...
+        "nt" => {
+          "z_score" => 100,
+        },
+      }
+      expect(JSON.parse(@report)["counts"]["1"]["1"]).to include_json(expected)
+
+      # Since NR is missing from both the sample and the background model, it won't be returned
+      # in the report service. The frontend should fill in the NR row with default 0 values.
+      unexpected = { "nr" => {} }
+      expect(JSON.parse(@report)["counts"]["1"]["1"]).not_to include_json(unexpected)
```
tfrcarvalho: Would it work to use `.not_to include("nr")` here? Seems easier to understand.

PR #2767, opened 2019-11-21
Adds filters, search, links and sidebars to refactored report page

tfrcarvalho:
# Description

* Adds all the filters, search and other options to the report page.
* Integrates taxon and sample details sidebars
* Persists options

# Notes

* Featured still guarded by flag

# Tests

* Tested manually all filters and persistence
* More tests once the whole refactor of report page is complete


app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -229,7 +231,22 @@ class ReportTableV2 extends React.Component {
             )}
           </div>
           <div className={cx(rowData.taxLevel == "species" && cs.speciesName)}>
-            {cellData}
+            <span
+              className={cx(cs.taxonName, cellData || cs.missingName)}
```
j-x-han: Names display properly when the report is first loaded, but the styling for the names seems to be flipped after the type is toggled. The names will be grayed out when set to `Scientific` instead of `Common`.

**When the page is refreshed:**
image

**After toggling the name type:**
image

image

app/assets/src/components/views/SampleViewV2/ReportFilters.jsx
```diff
@@ -0,0 +1,175 @@
...
+    onFilterChanged({ key, value });
+  };
+
+  render = () => {
+    const { backgrounds, sampleId, selected, view } = this.props;
+
+    return (
+      <div className={cs.filterList}>
+        <div className={cs.filterListElement}>
+          <SearchBox
```
j-x-han: Searching doesn't seem to work locally for me (even on the old version of the report page). Made a note to self to test on staging!

app/assets/src/components/views/SampleViewV2/SampleViewHeader.jsx
```diff
@@ -0,0 +1,185 @@
...
+                pipelineRun
+              )}`,
+              event
+            )
+          }
+        >
+          <span className={cs.pipelineRunVersion}>
+            Pipeline {renderVersion()}
+          </span>
+          <PipelineVersionSelect
```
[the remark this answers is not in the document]
tfrcarvalho: Added to notes. Will address in future PR.

app/assets/src/components/views/SampleViewV2/ReportFilters.jsx
```diff
@@ -0,0 +1,175 @@
...
+                subcat => (result[subcat] = category.name),
+                category.children || []
+              );
+            }, {})(CATEGORIES)}
+            selectedCategories={getOr(
+              [],
+              ["categories", "categories"],
+              selected
+            )}
+            selectedSubcategories={flatten(
```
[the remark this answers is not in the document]
tfrcarvalho: Yes, and is a very weird case. We made it generic before, maybe we should revisit in the future.

app/assets/src/components/views/SampleView/SampleViewControls.jsx
```diff
@@ -126,7 +125,7 @@ class SampleViewControls extends React.Component {
           direction="left"
         />
       );
-    } else if (canEdit) {
+    } else if (editable) {
       return <PrimaryButton onClick={this.deleteSample} text="Delete Sample" />;
```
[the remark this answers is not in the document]
tfrcarvalho: Most of the controls are broken. Will address in future PR.

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
tfrcarvalho: Good catch.
Fixed.

app/assets/src/components/views/SampleViewV2/ReportFilters.jsx
tfrcarvalho: Do you have elastic search running?

PR #2768, opened 2019-11-22
Add filter tags for taxon, thresholds and categories

tfrcarvalho:
# Description

Add filter tags for taxon, thresholds and categories

# Tests

* Set filters to add tags
* Remove filters to remove tags
* Verify that report data gets refreshed according to results


app/assets/src/components/views/SampleViewV2/ReportFilters.jsx
```diff
@@ -0,0 +1,275 @@
...
+                  },
+                })
+              }
+            />
+          </div>
+          <div className={cs.filterListElement}>
+            <ThresholdFilterDropdown
+              options={{
+                targets: THRESHOLDS,
+                operators: [">=", "<="],
```
j-x-han: Might have missed this on the previous PR, but threshold filters seem to be acting pretty weirdly...

**Filtering for score >= 10 affects the upper bound for score instead of lower bound**
image
image

**Filtering for score >= 1000 doesn't actually seem to filter any taxons out**
image
image

**Filtering for score <= 1000 behaves as if filtering for >= 1000 instead**
image
image

app/assets/src/components/views/SampleViewV2/ReportFilters.jsx
```diff
@@ -0,0 +1,275 @@
...
+
+  renderFilterTag = ({ key, label, path, value, idx }) => {
+    label = label || value;
+    return (
+      <FilterTag
+        className={cs.filterTag}
+        key={`${label}_filter_tag_${idx}`}
+        text={label}
+        onClose={() =>
+          this.handleRemoveFilter({
```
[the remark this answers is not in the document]
tfrcarvalho: You are right. It become a little hard to reason about.
Split the 3 cases... they do not have that much in common anyway.

app/assets/src/components/views/SampleViewV2/ReportFilters.jsx
tfrcarvalho: Thanks for catching. They were reversed. Fixed.

PR #2769, opened 2019-11-22
Wrap text inside the name cell.

tfrcarvalho:
# Description

Wrap text inside the name cell.

# Tests

* Check text in cell in taxon column wraps on white spaces when too long.


app/assets/src/components/views/SampleViewV2/report_table.scss
```diff
@@ -57,28 +57,38 @@
...
+    }
 
-.genusRow {
-  background-color: $off-white;
+    .missingName {
+      color: $medium-grey;
+    }
+
+    .countInfo {
+      // word-wrap: break-word;
```
j-x-han: Is this meant to be commented out?

app/assets/src/components/views/SampleViewV2/report_table.scss
```diff
@@ -57,28 +57,38 @@
...
     flex: 0 0 20px;
   }
 
-  .missingName {
-    color: $medium-grey;
-  }
-}
+  .taxonContainer {
+    .taxonName {
+      text-transform: capitalize;
```
j-x-han: Add `word-wrap` to the taxonName as well? Currently only the `countInfo` text is being wrapped.

app/assets/src/components/views/SampleViewV2/report_table.scss
tfrcarvalho: No. I was going to remove it. But i guess we can leave for long words.

tfrcarvalho: Not commented, but removed. `white-space: normal` takes care of the wrapping.
Break-word is for very long word which does not happen in this case.

app/assets/src/components/views/SampleViewV2/report_table.scss
tfrcarvalho: I will add it for now. But wrapping the name of the taxon might be undesired because it will push the - soon to be added - action icons outside of the cell.

PR #2770, opened 2019-11-22
Make CSS use fingerprinting hash

gregdingle:
# Description

To invalidate the new cloudfront cache, we want to use rails fingerprinting of static assets on both CSS and JS (already working). 

# Notes

* see https://guides.rubyonrails.org/asset_pipeline.html#what-is-fingerprinting-and-why-should-i-care-questionmark
* I also fixed a double slash // issue with font assets 

# Tests

* try locally with `  config.assets.debug = true`

* see hash in CSS filename

* see site load normally

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1,2 +1,1 @@
-@import url('bundle.min.css');
-@import url('https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700,800');
+//= require bundle.min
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/application.css
```diff
@@ -1,2 +1 @@
-@import url('bundle.min.css');
-@import url('https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700,800');
```
jshoe: Is the Open Sans font still downloaded from somewhere else? (since it's not usually a preinstalled font)

gregdingle: yes it is! in app/assets/src/styles/themes/_typography.scss

PR #2772, opened 2019-11-22
Generate downloadable CSV from the new report service

j-x-han:
# Description

Add the ability to download the new report as a CSV.

# Notes

* This includes a bug fix for species not getting sorted within each genus.

* The generated CSV is always sorted by aggregate score. Left a `TODO` to pass in the `sort_by` parameter; it might take some work to extract from `ReportTable` since sorting is currently done entirely on the frontend.

* The CSV format is slightly different from that of the CSV generated by the old version of the report due to the structure having changed. Most notably, `genus_name`, `family_taxid`, `is_phage` have been removed, and `contig`+`contig_r` columns have been added.

  * **Old column names**
  tax_id, name, common_name, tax_level, species_taxid, genus_taxid, genus_name, family_taxid, category_name, is_phage, NT_r, NT_rpm, NT_zscore, NT_percentidentity, NT_alignmentlength, NT_neglogevalue, aggregatescore, NT_r_pct, NT_rpm_bg, NT_maxzscore, NR_r, NR_rpm, NR_zscore, NR_percentidentity, NR_alignmentlength, NR_neglogevalue, NR_r_pct, NR_rpm_bg, NR_maxzscore

  * **New column names**
  tax_id, tax_level, genus_tax_id, name, common_name, category, nr_count, nr_percent_identity, nr_alignment_length, nr_e_value, nr_bg_mean, nr_bg_stdev, nr_contigs, nr_contig_r, nr_rpm, nr_z_score, nt_count, nt_percent_identity, nt_alignment_length, nt_e_value, nt_bg_mean, nt_bg_stdev, nt_contigs, nt_contig_r, nt_rpm, nt_z_score, max_z_score, agg_score, species_tax_ids

# Tests

* Verified that the `Download Report Table (.csv)` button downloads the correct report.
* Verified that contig and contig_r values in the CSV match the results in the table based on the selected min contig size.
* Verified that the download option on the old report table still works as expected.

### DIFF
--- a/app/assets/src/components/views/SampleView/SampleViewControls.jsx
+++ b/app/assets/src/components/views/SampleView/SampleViewControls.jsx
@@ -14,7 +14,7 @@ import {
 
 class SampleViewControls extends React.Component {
   downloadCSV = () => {
-    const { backgroundId, pipelineRun, sample } = this.props;
+    const { backgroundId, pipelineRun, sample, reportVersion } = this.props;
 
     let resParams = {};
     const stringer = require("querystring");
@@ -27,7 +27,11 @@ class SampleViewControls extends React.Component {
     let v = pipelineRun && pipelineRun.pipeline_version;
     if (v) resParams["pipeline_version"] = v;
 
-    let res = `/samples/${sample.id}/report_csv`;
+    // Need to know what version of the report table we're looking at and request accordingly
+    let res =
+      reportVersion === 1
+        ? `/samples/${sample.id}/report_csv`
+        : `/samples/${sample.id}/report_csv_v2`;
     res += `?${stringer.stringify(resParams)}`;
     location.href = res;
   };
@@ -141,6 +145,7 @@ SampleViewControls.propTypes = {
   pipelineRun: PropTypes.PipelineRun,
   editable: PropTypes.bool,
   view: PropTypes.string,
+  reportVersion: PropTypes.number,
 };
 
 export default SampleViewControls;
--- a/app/assets/src/components/views/SampleView/SampleViewControls.jsx
+++ b/app/assets/src/components/views/SampleView/SampleViewControls.jsx
@@ -14,7 +14,13 @@ import {
 
 class SampleViewControls extends React.Component {
   downloadCSV = () => {
-    const { backgroundId, pipelineRun, sample, reportVersion } = this.props;
+    const {
+      backgroundId,
+      pipelineRun,
+      sample,
+      reportVersion,
+      minContigSize,
+    } = this.props;
 
     let resParams = {};
     const stringer = require("querystring");
@@ -26,6 +32,9 @@ class SampleViewControls extends React.Component {
     // Set the right pipeline version.
     let v = pipelineRun && pipelineRun.pipeline_version;
     if (v) resParams["pipeline_version"] = v;
+
+    // Set the min contig size.
+    if (minContigSize) resParams["min_contig_size"] = minContigSize;
 
     // Need to know what version of the report table we're looking at and request accordingly
     let res =
@@ -146,6 +155,7 @@ SampleViewControls.propTypes = {
   editable: PropTypes.bool,
   view: PropTypes.string,
   reportVersion: PropTypes.number,
+  minContigSize: PropTypes.number,
 };
 
 export default SampleViewControls;
--- a/app/assets/src/components/views/SampleView/SampleViewControls.jsx
+++ b/app/assets/src/components/views/SampleView/SampleViewControls.jsx
@@ -14,13 +14,7 @@ import {
 
 class SampleViewControls extends React.Component {
   downloadCSV = () => {
-    const {
-      backgroundId,
-      pipelineRun,
-      sample,
-      reportVersion,
-      minContigSize,
-    } = this.props;
+    const { backgroundId, pipelineRun, sample, minContigSize } = this.props;
 
     let resParams = {};
     const stringer = require("querystring");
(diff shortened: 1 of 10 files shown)


app/assets/src/components/views/SampleView/SampleViewControls.jsx
```diff
@@ -27,7 +33,14 @@ class SampleViewControls extends React.Component {
     let v = pipelineRun && pipelineRun.pipeline_version;
     if (v) resParams["pipeline_version"] = v;
 
-    let res = `/samples/${sample.id}/report_csv`;
+    // Set the min contig size.
+    if (minContigSize) resParams["min_contig_size"] = minContigSize;
+
+    // Need to know what version of the report table we're looking at and request accordingly
+    let res =
+      reportVersion === 1
```
tfrcarvalho: Instead of using this parameter just use the feature flag that was already injected in the context.
Example of use in: https://github.com/chanzuckerberg/idseq-web/blob/390ae2d2a5b41220c601ebf72584900ecc842d03/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx#L609
```
<UserContext.Consumer>
              {currentUser => 
	if (currentUser.allowedFeatures.includes(REPORT_PAGE_V2))
```
You need to add a new feature constant.

app/assets/src/components/views/SampleViewV2/SampleViewHeader.jsx
```diff
@@ -158,6 +159,8 @@ export default function SampleViewHeader({
           pipelineRun={pipelineRun}
           editable={editable}
           view={view}
+          reportVersion={2}
```
tfrcarvalho: AS described before, you will not need this.

app/helpers/reports_helper.rb
```diff
@@ -108,4 +108,56 @@ def self.fake_genus(tax_id, tax_info)
...
+  def report_csv_from_params_v2(sample, params)
+    background_id = params[:background_id] || sample.default_background_id
+    background_id = background_id.to_i
+    pipeline_run = select_pipeline_run(sample, params[:pipeline_version])
+    pipeline_run_id = pipeline_run ? pipeline_run.id : nil
+    return "" if pipeline_run_id.nil? || pipeline_run.total_reads.nil? || pipeline_run.adjusted_remaining_reads.nil?
+    # TODO: pass in sort_by to PipelineReportService from params
+    tax_details = JSON.parse(PipelineReportService.call(pipeline_run_id, background_id, true, params[:min_contig_size]))
+
+    rows = []
```
tfrcarvalho: Why not do this on the report service? 
Instead of just exporting JSON we can export CSV-ready hash.

app/helpers/reports_helper.rb
```diff
@@ -108,4 +108,56 @@ def self.fake_genus(tax_id, tax_info)
     tax_info[:genus_tax_id] = fake_genus_id
     fake_genus_info
   end
+
+  def report_csv_from_params_v2(sample, params)
+    background_id = params[:background_id] || sample.default_background_id
```
tfrcarvalho: Please use the same parameters and the same code as report_v2 for selecting the background and deciding to output something.
Since we will move the CSV formatting below to the pipeline report service, we also do not need an helper for this.

app/helpers/reports_helper.rb
```diff
@@ -108,4 +108,56 @@ def self.fake_genus(tax_id, tax_info)
...
+    background_id = background_id.to_i
+    pipeline_run = select_pipeline_run(sample, params[:pipeline_version])
+    pipeline_run_id = pipeline_run ? pipeline_run.id : nil
+    return "" if pipeline_run_id.nil? || pipeline_run.total_reads.nil? || pipeline_run.adjusted_remaining_reads.nil?
```
tfrcarvalho: We do not do this verification on report_v2 - should we? IN any case we should probably return an error in this case.

app/assets/src/components/views/SampleView/SampleViewControls.jsx
j-x-han: Removed the parameter, changed to redirecting on the server-side.

app/helpers/reports_helper.rb
j-x-han: Added the check and error logs to the beginning of report generation.

j-x-han:
Fixed the bug where species were not getting sorted within each genus.

app/helpers/reports_helper.rb
```diff
@@ -108,4 +108,15 @@ def self.fake_genus(tax_id, tax_info)
     tax_info[:genus_tax_id] = fake_genus_id
     fake_genus_info
   end
+
+  def flat_hash(h, f = [], g = {})
+    return g.update(f => h) unless h.is_a? Hash
+    h.each { |k, r| flat_hash(r, f + [k], g) }
+    g
+    # example: turn    { :a => { :b => { :c => 1,
```
MarkAZhang: Thanks for the comment!! Could we move it above the function?

app/services/pipeline_report_service.rb
```diff
@@ -274,6 +292,32 @@ def merge_contigs(contigs, counts_by_tax_level)
     end
   end
 
+  def merge_contigs_csv(contigs, counts_by_tax_level)
```
MarkAZhang: Instead of adding a separate function that duplicates a lot from the original `merge_csv` function, would it be possible to just pass the boolean `csv` to the `merge_csv` function and do some `if csv` conditionals? This would avoid duplicating the code.

app/helpers/reports_helper.rb
```diff
@@ -108,4 +108,15 @@ def self.fake_genus(tax_id, tax_info)
     tax_info[:genus_tax_id] = fake_genus_id
     fake_genus_info
   end
+
+  def flat_hash(h, f = [], g = {})
```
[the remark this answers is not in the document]
j-x-han: Done! Also changed the old `generate_report_csv` in `report_helper.rb` to call it from this new util file rather than as a private method.

PR #2773, opened 2019-11-23
Add CORS accept header for Cloudfront

gregdingle:
# Description

After considerable debugging, @davidrissato and I determined that Cloudfront was asking for font files with a different origin, so our rails server blocked the requests by default CORS policy. This PR adds a whitelist through the gem `rack-cors`. 

image

# Tests

* open localhost
* see accept origin header in responses of current domain

### DIFF
--- a/config/initializers/rack_cors.rb
+++ b/config/initializers/rack_cors.rb
@@ -7,9 +7,10 @@ if defined? Rack::Cors
       origins [
         # SERVER_DOMAIN should be set to the current env root web address
         # such as https://staging.idseq.net/
-        ENV["SERVER_DOMAIN"] || "",
+        ENV["SERVER_DOMAIN"],
+        *Rails.application.config.allowed_cors_origins
         # All other domains should be added in env config files such as prod.rb.
-      ] + (Rails.application.config.allowed_cors_origins || [])
+      ].compact
       resource '/assets/*'
     end
   end
--- a/config/initializers/rack_cors.rb
+++ b/config/initializers/rack_cors.rb
@@ -1,0 +1,25 @@
+## Configure Rack CORS Middleware, so that CloudFront can serve our assets.
+## See https://github.com/cyu/rack-cors
+## And see https://stackoverflow.com/a/36585871
+if defined? Rack::Cors
+  Rails.configuration.middleware.insert_before 0, Rack::Cors do
+    allow do
+      origins [
+        # SERVER_DOMAIN should be set to the current env root web address
+        # such as https://staging.idseq.net/
+        ENV["SERVER_DOMAIN"] || "",
+        # Add all expected origins to be safe
+        "http://localhost:3000",
+        "http://127.0.0.1:3000",
+        "https://idseq.net",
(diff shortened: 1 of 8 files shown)


config/initializers/rack_cors.rb
```diff
@@ -0,0 +1,25 @@
...
+        # Add all expected origins to be safe
+        "http://localhost:3000",
+        "http://127.0.0.1:3000",
+        "https://idseq.net",
+        "https://www.idseq.net",
+        "https://staging.idseq.net",
+        "https://www.staging.idseq.net",
+        # Add CloudFront domains
+        "https://assets.idseq.net",
+        "https://assets.staging.idseq.net",
```
MarkAZhang: A little surprised that you need these `assets` domains, since the user's browser is never at these domains (I think) even though the asset files might be served from there. Are you sure you need them?

MarkAZhang:
Awesome! Excited this is happening.

config/initializers/rack_cors.rb
```diff
@@ -0,0 +1,25 @@
+## Configure Rack CORS Middleware, so that CloudFront can serve our assets.
+## See https://github.com/cyu/rack-cors
+## And see https://stackoverflow.com/a/36585871
+if defined? Rack::Cors
+  Rails.configuration.middleware.insert_before 0, Rack::Cors do
+    allow do
+      origins [
+        # SERVER_DOMAIN should be set to the current env root web address
+        # such as https://staging.idseq.net/
+        ENV["SERVER_DOMAIN"] || "",
```
davidrissato: These configurations will also be applied to the API itself, so I'm wondering if we should configure these things based on the environment.

You could do something like:

```diff
# config/environments/development.rb
+ config.allowed_cors_origins = ["http://localhost:3000", "http://127.0.0.1:3000/"]

# config/environments/staging.rb
+ config.allowed_cors_origins = ["https://staging.idseq.net", "https://www.staging.idseq.net", "https://assets.staging.idseq.net"]
```

And here in `rack_cors.rb` you could use:
```ruby
Rails.configuration.middleware.insert_before 0, Rack::Cors do
    allow { Rails.application.config.allowed_cors_origins }
end
```

config/initializers/rack_cors.rb
gregdingle: Yes, we were surprised too. The browser sets the origin to `assets.idseq.net` when fetching resources referenced in `url()` statements in CSS files hosted at `assets.idseq.net`, following the spec.

gregdingle: All other assets had origin of `idseq.net`.

config/initializers/rack_cors.rb
```diff
@@ -0,0 +1,16 @@
...
```
davidrissato: Two suggestions:
1) instead of assigning an empty string here, just leave this entry return `nil` and then compact the array
2) You could use the spread operator here

Ex:
```ruby
[
  ENV["SERVER_DOMAIN"], 
  *Rails.application.config.allowed_cors_origins
].compact
```

PR #2774, opened 2019-11-25
[Report Page] Report filter and truncation stats row + clear all filters

tfrcarvalho:
# Description

* Add stats row - below filters - with information about the filtered rows and truncated input.
* Add "Clear All Filters" link to the same row.

# Tests

* Verify the numbers reported on the stats rows
* Verify that the link clear all filters removes all filters

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
+++ b/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
@@ -1,5 +1,6 @@
 import React from "react";
 import {
+  compact,
   entries,
   every,
   find,
@@ -12,6 +13,7 @@ import {
   pull,
   set,
   some,
+  sum,
   values,
 } from "lodash/fp";
 import deepEqual from "fast-deep-equal";
@@ -53,6 +55,7 @@ export default class SampleViewV2 extends React.Component {
         project: null,
         projectSamples: [],
         reportData: [],
+        reportMetadata: {},
         sample: null,
         sidebarMode: null,
         sidebarVisible: false,
@@ -164,9 +167,10 @@ export default class SampleViewV2 extends React.Component {
 
     this.setState({
       reportData,
+      reportMetadata: rawReportData.metadata,
       filteredReportData,
       selectedOptions: Object.assign({}, selectedOptions, {
-        background: rawReportData.backgroundId,
+        background: rawReportData.metadata.backgroundId,
       }),
     });
   };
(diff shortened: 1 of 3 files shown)


app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -572,6 +576,93 @@ export default class SampleViewV2 extends React.Component {
...
+      filtered += genusRow.filteredSpecies.length;
+    });
+    return { total, filtered };
+  };
+
+  filteredMessage = () => {
+    const { total, filtered } = this.countReportRows();
+    return (
+      filtered != total &&
+      `${filtered} rows passing the above filters, out of ${total} total rows.`
```
j-x-han: The filtered rows message persists after clicking `Clear All Filters` and shows that not all rows are passing, despite no filters being selected.

image

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -572,6 +576,93 @@ export default class SampleViewV2 extends React.Component {
...
+      total += genusRow.species.length;
+      filtered += genusRow.filteredSpecies.length;
+    });
+    return { total, filtered };
+  };
+
+  filteredMessage = () => {
+    const { total, filtered } = this.countReportRows();
+    return (
+      filtered != total &&
```
j-x-han: The current version of the report displays the filtered rows message even if no filters are selected. Do we want to carry this over so a user can immediately see how many rows are in the entire table?
image

app/assets/src/components/views/SampleViewV2/sample_view_v2.scss
```diff
@@ -26,6 +28,25 @@
   .reportTable {
     flex: 1 0 auto;
   }
+
+  .statsRow {
+    @include font-body-xxs;
+
```
MarkAZhang: Nit: Remove empty line.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
MarkAZhang: Perhaps we can just say `${total} total rows` if there are no filters and leave off the filter-related text.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -572,6 +576,93 @@ export default class SampleViewV2 extends React.Component {
...
+    );
+  };
+
+  truncatedMessage = () => {
+    const {
+      reportMetadata: { truncatedReadsCount },
+    } = this.state;
+    return (
+      truncatedReadsCount &&
+      `Overly large input was truncated to ${truncatedReadsCount} reads.`
```
MarkAZhang: Should we say "truncated to ${} reads prior to host and quality filters"? Otherwise, I could see it being confusing, especially in combination with the `subsamplingMessage`.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: Done

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: I added "Initial input" to explicitly say its the first. Referring to the filters seems a bit repetitive.
We can bring it up with other and adjust later.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: Fixed

PR #2776, opened 2019-11-25
[Report Page] Add expand all icon to expand all genus rows on the header

tfrcarvalho:
# Description

* Add expand all icon to expand all genus rows on the header
* Icon should also reflect the state when all the rows are expanded or collapsed even if they were expanded/collapsed individually.

# Tests

* Verify that expanding and collapsing from the header expands and collapses all rows
* Apply a few filters, expand all rows individually - verify that the header's expand/collapse icon goes to the collapsable position

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/ReportTable.jsx
+++ b/app/assets/src/components/views/SampleViewV2/ReportTable.jsx
@@ -1,6 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { getOr, orderBy } from "lodash/fp";
+import { getOr, map, orderBy } from "lodash/fp";
 
 import { Table } from "~/components/visualizations/table";
 import { defaultTableRowRenderer } from "react-virtualized";
@@ -22,11 +22,20 @@ class ReportTable extends React.Component {
     super(props);
 
     this.state = {
+      expandAllOpened: false,
       expandedGenusIds: new Set(),
       dbType: this.props.initialDbType,
     };
 
     this.columns = [
+      {
+        cellRenderer: this.renderExpandIcon,
+        className: cs.expandHeader,
+        dataKey: "expanded",
+        headerClassName: cs.expandCell,
+        headerRenderer: this.renderExpandIconHeader,
+        width: 20,
+      },
       {
         cellRenderer: this.renderName,
         className: cs.nameCell,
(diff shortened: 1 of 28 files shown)


app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -354,10 +390,36 @@ class ReportTable extends React.Component {
...
+        expandedGenusIds: new Set(),
+        expandAllOpened: false,
+      });
+    } else {
+      this.setState({
+        expandedGenusIds: new Set(map("taxId", data)),
+        expandAllOpened: true,
+      });
+    }
+    event.stopPropagation();
```
MarkAZhang: Can you add a comment around why `event.stopPropagation` is necessary? These calls can be hard to understand so it's good to explain them.

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -354,10 +390,36 @@ class ReportTable extends React.Component {
...
   toggleExpandGenus = ({ taxId }) => {
-    const { expandedGenusIds } = this.state;
+    const { data } = this.props;
+    const { expandedGenusIds, expandAllOpened } = this.state;
     expandedGenusIds.delete(taxId) || expandedGenusIds.add(taxId);
 
-    this.setState({ expandedGenusIds: new Set(expandedGenusIds) });
+    const newExpandedAllOpened =
+      expandedGenusIds.size === data.length ||
+      (!!expandedGenusIds.size && expandAllOpened);
```
MarkAZhang: I don't really understand why `(!!expandedGenusIds.size && expandAllOpened)` is necessary. Can you clarify?

app/assets/src/components/views/SampleViewV2/report_table.scss
```diff
@@ -19,6 +19,19 @@
   text-align: left;
 }
 
+.expandIcon {
+  font-size: $font-size-h5;
+  border-left: 5px solid transparent;
```
MarkAZhang: Does `margin-left: 5px` not work here?

app/assets/src/components/views/SampleViewV2/report_table.scss
```diff
@@ -92,11 +101,13 @@
...
-.speciesRow {
-  border-left: 5px solid transparent;
+.speciesRow,
+.row {
+  // border-left: 5px solid transparent;
   &.highlighted {
     background-color: $primary-lightest;
-    border-left: 5px solid $primary-lighter;
+    // border-left: 5px solid $primary-lighter;
+    box-shadow: inset 5px 0 $primary-lighter;
```
MarkAZhang: Was `border-left` not working? `box-shadow` seems a little unconventional. Also remove commented lines.

MarkAZhang:
Also, can you include a screenshot in future PRs with UI changes? (not needed for this one)

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
tfrcarvalho: Actually it is not even needed with the final solution of having a dedicated column for this. Removed.

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
tfrcarvalho: It guarantees that individually closing/opening leads to the same global state.
Rewrote in a less condensed format to be more understandable.

app/assets/src/components/views/SampleViewV2/report_table.scss
tfrcarvalho: Margin no. But actually padding-left works and removes the need fox box-shadow, since we not longer have two borders fighting for the corner.

app/assets/src/components/views/SampleViewV2/report_table.scss
tfrcarvalho: See above.

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -354,10 +390,38 @@ class ReportTable extends React.Component {
...
+      expandedGenusIds: new Set(expandedGenusIds),
+      expandAllOpened: newExpandedAllOpened,
+    });
+  };
+
+  toggleExpandAll = () => {
+    const { data } = this.props;
+    const { expandAllOpened } = this.state;
+
+    if (expandAllOpened) {
```
j-x-han: Could `expandedGenusIds` be sorted? Currently if you have the table sorted by a column, you lose that sorting when toggling expand all rows.

(Alternatively I suppose a user could expand/close all rows first and then sort. 😛)

tfrcarvalho: Fixed.

tfrcarvalho:
> Small style bug I just noticed: when expanding a single row, all the columns jump slightly to the left

Fixed with an "enhanced coding" - aka, hack. :P

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -364,10 +401,38 @@ class ReportTable extends React.Component {
...
 
   toggleExpandGenus = ({ taxId }) => {
-    const { expandedGenusIds } = this.state;
+    const { data } = this.props;
+    const { expandedGenusIds, expandAllOpened } = this.state;
     expandedGenusIds.delete(taxId) || expandedGenusIds.add(taxId);
 
-    this.setState({ expandedGenusIds: new Set(expandedGenusIds) });
+    let newExpandedAllOpened = expandAllOpened;
+    if (expandedGenusIds.size === data.length) {
```
MarkAZhang: Thanks! The logic makes more sense now.

Nit:
I notice that the value of expandAllOpened depends not only on how many genuses are opened, but also on the previous value of expandAllOpened.

If there are 10 rows, and 5 are opened, and I click on the expandAll button, the behavior (whether all rows open or close) depends on whether I previously clicked the expandAll button. 

I feel like the behavior should just depend on the number of rows opened. So if there are 5/10 rows opened, clicking expandAll should always open all rows. It should only close all rows if 10/10 rows are opened.

I'll leave it up to you, but not knowing whether the button will expand all or collapse all feels more confusing to me.

tfrcarvalho: It feels more intuitive to me that if the users click in the expand/collapse all, that the next time they click it should have the opposite behavior (unless that behavior makes no sense because the rows are all expanded or all closed).
We can leave that as UX topic for @happyimadesignr to chime on once she does here design review.

PR #2777, opened 2019-11-25
[Report Page] Save options to the URL

tfrcarvalho:
# Description

* Save options to the URL as soon as options change
* Replaces current history

# Tests

* Change options and verify that url changes accordingly.

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
+++ b/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
@@ -406,9 +406,6 @@ export default class SampleViewV2 extends React.Component {
   };
 
   updateHistoryAndPersistOptions = () => {
-    // save new options to local storage
-    const { selectedOptions } = this.state;
-
     const urlState = pick(keys(URL_FIELDS), this.state);
     let localState = mapValuesWithKey((options, key) => {
       return omit(options.excludePaths || [], this.state[key]);
--- a/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
+++ b/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
@@ -393,9 +393,6 @@ export default class SampleViewV2 extends React.Component {
     // save new options to local storage
     const { selectedOptions } = this.state;
 
-    // remove exceptions to persistent options
-    const { taxon, ...localStatePersistedOptions } = selectedOptions;
-
     const urlState = pick(keys(URL_FIELDS), this.state);
     let localState = mapValuesWithKey((options, key) => {
       return omit(options.excludePaths || [], this.state[key]);
@@ -409,12 +406,7 @@ export default class SampleViewV2 extends React.Component {
 
     history.replaceState(urlState, `SampleView`, `${urlQuery}`);
 
-    localStorage.setItem(
-      "SampleViewOptions",
-      JSON.stringify({
-        selectedOptions: localStatePersistedOptions,
-      })
-    );
+    localStorage.setItem("SampleViewOptions", JSON.stringify(localState));
   };
 
   handleOptionChanged = ({ key, value }) => {
--- a/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
+++ b/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
@@ -8,7 +8,9 @@ import {
   get,
   keys,
   map,
+  mapValues,
   merge,
+  omit,
   pick,
   pull,
   set,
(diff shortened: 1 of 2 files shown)


app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -369,15 +379,33 @@ export default class SampleViewV2 extends React.Component {
...
+    const urlState = pick(["view", "selectedOptions"], this.state);
+
+    // Saving on URL enables sharing current view with other users
+    let urlQuery = this.urlParser.stringify(urlState);
+    if (urlQuery) {
+      urlQuery = `?${urlQuery}`;
+    }
+
+    // History state may include some small fields that enable direct loading of previous pages
+    // from browser history without having to request those fields from server (e.g. project)
```
MarkAZhang: Not sure what this means. Unclear what the use case for "push" might be. Can you clarify?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -42,7 +44,14 @@ export default class SampleViewV2 extends React.Component {
   constructor(props) {
     super(props);
 
-    let localState = this.loadState(localStorage, "SampleViewOptions");
+    this.urlParser = new UrlQueryParser({
+      pipelineVersion: "string",
+      selectedOptions: "object",
```
MarkAZhang: Should `view` be included here? I notice it is picked in `updateHistoryAndPersistOptions`.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -369,15 +379,33 @@ export default class SampleViewV2 extends React.Component {
...
-  persistReportOptions = () => {
+  updateHistoryAndPersistOptions = (action = "push") => {
     // save new options to local storage
     const { selectedOptions } = this.state;
+
     // remove exceptions to persistent options
-    const { taxon, ...persistentReportOptions } = selectedOptions;
+    const { taxon, ...localStatePersistedOptions } = selectedOptions;
+
+    const urlState = pick(["view", "selectedOptions"], this.state);
```
MarkAZhang: Should pipelineVersion be included in this `pick`? It's defined in the UrlQueryParser.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -60,7 +69,8 @@ export default class SampleViewV2 extends React.Component {
         view: "table",
         selectedOptions: this.defaultSelectedOptions(),
       },
-      localState
+      localState,
+      urlState
```
MarkAZhang: This part feels hard to read. It's hard to know which values are stored in localState and which ones are stored in urlState (looks like there's a lot of overlap). It looks like you have to cross-reference multiple functions to get the full list of fields stored in each. Is there any way to organize all that info in one place for better readability?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -42,7 +44,14 @@ export default class SampleViewV2 extends React.Component {
...
 
-    let localState = this.loadState(localStorage, "SampleViewOptions");
+    this.urlParser = new UrlQueryParser({
+      pipelineVersion: "string",
+      selectedOptions: "object",
+    });
+    this.urlOptions = this.urlParser.parse(location.search);
+
+    const urlState = pick("selectedOptions", this.urlOptions);
+    const localState = this.loadState(localStorage, "SampleViewOptions");
```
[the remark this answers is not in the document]
tfrcarvalho: I do not think this single line justifies the an extra API just for this... But since this is the second time we do it, I'm thinking of ways of generalizing the whole pushing to browsing and to state in the future.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: Did the change with fields reference in constants at the top.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: It can be as it will be needed soon.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: It was more of a generic option for whatever we might not want to undo.
But removed for now to not include possible future needs.

PR #2781, opened 2019-11-25
Import tooltips for new report's column headers

j-x-han:
# Description

Adds the more detailed tooltips with Learn More links to the new report table view.

image

# Notes

* Taxon column displays a black basic popup; the previous version of the report page has no tooltip at all for this header.
* NT/NR toggle has no tooltip for the moment; to discuss with @happyimadesignr to see what she wants to do with that column?

# Tests

* Verified that the detailed tooltips with links to the help center work as expected.


j-x-han:
@MarkAZhang By default, BaseTable will add a BasicPopup for a column header that just displays the column's name. I think the reason why this doesn't happen on the old version of the report is because it implements its own table instead of wrapping BaseTable.
image

PR #2787, opened 2019-11-26
Fetch unmapped reads count from assembly

gregdingle:
# Description

See IDSEQ-1750. As reported by @katrinakalantar : The unmapped reads field currently displays the # of reads that are unmapped after the initial gsnap/rapsearch steps. The sample report table shows metrics including reads that are "mapped" by assembly into a contig. The "unmapped reads" value should represent the number of reads that are totally unmapped at the END of the pipeline.

This PR adds code to fetch the unmapped count after assembly from the existing file `refined_annotated_out.count`. The final count should be set by the pipeline monitor after all stages have finished.

# Notes

* The count ultimately comes from the lines of the `unidentified_fasta` file. 
```
# idseq_dag/steps/generate_annotated_fasta.py
self.counts_dict["unidentified_fasta"] = count.reads_in_group([self.output_files_local()[1]])
```

* One confusing thing I found was in the pipeline output files was that `additional_files_to_upload` appear to all go into root pipeline run dir, not with other files of the same steps in `/assembly`. Seems like a poor design. 

* The reason why the `stats.json` file exists is a mystery to me. Why not use the database? In fact, it appears we do both. We store the same data in `job_stats_attributes`. There is also a comment by @boris-dimitrov : `# TODO:  S3 is a middleman between these two functions;  load_stats shouldn't wait for S3`

* I refactored a bunch of lines to use `supports_assembly?`

# Tests

I couldn't think of a good way to run automated tests on this so I ran manual tests in the console. 

I will also check the values in staging of the benchmark run after alignment then again after assembly. 

```
pr = PipelineRun.where(
  "sample_id < 12640 AND unmapped_reads > 0 AND pipeline_version > 3.1 AND assembled = 1"
).last
pr.unmapped_reads # 21

# test that reads_after is returned when file does not exist
pr.fetch_unmapped_reads([
  {
    "reads_after": 8262,
    "task": "unidentified_fasta"
  }
]) # 8262

# test that remote file count is returned
pr.fetch_unmapped_reads([
  {
    "reads_after": 8262,
    "task": "unidentified_fasta"
  }
], 
"s3://idseq-samples-prod/samples/744/34488/postprocess/3.13/refined_annotated_out.count"
) # 15812

```

### DIFF
--- a/app/helpers/pipeline_runs_helper.rb
+++ b/app/helpers/pipeline_runs_helper.rb
@@ -250,4 +250,9 @@ module PipelineRunsHelper
 
     return valid_pipeline_runs
   end
+
+  def aws_s3_read_json(path)
+    raw = Syscall.run("aws", "s3", "cp", path, "-")
+    JSON.parse(raw)
+  end
 end
--- a/app/helpers/pipeline_runs_helper.rb
+++ b/app/helpers/pipeline_runs_helper.rb
@@ -250,9 +250,4 @@ module PipelineRunsHelper
 
     return valid_pipeline_runs
   end
-
-  def aws_s3_read_json(path)
-    raw = Syscall.run("aws", "s3", "cp", path, "-")
-    JSON.parse(raw)
-  end
 end
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -45,6 +45,7 @@ class PipelineRun < ApplicationRecord
   SORTED_TAXID_ANNOTATED_FASTA_FAMILY_NR = 'taxid_annot_sorted_family_nr.fasta'.freeze
 
   DAG_ANNOTATED_FASTA_BASENAME = 'taxid_annot.fasta'.freeze
+  DAG_ANNOTATED_COUNT_BASENAME = 'annotated_out.count'.freeze
   DAG_UNIDENTIFIED_FASTA_BASENAME = 'unidentified.fa'.freeze
   UNIDENTIFIED_FASTA_BASENAME = 'unidentified.fasta'.freeze
   MULTIHIT_FASTA_BASENAME = 'accessions.rapsearch2.gsnapl.fasta'.freeze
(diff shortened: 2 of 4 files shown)


app/models/pipeline_run.rb
```diff
@@ -1019,10 +1041,7 @@ def compile_stats_file
     end
 
     # Load unidentified reads
-    unidentified = all_counts.detect { |entry| entry.value?("unidentified_fasta") }
-    if unidentified
-      self.unmapped_reads = unidentified[:reads_after]
-    end
+    self.unmapped_reads = fetch_unmapped_reads(all_counts) || unmapped_reads
```
cdebourcy: what does `|| unmapped_reads` do here?

gregdingle: it keeps the current value if `fetch_unmapped_reads` returns nil because of an error

app/models/pipeline_run.rb
```diff
@@ -1038,6 +1057,32 @@ def compile_stats_file
...
+  def fetch_unmapped_reads(
+    all_counts,
+    s3_path = "#{postprocess_output_s3_path}/#{ASSEMBLY_PREFIX}#{DAG_ANNOTATED_COUNT_BASENAME}"
+  )
+    unmapped_reads = nil
+    unidentified = all_counts.detect { |entry| entry.value?("unidentified_fasta") }
+    if unidentified
+      unmapped_reads = unidentified[:reads_after]
+
+      if supports_assembly? && finalized?
```
cdebourcy: curious, why is this clause inside the `if unidentified` block?

cdebourcy:
This is cool!

I believe you need to change the corresponding file download as well:
image

app/models/pipeline_run.rb
cdebourcy: interesting. Makes sense - so `fetch_unmapped_reads  -> nil` is both the "not yet available" case (for which `|| unmapped_reads` doesn't change anything) and the "error" case (for which `|| unmapped_reads` does do something)

app/models/pipeline_run.rb
MarkAZhang: I guess the idea is that if the alignment unmapped read count isn't available, then there's no way the assembly umapped read count is available, so we can skip it. Is it possible for the alignment one to be missing, but the assembly one to be present?

app/models/pipeline_run.rb
```diff
@@ -1038,6 +1057,32 @@ def compile_stats_file
     save
   end
 
+  # Fetch the unmapped reads count from alignment stage then refined counts from
+  # assembly stage, as each becomes available. Prior to Dec 2019, the count was
+  # only fetched from alignment.
```
MarkAZhang: Could you add a comment somewhere about how `all_counts` contains all the `.count` files from `/results`, but we need to fetch this one `.count` file from `/postprocess`? It was a little hard to understand why you needed to fetch another file in this method.

app/models/pipeline_run.rb
MarkAZhang: I'm not sure the `finalized?` should be there. It seems possible for the file to exist even though the pipeline isn't done yet, and I think we'd want to display the value once it's ready.

gregdingle: Yes just like mark said.

gregdingle: Yes just like mark said.

app/helpers/pipeline_runs_helper.rb
```diff
@@ -250,4 +250,9 @@ def get_succeeded_pipeline_runs_for_samples(samples, strict = false)
 
     return valid_pipeline_runs
   end
+
+  def aws_s3_read_json(path)
```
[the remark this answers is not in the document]
gregdingle: Sure I’ll take a look at those. In general, it seems like we should be using the ruby API not sys calls.

app/models/pipeline_run.rb
gregdingle: Yes I added that as an afterthought. I wanted to avoid the result monitor polling for a file that’s not there most of the time. Is there some state var that is closely coupled? Assembly done?

MarkAZhang: Not that I know of, but you might be able to write an assembly_done method by checking whether pipeline run stage step 3 is complete. Will leave that up to you.

PR #2788, opened 2019-11-27
Migrate new taxid columns in contig table.

MarkAZhang:
# Description

Add data migration Resque task to populate contig taxid columns.

This data migration will need to be run manually after the deployment with `Resque.enqueue(PopulateContigSpeciesTaxids)`. It should interfere minimally with the service (it acquires a write lock on the current contigs it is modifying, but idseq-web doesn't modify contigs after they are loaded so it should be unaffeted). The data migration may take a while to run and will run in a Resque worker.

Also update the db_load_contig_count code to populate the new columns.

# Notes

This data migration will need to be run manually after the deployment with `Resque.enqueue(PopulateContigSpeciesTaxids)`.

# Tests

* Verifed that Resque task performs correctly. Will need to run this again on staging and verify that it works, and that it doesn't degrade site performance while it's running. If the performance isn't great or it turns out to be buggy, we can always choose not to run the data migration until we push out a fix. (it doesn't happen automatically)
* Verified that db_load_contig_counts loads the contigs properly with the new taxids.

### DIFF
--- a/app/jobs/data_migrations/populate_contig_species_taxids.rb
+++ b/app/jobs/data_migrations/populate_contig_species_taxids.rb
@@ -5,8 +5,6 @@ class PopulateContigSpeciesTaxids
   @queue = :data_migration
   def self.perform
     start_time = Time.now.to_f
-    missing_species_id = TaxonLineage::MISSING_SPECIES_ID
-    missing_genus_id = TaxonLineage::MISSING_GENUS_ID
 
     # Just select the columns that we need. In particular, don't select the sequence column, which is very large.
     # The "lock" specifies that whenever you make a SELECT query,
@@ -32,10 +30,10 @@ class PopulateContigSpeciesTaxids
           begin
             if contig.lineage_json.present?
               lineage_json = JSON.parse(contig.lineage_json)
-              contig.species_taxid_nt = lineage_json.dig("NT", 0) || missing_species_id
-              contig.species_taxid_nr = lineage_json.dig("NR", 0) || missing_species_id
-              contig.genus_taxid_nt = lineage_json.dig("NT", 1) || missing_genus_id
-              contig.genus_taxid_nr = lineage_json.dig("NR", 1) || missing_genus_id
+              contig.species_taxid_nt = lineage_json.dig("NT", 0) || nil
+              contig.species_taxid_nr = lineage_json.dig("NR", 0) || nil
+              contig.genus_taxid_nt = lineage_json.dig("NT", 1) || nil
+              contig.genus_taxid_nr = lineage_json.dig("NR", 1) || nil
             end
           rescue => e
             Rails.logger.error(e)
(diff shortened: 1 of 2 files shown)


app/jobs/data_migrations/populate_contig_species_taxids.rb
```diff
@@ -0,0 +1,72 @@
...
+            end
+          rescue => e
+            Rails.logger.error(e)
+            failed_ids << contig.id
+          end
+        end
+
+        # Use active-import to update multiple contigs in one query.
+        # Just update the taxid fields.
+        results = Contig.import loaded_contigs, validate: false, on_duplicate_key_update: [:species_taxid_nt, :species_taxid_nr, :genus_taxid_nt, :genus_taxid_nr]
```
tfrcarvalho: nit: use parentheses?

app/jobs/data_migrations/populate_contig_species_taxids.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    unless failed_ids.empty?
+      Rails.logger.info("#{failed_ids.length} failed ids: #{failed_ids}")
+    end
+    Rails.logger.info(format("Updated %d rows in %3.1f seconds", total_updated, (Time.now.to_f - start_time)))
+  rescue => e
+    # Eventually, this data migration may no longer be compatible with the database.
+    Rails.logger.error(e)
+    Rails.logger.error("populate_contig_species_taxids failed. This data migration may be out of date with the db schema. Consider deleting.")
```
tfrcarvalho: 👍

app/models/pipeline_run.rb
```diff
@@ -542,16 +542,31 @@ def db_load_contigs(contig2taxid)
     taxon_lineage_map = {}
     TaxonLineage.where(taxid: taxid_list.uniq).order(:id).each { |t| taxon_lineage_map[t.taxid.to_i] = t.to_a }
 
+    # A lambda allows us to access variables in the enclosing scope, such as contig2taxid.
+    get_contig_hash = lambda do |header, sequence|
+      read_count = contig_stats_json[header] || 0
+      lineage_json = get_lineage_json(contig2taxid[header], taxon_lineage_map)
+      species_taxid_nt = lineage_json.dig("NT", 0) || TaxonLineage::MISSING_SPECIES_ID
```
tfrcarvalho: As discussed, lets not use MISSING ids as it can lead to errors when merging with taxon counts.

app/models/pipeline_run.rb
```diff
@@ -542,16 +542,31 @@ def db_load_contigs(contig2taxid)
...
+      genus_taxid_nr = lineage_json.dig("NR", 1) || nil
+
+      {
+        name: header, sequence: sequence, read_count: read_count, lineage_json: lineage_json.to_json,
+        species_taxid_nt: species_taxid_nt, species_taxid_nr: species_taxid_nr,
+        genus_taxid_nt: genus_taxid_nt, genus_taxid_nr: genus_taxid_nr,
+      }
+    end
+
     File.open(contig_fasta, 'r') do |cf|
```
gregdingle: maybe out of scope for this PR, but it seems like a fasta file parser should be in a util module. 

also, I'm surprised the pipeline does not do the parsing and generate the structured info we need here in a json file

app/jobs/data_migrations/populate_contig_species_taxids.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+
+      # Regularly log progress updates.
+      if total_updated % 100_000 == 0
+        Rails.logger.info("populate_contig_species_taxids: #{total_updated} contigs processed...")
+      end
+    end
+
+    unless failed_ids.empty?
+      Rails.logger.info("#{failed_ids.length} failed ids: #{failed_ids}")
```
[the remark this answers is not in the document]
MarkAZhang: Yes, was planning to when running the migration.

PR #2790, opened 2019-11-27
[Report Page] Add alerts for in progress/failed samples

j-x-han:
# Description

Add the alerts/statuses for samples that are in progress, failed, or completed with issues to the new report page. Example screenshots of these states below:

Screen Shot 2019-11-26 at 1 12 55 PM
Screen Shot 2019-11-26 at 1 12 35 PM
Screen Shot 2019-11-26 at 1 51 45 PM
Screen Shot 2019-11-26 at 2 54 45 PM
Screen Shot 2019-11-26 at 3 26 30 PM
Screen Shot 2019-11-26 at 1 12 46 PM

# Notes

* Most of the `renderSampleMessage` [code](https://github.com/chanzuckerberg/idseq-web/blob/476581d1259c0e8eee4b54e38ae26fda58ee9774/app/assets/src/components/views/SampleView/SampleView.jsx#L305) and [styling](https://github.com/chanzuckerberg/idseq-web/blob/476581d1259c0e8eee4b54e38ae26fda58ee9774/app/assets/src/components/views/SampleView/sample_view.scss#L109) was migrated over from the old version of the report page. The styles were replaced with existing spacing variables where possible.

* Samples in the `waiting` status will occasionally throw the following error... It seems that the `rawReportData` isn't fetched so the `pipelineRunInfo` fails to be populated. The error seems to be intermittent, since it appears/disappears when refreshing. Currently investigating, fix may come in a future PR.
image

# Tests

* Verified that samples in various states of in progress, completed, or failed displayed alerts as expected.
* Verified that links in the alerts direct users to the expected page.

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
+++ b/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
@@ -20,6 +20,7 @@ import {
   values,
 } from "lodash/fp";
 import deepEqual from "fast-deep-equal";
+import cx from "classnames";
 
 import {
   getBackgrounds,
@@ -31,6 +32,7 @@ import { getAmrData } from "~/api/amr";
 import { UserContext } from "~/components/common/UserContext";
 import { AMR_TABLE_FEATURE } from "~/components/utils/features";
 import { logAnalyticsEvent, withAnalytics } from "~/api/analytics";
+import { sampleErrorInfo } from "~/components/utils/sample";
 import DetailsSidebar from "~/components/common/DetailsSidebar";
 import NarrowContainer from "~/components/layout/NarrowContainer";
 import PropTypes from "~/components/utils/propTypes";
@@ -39,6 +41,9 @@ import SampleViewHeader from "./SampleViewHeader";
 import Tabs from "~/components/ui/controls/Tabs";
 import UrlQueryParser from "~/components/utils/UrlQueryParser";
 import AMRView from "~/components/AMRView";
+import BacteriaIcon from "~ui/icons/BacteriaIcon";
+import AlertIcon from "~ui/icons/AlertIcon";
+import LoadingIcon from "~/components/ui/icons/LoadingIcon";
 
 import ReportFilters from "./ReportFilters";
 import cs from "./sample_view_v2.scss";
@@ -73,6 +78,7 @@ export default class SampleViewV2 extends React.Component {
         backgrounds: [],
         currentTab: "Report",
         filteredReportData: [],
+        pipelineRunInfo: {},
         pipelineRun: null,
         project: null,
         projectSamples: [],
@@ -202,6 +208,7 @@ export default class SampleViewV2 extends React.Component {
       selectedOptions: Object.assign({}, selectedOptions, {
         background: rawReportData.metadata.backgroundId,
       }),
+      pipelineRunInfo: rawReportData.pipelineRunInfo,
     });
   };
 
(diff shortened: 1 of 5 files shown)


app/services/pipeline_report_service.rb
```diff
@@ -67,13 +67,39 @@ def call
...
 
   def generate
-    pipeline_run = PipelineRun.find(@pipeline_run_id)
     if @pipeline_run_id.nil?
-      Rails.logger.error("Pipeline run doesn't exist.")
-      return ""
-    elsif pipeline_run.total_reads.nil? || pipeline_run.adjusted_remaining_reads.nil?
-      Rails.logger.error("Pipeline run #{@pipeline_run_id} has no reads.")
-      return ""
+      return JSON.dump(
```
tfrcarvalho: Let's add this to a function called `get_pipeline_status` which can get the pipeline run.

app/services/pipeline_report_service.rb
```diff
@@ -67,13 +67,39 @@ def call
...
+        pipelineRunInfo: {
+          pipelineRunComplete: false,
```
tfrcarvalho: What do you thing of restructuring this to a ```
status: oneof([COMPLETE, FAILED, WAITING])
errorType: oneof([USER, PIPELINE, UNKNOWN]),
statusMessage: "Error message"
```

File: app/services/pipeline_report_service.rb [hunk 3]
```diff
@@ -67,13 +67,39 @@ def call
...
+          pipelineRunFailed: false,
+          sampleStatus: "Waiting to Start or Receive Files",
+          errorMessage: nil,
+          knownUserError: nil,
+        },
+        metadata: {},
```
tfrcarvalho: Maybe use the metadata to show the status?

app/services/pipeline_report_service.rb
```diff
@@ -67,13 +67,39 @@ def call
...
+        counts: {},
```
tfrcarvalho: Do we need to return these keys?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -760,39 +878,7 @@ export default class SampleViewV2 extends React.Component {
...
-                <ReportTable
-                  data={filteredReportData}
-                  onTaxonNameClick={this.handleTaxonClick}
-                />
-              </div>
-            </div>
-          )}
-          {currentTab === "Antimicrobial Resistance" &&
-            amrData && <AMRView amr={amrData} />}
+          {this.renderTab()}
```
tfrcarvalho: Lets keep the tab check here and call renderReport(). 
I am also considering moving the renderReport into a Report component later on.

app/services/pipeline_report_service.rb
j-x-han: Done.

app/services/pipeline_report_service.rb
j-x-han: Combined `pipelineRunComplete` and `pipelineRunFailed` into `pipelineRunStatus`.

It looks like `known_user_error` is an existing field for pipeline runs, so I left that one as is for consistency. I didn't change `errorMessage` either because it's specific to errors, not all status messages.

app/services/pipeline_report_service.rb
j-x-han: The status is actually corresponding to the job status of the pipeline run, not the sample itself, so I felt it would fit in better with `pipelineRunInfo`. Renamed to `jobStatus` for more clarity.

app/services/pipeline_report_service.rb
j-x-han: `SampleViewV2` attempts to parse/iterate through these keys, so the frontend will throw some errors if they don't exist. I could alternatively add some checks to `fetchSampleReportData` to avoid those errors. Which way is preferred?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
j-x-han: Done.

app/services/pipeline_report_service.rb
tfrcarvalho: All this metadata is specific for the pipeline run (not sample). Don't forget  this is a pipeline run report service.

app/services/pipeline_report_service.rb
tfrcarvalho: I would prefer that the frontend adjusts.

j-x-han: Done.

app/services/pipeline_report_service.rb
j-x-han: Removed pipelineRunInfo and moved everything into metadata as discussed!

PR #2791, opened 2019-11-27
[Report Page] Fix bugs in validate names

j-x-han:
# Description

Discovered a typo and a null error bug in `validate_names` when switching to older pipeline runs for a sample.

The new report page should no longer error if the sample contains a species without a name; it should now display `Non-Species-Specific Reads In Genus >genus name/taxid<` as expected.

# Tests

* Verified that report page does not error if a species has a missing name.


j-x-han:
I'll make a note to add test cases for missing names!

PR #2794, opened 2019-11-28
[Report Page] Add hover actions.

tfrcarvalho:
# Description

* Adds hover actions to the report page v2
* Fix a bug where it was showing an error when sample fails on experimental stage

# Tests

* Tested each of the actions

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/ReportTable.jsx
+++ b/app/assets/src/components/views/SampleViewV2/ReportTable.jsx
@@ -1,34 +1,47 @@
 import React from "react";
-import PropTypes from "prop-types";
-import { getOr, map, orderBy } from "lodash/fp";
+import { compact, getOr, keys, map, orderBy } from "lodash/fp";
+import cx from "classnames";
 
 import { Table } from "~/components/visualizations/table";
 import { defaultTableRowRenderer } from "react-virtualized";
-import cx from "classnames";
-import { withAnalytics } from "~/api/analytics";
-import TableRenderers from "~/components/views/discovery/TableRenderers";
-import InsightIcon from "~ui/icons/InsightIcon";
 import { getCategoryAdjective } from "~/components/views/report/utils/taxon";
-import { REPORT_TABLE_COLUMNS } from "../report/ReportTable/constants";
+import { withAnalytics } from "~/api/analytics";
+import {
+  pipelineVersionHasAssembly,
+  pipelineVersionHasCoverageViz,
+} from "~/components/utils/sample";
+import { UserContext } from "~/components/common/UserContext";
+import InsightIcon from "~ui/icons/InsightIcon";
+import TableRenderers from "~/components/views/discovery/TableRenderers";
+import PhyloTreeChecks from "~/components/views/phylo_tree/PhyloTreeChecks";
+import PropTypes from "~/components/utils/propTypes";
+import PhyloTreeCreationModal from "~/components/views/phylo_tree/PhyloTreeCreationModal";
 
+import { REPORT_TABLE_COLUMNS } from "../report/ReportTable/constants";
+import HoverActions from "../report/ReportTable/HoverActions";
 import cs from "./report_table.scss";
 
 // Values for null values when sorting ascending and descending
 // for strings - HACK: In theory, there can be strings larger than this
 const STRING_NULL_VALUES = ["", "zzzzzzzzz"];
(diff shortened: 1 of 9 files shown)


app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -340,6 +354,137 @@ class ReportTable extends React.Component {
...
+    const phyloTreeEnabled =
+      phyloTreeAllowed &&
+      rowData.taxId > 0 &&
+      PhyloTreeChecks.passesCreateCondition(
+        getOr(0, "nt.count", rowData),
+        getOr(0, "nr.count", rowData)
+      );
+
+    const analyticsContext = {
+      projectId: projectId,
```
j-x-han: Previous version of the report includes `projectName` as well, do we want to keep that?

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -462,20 +607,54 @@ class ReportTable extends React.Component {
...
+          rowClassName={cs.row}
+          rowRenderer={this.rowRenderer}
+          sortable={true}
+          sortedHeaderClassName={cs.sortedHeader}
+        />
+        {phyloTreeModalParams && (
+          <UserContext.Consumer>
+            {currentUser => (
+              <PhyloTreeCreationModal
+                admin={currentUser.admin}
```
j-x-han: Some console errors when selecting phylo tree: 
image

app/services/pipeline_report_service.rb
```diff
@@ -241,7 +247,11 @@ def fetch_taxon_counts(_pipeline_run_id, _background_id)
                                          tax_level: [TaxonCount::TAX_LEVEL_SPECIES, TaxonCount::TAX_LEVEL_GENUS]
                                        )
                                        .where.not(
-                                         tax_id: [TaxonLineage::BLACKLIST_GENUS_ID, TaxonLineage::HOMO_SAPIENS_TAX_ID]
+                                         tax_id: [
+                                           TaxonLineage::BLACKLIST_GENUS_ID,
+                                           TaxonLineage::HOMO_SAPIENS_TAX_ID,
```
j-x-han: It looks like `HUMAN_TAX_IDS` ([9605, 9606]) includes `HOMO_SAPIENS_TAX_ID` (9606), is there a reason why we have both?

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
tfrcarvalho: No, it was conscious decision.

app/services/pipeline_report_service.rb
tfrcarvalho: Actually no. Should have only HUMAN_TAX_IDS. Good catch.

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -1,34 +1,47 @@
...
   constructor(props) {
     super(props);
 
+    const { pipelineVersion } = props;
+
     this.state = {
       expandAllOpened: false,
       expandedGenusIds: new Set(),
       dbType: this.props.initialDbType,
+      // hover actions state
```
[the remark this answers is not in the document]
tfrcarvalho: Removed.

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -462,20 +607,54 @@ class ReportTable extends React.Component {
...
+
+    if (xsrfCookies.length === 0) {
+      return null;
+    }
+    return decodeURIComponent(xsrfCookies[0].split("=")[1]);
+  };
+
   render = () => {
-    const { rowHeight } = this.props;
+    const { projectId, projectName, rowHeight } = this.props;
```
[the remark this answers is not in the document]
tfrcarvalho: Thanks, fixed.

app/services/pipeline_report_service.rb
tfrcarvalho: Will change `HOMO_SAPIENS_TAX_ID` to the correct definition. We should deprecate `HUMAN_TAX_IDS`.

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
tfrcarvalho: Fixed.

PR #2795, opened 2019-11-28
[Report Page] Add analytics

j-x-han:
# Description

Add back some analytics to the new report page that were missing after the refactor.

# Notes

A second pass should be done to make sure all analytics were carried over once all the features are in. Analytics that are known to be missing since they're dependent on features in pending/future PRs:
* row actions
* switching between table/tree view

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/ReportTable.jsx
+++ b/app/assets/src/components/views/SampleViewV2/ReportTable.jsx
@@ -5,7 +5,7 @@ import { getOr, map, orderBy } from "lodash/fp";
 import { Table } from "~/components/visualizations/table";
 import { defaultTableRowRenderer } from "react-virtualized";
 import cx from "classnames";
-import { withAnalytics } from "~/api/analytics";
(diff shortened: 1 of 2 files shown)


app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -326,13 +326,29 @@ class ReportTable extends React.Component {
...
         >
           {cellData ? cellData[0] : "-"}
         </div>
         <div
           className={cx(cs.stackElement, dbType == "nr" || cs.lowlightValue)}
-          onClick={onClick ? () => onClick[1]("nr") : null}
+          onClick={
+            onClick
+              ? () =>
+                  withAnalytics(
```
tfrcarvalho: The withAnalytics is supposed to receive a function. 
You can do 
```onClick={withAnalytics(
	() => onClick && onClick[1]("nr"),
	"ReportTable_count-type-nr_clicked"
```
Also, you have to add this to `renderNtNrSelector` function (not `renderNtNrValues`). I think the onClick here was a mistake.

PR #2799, opened 2019-12-02
[IDSEQ-1549] Add barebones Omniauth error page

jshoe:
### Description
- Ran into this while implementing the "expire admin passwords" task. Our current error behavior is to always hit the `auth0_signout_url` endpoint, but we don't have a good flash message on that page.
- Added an error page using `BlankScreenMessage`.

### Notes
- I could also add a timed redirect back to the login page (?).

### Tests
- Output from `new UnauthorizedError` on the Auth0 side:

Screen Shot 2019-12-02 at 2 36 34 PM

- Regular behavior for other errors like `access_denied` or `login_required` with the refresh token expirations.

Dec-02-2019 14-37-52


app/controllers/auth0_controller.rb
```diff
@@ -41,6 +42,20 @@ def failure
     logout
   end
 
+  # Handle omniauth errors coming from Auth0.
+  def omniauth_failure
+    error = params["error"]
+
+    # Display 'unauthorized' errors but go to `failure` endpoint for all others.
+    if error.present? && error == AUTH0_UNAUTHORIZED
+      description = params["error_description"]
```
davidrissato: I believe we need sanitize this input and remove any HTML tags from it to avoid script injections.

jshoe: added a sanitize 👍

app/controllers/concerns/parameter_sanitization.rb
```diff
@@ -15,4 +15,9 @@ def sanitize_title_name(title)
     # Allow letters, numbers, underscores, dashes, and spaces
     return title.gsub(/[^A-Za-z0-9_\- ]/, ' ').strip
   end
+
+  def sanitize_description(description)
+    # Allow letters, numbers, underscores, dashes, spaces, colons, and periods.
+    description.gsub(/[^A-Za-z0-9_\-:. ]/, ' ').strip
```
davidrissato: In this case we may want more loose rules. I suggest using this approach:
```ActionController::Base.helpers.sanitize("Should remove <b>tags</b>", tags: [])```

PR #2802, opened 2019-12-03
Enable/redirect users with report_v2 flag to new report page

j-x-han:
# Description

Automatically redirect users with the `report_v2` feature flag to the new report page when viewing a sample.

There is a link at the top of the page to navigate to the legacy version of the report.
image

Users can also view the old version of the report by adding `/legacy` after the sample id in the url, e.g. `/samples/{sample-id}/legacy`.

# Notes

* Minor bug fix in `reports_helper.rb`: check if species name exists before assigning a name to fake genus.

# Tests

* Verify sample report pages display the new version by default.
* Verify adding `/legacy` will bring users to the old version of the report page.
* Verify users without the `report_v2` flag will still see the old report page by default.
* Verify users without the `report_v2` flag will be redirected to the home page if they attempt to access the new report page directly (`/samples/{sample-id}/show_v2`).

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -48,7 +48,7 @@ class SamplesController < ApplicationController
   before_action :check_owner, only: OWNER_ACTIONS
   before_action :check_access
   before_action only: :show_v2 do
-    allowed_feature_required("report_v2", true)
+    allowed_feature_required("report_v2")
   end
   before_action only: :amr do
     allowed_feature_required("AMR")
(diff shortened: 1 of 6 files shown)


app/controllers/samples_controller.rb
```diff
@@ -680,56 +680,65 @@ def save_metadata_v2
...
   # GET /samples/1.json
   def show
-    @pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
-    @amr_counts = nil
-    can_see_amr = (current_user.admin? || current_user.allowed_feature_list.include?("AMR"))
-    if can_see_amr && @pipeline_run
-      amr_state = @pipeline_run.output_states.find_by(output: "amr_counts")
-      if amr_state.present? && amr_state.state == PipelineRun::STATUS_LOADED
-        @amr_counts = @pipeline_run.amr_counts
+    if !params[:legacy] && (current_user.admin? || current_user.allowed_feature_list.include?("report_v2"))
```
tfrcarvalho: Let's just do feature flag and then we add the flag to admins. I just want to make sure we can revert this quickly without an extra push.

j-x-han: Removed admin access and added link to legacy page.

PR #2803, opened 2019-12-03
IDSEQ-1555 - Auth0 User Management

davidrissato:
# Description

IDSEQ-1555 - Synchronize auth0 users info when editing, creating and deleting users using admin user console.

# Tests

* Manual

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -71,7 +71,7 @@ class UsersController < ApplicationController
   def destroy
     @user.destroy!
 
-    # Deletion user from Auth0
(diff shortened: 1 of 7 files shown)


app/controllers/users_controller.rb
```diff
@@ -63,7 +69,11 @@ def update
   # DELETE /users/1
   # DELETE /users/1.json
   def destroy
-    @user.destroy
+    @user.destroy!
+
+    # Deletion user from Auth0
```
jshoe: nit: "Delete user from Auth0"

PR #2804, opened 2019-12-03
Add taxon contig select for selecting taxa for contig bulk download.

MarkAZhang:
# Description

This adds the dropdown for selecting taxa.
Screen Shot 2019-12-03 at 1 36 58 PM

# Notes
* An upcoming refactor will refactor the existing dropdown for selecting taxon for the reads non-host download into its own component. That component is quite similar to this one.

# Tests
* Verified that the select only shows taxon with contigs in the selected samples (as opposed to reads).
* Wrote model and controller tests.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -1378,9 +1378,9 @@ class SamplesController < ApplicationController
 
     # User should not be querying for unviewable samples.
     if samples.length != sample_ids.length
-      LogUtil.log_err_and_airbrake("Get taxa with reads error: Unauthorized access of samples")
+      LogUtil.log_err_and_airbrake("Get taxa with contigs error: Unauthorized access of samples")
       render json: {
-        error: "There was an error fetching the taxa with reads for samples.",
+        error: "There was an error fetching the taxa with contigs for samples.",
       }, status: :unauthorized
       return
     end
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -29,7 +29,7 @@ class SamplesController < ApplicationController
 
   OTHER_ACTIONS = [:create, :bulk_new, :bulk_upload, :bulk_upload_with_metadata, :bulk_import, :new, :index, :index_v2, :details,
                    :dimensions, :all, :show_sample_names, :cli_user_instructions, :metadata_fields, :samples_going_public,
-                   :search_suggestions, :stats, :upload, :validate_sample_files, :taxa_with_reads_suggestions, :uploaded_by_current_user,].freeze
+                   :search_suggestions, :stats, :upload, :validate_sample_files, :taxa_with_reads_suggestions, :uploaded_by_current_user, :taxa_with_contigs_suggestions,].freeze
   OWNER_ACTIONS = [:raw_results_folder].freeze
   TOKEN_AUTH_ACTIONS = [:create, :update, :bulk_upload, :bulk_upload_with_metadata].freeze
 
@@ -1346,7 +1346,7 @@ class SamplesController < ApplicationController
 
   # GET /samples/taxa_with_reads_suggestions
   # Get taxon search suggestions, where taxa are restricted to the provided sample ids.
-  # Also include, for each taxon, the number of samples that contain the taxon.
+  # Also include, for each taxon, the number of samples that contain reads for the taxon.
   def taxa_with_reads_suggestions
     sample_ids = (params[:sampleIds] || []).map(&:to_i)
     query = params[:query]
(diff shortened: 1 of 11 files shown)


app/controllers/samples_controller.rb
```diff
@@ -1352,8 +1352,33 @@ def taxa_with_reads_suggestions
       return
     end
 
-    taxon_list = taxon_search(query, ["species", "genus"], samples: samples)
+    taxon_list = taxon_search(query, ["species", "genus"])
```
MarkAZhang: This slightly optimizes the reads endpoint.

It turns I don't need the `samples: samples` filter here, since when I calculate the counts, I make a very similar SQL query. Removing the filter removes a redundant SQL query.

app/helpers/samples_helper.rb
```diff
@@ -619,7 +619,50 @@ def augment_taxon_list_with_sample_count(taxon_list, samples)
...
+      end
+
+      # We just need the distinct pairs to do our counting, not the individual contigs.
+      Contig
+        .where(pipeline_run_id: pipeline_run_ids, "#{tax_id_field}": tax_ids)
+        .select("DISTINCT pipeline_run_id, #{tax_id_field}")
+        .each { |contig| pipeline_runs_by_taxid[contig[tax_id_field]] << contig.pipeline_run_id }
+    end
+
+    process_tax_ids.call("species", "nr")
```
MarkAZhang: I couldn't find an easy way to do this query without making four SQL queries. The queries themselves seem pretty fast (2ms for 8000 contigs) since the relevant columns are indexed.

app/models/pipeline_run.rb
```diff
@@ -11,6 +11,7 @@ class PipelineRun < ApplicationRecord
   accepts_nested_attributes_for :pipeline_run_stages
   has_and_belongs_to_many :backgrounds
   has_and_belongs_to_many :phylo_trees
+  has_and_belongs_to_many :bulk_downloads
```
MarkAZhang: I had forgotten this line earlier. If you don't have this, you can't destroy pipeline_runs (the foreign keys in bulk_downloads_pipeline_runs table will complain)

app/controllers/samples_controller.rb
tfrcarvalho: Do you filter by taxon counts too?
I believe that the samples filter constrains by taxon_counts, so you will not see tax ids that have contigs but not taxon counts.
If you remove the filter you will see all contigs (even if there are no taxon counts). Is this what you want (could be confusing for the end user)?

app/helpers/samples_helper.rb
tfrcarvalho: 2ms is still surprisingly fast!

app/helpers/samples_helper.rb
```diff
@@ -619,7 +619,50 @@ def augment_taxon_list_with_sample_count(taxon_list, samples)
...
+    # Separate genus and species taxids. We query them separately to try to get some extra performance.
+    species_tax_ids = get_tax_ids_by_level.call("species")
+    genus_tax_ids = get_tax_ids_by_level.call("genus")
+
+    pipeline_run_ids = get_succeeded_pipeline_runs_for_samples(samples).pluck(:id)
+
+    # We maintain a hash of taxid => array of pipeline run ids. We de-dupe the pipeline run ids at the very end.
+    pipeline_runs_by_taxid = Hash.new { |h, k| h[k] = [] } # Keys are auto-initialized to the empty array.
+
+    process_tax_ids = lambda do |level, type|
```
tfrcarvalho: note: I would not abuse of lambdas... seems like a a normal method would be good here.

app/controllers/samples_controller.rb
MarkAZhang: I filter by taxon counts for the taxon count endpoint and contigs for the contig endpoint. I think seeing all contigs even if there are no taxon counts is fine, since the user is downloading contigs, not reads. I think it would be wrong to hide a taxon with contigs but no taxon counts. (although I think this is very rare)

app/helpers/samples_helper.rb
MarkAZhang: The method uses `species_tax_ids` and `genus_tax_ids`, which are from the function scope. Is there a downside to using lambdas? I typically only use them for nested functions.

tfrcarvalho: IMO, just readability, since they are in the middle of another function. It's like code interleaved with functions, makes it hard to read.

app/controllers/samples_controller.rb
tfrcarvalho: The prob is they do not see these contigs anywhere else, which seems to be confusing to the user. 
I guess the long term solution is that we need to show them somewhere in the report.

MarkAZhang: Ahh I see what you mean. The long-term solution sounds good.

app/helpers/samples_helper.rb
MarkAZhang: That makes sense. I can make it a convention to move lambdas to the top of the function, instead of interleaving them.

I tend to use lambdas for cases where the lambda code is specific to the containing function. You could extract the lambda code outside the parent function, but that can also hurt readability because now the fact that the lambda is intended to only be used within the parent function is less clear.

MarkAZhang:
Hey @tfrcarvalho, I'll move the lambdas to the top of the function for readability. Let me know if you have any other comments. A couple of my changes (including a follow-up refactor) depend on this.

PR #2809, opened 2019-12-04
[Report Page] Fix tag_pathogens

j-x-han:
# Description

* The new report page was crashing for certain samples.
  *  `tag_pathogens` was checking the entire lineage of a taxon to see any of the names in the lineage were contained in the pathogen list. However, some tax ids for a given pipeline run are missing a taxon lineage defined for the corresponding taxon lineage version (e.g staging pipeline run 10339, tax id 171), so an error would occur.
  * `tag_pathogens` actually doesn't need to check the entire lineage since only species and genus are displayed, so the lineage check was removed to fix the bug.
* Pathogens on the genus level will now also be labelled appropriately. (Previously, all the species within the genus would be labelled rather than the genus row itself.)

image

# Notes

The TaxonLineage issue is also mentioned in [this doc](https://czi.quip.com/u5ajAc1o5Chq/Report-Page-Optimization).

# Tests

* Verified that new report page doesn't crash/load forever.
* Verified that pathogen labels show up on the genus level as expected.

### DIFF
--- a/app/services/pipeline_report_service.rb
+++ b/app/services/pipeline_report_service.rb
@@ -475,23 +475,20 @@ class PipelineReportService
   end
 
   def tag_pathogens(counts_by_tax_level)
-    counts_by_tax_level[TaxonCount::TAX_LEVEL_SPECIES].each do |_species_tax_id, species_info|
-      pathogen_tags = []
-      TaxonLineage::PRIORITY_PATHOGENS.each do |category, pathogen_list|
-        pathogen_tags |= [category] if pathogen_list.include?(species_info[:name])
-      end
-      best_tag = pathogen_tags[0] # first element is highest-priority element (see PRIORITY_PATHOGENS documentation)
-      species_info['pathogenTag'] = best_tag
-    end
-    counts_by_tax_level[TaxonCount::TAX_LEVEL_GENUS].each do |_genus_tax_id, genus_info|
-      pathogen_tags = []
-      TaxonLineage::PRIORITY_PATHOGENS.each do |category, pathogen_list|
-        pathogen_tags |= [category] if pathogen_list.include?(genus_info[:name])
-      end
-      best_tag = pathogen_tags[0] # first element is highest-priority element (see PRIORITY_PATHOGENS documentation)
-      genus_info['pathogenTag'] = best_tag
-    end
+    get_best_pathogen_tag(counts_by_tax_level[TaxonCount::TAX_LEVEL_SPECIES])
+    get_best_pathogen_tag(counts_by_tax_level[TaxonCount::TAX_LEVEL_GENUS])
     counts_by_tax_level
+  end
+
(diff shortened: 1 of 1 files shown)


app/services/pipeline_report_service.rb
```diff
@@ -474,22 +474,23 @@ def find_taxa_to_highlight(sorted_genus_tax_ids, counts_by_tax_level)
...
   end
 
-  def tag_pathogens(counts_by_tax_level, lineage_by_tax_id)
-    counts_by_tax_level[TaxonCount::TAX_LEVEL_SPECIES].each do |species_tax_id, species_info|
-      lineage = lineage_by_tax_id[species_tax_id] ? lineage_by_tax_id[species_tax_id] : lineage_by_tax_id[species_info[:genus_tax_id]]
-      name_columns = lineage.keys.select { |cn| cn.include?("_name") }
+  def tag_pathogens(counts_by_tax_level)
+    counts_by_tax_level[TaxonCount::TAX_LEVEL_SPECIES].each do |_species_tax_id, species_info|
       pathogen_tags = []
       TaxonLineage::PRIORITY_PATHOGENS.each do |category, pathogen_list|
```
tfrcarvalho: We should make a task of moving this data structure to a hash keyed by the name...

app/services/pipeline_report_service.rb
```diff
@@ -474,22 +474,23 @@ def find_taxa_to_highlight(sorted_genus_tax_ids, counts_by_tax_level)
...
-        pathogen_tags |= [category] if pathogen_list.include?(genus_name)
-        name_columns.each do |col|
-          pathogen_tags |= [category] if pathogen_list.include?(lineage[col])
-        end
       end
       best_tag = pathogen_tags[0] # first element is highest-priority element (see PRIORITY_PATHOGENS documentation)
       species_info['pathogenTag'] = best_tag
     end
+    counts_by_tax_level[TaxonCount::TAX_LEVEL_GENUS].each do |_genus_tax_id, genus_info|
+      pathogen_tags = []
```
tfrcarvalho: Can we put the content of this block on a method e.g. `get_best_pathogen_tag` and reuse it for species and genus?

j-x-han: Done

app/services/pipeline_report_service.rb
j-x-han: Do you mean restructuring `counts_by_tax_level` to look like {"species": {...}, "genus": {...}} rather than {1: {...}, 2: {...}}? I'll create a task/add a TODO comment.

tfrcarvalho: You should do that too! 
But I actually meant the `TaxonLineage::PRIORITY_PATHOGENS` should be refactored to hash of `{<name_or_tax_id>: <pathogen_tag>}`.
Sorry, should have been more clear...

PR #2811, opened 2019-12-04
[Report Page] Adapts taxonomic tree view to report page format changes.

tfrcarvalho:
# Description

* Create a new version of TaxonTreeVis that format changes.
* New report type (table/tree) selector component
* UI adjustments: 
	- hide uncategorized nodes
	- show non-specific data as in the report

# Notes

* The file: `app/assets/src/components/views/SampleViewV2/TaxonTreeVis.jsx` should be compared against `app/assets/src/components/views/TaxonTreeVis.jsx`. 
	- It made sense for the refactor to make this view a component of SampleViewV2
	- The previous file is going to be deprecated and removed upon release of the new report page

# Tests

* Interact with the page to check: 
	- information showing in the correct levels and matching the report table
	- collapse/expand behavior
	- on hover nodes behavior
	- on click taxon behavior
	- pathogen tags
	- highlighted branches match report

### DIFF
--- a/app/assets/src/components/visualizations/TidyTree.js
+++ b/app/assets/src/components/visualizations/TidyTree.js
@@ -85,27 +85,11 @@ export default class TidyTree {
   }
 
   setTree(nodes) {
-    console.log("nodes", nodes);
     let stratifier = stratify()
       .id(node => node.id)
       .parentId(node => node.parentId);
 
-    // const a = nodes.filter(node => node.id === "_" || node.parentId === "_");
-    // a.forEach(node => {
-    //   if (node.parentId === a[0].id) {
-    //     console.log("vvv HERE vvv");
-    //     console.log("vvv HERE vvv");
-    //     console.log("vvv HERE vvv");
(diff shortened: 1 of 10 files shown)


app/assets/src/components/visualizations/TidyTree.js
```diff
@@ -225,16 +225,19 @@ export default class TidyTree {
       this.options.onCollapsedStateChange(node);
   }
 
-  expandCollapsedWithFewChildren(node) {
+  expandCollapsedWithFewChildrenOrNoName(node) {
```
j-x-han: I'm not able to expand an "Uncategorized" node after collapsing it (although I checked the old version of the taxon tree and it also behaves poorly there.)

Example on sample 13423:

image

image

It seems to happen with some other nodes as well, but I was able to replicate it most consistently with the "Uncategorized" ones.

tfrcarvalho: Fixed.

PR #2813, opened 2019-12-04
Enable JS code splitting in dev

gregdingle:
# Description

This moves the prod webpack config into common so it is active in local dev.  The bundle splitting should help. 

# Notes

* Another benefit is the simplicity of more shared config so we will catch any config bugs sooner
* Minification is skipped with cheap source maps. See https://webpack.js.org/plugins/uglifyjs-webpack-plugin/#sourcemap . 

# Tests

delete all assets in `dist`
load site, see error
`npm start`
load site, see it working normally

add a `throw` in code
see that source maps work
image

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -56,6 +56,7 @@ Rails.application.configure do
   # Debug mode disables concatenation and preprocessing of assets.
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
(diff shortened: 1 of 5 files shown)


tfrcarvalho:
There is another metric you should look at, which is how long does webpack takes to regenerate the bundles in watch mode. 
I think that was the main reason the source map type (devtool) was changed.

PR #2814, opened 2019-12-05
Add combined sample taxon results bulk download.

MarkAZhang:
# Description

Add download type with UI and back-end code.

Screen Shot 2019-12-04 at 5 13 53 PM

The background field only shows up when a zscore metric is selected.

Screen Shot 2019-12-04 at 5 13 58 PM

# Notes
* Generalized code to handle conditional fields in bulk downloads.
* Changed heatmap results slightly to make it easier to link the returned pipeline runs with the provided samples.
* Generates a single .csv file.

# Tests
* Helper and model tests for back-end.
* Verified that bulk download produces expected results.
* Verify that the heatmap still produces the same results as before (since we changed the heatmap code slightly)
* Verify that conditional fields behave as expected, including validation and resetting of fields.

### DIFF
--- a/app/assets/src/components/views/bulk_download/ChooseStep.jsx
+++ b/app/assets/src/components/views/bulk_download/ChooseStep.jsx
@@ -9,7 +9,6 @@ import {
   map,
   isUndefined,
   orderBy,
-  isNumber,
   reject,
 } from "lodash/fp";
 import cx from "classnames";
@@ -23,6 +22,7 @@ import {
   getBackgrounds,
   getTaxaWithReadsSuggestions,
   uploadedByCurrentUser,
+  getHeatmapMetrics,
 } from "~/api";
 import PrimaryButton from "~/components/ui/controls/buttons/PrimaryButton";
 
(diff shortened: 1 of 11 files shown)


app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -201,23 +236,54 @@ class ChooseStep extends React.Component {
...
-    switch (field.type) {
-      case "file_format":
+    // Handle rendering conditional fields.
+
+    // For the file format field, render a placeholder. This is a special case.
+    const fileFormatConditionalField = CONDITIONAL_FIELDS[0];
+    if (
+      field.type === fileFormatConditionalField.field &&
+      downloadType.type === fileFormatConditionalField.downloadType &&
+      !fileFormatConditionalField.triggerValues.includes(
```
jshoe: maybe there could be a helper function that takes in `fileFormatConditionalField` / `conditionalField`  to evaluate these conditionals since it's a little repetitive

app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -201,23 +236,54 @@ class ChooseStep extends React.Component {
...
-          return (
-            <div className={cs.field} key={field.type}>
-              <div className={cs.label}>{field.display_name}:</div>
-              <div className={cs.forcedOption}>.fasta</div>
-              <div className={cs.info}>
-                Note: Only .fasta is available when selecting one taxon.
-              </div>
-            </div>
-          );
+          abortRenderOption = true;
```
jshoe: since you can `return` on the first `true`, you can actually just use `CONDITIONAL_FIELDS.every()` to break earlier, right?

app/helpers/bulk_downloads_helper.rb
```diff
@@ -56,4 +56,82 @@ def format_bulk_download(bulk_download, with_pipeline_runs = false)
     end
     formatted_bulk_download
   end
+
+  # This method models HeatmapHelper.sample_taxons_dict when fetching metrics. It uses the necessary logic from that function.
+  def self.generate_combined_sample_taxon_results_csv(samples, background_id, metric)
```
jshoe: are there other parts of this that could still be extracted into more shared helpers with `HeatmapHelper.sample_taxons_dict`? Just b/c these code paths tend to get out of sync very quickly

jshoe: or maybe moving just the metric formatting block into another function?

app/assets/src/components/views/bulk_download/ChooseStep.jsx
MarkAZhang: Yes! Thanks for the suggestion.

app/helpers/bulk_downloads_helper.rb
MarkAZhang: This function doesn't use that much from `sample_taxons_dict`, despite it being the inspiration for this function. I only use `taxon_counts_cleanup` and `only_species_level_counts`.

I did add some additional comments and move the metric calculation into a separate method.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
MarkAZhang: Did a fair amount of refactoring. Even simplified one of the props (`selectedFields`).

PR #2815, opened 2019-12-05
Add contigs non host single taxon bulk download.

MarkAZhang:
# Notes

* Also fix a bug with contigs non-host all taxon bulk download. Execution-type was being misclassified.

# Tests

* Added model test.
* Verified manually that bulk download works as expected.

### DIFF
--- a/app/models/bulk_download.rb
+++ b/app/models/bulk_download.rb
@@ -307,6 +307,10 @@ class BulkDownload < ApplicationRecord
                                     # We avoid pre-fetching "taxon_byteranges" here, even though it has to be fetched once per pipeline_run in the code below,
                                     # because of the potentially large number of taxon_byteranges per pipeline_run.
                                     pipeline_runs.includes(:sample)
+                                  elsif download_type == CONTIGS_NON_HOST_BULK_DOWNLOAD_TYPE
+                                    # Pre-fetching contigs here doesn't prevent multiple queries for contig in get_contigs_for_taxid.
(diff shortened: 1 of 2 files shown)


app/models/bulk_download.rb
```diff
@@ -361,6 +365,16 @@ def write_output_files_to_s3_tar_writer(s3_tar_writer)
               "reads_nonhost_#{get_param_display_name('taxa_with_reads')}.fasta",
             reads_nonhost_for_taxid_fasta
           )
+        elsif download_type == CONTIGS_NON_HOST_BULK_DOWNLOAD_TYPE
+          contigs = pipeline_run.get_contigs_for_taxid(get_param_value("taxa_with_contigs").to_i)
+          contigs_nonhost_for_taxid_fasta = ''
+          contigs.each { |contig| contigs_nonhost_for_taxid_fasta += contig.to_fa }
+
+          s3_tar_writer.add_file_with_data(
+            "#{get_output_file_prefix(pipeline_run.sample, cleaned_project_names)}" \
```
jshoe: maybe separate this line for readability?

spec/models/bulk_download_spec.rb
```diff
@@ -532,6 +532,23 @@ def create_taxon_lineage(taxid)
       expect(bulk_download.status).to eq(BulkDownload::STATUS_ERROR)
     end
 
+    it "correctly generates download file for download type contigs_non_host for single taxa" do
```
jshoe: do you mean `for a single taxon`?

PR #2817, opened 2019-12-05
[Report Page] Add instrumentation to old report page

tfrcarvalho:
# Description

* Add instrumentation to the old report page
* Before it was done adhoc directly on staging, but might be worth seeing a comparison in prod

# Tests

* Verified that the report page still works as expected.

### DIFF
--- a/app/lib/report_helper.rb
+++ b/app/lib/report_helper.rb
@@ -84,6 +84,14 @@ module ReportHelper
     path + "?" + kvs.to_param
   end
 
+  def self.instrumented_report_info_json(pipeline_run, background_id)
+    @timer = Timer.new("report_info")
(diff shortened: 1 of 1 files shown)


app/lib/report_helper.rb
```diff
@@ -739,35 +756,43 @@ def self.taxonomy_details(pipeline_run_id, background_id, scoring_model, sort_by
...
     # These counts are shown in the UI on each genus line.
     count_species_per_genus!(tax_2d)
+    @timer.split("count_species_per_genus")
 
     # Generate lineage info.
     unfiltered_ids = []
     tax_2d.each do |tid, _|
       unfiltered_ids << tid if tid > 0
     end
+    @timer.split("reformating")
```
j-x-han: nit: "reformatting"

PR #2820, opened 2019-12-06
[IDSEQ-1769] Add account activation emails instead of manual password resets + account not found email

jshoe:
### Description
- JIRA: https://jira.czi.team/browse/IDSEQ-1769
- One of the missing pieces of functionality in adding Auth0 users was needing to send a manual password reset email (to let them log in), which could have confusing wording to a brand new user. This just sends that automatically with better wording.
- The form will now have a `Send activation email` box that will be checked by default. The admin can still uncheck it if they'd rather send a personalized welcome later (and will manually send a password reset).
- Also bundled an extra improvement to send an `account not found` email when users try to `reset password` on a non-existent account.

### Notes
- Email wording/styling produced with Jenn/Olivia.
- Got an error saying I could only use inline styles on these emails, interestingly.
- The Auth0 login page "Forgot Password" links will now also link to the one on idseq-web. This was their recommended approach for adding more functionality like an "account not found" email handler because their API endpoint isn't designed to respond to account presence/absence. Couldn't find an Auth0 rule set that would do that.

Screen Shot 2019-12-04 at 4 29 33 PM

Screen Shot 2019-12-05 at 4 14 30 PM

Screen Shot 2019-12-05 at 4 12 57 PM

### Tests
- Tested the flows with/without admin, with/without activation emails, with/without user accounts present.

### DIFF
--- a/app/views/user_mailer/new_auth0_user_new_project.html.erb
+++ b/app/views/user_mailer/new_auth0_user_new_project.html.erb
@@ -5,7 +5,7 @@
 <p></p>
 <p><b><%= link_to 'Activate Your Account', @reset_password_url %></b></p>
 <p>Please note that this activation link expires after 48 hours.</p>
-<p>If you missed that window, please <%= link_to 'request a new password reset email', root_url + "users/password/new" %>.</p>
+<p>If the link has expired, please <%= link_to 'click here to request a new one', root_url + "users/password/new" %>.</p>
 <p>
   Thanks,<br/> The IDseq Team
 </p>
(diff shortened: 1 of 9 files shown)


app/views/user_mailer/new_auth0_user_new_project.html.erb
```diff
@@ -3,6 +3,9 @@
 <p><%= @sharing_user_name %> (<%= @sharing_user_email %>) has invited you to IDseq project
   <b><a href="https://idseq.net/my_data?projectId=<%=(@shared_project_id)%>"><%= @shared_project_name %></a></b>.</p>
 <p></p>
-<p><b><%= link_to 'Activate your account', @reset_password_url %></b></p>
+<p><b><%= link_to 'Activate Your Account', @reset_password_url %></b></p>
```
jshoe: Updated this one to be consistent with the design choices we agreed on, like capitalizing "Activate Your Account", changing wording.

PR #2822, opened 2019-12-06
Update ecs task name for bulk downloads.

MarkAZhang:
# Description

By using the same task name for all bulk download tasks, we don't create a new ecs log group for every bulk download task (which is how aegea currently uses task names)

Note that we now create a separate log stream inside the same log group (bulk_downloads):

Screen Shot 2019-12-05 at 6 52 15 PM

The task name is also used for other things, such as the task definition name. Now, each new bulk download adds a "revision" to the task definition, which works fine.

Screen Shot 2019-12-05 at 6 52 27 PM

# Notes
* Also added a `log_stream_name` function for convenience.

# Tests

* Verified the above behaviors manually.


MarkAZhang:
Also cleaned up all the `bulk_downloads_X` in both the log groups and the ecs task definitions.

PR #2823, opened 2019-12-06
IDSEQ-1883 - Remove some bad test patterns

davidrissato:
# Description

There were multiple test cases invoking Devise sign_in endpoint directly, instead of using a centralized helper. That was making extremely hard to make changes to the authentication mechanism due to all rippling effects in our test code.

Additionally, the user names that are being used on our test fixtures are very simple, like `:one` and `:two`, hiding important details about their roles as admins.

Changes:
- Tests are now invoking a test helper method to sign in users, instead of direct invoking Devise sign_in endpoint
- Change fixtures user names to something more expressive like `:admin_one` and `:regular_user`.

# Notes

I also noticed most tests are using user `:one`, an admin, even for operations that are not restricted to admins. Probably unintentional (this is why I renamed the user).  But that means these test case scenarios are not protecting us about errors users may eventually see.

# Tests

* Automated tests


test/controllers/admin_controller_test.rb
```diff
@@ -2,8 +2,8 @@
 
 class AdminControllerTest < ActionDispatch::IntegrationTest
   setup do
-    @user = users(:one)
-    post user_session_path, params: { 'user[email]' => @user.email, 'user[password]' => 'password' }
+    @edited_user = users(:regular_user)
```
MarkAZhang: Nit: could we re-name `edited_user` to `user_to_modify`? Modify since we are both editing and deleting, and using the future tense since we will be modifying in the test.

test/controllers/application_controller_test.rb
```diff
@@ -2,17 +2,16 @@
 
 class ApplicationControllerTest < ActionDispatch::IntegrationTest
   setup do
-    @user = users(:one)
-    @user_params = { 'user[email]' => @user.email, 'user[password]' => 'password' }
+    @user = users(:admin_one)
```
MarkAZhang: Is there a reason to do `@user = users(:admin_one)` and then later `sign_in @user` instead of just `sign_in :admin_one` in one line?

If there is a reason to do it in two lines, perhaps we can re-name `@user` to `@user_admin` for clarify in the test.

test/test_helper.rb
```diff
@@ -15,7 +15,9 @@ def access_sample_with_background(background, sample)
   end
 
   def sign_in(user)
-    @user = users(user)
-    post user_session_path, params: { 'user[email]' => @user.email, 'user[password]' => 'password' }
+    unless user.instance_of? User
```
MarkAZhang: Is there a reason we can't require `sign_in` to be called in the same everywhere? Either

```
user = users(:admin_one)
sign_in(user)
```

or 

```
sign_in :admin_one
```

instead of allowing both?

davidrissato:
@MarkAZhang asked:
> I wonder why you can't just require all callers to do sign_in :admin_one in one line.
We can do that, but I didn't want to enter in details on how these tests are implemented right now, so I kept the previous behavior. Also, I see advantages on having the user in a variable as a DRY pattern.

> I would rename @user to @user_admin if you continue to do it in two lines.
I would say most tests are using an admin indiscriminately, when in fact that should be tested with a regular user. Renaming @user to @user_admin would just reinforce this idea.
So I just renamed these users for now to make it more evident, but reviewing all tests should be a scope for a separate ticket.

test/controllers/admin_controller_test.rb
davidrissato: done

davidrissato:
@gregdingle wrote:
> LGTM.
> 
> Good explanation and notes. The users "one" and "two" had been bothering me as well. And how about the user "joe"? Do we like joe?

I heard this morning "joe" is a well known "regular user" profile. I personally think `sign_in :regular_user` to be more expressive on this intent, but I can be flexible on that. For me, that `:one` user being an admin was the most critical issue.

tfrcarvalho:
> So I just renamed these users for now to make it more evident, but reviewing all tests should be a scope for a separate ticket.

Reviews of tests should also focus on moving these to rspec.

PR #2826, opened 2019-12-06
Add max sample app config field for bulk downloads.

MarkAZhang:
# Description

Adds value in AppConfig for controlling the max samples per bulk download.

If user exceeds max samples, when they try to click on the button, they get a temporary tooltip:

Screen Shot 2019-12-05 at 8 38 20 PM

Also added validation on the back-end. The user should never get this message (I had to disable the temporary tooltip above to get it), but it verifies that the back-end validation is working.

Screen Shot 2019-12-05 at 8 50 23 PM

# Notes

* Also removed the double download icons on SamplesView, in preparation for launch.

# Tests

* Verified manually that front-end and back-end validation behave as expected.
* Added controller test.

### DIFF
--- a/app/assets/src/components/views/samples/SamplesView.jsx
+++ b/app/assets/src/components/views/samples/SamplesView.jsx
@@ -474,7 +474,14 @@ class SamplesView extends React.Component {
 
   handleBulkDownloadModalOpen = () => {
     const { maxSamplesBulkDownload } = this.context || {};
-    if (this.props.selectedSampleIds.size > parseInt(maxSamplesBulkDownload)) {
+    if (!maxSamplesBulkDownload) {
+      this.setState({
+        bulkDownloadButtonTempTooltip:
+          "Unexpected issue. Please contact us for help.",
+      });
+    } else if (
+      this.props.selectedSampleIds.size > parseInt(maxSamplesBulkDownload)
+    ) {
       this.setState({
         bulkDownloadButtonTempTooltip: `No more than ${maxSamplesBulkDownload} samples allowed in one download.`,
       });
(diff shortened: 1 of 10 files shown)


app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -459,7 +473,14 @@ class SamplesView extends React.Component {
   };
 
   handleBulkDownloadModalOpen = () => {
-    this.setState({ bulkDownloadModalOpen: true });
+    const { maxSamplesBulkDownload } = this.context || {};
+    if (this.props.selectedSampleIds.size > parseInt(maxSamplesBulkDownload)) {
+      this.setState({
+        bulkDownloadButtonTempTooltip: `No more than ${maxSamplesBulkDownload} samples allowed in one download.`,
```
jshoe: this might end up trying to compare `> NaN` and being false, right?

app/controllers/bulk_downloads_controller.rb
```diff
@@ -23,6 +24,14 @@ def create
     # Convert sample ids to pipeline run ids.
     begin
       sample_ids = create_params[:sample_ids]
+
+      # Max samples check.
+      max_samples_allowed = get_app_config(AppConfig::MAX_SAMPLES_BULK_DOWNLOAD)
+
+      if sample_ids.length > Integer(max_samples_allowed)
```
jshoe: I think this is in danger of TypeError for `Integer(nil)` as well?

app/assets/src/components/views/samples/SamplesView.jsx
MarkAZhang: The value should be set, but I've added a guard just in case.

app/controllers/bulk_downloads_controller.rb
MarkAZhang: Same here.

PR #2827, opened 2019-12-06
[Release Fix] Fix pipeline run status check

j-x-han:
# Description

The pipeline run status check was incorrect, resulting in an error rather than displaying an alert warning the user that their pipeline run had failed when attempting to view the report page.

# Notes

Related JIRA task for future cleanup of pipeline run statuses: https://jira.czi.team/browse/IDSEQ-1890

# Tests

* Verified that samples with failed pipeline runs now correctly display an alert rather than causing the page to error and crash.

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
+++ b/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
@@ -965,7 +965,10 @@ export default class SampleViewV2 extends React.Component {
       selectedOptions,
       view,
     } = this.state;
-    if (reportMetadata.pipelineRunStatus === "COMPLETE") {
+    if (
+      reportMetadata.pipelineRunStatus === "COMPLETE" &&
+      !reportMetadata.hasErrors
+    ) {
       return (
         <div className={cs.reportViewContainer}>
           <div className={cs.reportFilters}>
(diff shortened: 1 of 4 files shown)


app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -965,7 +965,10 @@ export default class SampleViewV2 extends React.Component {
       selectedOptions,
       view,
     } = this.state;
-    if (reportMetadata.pipelineRunStatus === "COMPLETE") {
+    if (
+      reportMetadata.pipelineRunStatus === "COMPLETE" &&
+      !reportMetadata.hasErrors
```
tfrcarvalho: We don't want to do this because on experimental errors. In that case we need to show the report.

app/services/pipeline_report_service.rb
```diff
@@ -236,6 +236,13 @@ def get_pipeline_status(pipeline_run_id)
     ]
   end
 
+  def pipeline_run_succeeded?(pipeline_run)
+    # This condition is carried over from the previous version of the report page (report_helper.rb ln95).
```
tfrcarvalho: nit: remove the line... people should be able to track it from the git history. The line will become obsolete with any changes to the other file.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
j-x-han: Replaced with `pipelineRunReportAvailable`

app/services/pipeline_report_service.rb
j-x-han: Done

PR #2830, opened 2019-12-06
Reduce fetches of unmapped reads count

gregdingle:
# Description

The new refined unmapped count is working in staging, but it is fetching from s3 way more than needed. The result monitor fetched the same file over 100 times. 

After discussion with @boris-dimitrov , I decided to limit the fetching to the stage during which the file is created. 

# Notes

I also added a bunch of TODOs related to https://jira.czi.team/browse/IDSEQ-1912. 

# Tests

In rails console:

Open a stage 3 pr run stage.  
Run `fetch_unmapped_reads`
See fetch .

Open a stage 4 pr run stage.  
Run `fetch_unmapped_reads`
See no fetch.

On deploy to staging, check frequency in logs.

`aegea grep Fetch --start-time=-1d ecs-logs-staging`

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -1098,7 +1098,7 @@ class PipelineRun < ApplicationRecord
       unmapped_reads = unidentified[:reads_after]
 
       # TODO: (gdingle): rename to stage_number. See https://jira.czi.team/browse/IDSEQ-1912.
-      if supports_assembly? && active_stage.step_number == 3 # assembly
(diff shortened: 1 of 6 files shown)


app/models/pipeline_run.rb
```diff
@@ -333,6 +334,7 @@ def completed?
   end
 
   def active_stage
+    # TODO: (gdingle):
```
jshoe: you forgot the comment on this one

gregdingle: good catch!

PR #2831, opened 2019-12-07
Protect omniauth_failure endpoint from missing params

jshoe:
### Description
- This should never happen in a deployed env but was encountered in local development. Fixes a potential `nil.to_sym` call which would throw an exception.

### Notes
- In this case, the user will be directed to the `failure` endpoint and logged out.

# Tests
- Visit a link such as: `http://localhost:3000/auth/auth0/callback` and see the redirection.

### DIFF
--- a/app/controllers/auth0_controller.rb
+++ b/app/controllers/auth0_controller.rb
@@ -47,6 +47,9 @@ class Auth0Controller < ApplicationController
   # Handle omniauth errors coming from Auth0.
   def omniauth_failure
     # Error and error_description come from Auth0. Ex: unauthorized and password_expired.
+    unless params["error"] && params["error_description"]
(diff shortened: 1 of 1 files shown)


app/controllers/auth0_controller.rb
```diff
@@ -47,8 +47,8 @@ def failure
   # Handle omniauth errors coming from Auth0.
   def omniauth_failure
     # Error and error_description come from Auth0. Ex: unauthorized and password_expired.
-    error_type = params["error"].to_sym
-    error_code = params["error_description"].to_sym
+    error_type = (params["error"] || "").to_sym
+    error_code = (params["error_description"] || "").to_sym
```
tfrcarvalho: Should we add our own error message in this case?

PR #2833, opened 2019-12-09
[Report Page] Add caching to new report page

j-x-han:
# Description

Adds caching for the new pipeline report page.

Follows the caching pattern used in the previous report: once a pipeline run has finished, reports against all backgrounds will be precached.

# Notes

`PipelineReportService.call` was modified to take a `pipeline_run` instead of a `pipeline_run.id` to reduce some redundancy in the code. Calls to the function were updated accordingly.

When porting the cache-related code in `report_info` into `report_v2`, calls to `MetricUtil.put_metric_now` were replaced with `MetricUtil.log_analytics_event`.

# Tests

* Enabled caching in development and verified that the report information would be fetched from the cache rather than re-computed from scratch.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -173,4 +173,25 @@ class ApplicationController < ActionController::Base
     yield
     @timer.publish
   end
+
+  def fetch_from_cache(skip_cache, cache_key, httpdate, event_name)
+    if skip_cache
+      yield
+    else
+      MetricUtil.log_analytics_event(event_name + "_cache_requested", current_user) unless skip_cache
+      # This allows 304 Not Modified to be returned so that the client can use its
+      # local cache and avoid the large download.
+      response.headers["Last-Modified"] = httpdate
+      # This is a custom header for testing and debugging
+      response.headers["X-IDseq-Cache"] = 'requested'
+      response.headers["X-IDseq-Cache-Key"] = cache_key
+      Rails.logger.info("Requesting #{cache_key}")
+
+      Rails.cache.fetch(cache_key, expires_in: 30.days) do
+        MetricUtil.log_analytics_event(event_name + "_cache_missed", current_user)
+        response.headers["X-IDseq-Cache"] = 'missed'
+        yield
+      end
+    end
+  end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -178,7 +178,7 @@ class ApplicationController < ActionController::Base
     if skip_cache
       yield
     else
-      MetricUtil.log_analytics_event(event_name + "_cache_requested", current_user) unless skip_cache
+      MetricUtil.log_analytics_event(event_name + "_cache-requested", current_user) unless skip_cache
       # This allows 304 Not Modified to be returned so that the client can use its
       # local cache and avoid the large download.
       response.headers["Last-Modified"] = httpdate
@@ -188,7 +188,7 @@ class ApplicationController < ActionController::Base
       Rails.logger.info("Requesting #{cache_key}")
 
       Rails.cache.fetch(cache_key, expires_in: 30.days) do
-        MetricUtil.log_analytics_event(event_name + "_cache_missed", current_user)
+        MetricUtil.log_analytics_event(event_name + "_cache-missed", current_user)
         response.headers["X-IDseq-Cache"] = 'missed'
         yield
       end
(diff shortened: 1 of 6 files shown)


app/controllers/samples_controller.rb
```diff
@@ -800,19 +800,39 @@ def samples_going_public
...
-          jobStatus: "Waiting to Start or Receive Files",
-          errorMessage: nil,
-          knownUserError: nil,
-        },
-      }
-    end
+    background_id = get_background_id(@sample, params[:background])
+
+    skip_cache = params[:skip_cache] || false
+    MetricUtil.log_analytics_event("samples.cache.requested", current_user) unless skip_cache
```
jshoe: For the `log_analytics_event` calls, could you changes the names to the other format? https://czi.quip.com/67RCAIiHN0Qc/IDseq-product-analytics-How-to-log

app/models/pipeline_run.rb
```diff
@@ -1577,6 +1578,22 @@ def precache_report_info!
...
+      cache_key = PipelineReportService.report_info_cache_key(
+        "/samples/#{sample.id}/report_v2",
+        params.merge(background_id: background_id)
+      )
+      Rails.logger.info("Precaching #{cache_key} with background #{background_id}")
+      Rails.cache.fetch(cache_key, expires_in: 30.days) do
+        PipelineReportService.call(self, background_id)
+      end
+
+      MetricUtil.log_analytics_event("samples.cache.precached", current_user)
```
jshoe: also not sure who `current_user` would be in this example, can probably leave it blank

app/controllers/samples_controller.rb
```diff
@@ -800,19 +800,39 @@ def samples_going_public
...
+    report_info_params = pipeline_run.report_info_params
+
+    cache_key = PipelineReportService.report_info_cache_key(
+      request.path,
+      params
+        .permit(report_info_params.keys)
+        .merge(pipeline_run_id: pipeline_run.id)
+    )
+
+    json =
```
jshoe: It feels hard to read this conditional assignment with so much in the `else` block. did Rubocop force this? Maybe have it as something like `unless skip_cache, compute all the headers`?

app/controllers/samples_controller.rb
j-x-han: Changed!

app/models/pipeline_run.rb
j-x-han: `log_analytics_event` requires a user to be passed in, so I replaced it with `nil`.

app/controllers/samples_controller.rb
j-x-han: At @tfrcarvalho 's request, I pulled the cache logic out into its own function so the block should be cleaner now!

app/services/pipeline_report_service.rb
```diff
@@ -67,12 +67,13 @@ def call
   end
 
   def generate
-    pipeline_run, metadata = get_pipeline_status(@pipeline_run_id)
+    pipeline_run, metadata = get_pipeline_status(@pipeline_run)
```
tfrcarvalho: This makes no sense if you are sending in a pipeline_run now. We are returning a pipeline run and sending it as input?

app/controllers/samples_controller.rb
```diff
@@ -800,19 +800,23 @@ def samples_going_public
...
+    cache_key = PipelineReportService.report_info_cache_key(
+      request.path,
+      params
+        .permit(report_info_params.keys)
+        .merge(pipeline_run_id: pipeline_run.id)
+    )
+    httpdate = Time.at(report_info_params[:report_ts]).utc.httpdate
+
+    json =
+      fetch_from_cache(skip_cache, cache_key, httpdate, "PipelineReport") do
```
tfrcarvalho: `fetch_from_or_store_in_cache`?

app/controllers/application_controller.rb
```diff
@@ -173,4 +173,25 @@ def instrument_with_timer
     yield
     @timer.publish
   end
+
+  def fetch_from_cache(skip_cache, cache_key, httpdate, event_name)
```
tfrcarvalho: Add a comment explaining this method.

app/services/pipeline_report_service.rb
j-x-han: Removed pipeline_run as an output.

app/controllers/samples_controller.rb
j-x-han: Done.

app/controllers/application_controller.rb
j-x-han: Done.

PR #2834, opened 2019-12-09
Remove remaining references to old sample metadata fields.

MarkAZhang:
# Description

Remove remaining references to fields, in preparation for removing the fields via migration.

# Notes

* Migration will be in follow-up PR, which will wait until the next deploy.

# Tests

* Verify that the Phylo Tree creation modal works with the new code.

Screen Shot 2019-12-09 at 12 23 33 PM

* Verify that you can still upload samples and update the sample name and sample notes via the sidebar.

* Migrated any remaining values in these fields in prod to the Metadata table. See: https://czi.quip.com/v6waAR0JgdLZ/Old-sample-metadata-migration#JSKACAZ6OCT

### DIFF
--- a/app/controllers/phylo_trees_controller.rb
+++ b/app/controllers/phylo_trees_controller.rb
@@ -288,7 +288,7 @@ class PhyloTreesController < ApplicationController
       end
       if metadata_by_sample_id[sp["sample_id"]]
         sp["sample_type"] = metadata_by_sample_id[sp["sample_id"]][:sample_type]
-        sp["collection_location"] = metadata_by_sample_id[sp["sample_id"]][:collection_location_v2] || metadata_by_sample_id[sp["sample_id"]][:collection_location]
+        sp["collection_location"] = metadata_by_sample_id[sp["sample_id"]][:collection_location_v2]
       end
     end
 
--- a/app/controllers/phylo_trees_controller.rb
+++ b/app/controllers/phylo_trees_controller.rb
@@ -259,12 +259,11 @@ class PhyloTreesController < ApplicationController
       select
         samples.name,
         samples.project_id,
-        samples.sample_tissue,
-        samples.sample_location,
         samples.created_at,
         host_genomes.name as host,
         projects.name as project_name,
-        pipeline_runs.id as pipeline_run_id
+        pipeline_runs.id as pipeline_run_id,
+        samples.id as sample_id
       from pipeline_runs, projects, samples, host_genomes
       where
         pipeline_runs.id in (#{pipeline_run_ids.join(',')}) and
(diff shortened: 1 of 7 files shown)


app/controllers/phylo_trees_controller.rb
```diff
@@ -278,12 +277,19 @@ def sample_details_json(pipeline_run_ids, taxid)
...
+
     samples_projects.each do |sp|
       sp["taxid_reads"] ||= {}
       %w[NT NR].each do |count_type|
         key = "#{sp['pipeline_run_id']},#{count_type}"
         sp["taxid_reads"][count_type] = (taxon_counts[key] || []).count # count is a column of taxon_counts indicating number of reads
       end
+      if metadata_by_sample_id[sp["sample_id"]]
+        sp["sample_type"] = metadata_by_sample_id[sp["sample_id"]][:sample_type]
+        sp["collection_location"] = metadata_by_sample_id[sp["sample_id"]][:collection_location_v2] || metadata_by_sample_id[sp["sample_id"]][:collection_location]
```
MarkAZhang: @jshoe Is this the right way to fetch the location? I noticed that there were old samples that had a value for collection_location but not for collection_location_v2, which is why I did it like this.

lib/tasks/pipeline_monitor.rake
```diff
@@ -197,7 +197,6 @@ class CheckPipelineRuns
       project_id: bm_project.id,
       user_id: bm_user.id,
       input_files_attributes: input_files_attributes,
-      sample_organism: known_organisms,
```
MarkAZhang: Interestingly, the benchmark sample still adds known_organisms into sample_organism. This information is also added to `sample_notes` (in a different format) so it seemed safe to just remove this.

app/helpers/samples_helper.rb
```diff
@@ -62,38 +62,6 @@ def generate_sample_list_csv(formatted_samples)
     end
   end
 
-  # Load bulk metadata from a CSV file
-  # NOTE: Succeeded by Metadatum#load_csv_from_s3 for new Metadatum objects.
-  def populate_metadata_bulk(csv_s3_path)
```
gregdingle: yay 👏

app/controllers/phylo_trees_controller.rb
jshoe: The `by_sample_ids` should be correct because it'll also grab a location field value. However, you should only have to fetch the `collection_location_v2` field. Do you want to send me the examples that seemed to be missing `collection_location_v2`?

jshoe:
looks like there was a RuboCop nit in Travis btw

PR #2835, opened 2019-12-10
[IDSEQ-1751] Fix case-insensitivity in map marker clicks on locations

jshoe:
### Description
- There was a small bug where the comparisons on discovery map marker sidebar loads were not case sensitive, so "Japan" and a plain text entry "japan" would not be distinguished, and the sidebar counts wouldn't match the marker.

### Notes
- This adds the 'BINARY' keyword to do a case-sensitive comparison instead. There should not be a noticeable perf difference because the previous string comparison was not indexed either.
- The alternative approach would be having a stronger comparison on just the location_id, but the code for sidebar loading also accepts plain text for filtering by the user-inputted plain text values so that would be a larger change in design.

### Tests
- See screenshots here: https://docs.google.com/document/d/1tVt8gk4LnFR-RvYu0eGa97IgKkpb0i6k3bLhtomVV9E/edit

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -148,8 +148,8 @@ class SamplesController < ApplicationController
 
   def index_v2
     # this method is going to replace 'index' once we fully migrate to the
-    # discovery views (old one was kept to avoid breaking the current interface
-    # without sacrificing speed of development)
+    # discovery views (old one was kept to avoid breaking the current inteface
+    # without sacrificing speed of development and avoid breaking the current interface)
     domain = params[:domain]
     order_by = params[:orderBy] || :id
     order_dir = params[:orderDir] || :desc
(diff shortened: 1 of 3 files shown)


app/controllers/samples_controller.rb
```diff
@@ -148,7 +148,7 @@ def index
 
   def index_v2
     # this method is going to replace 'index' once we fully migrate to the
-    # discovery views (old one was kept to avoid breaking the current inteface
+    # discovery views (old one was kept to avoid breaking the current interface
     # without sacrificing speed of development and avoid breaking the current interface)
```
tfrcarvalho: nit: since you are fixing comments, can you remove the last "and avoid breaking..." since its repeated?

jshoe: done 👍

PR #2836, opened 2019-12-10
[Bug fix] Add nil check to cleanup_missing_genus_counts

j-x-han:
# Description

Discovered a bug where `cleanup_missing_genus_counts` would encounter a nil error if there are no `species_counts`.

Example: https://idseq.net/samples/33729/ has 2 adjusted remaining reads. The legacy page https://idseq.net/samples/33729/legacy displays an empty table, but the new version of the report page "loads" forever because of the backend error.

A check was added to avoid the nil error.

# Tests

* Verified that the report won't load forever/error on the backend if `cleanup_missing_genus_counts` receives `nil` as an input.

### DIFF
--- a/app/helpers/reports_helper.rb
+++ b/app/helpers/reports_helper.rb
@@ -85,12 +85,14 @@ module ReportsHelper
     fake_genera = []
     missing_genera = Set.new
     taxids_with_missing_genera = Set.new
-    species_counts.each do |tax_id, tax_info|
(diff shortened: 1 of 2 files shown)


j-x-han:
Turns out the nil error would also occur in logic later on in the pipeline_report_service, so I moved the nil check up in the logic.

@tfrcarvalho 

image

PR #2842, opened 2019-12-11
IDSEQ-1558 - Implement sign_in metrics

davidrissato:
# Description

IDSEQ-1558
Implement database sign_in metrics to match previous Devise behavior

# Notes

I also fixing a tiny in auth0 helper that occurred when refreshing failed callback url and warden env was not set. That issue was probably not affecting any users, but it was annoying during my tests.

# Tests

* Manual and automated test

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -120,6 +120,27 @@ class User < ApplicationRecord
     projects.exists?(project_id)
   end
 
+  # Update login trackable fields
+  def update_tracked_fields!(request)
+    # This method has been adapted from Devise trackable module to keep previous behavior (IDSEQ-1558 / IDSEQ-1720)
+    # See: https://github.com/plataformatec/devise/blob/715192a7709a4c02127afb067e66230061b82cf2/lib/devise/models/trackable.rb#L20
+    return if new_record?
+
+    old_current = current_sign_in_at
+    new_current = Time.now.utc
+    self.last_sign_in_at     = old_current || new_current
+    self.current_sign_in_at  = new_current
+
+    old_current = current_sign_in_ip
+    new_current = request.remote_ip
+    self.last_sign_in_ip     = old_current || new_current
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -120,6 +120,27 @@ def owns_project?(project_id)
...
+    self.current_sign_in_at  = new_current
+
+    old_current = current_sign_in_ip
+    new_current = request.remote_ip
+    self.last_sign_in_ip     = old_current || new_current
+    self.current_sign_in_ip  = new_current
+
+    self.sign_in_count ||= 0
+    self.sign_in_count += 1
+    save(validate: false)
```
jshoe: since you named it as `update_tracked_fields!`, do you also want `save` to be `save!` (?) or is that kind of dangerous

davidrissato: TL;DR: I believe we should keep this way.

Long answer:
I noticed that when I borrowed this logic from Devise.

I actually tried to be as close as I could to the original implementation (I even used the same method name). I believe the idea behind that `!` is there to indicate this method saves the record when invoked, not necessarily the way it fails.

Actually, I don't fully agree with their implementation. If I were to implement these counters from scratch I would rather have created a separate table to keep them. That's because adding fields to the user object model may have a lot of undesired side effects. For example, user model may have save hooks attached that could eventually fail, or be triggered unnecessarily. Also, the fact these fields are part of the user model affects the value of `updated_at` field, which is totally wrong in my opinion. You can even notice the `validate: false` is in their solution to mitigate some of the inconveniences that wouldn't exist if these values were stored in a separate table.

Also these counters are not exactly super precise (it doesn't take into account users returning to the system in the next day, for example). We may want to review this logic in the future.

For all these reasons, I believe we should just pass if `save` fails.

jshoe: thanks for explaining! 👍

PR #2843, opened 2019-12-11
[Release fix] Fix caching for new report page

j-x-han:
# Description

If a user visited the report page of a sample whose pipeline run was still in progress, the in-progress report would be cached. The `cache_key` was also being computed incorrectly, resulting in that user always seeing the in progress report even after the pipeline run was completed.

This fix ensures that the report won't be cached until the pipeline run is ready and fixes the `cache_key` computation.

# Tests

* Cleared local cache and checked that in-progress samples would not have their reports cached.
* Cleared local cache and checked that the requested cache_keys were computed as expected.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -803,7 +803,7 @@ class SamplesController < ApplicationController
     background_id = get_background_id(@sample, params[:background])
 
     # If the pipeline run hasn't completed yet, don't cache the report
-    pipeline_run_completed = pipeline_run.all_output_states_terminal? && pipeline_run.all_output_states_loaded? && !pipeline_run.compiling_stats_failed
+    pipeline_run_completed = pipeline_run.all_output_states_terminal? && pipeline_run.all_output_states_loaded?
     skip_cache = params[:skip_cache] || !pipeline_run_completed || false
     report_info_params = pipeline_run.report_info_params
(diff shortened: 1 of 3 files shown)


jshoe:
looks like snyk had an error. you can try repushing or rebasing to see if that fixes it

app/controllers/samples_controller.rb
```diff
@@ -802,13 +802,25 @@ def report_v2
     pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
     background_id = get_background_id(@sample, params[:background])
 
-    skip_cache = params[:skip_cache] || false
+    # Don't cache the response until the pipeline run is report-ready
+    # so the displayed pipeline run status will be updated correctly.
+    skip_cache = !pipeline_run.report_ready? || params[:skip_cache] || false
```
tfrcarvalho: Do you think we still need this with the report ts fix?

tfrcarvalho: I guess we can keep it.

j-x-han: The `report_ts` fix should prevent the report from always showing the same loading screen, but I don't think it makes much sense to cache the report unless it's ready anyways.

tfrcarvalho: Let's keep it.

PR #2846, opened 2019-12-12
[Hot fix] Truncate long file names for bulk downloads.

MarkAZhang:
# Description

Turns out there are restrictions to the length of file names when using ruby's tar writer library.
The max file name is 100 chars (including extensions), and the max file path is 256 chars.
See: https://github.com/rubygems/rubygems/issues/1376

This PR re-organizes the file name and truncates where necessary.

# Notes

* Previously, the project and sample name were both in the file name and all files were in the root directory. This is changed so that files now go in a directory with the project name.
* The sample name has the biggest restriction. We truncate the sample name to 65 chars, leaving 10 chars for the sample id (which ensures the file name is unique). We also include a file suffix which is 25 chars, for a total of 100 chars.
* The file suffix is important because it describes the contents of the file. It is something like "reads_nh.fasta" (nh for nonhost). It includes the file extension.
* Some stats: The average sample name is 27 characters. 4% of samples have names longer than 65 chars.

* Will add tests that verify that the file suffix is 25 chars or less for all bulk downloads in a subsequent PR. Requires a little refactoring, and didn't want to include that in the hot fix.

# Tests

* Verified that bulk download succeeds for a sample with a very long name.
* Updated tests.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -61,7 +61,7 @@ class Project < ApplicationRecord
   end
 
   def cleaned_project_name
-    "project-#{name.downcase.split(' ').join('_')}"
+    name.downcase.split(' ').join('_').to_s
   end
 
   def host_gene_counts_from_params(params)
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -61,7 +61,7 @@ class Project < ApplicationRecord
   end
 
   def cleaned_project_name
-    name.downcase.split(' ').join('_').to_s
+    name.downcase.split(' ').join('_')
   end
 
   def host_gene_counts_from_params(params)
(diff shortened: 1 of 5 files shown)


app/models/project.rb
```diff
@@ -61,7 +61,7 @@ def samples
   end
 
   def cleaned_project_name
-    "project-#{name.downcase.split(' ').join('_')}"
+    name.downcase.split(' ').join('_').to_s
```
tfrcarvalho: nit: not worth maintaining a prefix to indicate a project? e.g. `pr`?

tfrcarvalho: Humm i guess it makes sense for any prefixes to be added outside.

app/models/project.rb
```diff
@@ -61,7 +61,7 @@ def samples
   end
 
   def cleaned_project_name
-    "project-#{name.downcase.split(' ').join('_')}"
+    name.downcase.split(' ').join('_')
```
MarkAZhang: The `project-` prefix seems unnecessary and just makes the name less readable. This cleaned project name is only being used in a couple of places, to generate the file name in a couple of single sample downloads.

PR #2859, opened 2019-12-12
[Hot fix] Fix call to PipelineReportService in sample taxon report download..

MarkAZhang:
# Description

The PipelineReportService API has been changed so that you pass in a PipelineRun object instead of a pipeline run id. However, this API change is currently in staging, not prod. My recent hot fix used the new API, so the hot fix failed on prod.

This hot fix only needs to be applied to prod, not to staging or master (because the API is correct in those cases).

This is a case where doing the hot fix on prod branch first may have caught this issue.


MarkAZhang:
Going to go ahead and merge/deploy this so it's ready by the 10am user test.

PR #2865, opened 2019-12-13
[Report page] Change contigs query to use new columns

j-x-han:
# Description

Update the contigs query used for the new report page to use the new columns (species_taxid_nt, species_taxid_nr, genus_taxid_nt, genus_taxid_nr) rather than lineage_json.

Based on a few local tests, this cuts the time for the query into roughly  third of what it was previously.

# Tests

* Verified that the contigs are still calculated and displayed as expected.

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -1444,12 +1444,12 @@ class PipelineRun < ApplicationRecord
   end
 
   def get_summary_contig_counts_v2(min_contig_size)
-    # key: taxid:count_type, value: read_counts: count
+    # Stores the number of contigs that match a given taxid, count_type (nt or nr), and read_count (number of reads aligned to that contig).
     # Create and store default values for the hash if the key doesn't exist yet
     summary_dict = Hash.new do |summary, taxid|
-      summary[taxid] = Hash.new do |taxid_, count_type| # rubocop forces different variable names
-        taxid_[count_type] = Hash.new do |count_type_, read_count|
-          count_type_[read_count] = 0
+      summary[taxid] = Hash.new do |taxid_hash, count_type| # rubocop forces different variable names
+        taxid_hash[count_type] = Hash.new do |count_type_hash, read_count|
+          count_type_hash[read_count] = 0
         end
       end
     end
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -1444,23 +1444,25 @@ class PipelineRun < ApplicationRecord
   end
 
   def get_summary_contig_counts_v2(min_contig_size)
-    summary_dict = {} # key: count_type:taxid , value: contigs, contig_reads
-    contig_taxids = contigs.select("read_count, species_taxid_nt, species_taxid_nr, genus_taxid_nt, genus_taxid_nr")
-                           .where("read_count >= ?", min_contig_size)
-                           .where("lineage_json IS NOT NULL")
-    contig_taxids.each do |c|
-      count_types_to_taxids = {
-        "nt": [c.species_taxid_nt, c.genus_taxid_nt].compact,
-        "nr": [c.species_taxid_nr, c.genus_taxid_nr].compact,
-      }
-      count_types_to_taxids.each do |count_type, taxids|
-        taxids.each do |taxid|
-          summary_per_tax = summary_dict[taxid] ||= {}
-          summary_per_tax_and_type = summary_per_tax[count_type.downcase] ||= {}
-          summary_per_tax_and_type[c.read_count] ||= 0
-          summary_per_tax_and_type[c.read_count] += 1
+    # key: taxid:count_type, value: read_counts: count
+    # Create and store default values for the hash if the key doesn't exist yet
+    summary_dict = Hash.new do |summary, taxid|
+      summary[taxid] = Hash.new do |tax_id, count_type| # rubocop forces different variable names
+        tax_id[count_type] = Hash.new do |counttype, read_count|
+          counttype[read_count] = 0
         end
       end
+    end
+    contig_taxids = contigs.where("read_count >= ?", min_contig_size)
+                           .where("lineage_json IS NOT NULL")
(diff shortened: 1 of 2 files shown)


app/models/pipeline_run.rb
```diff
@@ -1445,10 +1445,15 @@ def get_summary_contig_counts(min_contig_size)
 
   def get_summary_contig_counts_v2(min_contig_size)
     summary_dict = {} # key: count_type:taxid , value: contigs, contig_reads
-    contig_lineages(min_contig_size).each do |c|
-      lineage = JSON.parse(c.lineage_json)
-      lineage.each do |count_type, taxid_arr|
-        taxids = taxid_arr[0..1]
+    contig_taxids = contigs.select("read_count, species_taxid_nt, species_taxid_nr, genus_taxid_nt, genus_taxid_nr")
```
tfrcarvalho: You might want to try pluck instead of select.
Creating the activerecord model object takes time and is not necessary here.

app/models/pipeline_run.rb
```diff
@@ -1445,10 +1445,15 @@ def get_summary_contig_counts(min_contig_size)
...
+                           .where("read_count >= ?", min_contig_size)
+                           .where("lineage_json IS NOT NULL")
+    contig_taxids.each do |c|
+      count_types_to_taxids = {
```
tfrcarvalho: This construct seem unnecessary... we should be able to do something like

```
# before the loop, to always initialize new keys with 0
summary_per_tax = Hash.new(0) 

# in the loop
read_count, species_taxid_nt, species_taxid_nr, genus_taxid_nt, genus_taxid_nr = c

summary_dict[species_taxid_nt]["nt"][read_count] += 1 if species_taxid_nt
summary_dict[species_taxid_nr]["nr"][read_count] += 1 if species_taxid_nr

```

app/models/pipeline_run.rb
j-x-han: Done.

app/models/pipeline_run.rb
j-x-han: Done (initializing to `Hash.new(0)` became a longer code block to ensure the default values would actually be [stored in the hash](https://ruby-doc.org/core-2.5.1/Hash.html#method-c-new) and rubocop enforced some styling).

j-x-han:
Checked a few samples locally and it seems that this decreases the time of the query to roughly a third of what it was previously. Will check again on staging and prod.

app/models/pipeline_run.rb
```diff
@@ -1444,24 +1444,26 @@ def get_summary_contig_counts(min_contig_size)
...
-        taxids.each do |taxid|
-          summary_per_tax = summary_dict[taxid] ||= {}
-          summary_per_tax_and_type = summary_per_tax[count_type.downcase] ||= {}
-          summary_per_tax_and_type[c.read_count] ||= 0
-          summary_per_tax_and_type[c.read_count] += 1
+    # key: taxid:count_type, value: read_counts: count
+    # Create and store default values for the hash if the key doesn't exist yet
+    summary_dict = Hash.new do |summary, taxid|
+      summary[taxid] = Hash.new do |tax_id, count_type| # rubocop forces different variable names
+        tax_id[count_type] = Hash.new do |counttype, read_count|
```
tfrcarvalho: nit: add an underscore to the end to keep the same name intact, i.e. `count_type_`

app/models/pipeline_run.rb
```diff
@@ -1444,19 +1444,26 @@ def get_summary_contig_counts(min_contig_size)
...
-      lineage.each do |count_type, taxid_arr|
-        taxids = taxid_arr[0..1]
-        taxids.each do |taxid|
-          summary_per_tax = summary_dict[taxid] ||= {}
-          summary_per_tax_and_type = summary_per_tax[count_type.downcase] ||= {}
-          summary_per_tax_and_type[c.read_count] ||= 0
-          summary_per_tax_and_type[c.read_count] += 1
+    # key: taxid:count_type, value: read_counts: count
+    # Create and store default values for the hash if the key doesn't exist yet
+    summary_dict = Hash.new do |summary, taxid|
```
MarkAZhang: Nit: Instead of naming them `taxid` and `tax_id`, can we do `tax_id` and `tax_id_hash`?
```
summary_dict = Hash.new do |summary_hash, tax_id|
  summary_hash[tax_id] = Hash.new do |tax_id_hash, count_type|
    tax_id_hash[count_type] = Hash.new do |count_type_hash, read_count|
```

MarkAZhang: I think it makes it slightly more legible. This code block seems particularly confusing to read (with respect to the complexity of what it's actually doing), which is why I'm being nit-picky here.

app/models/pipeline_run.rb
```diff
@@ -1444,19 +1444,26 @@ def get_summary_contig_counts(min_contig_size)
...
-    contig_lineages(min_contig_size).each do |c|
-      lineage = JSON.parse(c.lineage_json)
-      lineage.each do |count_type, taxid_arr|
-        taxids = taxid_arr[0..1]
-        taxids.each do |taxid|
-          summary_per_tax = summary_dict[taxid] ||= {}
-          summary_per_tax_and_type = summary_per_tax[count_type.downcase] ||= {}
-          summary_per_tax_and_type[c.read_count] ||= 0
-          summary_per_tax_and_type[c.read_count] += 1
+    # key: taxid:count_type, value: read_counts: count
```
MarkAZhang: You probably didn't write this comment, but just noting that I can't figure out what this means.
I think a better comment would be:
```
Stores the number of contigs that match a given taxid, count_type (nt or nr), or read_count (number of reads aligned to that contig.
```

app/models/pipeline_run.rb
j-x-han: Done

app/models/pipeline_run.rb
j-x-han: Done! Thanks for the clear phrasing 😄

PR #2867, opened 2019-12-13
[Report Page] Parallelize contig and taxon counts requests

tfrcarvalho:
# Description

* Parallelize requests to the database to fetch taxa (twice) and contigs
* Left as default but optional for now for performance analysis
* Rough comparison pointed to an improvement of 25-35% of the time taken by those three tasks in sequence (not total time). Further analysis to do in prod.

# Tests

* Create a report and verify that is loaded correctly


jshoe:
looks like there's a Rubocop nit..

PR #2870, opened 2019-12-13
Update and group options in user menu

MarkAZhang:
# Description

Update user menu.

Screen Shot 2019-12-12 at 7 25 20 PM

@happyimadesignr for final approval.

### DIFF
--- a/app/assets/src/components/common/Header.jsx
+++ b/app/assets/src/components/common/Header.jsx
@@ -159,7 +159,7 @@ const UserMenuDropDown = ({
               className={cs.option}
               href="/bulk_downloads"
               onClick={() =>
-                logAnalyticsEvent("Header_dropdown-dropdowns-option_clicked")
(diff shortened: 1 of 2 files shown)


app/assets/src/components/common/Header.jsx
```diff
@@ -149,21 +149,28 @@ const UserMenuDropDown = ({
...
-          key="3"
+          key="1"
           text={
-            <a className={cs.option} href="/users/new">
-              Create User
+            <a
+              className={cs.option}
+              href="/bulk_downloads"
+              onClick={() =>
+                logAnalyticsEvent("Header_dropdown-dropdowns-option_clicked")
```
jshoe: I see this was moved code, but you probably meant `downloads-option_clicked`

MarkAZhang: Great catch!!

PR #2878, opened 2019-12-13
Revert aegea to 2.8.3

MarkAZhang:
We are having issues in prod that are most likely related to the recent aegea version bump. Reverting the aegea version.


MarkAZhang:
The problem was resolved without needing to submit this.

PR #2883, opened 2019-12-16
[Release fix] Display report after alignment is finished

j-x-han:
# Description

The new report page currently attempts to display the report as soon as `adjusted_remaining_reads` are available, so an empty table will show up while alignment/post-processing might still be in progress.

This change ensures that the report won't be displayed while alignment is still in progress and cleans up some other redundant conditions.

# Notes

* Will upload a sample on staging to ensure the report page will be displayed if Experimental is in progress.


app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -1074,7 +1070,7 @@ export default class SampleViewV2 extends React.Component {
               pipelineRun={pipelineRun}
               project={project}
               projectSamples={projectSamples}
-              reportPresent={reportMetadata.report_ready !== false}
+              reportPresent={reportMetadata.reportAvailable !== false}
```
gregdingle: by the way, `!== false` seems fragile. why not `=== true`?

app/services/pipeline_report_service.rb
```diff
@@ -259,23 +257,10 @@ def get_pipeline_status(pipeline_run)
       errorMessage: pipeline_run.error_message,
       knownUserError: pipeline_run.known_user_error,
       jobStatus: pipeline_run.job_status_display,
-      pipelineRunReportAvailable: pipeline_run_report_available?(pipeline_run),
+      reportAvailable: pipeline_run && pipeline_run.report_ready?,
```
gregdingle: sidenote: I prefer us to use consistent names in JS and ruby for traceability unless there is a good reason not to. why not `reportReady`?

PR #2884, opened 2019-12-17
Fetch bulk download file size after success.

MarkAZhang:
# Description

Add the file size of the generated file to the bulk download list.

Screen Shot 2019-12-16 at 6 50 26 PM

Fetch the file size from S3 using "head-object" call (which just fetches metadata) whenever a bulk download 'succeeds'. If the file size fetch fails, we log an error, but otherwise ignore the error.

# Notes

* Also rewrite the human-readable-file-size method. Fix a bug in it, and also make it look more like the file sizes on iOS.

# Tests

* Updated bulk download controller tests to account for file size.
* Added tests for human readable file size.
* Verified for both ECS and Resque style downloads that the file size is set correctly.


spec/lib/string_util_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "rails_helper"
+
+RSpec.describe StringUtil do
```
gregdingle: thanks for adding this... such lib functions should always have tests

app/lib/string_util.rb
```diff
@@ -1,15 +1,22 @@
 class StringUtil
+  FILE_SIZE_UNITS = ["B", "KB", "MB", "GB", "TB"].freeze
+  FILE_SIZE_UNIT_FACTOR = 1024.0
   def self.human_readable_file_size(size, decimal_places: 1)
-    file_unit = ""
+    if size.nil?
```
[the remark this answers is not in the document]
gregdingle: yea, I'm surprised we decided to reinvent the wheel here... deserves a comment at least

db/migrate/20191216200333_add_file_size_to_bulk_download.rb
```diff
@@ -0,0 +1,5 @@
+class AddFileSizeToBulkDownload < ActiveRecord::Migration[5.1]
+  def change
+    add_column :bulk_downloads, :output_file_size, :integer
```
gregdingle: please add a comment describing what this should contain and when `nil` is a valid value

spec/lib/string_util_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    end
+
+    it "allows customizable decimal places" do
+      expect(StringUtil.human_readable_file_size(12_345, decimal_places: 3)).to eq("12.056 KB")
+    end
+
+    it "handles big numbers well" do
+      expect(StringUtil.human_readable_file_size(1_024**5 * 100)).to eq("102400 TB")
+    end
+  end
```
gregdingle: it would be nice to test 0 size as well

db/migrate/20191216200333_add_file_size_to_bulk_download.rb
gregdingle: note these comments will be passed on to `schema.rb` for future reference, and to SQL definition

app/lib/string_util.rb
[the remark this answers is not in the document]
MarkAZhang: Thanks for finding that function! I vastly prefer using that function instead.

db/migrate/20191216200333_add_file_size_to_bulk_download.rb
MarkAZhang: Done!

PR #2886, opened 2019-12-17
Tweak text in bulk downloads display name and taxon options.

MarkAZhang:
# Description

Correct taxa/taxon usage for contigs and reads non-host downloads.

Screen Shot 2019-12-16 at 8 03 53 PM

Append the single taxon to the display name for contigs and reads non-host downloads that specified a single taxon.

Screen Shot 2019-12-16 at 8 04 04 PM

# Tests

* Added tests for download_display_name.

### DIFF
--- a/app/assets/src/components/views/bulk_download/ChooseStep.jsx
+++ b/app/assets/src/components/views/bulk_download/ChooseStep.jsx
@@ -280,11 +280,11 @@ class ChooseStep extends React.Component {
       case "taxa_with_reads":
         dropdownOptions = [
           {
-            text: "All Taxon",
+            text: "All Taxa",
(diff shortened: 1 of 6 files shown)


app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -280,11 +280,11 @@ class ChooseStep extends React.Component {
       case "taxa_with_reads":
         dropdownOptions = [
           {
-            text: "All Taxon",
+            text: "All Taxa",
```
MarkAZhang: Will consolidate this dropdown with TaxonContigSelect in a follow-up PR.

PR #2888, opened 2019-12-17
[Release fix] OmniAuth/Warden initialization order

davidrissato:
# Description

Fix an issue with authentication when `OmniAuth` middleware was invoking the `failure` method. 

`warden` middleware was not loaded yet, and `logout` was failing to remove the previous user session.

This fix consist in swapping the loading order of OmniAuth and Warden initialization, and guaranteeing the session will be reseted if warden fails to do so.

# Tests

* Manual


app/helpers/auth0_helper.rb
```diff
@@ -72,9 +72,16 @@ def auth0_session
...
     session.delete(:auth0_credentials)
-    warden&.logout(:auth0_user)
-    # logout users from previous Devise scope
-    warden&.logout(:user)
+    begin
+      warden.logout(:auth0_user)
+      # logout users from previous Devise scope
+      warden.logout(:user)
+    rescue
+      # if warden middleware is not present for any reason
```
jshoe: should we still log this `rescue` block?

davidrissato: Yes, may be a good idea. Let me add it

PR #2894, opened 2019-12-17
Make water_control a required metadata field

gregdingle:
# Description

This updates: 
* water_control metadata field in database
* unit test of required fields
* CSS of metadata form

image

# Notes

* also adds some validation according to existing docs
* refactored existing integration tests a bit
* question @MarkAZhang : for a metadata field to be applicable for a sample, must it be an attribute of the project AND the host genome? or either project OR host genome? this should be added to the docs
* question @MarkAZhang : what determines the rendered order of the metadata fields as columns?
* existing samples and projects will not be affected; they will not have water_control required
* dictionary auto updates

image

* and so does metadata edit
image

* and so does CSV download

image

# Tests

## unit tests
* migrate
* rails t

## new project
* open metadata upload screen in new project
* see new column

## old project
* open in old project
* do not see old column

### DIFF
--- a/app/assets/src/components/common/metadata_manual_input.scss
+++ b/app/assets/src/components/common/metadata_manual_input.scss
@@ -54,7 +54,7 @@
       }
 
       &.lessWidth {
-        width: 80px;
+        width: 110px;
       }
     }
 
--- a/app/assets/src/components/common/metadata_manual_input.scss
+++ b/app/assets/src/components/common/metadata_manual_input.scss
@@ -2,6 +2,15 @@
 @import "~styles/themes/colors";
 @import "~styles/themes/elements";
 
+// NOTE: Omit px so value can be exported for parseInt
+$metadata-input-width: 95;
+$metadata-input-extra-width: 225;
+
+:export {
+  metadataInputWidth: $metadata-input-width;
+  metadataInputExtraWidth: $metadata-input-extra-width;
+}
+
 .metadataManualInput {
   .columnPicker {
     /* Override semantic ui */
@@ -47,12 +56,10 @@
 
     .input {
       margin: $space-xxs $space-m $space-xxs 0;
-      // The width of the input needs to be about 15px smaller than the width of
-      // the column to maintain some padding. See also getColumnWidth.
-      width: 95px;
+      width: $metadata-input-width + 0px;
 
       &.extraWidth {
-        width: 225px;
+        width: $metadata-input-extra-width + 0px;
       }
     }
 
--- a/app/assets/src/components/common/metadata_manual_input.scss
+++ b/app/assets/src/components/common/metadata_manual_input.scss
@@ -47,10 +47,14 @@
 
     .input {
       margin: $space-xxs $space-m $space-xxs 0;
-      width: 130px;
+      width: 110px; // see also getColumnWidth
 
       &.extraWidth {
         width: 240px;
+      }
+
+      &.lessWidth {
+        width: 80px;
       }
     }
 
--- a/app/assets/src/components/common/metadata_manual_input.scss
+++ b/app/assets/src/components/common/metadata_manual_input.scss
@@ -52,10 +52,6 @@
       &.extraWidth {
         width: 240px;
       }
-
-      &.lessWidth {
-        width: 110px;
-      }
     }
 
     .noInput {
--- a/app/assets/src/components/common/metadata_manual_input.scss
+++ b/app/assets/src/components/common/metadata_manual_input.scss
@@ -47,10 +47,12 @@
 
     .input {
       margin: $space-xxs $space-m $space-xxs 0;
-      width: 110px; // see also getColumnWidth
+      // The width of the input needs to be about 15px smaller than the width of
+      // the column to maintain some padding. See also getColumnWidth.
+      width: 95px;
 
       &.extraWidth {
-        width: 240px;
+        width: 225px;
       }
     }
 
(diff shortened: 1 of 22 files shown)


app/helpers/test_helper.rb
```diff
@@ -3,7 +3,8 @@
 
 module TestHelper
   # Consider this a fixture for tests. Needs to integrate with dag_json for public_sample_run_stage.
-  TEST_RESULT_FOLDER = %w[
+  # Avoid "warning: already initialized constant" by conditional assignment
```
gregdingle: It looks like this was being cause by including `test_helper` more than once, but I'm not sure how to fix that

app/models/metadata_field.rb
```diff
@@ -3,10 +3,24 @@ class MetadataField < ApplicationRecord
...
+    NUMBER_TYPE,
+    DATE_TYPE,
+    LOCATION_TYPE,
+  ], }
+
+  # NOTE: not sure why these columns were not created as booleans
+  validates :force_options, inclusion: { in: [0, 1] }
+  validates :is_core, inclusion: { in: [0, 1] }
+  validates :is_default, inclusion: { in: [0, 1] }
+  validates :is_required, inclusion: { in: [0, 1] }
```
MarkAZhang: Thanks for adding all of these additional validations!

Regarding the "booleans", I recall the reason why I didn't make them booleans was because I observed that similar previous columns were also not booleans and I just followed what was done previously (I was worried that there would be additional things to figure out when using booleans for the first time). I would definitely support a convention to use boolean columns in the future.

db/migrate/20191217193112_update_water_control_metadata_field.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+  end
+
+  def down
+    water_control = MetadataField.find_by(name: "water_control")
+    if water_control
+      water_control.update(is_required: 0)
+    end
+  end
+end
```
MarkAZhang: It's not clear to me that this is the right approach to making water control "required", as opposed to simply running `MetadataField.find_by(name: "water_control").update(is_required: 1)` in the Rails console. Although I know this is how we've often done it in the past. 

As long as the application doesn't depend on water_control being required, this feels like something an admin would configure either in an admin dashboard or via the rails console, as opposed to something to put in a migration. What are your thoughts?

app/assets/src/components/common/metadata_manual_input.scss
```diff
@@ -47,7 +47,7 @@
 
     .input {
       margin: $space-xxs $space-m $space-xxs 0;
-      width: 130px;
+      width: 110px; // see also getColumnWidth
 
       &.extraWidth {
```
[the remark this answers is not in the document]
MarkAZhang: There's actually a bug with these values (and the bug existed prior to your change). See screenshot. This occurs when you add a bunch of columns to the MetadataManualInput.

Screen Shot 2019-12-18 at 12 26 33 PM

To prevent this, the width of the input needs to be smaller than the width of the column. Perhaps by about 15px. So if the column is 110, the width of the input should be 95px.

There's a way to export variables from `.scss` files, which you can use to share the constants between scss and javascript. See the `:export` block in `pipeline_viz.scss`.  By sharing a constant for the "buffer" (e.g. 15px), this can help prevent similar bugs in the future.

MarkAZhang:
Regarding your questions:
* It's project AND host genome.
* I believe the ordering is simply based on the metadata_field_id at the moment.
* Regarding the auto-generated CLI metadata template, it seemed like a nice to have, but not clear what it's good for. I'm happy to remove it, but it should also auto-update to include water control once you make it required. (let me know if this isn't the case). So you shouldn't need to worry about it.

app/models/metadata_field.rb
gregdingle: Ok good to know. The advantage from what I read is that rails will convert mysql tinyint into `true` and `false`.

db/migrate/20191217193112_update_water_control_metadata_field.rb
gregdingle: I was just trying to follow the precedent of declaring the required metadata in migrations. 

I think the best approach is to use `seeds.rb` which is what that is designed for. Then we would have an easy to read document of built-in metadata. I'm just not sure how painful converting would be.

gregdingle:
Thanks for the review. 

> it should also auto-update to include water control once you make it required. (let me know if this isn't the case).

The CLI template does not auto update. It is filled with a bunch of fields and example data from I don't know where.

> I believe the ordering is simply based on the metadata_field_id at the moment.
Ok we may want to specify an ordering. We could do it simply by adding an `position` col to the db table, then order by that by default. I'll leave that for a follow up PR. 


> So if the column is 110, the width of the input should be 95px
Thanks for catching this!

> export variables from .scss files
I will look into that, thanks!

gregdingle:
Tested CSS when columns overflow:

image

I'll look into SCSS export constant.

gregdingle:
> The CLI template does not auto update. It is filled with a bunch of fields and example data from I don't know where.

Correction: the CLI template has `water_control` after all , hidden among the 30 columns. I think our template should not have 30 columns... it seems really intimidating!

app/assets/src/components/common/metadata_manual_input.scss
```diff
@@ -47,10 +56,10 @@
 
     .input {
       margin: $space-xxs $space-m $space-xxs 0;
-      width: 110px; // see also getColumnWidth
+      width: $metadata-input-width + 0px;
```
MarkAZhang: Is the reason you need to add `0px` so that it will be interpreted as a "px" value?

gregdingle: yes that's correct. I looked it up 📖

PR #2897, opened 2019-12-18
[Report Page] Fix status for samples complete with insufficient reads

j-x-han:
# Description

Currently, the PipelineReportService will attempt to compute the report results if the pipeline run results have been finalized even if it failed, which can result in errors (see https://idseq.net/samples/36613 and https://idseq.net/samples/36613/report_v2). Instead, the report should immediately return if an error (such as insufficient reads) has occurred in the pipeline run.

# Tests

* Verified locally that the report page would display the appropriate warning message for a sample with insufficient reads.
* Pasted the new PipelineReportService code into the rails console and verified that `PipelineReportService.call` on the sample listed above generated a report instead running into errors.

### DIFF
--- a/app/services/pipeline_report_service.rb
+++ b/app/services/pipeline_report_service.rb
@@ -69,7 +69,10 @@ class PipelineReportService
 
   def generate
     metadata = get_pipeline_status(@pipeline_run)
-    unless @pipeline_run && (@pipeline_run.finalized? || @pipeline_run.report_ready?)
+    # Only generate the report if the pipeline has initialized and has loaded
+    # taxon_counts (is report_ready), or if its results are finalized without errors.
+    # Otherwise just return the metadata, which includes statuses and error messages to display.
+    unless @pipeline_run && (@pipeline_run.report_ready? || (@pipeline_run.finalized? && !@pipeline_run.failed?))
       return JSON.dump(
         metadata: metadata
       )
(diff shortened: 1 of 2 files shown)


app/services/pipeline_report_service.rb
```diff
@@ -69,7 +69,10 @@ def call
 
   def generate
     metadata = get_pipeline_status(@pipeline_run)
-    unless @pipeline_run && (@pipeline_run.finalized? || @pipeline_run.report_ready?)
+    # Only generate the report if the pipeline has initialized and has loaded
+    # taxon_counts (is report_ready), or if its results are finalized without errors.
+    # Otherwise just return the metadata, which includes statuses and error messages to display.
+    unless @pipeline_run && (@pipeline_run.report_ready? || (@pipeline_run.finalized? && !@pipeline_run.failed?))
```
gregdingle: Good comment but it points to the need for some refactoring of the concepts of "report ready" and-or "finalized". It seems like we should:
1. define a new pipeline state of `complete_with_issue`, as we show elsewhere in the UI as "complete - issue", and includes all the "statuses and error messages to display" in your comment
2. redefine `report_ready` to mean taxon counts loaded **or** `complete_with_issue`

j-x-han: Agreed. There's a **lot** going on with pipeline run statuses right now that makes handling these cases pretty tricky -- I've made a lot of PRs related to this since someone seems to find a new edge case every time! 😅
I filed a [JIRA ticket](https://jira.czi.team/browse/IDSEQ-1890) as a reminder to go back and do a more complete refactor of the statuses, which should include what it means for a pipeline run to be "report ready" or "finalized." I'll add a comment linking to the ticket!

gregdingle: It sounds like you prefer to merge this as-is so I'll approve.

PR #2900, opened 2019-12-18
Add user settings and user-facing settings page.

MarkAZhang:
# Description

Add user preferences.

Currently gated on bulk downloads because that's the only user preference.

Includes a page for users to modify their preferences. Includes validation for the "boolean" data type. 

User preferences page
Screen Shot 2019-12-18 at 2 36 03 PM

Dropdown menu (once bulk downloads is enabled)
Screen Shot 2019-12-18 at 2 35 49 PM

When no user preferences can be found (shouldn't happen because this feature is gated on bulk downloads)
Screen Shot 2019-12-18 at 2 26 06 PM

Error message if something goes wrong.
Screen Shot 2019-12-18 at 2 22 45 PM

# Notes
Some notable design decisions:
* Instead of having a JSON object for all preferences or having a single row with different preferences as columns, we store preference name/preference value/user pairs in each row. This allows you to add new preferences without database schema changes. It also allows for future queries where you need all users with a certain preference (which would be hard with a JSON object)
* We hard-code the metadata for the different user preferences instead of storing them in a UserPreferencesField table. Since the code will depend on certain user preferences being a certain way, it seems correct to put the metadata for these preferences in code. For example, if the code expects a certain user preference to be a boolean, but the data type is stored in the database, then there's no way to guarantee that this is the case.
* Right now, the "value" column of the database is a "string". If you need specific behavior for certain user preferences, you will need to add a "data_type" and write the validation for it.
* If a user doesn't have anything set for a particular user preference, the default value will be used.
* Still need to write tests. Wanted to get a round of feedback on the table schema first (in case any of the above is controversial)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -21,6 +21,7 @@ class User < ApplicationRecord
   has_many :phylo_trees, dependent: :destroy
   has_many :backgrounds, dependent: :destroy
   has_many :bulk_downloads, dependent: :destroy
+  has_many :user_settings, dependent: :destroy
 
   validates :email, presence: true
   validates :name, presence: true, format: {
@@ -119,6 +120,46 @@ class User < ApplicationRecord
 
   def owns_project?(project_id)
     projects.exists?(project_id)
+  end
+
+  def get_user_setting(key)
+    user_setting = user_settings.find_by(key: key)
+
+    return user_setting.value unless user_setting.nil?
+
+    return UserSetting::METADATA[key][:default]
+  end
+
+  def save_user_setting(key, value)
+    user_setting = user_settings.find_or_initialize_by(key: key)
+
+    user_setting.value = value
+    user_setting.save!
+  end
+
+  # Remove any user settings gated on allowed_features that the user doesn't have access to.
+  def viewable_user_setting_keys
+    parsed_allowed_feature_list = allowed_feature_list
+    UserSetting::METADATA.select do |_key, metadata|
+      metadata[:required_allowed_feature].nil? || parsed_allowed_feature_list.include?(metadata[:required_allowed_feature])
+    end.keys
+  end
+
+  def viewable_user_settings
+    # Fetch viewable user settings.
+    existing_user_settings = user_settings
+                             .where(key: viewable_user_setting_keys)
+                             .map { |setting| [setting.key, setting.value] }
+                             .to_h
+
+    # Fill in all missing user settings with the default value.
+    viewable_user_setting_keys.each do |key|
+      if existing_user_settings[key].nil?
+        existing_user_settings[key] = UserSetting::METADATA[key][:default]
+      end
+    end
+
+    existing_user_settings
   end
 
   # Update login trackable fields
--- a/app/assets/src/api/index.js
+++ b/app/assets/src/api/index.js
@@ -383,6 +383,15 @@ const uploadedByCurrentUser = async sampleIds => {
 
 const getHeatmapMetrics = () => get("/visualizations/heatmap_metrics.json");
 
+const getUserSettingMetadataByCategory = () =>
+  get("/user_settings/metadata_by_category");
+
+const updateUserSetting = (key, value) =>
+  postWithCSRF("user_settings/update", {
+    key,
+    value,
+  });
+
 export {
   bulkImportRemoteSamples,
   bulkUploadRemoteSamples,
@@ -414,6 +423,7 @@ export {
   getSummaryContigCounts,
   getTaxonDescriptions,
   getTaxonDistributionForBackground,
+  getUserSettingMetadataByCategory,
   getVisualizations,
   markSampleUploaded,
   saveProjectDescription,
@@ -430,4 +440,5 @@ export {
   getTaxaWithReadsSuggestions,
   getTaxaWithContigsSuggestions,
   uploadedByCurrentUser,
+  updateUserSetting,
 };
--- a/app/assets/src/components/common/Header.jsx
+++ b/app/assets/src/components/common/Header.jsx
@@ -223,7 +223,7 @@ const UserMenuDropDown = ({
     allowedFeatures.includes("bulk_downloads") &&
       userDropdownItems.push(
         <BareDropdown.Item
-          key="1"
+          key="downloads"
           text={
             <a
               className={cs.option}
@@ -238,9 +238,29 @@ const UserMenuDropDown = ({
         />
       );
 
+    adminUser &&
+      userDropdownItems.push(
+        <BareDropdown.Item
+          key="user_settings"
+          text={
+            <a
+              className={cs.option}
+              href="/user_settings"
+              onClick={() =>
+                logAnalyticsEvent(
+                  "Header_dropdown-user-settings-option_clicked"
+                )
+              }
+            >
+              Settings
+            </a>
+          }
+        />
+      );
+
     userDropdownItems.push(
       <BareDropdown.Item
-        key="2"
+        key="help"
         text={
           <ExternalLink
             className={cs.option}
@@ -254,7 +274,7 @@ const UserMenuDropDown = ({
         }
       />,
       <BareDropdown.Item
-        key="3"
+        key="feedback"
         text={
           <a
             className={cs.option}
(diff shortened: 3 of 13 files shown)


app/assets/src/components/views/UserPreferencesView.jsx
```diff
@@ -0,0 +1,134 @@
...
+        key,
+        value,
+        this.state.modifiedUserPreferences
+      ),
+    });
+
+    try {
+      await updateUserPreference(key, value);
+    } catch {
+      // Show an error if the update fails.
```
jshoe: maybe still have some sort of backend error reach the frontend for debuggability? `console.error`?

app/assets/src/components/views/UserPreferencesView.jsx
```diff
@@ -0,0 +1,134 @@
...
+export default class UserPreferencesView extends React.Component {
+  state = {
+    userPreferenceMetadata: null,
+    // User preference values that were modified on this page and not reflected in the UserContext.
+    modifiedUserPreferences: {},
+    error: "",
+  };
+
+  async componentDidMount() {
+    const userPreferenceMetadata = await getUserPreferenceMetadataByCategory();
```
jshoe: should we have some sort of error msg to show in case the `get` fails?

app/assets/src/components/views/user_preferences_view.scss
```diff
@@ -0,0 +1,27 @@
+@import "~styles/themes/elements";
```
jshoe: Is Jenn going to review this screen in general? or was it already in the mocks / discussed

app/models/user_preference.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  # value must be "true" or "false".
+  DATA_TYPE_BOOLEAN = "boolean".freeze
+
+  # Stores metadata for various user preferences.
+  USER_PREFERENCE_METADATA = {
+    RECEIVES_BULK_DOWNLOAD_SUCCESS_EMAILS => {
+      default: "true",
+      description: "Receive an email whenever a bulk download succeeds",
+      allowed_feature: "bulk_downloads",
```
jshoe: maybe `required_allowed_feature` or `prerequisite_feature`? It read a little confusingly to me here

app/models/user.rb
```diff
@@ -121,6 +122,46 @@ def owns_project?(project_id)
...
+
+    user_preference.value = value
+    user_preference.save!
+  end
+
+  # Remove any user preferences gated on allowed_features that the user doesn't have access to.
+  def viewable_user_preference_keys
+    parsed_allowed_feature_list = allowed_feature_list
+    UserPreference::USER_PREFERENCE_METADATA.select do |_key, metadata|
+      metadata[:allowed_feature].nil? || parsed_allowed_feature_list.include?(metadata[:allowed_feature])
```
jshoe: maybe `metadata[:allowed_feature].present?` in case people accidentally use `allowed_feature: ""` ?

gregdingle:
An idea I had just now is to make `user_preferences` replace `allowed_features`--in which case we would almost certainly want to call it `user_settings`, and make some only visible to admins. The advantages are that
* feature flags would be easier to change: an admin could navigate to a user's settings page and click a checkbox
* we could see and edit all the per-user global config in one table

MarkAZhang:
Hey Greg, that's a great idea regarding `allowed_features`. I'm all for calling it "User Settings".

The current implementation already allows you to specify a subset of preferences to appear on the user preferences page (by adding them to the display categories). However, there's currently no way to specify that a preference should be admin-only. That's a great suggestion! (though I might add it to a follow-up PR)

Awaiting your other comments and then I can make the changes.

It's not clear how much use this will get. I remember some folks (perhaps Jonathan?) mentioning that something like this has been needed for quite some time, which is why I wrote up a somewhat more complete implementation. But I feel like we should keep it simple at first and iterate on it as we go.

app/models/user_preference.rb
gregdingle: Agree with jonathan. 

Separately, I prefer less redundant name `UserPreference::METADATA`.

app/models/user.rb
```diff
@@ -121,6 +122,46 @@ def owns_project?(project_id)
...
 
+  def get_user_preference(key)
+    user_preference = user_preferences.find_by(key: key)
+
+    return user_preference.value unless user_preference.nil?
+
+    return UserPreference::USER_PREFERENCE_METADATA[key][:default]
+  end
+
+  def set_user_preference(key, value)
```
gregdingle: I'd prefer this named `save_user_preference` to denote the `save!` call. In general, I think it is better to let the caller decide when to commit changes but not a big deal.

app/controllers/user_preferences_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  def update
+    current_user.set_user_preference(params[:key], params[:value])
+
+    render json: {
+      status: "success",
+      key: params[:key],
+      value: current_user.get_user_preference(params[:key]),
+    }
+  rescue
+    render json: {
```
gregdingle: as a matter of convention, it would be good to know whether we should always log http errors

app/models/user.rb
```diff
@@ -121,6 +122,46 @@ def owns_project?(project_id)
...
+    user_preference = user_preferences.find_or_initialize_by(key: key)
+
+    user_preference.value = value
+    user_preference.save!
+  end
+
+  # Remove any user preferences gated on allowed_features that the user doesn't have access to.
```
gregdingle: good comment!

db/migrate/20191218014846_create_user_preferences.rb
```diff
@@ -0,0 +1,10 @@
+class CreateUserPreferences < ActiveRecord::Migration[5.1]
+  def change
+    create_table :user_preferences do |t| # A table of user preferences. Each user + preference combination is a separate row.
+      t.references :user, foreign_key: true
+      t.string :key, comment: "The name of the user preference, e.g. receives_bulk_download_success_emails"
+      t.string :value, comment: "This value of the user preference, e.g. true"
```
gregdingle: I would add "schema defined by data type in metadata"

db/migrate/20191218014846_create_user_preferences.rb
```diff
@@ -0,0 +1,10 @@
+class CreateUserPreferences < ActiveRecord::Migration[5.1]
+  def change
+    create_table :user_preferences do |t| # A table of user preferences. Each user + preference combination is a separate row.
+      t.references :user, foreign_key: true
```
gregdingle: nice `foreign_key` 😄

app/assets/src/components/views/user_preferences_view.scss
```diff
@@ -0,0 +1,27 @@
...
+  color: $medium-grey;
+}
+
+.category {
+  .title {
+    @include font-header-m;
+  }
+
+  .field {
+    padding: 10px;
```
gregdingle: `$space-default`?

app/assets/src/components/views/UserPreferencesView.jsx
gregdingle: revert the state?

MarkAZhang: That's a great idea, but I'd like to defer it because trying to revert state asynchronously introduces the possibility for race conditions and I'd like to keep the logic simple if possible. I think it's enough to have an error message to alert the user that something is wrong.

app/assets/src/components/views/user_preferences_view.scss
MarkAZhang: Good catch, thanks!

app/controllers/user_preferences_controller.rb
MarkAZhang: I think we should. Thanks for pointing it out here.

app/assets/src/components/views/user_preferences_view.scss
MarkAZhang: This is going to be admin-only for now, so no need to review.

MarkAZhang:
Since we are no longer doing the automated emails for bulk downloads, I added an example flag for demo purposes, and made the page admin-only.

Verified that non-admins don't see the preferences page.

PR #2902, opened 2019-12-18
Replace references to "tissue" with "sample type"

gregdingle:
# Description

The UI still shows "tissue" in places although it actually shows "sample type". And the code is full of "tissue" identifiers. This PR corrects the UI, and updates all ruby code, but not all JS, which proved to be too time consuming.  

whole mosquito

METADATA

LOCATION V

# Notes

* The "locationV2" has some similar special cases
* The way that names are remapped to camelCase from ruby to JS makes it really difficult to search and replace. 
* The lack of standardization or documentation on data structures makes for reading a lot of code.

# Tests
## specific unit tests that search by sample type
Unit test projects controller
Unit test samples controller

## all unit tests
rails t
rspec

## search suggestions
override get server search suggestions to test that code path
and hit endpoint http://localhost:3000/search_suggestions?categories[]=sample&categories[]=project&categories[]=tissue&query=abdomen&domain=my_data

## manual test data discovery, both samples and projects views
left filter nav
search box
right stats sidebar

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -104,7 +104,7 @@ class ProjectsController < ApplicationController
                              .references(:samples)
           # get aggregated lists of association values in string by using MySQL's GROUP_CONCAT (should update to JSON_ARRAYAGG when possible)
           group_concat_host = "GROUP_CONCAT(DISTINCT host_genomes.name SEPARATOR '::') AS hosts"
-          group_concat_sample_type = "GROUP_CONCAT(DISTINCT CASE WHEN metadata_fields.name = 'sample_type' THEN metadata.string_validated_value ELSE NULL END SEPARATOR '::') AS tissues"
+          group_concat_sample_type = "GROUP_CONCAT(DISTINCT CASE WHEN metadata_fields.name = 'sample_type' THEN metadata.string_validated_value ELSE NULL END SEPARATOR '::') AS sample_types"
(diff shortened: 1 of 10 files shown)


spec/controllers/projects_controller_spec.rb
```diff
@@ -3,6 +3,8 @@
 WebMock.allow_net_connect!
 
 RSpec.describe ProjectsController, type: :controller do
+  KLEBSIELLA_TAX_ID = 2
```
gregdingle: avoids double init warning

app/controllers/projects_controller.rb
```diff
@@ -123,7 +123,11 @@ def index
               project_hash["owner"] = (project_hash["uploaders"] || '').split('::')[0]
               project_hash["editable"] = current_user.admin? || project_hash["editable"] == 1
               project_hash.delete("uploaders")
-              ["locations", "hosts", "tissues"].each { |k| project_hash[k] = (project_hash[k] || '').split('::') }
+              ["locations", "hosts", "sample_types"].each { |k| project_hash[k] = (project_hash[k] || '').split('::') }
+              # Return as "tissue" for legacy compatibility. It's too hard to
+              # rename all JS instances of "tissue".
```
jshoe: should we have a TODO/JIRA for this?

gregdingle: For a minor code quality issue like this, I would leave it to the next person who works with this code to escalate to a ticket for team prioritization, according to their judgement. In response to the new team convention that every TODO should have a ticket, I'm starting to add mere comments when I identify minor issues. As in this case, just knowing that someone tried to fix something is valuable.

PR #2905, opened 2019-12-19
[Report Page] Fix bugs related to categories

j-x-han:
# Description

Fixing https://jira.czi.team/browse/IDSEQ-1950 and https://jira.czi.team/browse/IDSEQ-1951:

* Taxon tree will no longer crash if no taxons pass the selected filters (the tree will no longer be rendered unless there are taxons).
image

* Uncagtegorized taxons will now show up if you select that filter.
image

# Tests

* Applied filters until no taxons passed, selected taxon tree view, verified that page did not crash.
* Applied the "uncategorized" category filter and verified that the results matched that of the legacy page.

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
+++ b/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
@@ -332,11 +332,12 @@ export default class SampleViewV2 extends React.Component {
 
     // taxon's category was selected and its subcategories were not excluded
     if (
-      categories.has(row.category) &&
-      !some(
-        subcategory => subcategories.has(subcategory),
-        row.subcategories || []
-      )
+      (categories.has(row.category) &&
+        !some(
+          subcategory => subcategories.has(subcategory),
+          row.subcategories || []
+        )) ||
+      (categories.has("uncategorized") && row.category === null)
     ) {
       return true;
     }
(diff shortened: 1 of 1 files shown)


app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -332,11 +332,12 @@ export default class SampleViewV2 extends React.Component {
 
     // taxon's category was selected and its subcategories were not excluded
```
gregdingle: so categories is filter by inclusion and subcategories by exclusion? seems confusing!

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -1029,19 +1030,20 @@ export default class SampleViewV2 extends React.Component {
...
-                nameType={selectedOptions.nameType}
-                onTaxonClick={this.handleTaxonClick}
-                sample={sample}
-                taxa={filteredReportData}
-                useReportV2Format={true}
-              />
-            </div>
-          )}
+          {view == "tree" &&
+            filteredReportData.length > 0 && (
```
gregdingle: did we consider adding an empty state message to the user such as "no data passed the selected filters" ?

j-x-han: There's a small line underneath the filters that tells the user how many rows are passing the selected filters, which will say "0 rows passing the above filters, out of # total rows."

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
j-x-han: I think subcategories are also filtered by inclusion; there's a condition just above this one that explicitly checks the subcategory.

PR #2907, opened 2019-12-19
Add admin options for bulk downloads

MarkAZhang:
# Description

Add the following options for admins:
* Admins are exempt from the max sample limit.
* Admins can see bulk downloads from all users.
* Admins can see the name of the bulk download owner.

Screen Shot 2019-12-18 at 9 06 25 PM

* Admins can see the id of the bulk download, the execution type, and a log url for ecs downloads. 

Screen Shot 2019-12-18 at 10 16 54 PM

These options will make it easier for admins to address issues around bulk downloads.

# Tests

* Added tests for admin privileges. There are already existing tests to check that normal users don't have these privileges.
* Verify that the UI still behaves as before for normal users, but admin users can get around the max sample limit and see additional fields in bulk download details modal.

### DIFF
--- a/app/lib/aws_util.rb
+++ b/app/lib/aws_util.rb
@@ -1,6 +1,9 @@
 class AwsUtil
+  # This is currently used to generate log urls for admins. We assume deployment in us-west-2 region.
+  AWS_REGION = "us-west-2".freeze
+
   def self.get_cloudwatch_url(log_group, log_stream)
-    "https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2" \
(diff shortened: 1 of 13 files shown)


app/lib/aws_util.rb
```diff
@@ -0,0 +1,6 @@
+class AwsUtil
+  def self.get_cloudwatch_url(log_group, log_stream)
+    "https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2" \
```
jshoe: I feel like `us-west-2` should be in a global env var, but I don't see other usages with one. Or you could have one in this file?

PR #2909, opened 2019-12-19
Add a Rake task to export DAGs

kislyuk:
# Description

This is part of the work to separate pipeline/workflow logic from idseq-web.

When cutting over to the new pipeline infra, this task will be run to export the workflow definitions using a recent representative sample, then we will update https://github.com/chanzuckerberg/idseq-web/blob/master/app/models/pipeline_run_stage.rb#L215-L344 to dispatch workflows/pipelines by executing Step Functions with the given parameters/attributes (replacing the current approach of constructing the DAG and launching a Batch job).

### DIFF
--- a/lib/tasks/export_dags.rake
+++ b/lib/tasks/export_dags.rake
@@ -7,7 +7,8 @@ task :export_dags, [:sample_id] => :environment do |_, args|
       puts "Exported #{dag_name}.json"
     end
 
-    def aegea_batch_submit_command(*) end
+    def aegea_batch_submit_command(*)
+    end
 
     def extract_dag
       send(job_command_func)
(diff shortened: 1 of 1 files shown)


lib/tasks/export_dags.rake
```diff
@@ -0,0 +1,33 @@
...
+
+    def aegea_batch_submit_command(*) end
+
+    def extract_dag
+      send(job_command_func)
+    end
+  end
+
+  if args.sample_id?
+    Sample.where(id: args.sample_id).first.pipeline_runs.each do |pipeline_run|
```
jshoe: nit: instead of `where().first` you can use `find_by`

PR #2910, opened 2019-12-19
[Release fix] [Report page] Fix contig/contig r threshold filters

j-x-han:
# Description

Fix the contig/contig r threshold filters on the new report page. Also fix the comparison boundaries for threshold filters.

# Tests

* Applied contig and contig r filters on a sample, compared to the legacy page, and verified that the expected taxons show up.


j-x-han:
@jshoe I don't think so, the contigs structure and filters were handled slightly differently in the old report page.

PR #2911, opened 2019-12-21
Add restrictions to privacy-sensitive bulk downloads.

MarkAZhang:
# Description

* Add concept of admin-only and uploader-only download types.
* host gene counts is admin-only
* original input file is uploader-only
* Added validation in the create endpoint for these concepts.
* Added "admin-only" tag for admin-only downloads, so admins know.

What admins see:
Screen Shot 2019-12-20 at 6 02 47 PM

What normal users see:
Screen Shot 2019-12-20 at 6 02 57 PM

Error messages from back-end validation (when UI is bypassed):

Screen Shot 2019-12-20 at 6 18 43 PM

Screen Shot 2019-12-20 at 6 19 33 PM

# Tests
* Updated non-admin and admin tests.
* Verified manually that UI behaves as expected for non-admin and admin.

### DIFF
--- a/app/assets/src/components/ui/labels/StatusLabel.jsx
+++ b/app/assets/src/components/ui/labels/StatusLabel.jsx
@@ -8,9 +8,13 @@ import cs from "./status_label.scss";
 
 class StatusLabel extends React.Component {
   render() {
-    const { status, type, className, tooltipText } = this.props;
+    const { status, type, className, tooltipText, inline } = this.props;
     const label = (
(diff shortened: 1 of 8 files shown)


app/assets/src/components/ui/labels/StatusLabel.jsx
```diff
@@ -8,9 +8,13 @@ import cs from "./status_label.scss";
```
jshoe: Looks like there was a CI error btw

PR #2912, opened 2019-12-31
Fix bugs and edge cases in User Create/Update flows

jshoe:
### Description
Bundled fixes for a couple of issues and edge cases that were brought up:
(1) Hide the "send activation email" checkbox when you're updating a user (not making a new one).
(2) Clarify the forgot password workflow with a message (we had this message on the Auth0-hosted screen as well).
(3) Show user creation error messages properly.
(4) Add a validation check for email uniqueness (we had it on the index but not in the validator) and disallow uppercase characters.

### Notes
(3) Before we had the pattern for `if @user.save / else`, but that only handles validation checks, so exceptions could be thrown and not properly passed to the frontend. This was causing the frontend page to crash without explanation.

(4) We discovered that Auth0 automatically converts all emails to lowercase and treats them as case-insensitive, even when our input had capital letters (see: [link](https://community.auth0.com/t/creating-a-user-converts-email-to-lowercase/6678) and [link](https://community.auth0.com/t/api-is-converting-username-to-lowercase/7763)). This would make it impossible for the user to log in because of the inconsistency between our field and theirs.

Based on email specs, an email could technically be case-sensitive, but in reality it's very unlikely that an email provider would allow case-sensitive emails (e.g. bob@ as different from Bob@). We could either lowercase emails automatically as well, or warn on capital letters at creation time. I did the latter for user clarity and to be a little safer. I also checked staging/prod and did not find any other emails with capital letters, so this is an edge case.

### Tests
Screen Shot 2019-12-30 at 2 29 36 PM
Screen Shot 2019-12-30 at 3 00 44 PM
Screen Shot 2019-12-30 at 3 06 31 PM

### DIFF
--- a/app/assets/src/components/CreateUser.jsx
+++ b/app/assets/src/components/CreateUser.jsx
@@ -208,6 +208,7 @@ class CreateUser extends React.Component {
   }
 
   renderUserForm(submitFunc, funcName) {
+    const { selectedUser } = this.props;
     const { sendActivation } = this.state;
     return (
       <div className="user-form">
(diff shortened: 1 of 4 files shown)


jshoe:
@MarkAZhang (3) That's right :)
For the Forgot Password screen, we switched back to that one to support the additional feature of "emailing when they enter an email that doesn't have an account". So even on the Auth0 login page, the link goes to the legacy "forgot password" updated here 👍

PR #2926, opened 2020-01-08
Create sample type model

gregdingle:
# Description

As part of standardizing sample types, this creates a new model and populates it. 

# Notes

* A following PR will create a controller endpoint.

# Tests

Run migration, see no errors, see data in database
Check validation rules with initial data


spec/factories/sample_types.rb
```diff
@@ -0,0 +1,4 @@
+FactoryBot.define do
```
gregdingle: not sure if i should delete these stub files ... we have many that have not been deleted.

jshoe:
Looks like brakeman has a warning to consider

gregdingle:
> Looks like brakeman has a warning to consider

Yup, I had to learn about ruby regex quirks today.

spec/factories/sample_types.rb
jshoe: probably don't need if not being used yet

app/models/sample_type.rb
```diff
@@ -0,0 +1,23 @@
+# From https://docs.google.com/spreadsheets/d/1_hPkQe5LI0Zw_C0Ls4HVCEDsc_FNNVOaU_aAfoaiZRE/
```
jshoe: maybe include a title to search for in case the doc moves?

gregdingle:
Sorry I didn't understand that you had comments coming. I will address them in the next PR coming shortly.

PR #2929, opened 2020-01-09
Live suggestions for sample type input

gregdingle:
# Description

This changes the sample type input to be a live autocomplete similar to the location input. The data comes from newly added `SampleType`. Unlike `location`, all possible suggestions are returned from the server and the filtering is done in JS alone, because of the small number of possible suggestions.

Also, there are two categories of match: "suggested" and "all". Suggested depends on the host genome, whether it is a human or insect or neither. This follows the framework created by @katrinakalantar at https://docs.google.com/spreadsheets/d/1_hPkQe5LI0Zw_C0Ls4HVCEDsc_FNNVOaU_aAfoaiZRE/edit?ts=5e152439#gid=0 . 

Plain text entries are still possible.

# Notes

* Although the suggested values come from a rails model, the selected values are stored as more generic `Metadatum`. 
* "OTHER" would be a more accurate description, but "ALL" is the word in the design at https://chanzuckerberg.invisionapp.com/share/A6V572ZSBU5#/screens/396544828 .
* I came up with the string matching algorithm based on intuition
* `MetadataInput` component is growing somewhat out of control because all metadata type specific props are being passed in: `isHuman`, `sampleTypes`, etc
* I added to the host genome wiki: https://github.com/chanzuckerberg/idseq-web/wiki/%5BDev%5D-How-to-update-host-genomes#web-app-update. "- (Uncommon) If your host is an insect, add it to the constant `INSECT_GENOME_IDS` so that only insect-appropriate sample types will be suggested for it."
* Many of the official types are too large for the input box so they are clipped. The input is the same width as before so I left it. 

# Tests

Try "apply all" 
Try saving and reloading

Try a variety of inputs. See expected results in suggested, all and plain text.

Check full word match, lowercase
image

Check insect body parts for insect host
image

Check "CSF" acronym match
image

Check second word in two words
image

Check prefix
image

### DIFF
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -398,7 +398,7 @@ MetadataManualInput.propTypes = {
   withinModal: PropTypes.bool,
   projectMetadataFields: PropTypes.object,
   hostGenomes: PropTypes.array,
-  sampleTypes: PropTypes.arrayOf(PropTypes.SampleTypeType).isRequired,
+  sampleTypes: PropTypes.arrayOf(PropTypes.SampleTypeProps).isRequired,
 };
 
 export default MetadataManualInput;
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -29,13 +29,6 @@ import MetadataInput from "./MetadataInput";
 const map = _fp.map.convert({ cap: false });
 
 class MetadataManualInput extends React.Component {
-  // List as of 2020-01-08.
-  INSECT_GENOME_IDS = [
-    2, // Mosquito
-    6, // Tick
-    16, // Bee
-  ];
-
   state = {
     selectedFieldNames: [],
     projectMetadataFields: null,
@@ -271,6 +264,9 @@ class MetadataManualInput extends React.Component {
         )
       : sample.host_genome_id;
 
+  getSampleHostGenome = sample =>
+    find(["id", this.getSampleHostGenomeId(sample)], this.state.hostGenomes);
+
   isHostGenomeIdValidForField = (hostGenomeId, field) =>
     // Special-case 'Host Genome' (the field that lets you change the Host Genome)
     field === "Host Genome" ||
@@ -326,6 +322,7 @@ class MetadataManualInput extends React.Component {
 
           // Only show a MetadataInput if this metadata field matches the sample's host genome.
           if (this.isHostGenomeIdValidForField(sampleHostGenomeId, column)) {
+            const hostGenome = this.getSampleHostGenome(sample);
             return (
               <div>
                 <MetadataInput
@@ -342,8 +339,8 @@ class MetadataManualInput extends React.Component {
                     });
                   }}
                   withinModal={this.props.withinModal}
-                  isHuman={sampleHostGenomeId === 1}
-                  isInsect={this.INSECT_GENOME_IDS.includes(sampleHostGenomeId)}
+                  isHuman={hostGenome.taxa_category === "human"}
+                  isInsect={hostGenome.taxa_category === "insect"}
                   sampleTypes={this.props.sampleTypes}
                 />
                 {this.props.samples.length > 1 &&
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -362,10 +362,12 @@ class MetadataManualInput extends React.Component {
   };
 
   getColumnWidth = column => {
+    // The width of the input needs to be about 15px smaller than the width of
+    // the column to maintain some padding. See also getColumnWidth.
     if (["Sample Name", "collection_location_v2"].includes(column)) {
-      return 240;
+      return parseInt(cs.metadataInputExtraWidth) + 15;
     } else {
-      return 145;
+      return parseInt(cs.metadataInputWidth) + 15;
     }
   };
 
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -29,6 +29,13 @@ import MetadataInput from "./MetadataInput";
 const map = _fp.map.convert({ cap: false });
 
 class MetadataManualInput extends React.Component {
+  // List as of 2020-01-08.
+  INSECT_GENOME_IDS = [
+    2, // Mosquito
+    6, // Tick
+    16, // Bee
+  ];
+
   state = {
     selectedFieldNames: [],
     projectMetadataFields: null,
@@ -336,6 +343,8 @@ class MetadataManualInput extends React.Component {
                   }}
                   withinModal={this.props.withinModal}
                   isHuman={sampleHostGenomeId === 1}
+                  isInsect={this.INSECT_GENOME_IDS.includes(sampleHostGenomeId)}
+                  sampleTypes={this.state.sampleTypes}
                 />
                 {this.props.samples.length > 1 &&
                   this.renderApplyToAll(sample, column)}
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -344,7 +344,7 @@ class MetadataManualInput extends React.Component {
                   withinModal={this.props.withinModal}
                   isHuman={sampleHostGenomeId === 1}
                   isInsect={this.INSECT_GENOME_IDS.includes(sampleHostGenomeId)}
-                  sampleTypes={this.state.sampleTypes}
+                  sampleTypes={this.props.sampleTypes}
                 />
                 {this.props.samples.length > 1 &&
                   this.renderApplyToAll(sample, column)}
@@ -398,6 +398,7 @@ MetadataManualInput.propTypes = {
   withinModal: PropTypes.bool,
   projectMetadataFields: PropTypes.object,
   hostGenomes: PropTypes.array,
+  sampleTypes: PropTypes.arrayOf(PropTypes.SampleTypeType).isRequired,
 };
 
 export default MetadataManualInput;
--- a/app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
+++ b/app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { groupBy } from "lodash/fp";
 
 import PropTypes from "~/components/utils/propTypes";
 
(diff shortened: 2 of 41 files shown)


app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
```diff
@@ -0,0 +1,93 @@
...
+    }
+    if (suggestedSampleTypes.all) {
+      results.all = {
+        name: "ALL",
+        results: suggestedSampleTypes.all.map(formatResult),
+      };
+    }
+    // Always lowercase plain text input to distinguish it from Title Case
+    // offical sample types, which are enforced in SampleType.rb.
+    // TODO (gdingle): enforce lowercasing in CSV upload as well.
```
gregdingle: not sure how best to do the same for CSV and CLI uploads in a simple way, and not sure if it's worth the trouble... @MarkAZhang ?

app/assets/src/components/common/MetadataManualInput.jsx
```diff
@@ -29,6 +29,13 @@ import MetadataInput from "./MetadataInput";
 const map = _fp.map.convert({ cap: false });
 
 class MetadataManualInput extends React.Component {
+  // List as of 2020-01-08.
+  INSECT_GENOME_IDS = [
```
tfrcarvalho: Did you consider including this info in the DB schema somehow (host genome model)? We might start missing updating this when we add new insect genomes.

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
```diff
@@ -0,0 +1,93 @@
...
+        return true;
+      }
+      return false;
+    };
+    const matchedSampleTypes = this.props.sampleTypes.filter(matchSampleTypes);
+
+    const isSuggested = sampleType => {
+      // Insects only get their own types
+      if (this.props.isInsect) {
+        return sampleType.insect_only ? "suggested" : "donotshow";
```
tfrcarvalho: These string should probably be constants.

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+import PropTypes from "~/components/utils/propTypes";
+
+import LiveSearchPopBox from "./LiveSearchPopBox";
+
+class SampleTypeSearchBox extends React.Component {
+  handleSearchTriggered = query => {
+    const matchSampleTypes = sampleType => {
+      // Match chars in any position. Good for acronyms.
+      const regex = new RegExp(query.split("").join(".*"), "gi");
```
tfrcarvalho: are we assuming single word? What happens with spaces and word order?

MarkAZhang:
`<MetadataInput>` is also used in the `MetadataTab` which is in the sample details sidebar. You should also verify that your new input works there as expected.

app/assets/src/components/common/MetadataManualInput.jsx
gregdingle: Yes, I considered it... and I think that it is reasonable to add a column such as `sample_type_taxa_name` to `host_genomes`. However, I decided against it for this PR because:
* we haven't had a chance to play with the suggestions yet and there may be bigger changes that result 
* the hard-coded precedent in the code of `isHuman` 
* the cost of a missed insect is low: the same options would be available in the "ALL" category instead of "SUGGESTED"

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
gregdingle: No problem. Side note: it's funny how we are more strict about constants in javascript than in python.

tfrcarvalho: I am guessing you meant ruby... when i notice the pattern i try to push for it on both. Not sure if I have been consistent though :P

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
gregdingle: Currently, spaces are considered just another character, so the query cannot contain more spaces than in the name. For example: "ab domen" will not match. "whole b" will match. 

Having thought about it now, I think ignoring spaces in the query is more intuitive. What do you think?

tfrcarvalho: Sorry, did not quite get the example: what does "whole b" match?

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
```diff
@@ -0,0 +1,94 @@
...
+    const matchedSampleTypes = this.props.sampleTypes.filter(matchSampleTypes);
+
+    const isSuggested = sampleType => {
+      // Insects only get their own types
+      if (this.props.isInsect) {
+        return sampleType.insect_only ? this.SUGGESTED : this.DONOTSHOW;
+      }
+      // Humans get any type
+      if (this.props.isHuman) {
+        return this.SUGGESTED;
```
MarkAZhang: Nit: Could we move the constants above the class and just have `SUGGESTED` instead of `this.SUGGESTED`?

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
```diff
@@ -0,0 +1,93 @@
...
+      // Match chars in any position. Good for acronyms.
+      const regex = new RegExp(query.split("").join(".*"), "gi");
+      if (regex.test(sampleType.name)) {
+        return true;
+      }
+      return false;
+    };
+    const matchedSampleTypes = this.props.sampleTypes.filter(matchSampleTypes);
+
+    const isSuggested = sampleType => {
```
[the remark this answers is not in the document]
gregdingle: Good point. `isSuggested` was leftover from when the function was returning a boolean!

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
gregdingle: Good point. I guess that is more idiomatic.

app/assets/src/components/utils/propTypes.js
```diff
@@ -141,5 +148,6 @@ export default {
   SampleUploadType,
   DownloadType,
   DownloadTypeParam,
+  SampleTypeType,
```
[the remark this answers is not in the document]
gregdingle: I think `SampleType` would be a mistake because of the ambiguity with `Sample`. `SampleTypeType` just seems funny to me, but how about `SampleTypeProps`?

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
tfrcarvalho: I see, "whole blood". 
I guess its more of a UX decision. IMO, I would probably split it on spaces and run the query independently for each word. Would love that "blood whole" would match too... (maybe there are other side effects im not seeing right now).

app/assets/src/components/common/MetadataManualInput.jsx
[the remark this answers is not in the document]
gregdingle: The only reasons are what I put above. FYI, I followed the precedent of `isHuman={sampleHostGenomeId === 1}`

db/migrate/20200108180542_create_sample_types.rb
```diff
@@ -27,6 +27,7 @@ def populate_sample_types!
       'CNS Infections': [
         { name: 'Brain' },
         { name: 'Cerebrospinal Fluid', human_only: true },
+        { name: 'Central Nervous System' },
```
gregdingle: I opportunistically fixed this following discussion with Katrina . It will work as long as this PR is deployed along with the previous 😸

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
gregdingle: No, I meant python, in the pipeline. It's full of magic strings. I think in ruby we consistently use constants.

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
gregdingle: I changed it to ignore the spaces. As for matching words in any order, it seems like overkill to me, but happy to be convinced otherwise.

gregdingle:
> MetadataInput is also used in the MetadataTab which is in the sample details sidebar. You should also verify that your new input works there as expected.

Good point. I'll check that now. 

> If the user clicks out of the input without picking something, the text stays in the input but nothing is actually selected (which isn't indicated). Might want to select-on-blur like the location box.

Good catch! This was actually a subtle bug. 

> Let's say you have both insect and human samples in your upload. You pick a human-only sample type for a human sample and click "Apply to All". This would apply the human-only sample type to all the insect samples as well. Not sure if this is an issue.

This would not be much of an issue in our current design, because the pairing of host and sample types is a suggestion only.

app/assets/src/components/ui/controls/SampleTypeSearchBox.jsx
tfrcarvalho: For this particular case, it probably is... for a more general function I would do it. So, up to you.

db/migrate/20200108180542_create_sample_types.rb
```diff
@@ -27,6 +27,7 @@ def populate_sample_types!
...
         { name: 'Ocular Fluid', human_only: true },
       ],
```
katrinakalantar: May make mores sense (for consistency across the sample groups) to use 
- "Central Nervous System" instead of "CNS Infections"
- "Respiratory Tract" instead of "Respiratory Infections" 
- "Reproductive Tract" instead of "Reproductive"

gregdingle:
@tfrcarvalho @MarkAZhang Updates: 

* I fixed the issues that @MarkAZhang raised
* I tested the input on the single sample page
image
* I went ahead and added a new column `taxa_category` to `host_genomes` as discussed, hooked up the UI, and I replaced pre-existing `isHuman` logic, so now we can support multiple "human" genomes if we want.

app/models/host_genome.rb
```diff
@@ -2,6 +2,13 @@ class HostGenome < ApplicationRecord
   has_many :samples, dependent: :restrict_with_exception
   has_and_belongs_to_many :metadata_fields
 
+  validates :taxa_category, inclusion: { allow_blank: true, in: [
+    # From "Sample Type Groupings" at
+    # https://docs.google.com/spreadsheets/d/1_hPkQe5LI0Zw_C0Ls4HVCEDsc_FNNVOaU_aAfoaiZRE/
+    "human",
+    "insect",
+  ], }, if: -> { respond_to?(:taxa_category) } # for migrations
```
MarkAZhang: Wow! Cool solution.

PR #2930, opened 2020-01-09
Validate bulk downloads field params in controller

MarkAZhang:
# Description

When creating a bulk download, validation that the params provided by the user are valid.
Only 4 of the bulk downloads require params.
If the user uses the BulkDownloadModal, it should be impossible for them to send an invalid set of parameters, but this provides an extra check on the back-end.

# Tests

* Verify that all bulk downloads can still be created.
* Write helper and controller tests.

### DIFF
--- a/app/models/bulk_download.rb
+++ b/app/models/bulk_download.rb
@@ -23,6 +23,47 @@ class BulkDownload < ApplicationRecord
   attr_accessor :params
 
   validates :status, presence: true, inclusion: { in: [STATUS_WAITING, STATUS_RUNNING, STATUS_ERROR, STATUS_SUCCESS] }
+  validate :params_checks
+
+  def params_checks
+    set_params
+    params = self.params
+
+    params_value = lambda do |key|
+      return nil if params.nil?
+      # Each field param should be structured as
+      # { displayName: "FOO", value: "BAR" }
+      # field_params[key] should be ActionController::Parameters object that responds to :keys
+      params[key].respond_to?(:keys) ? params[key]["value"] : nil
+    end
+
+    if download_type == BulkDownloadTypesHelper::SAMPLE_TAXON_REPORT_BULK_DOWNLOAD_TYPE
+      errors.add(:params, "background value must be an integer") unless params_value.call("background").is_a? Integer
+    end
+
+    if download_type == BulkDownloadTypesHelper::COMBINED_SAMPLE_TAXON_RESULTS_BULK_DOWNLOAD_TYPE
+      metric = params_value.call("metric")
+      errors.add(:params, "metrics value is invalid") unless HeatmapHelper::ALL_METRICS.pluck(:value).include?(metric)
+
+      if ["NT.zscore", "NR.zscore"].include?(metric)
+        errors.add(:params, "background value must be an integer") unless params_value.call("background").is_a? Integer
+      end
+    end
+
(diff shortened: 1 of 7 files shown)


MarkAZhang:
Hey Greg, that's a great point. I didn't think of that. Will change.

app/models/bulk_download.rb
```diff
@@ -23,6 +23,47 @@ class BulkDownload < ApplicationRecord
   attr_accessor :params
 
   validates :status, presence: true, inclusion: { in: [STATUS_WAITING, STATUS_RUNNING, STATUS_ERROR, STATUS_SUCCESS] }
+  validate :params_checks
+
+  def params_checks
```
tfrcarvalho: The "rails way" and leading to simpler, easy to test functions would be to validate each parameter independently.

app/models/bulk_download.rb
```diff
@@ -23,6 +23,47 @@ class BulkDownload < ApplicationRecord
...
+    set_params
+    params = self.params
+
+    params_value = lambda do |key|
```
tfrcarvalho: Isn't this just a `dig` (https://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig)?
i.e. `params.dig(key, "value") ?

app/models/bulk_download.rb
```diff
@@ -23,6 +23,47 @@ class BulkDownload < ApplicationRecord
...
+      return nil if params.nil?
+      # Each field param should be structured as
+      # { displayName: "FOO", value: "BAR" }
+      # field_params[key] should be ActionController::Parameters object that responds to :keys
+      params[key].respond_to?(:keys) ? params[key]["value"] : nil
```
tfrcarvalho: Generally: Isn't in better to just check if it is an Hash or subclass: `obj.is_a?(Hash)`? 
In theory, this accepts any object that has a keys method even if it does not do what you expect...

The params inside params are also ActionController::Parameters?

app/models/bulk_download.rb
```diff
@@ -23,6 +23,47 @@ class BulkDownload < ApplicationRecord
   attr_accessor :params
 
   validates :status, presence: true, inclusion: { in: [STATUS_WAITING, STATUS_RUNNING, STATUS_ERROR, STATUS_SUCCESS] }
+  validate :params_checks
+
+  def params_checks
+    set_params
```
tfrcarvalho: Params is typically to be used in a controller.... Doesnt this assume that it comes from a query?
What if we want to use BulkDownloads outside of an api call? Would this work?

MarkAZhang: I ended up re-writing how `params` is handled. Now there's a reader method that tries to load `params` from `params_json` if `params` isn't defined.

The previous approach was bad because you had to remember to call `set_params`. Now it happens transparently.

app/models/bulk_download.rb
MarkAZhang: It was a little trickier when it could be an `ActionController::Parameter`. I now convert from ACP to a Hash in the controller, so it's guaranteed to be a hash here. You're right, this method is no longer needed.

app/models/bulk_download.rb
MarkAZhang: Deleted this method.

app/models/bulk_download.rb
MarkAZhang: Discussed offline.

PR #2932, opened 2020-01-09
[Report page] Fix pathogen tags

j-x-han:
# Description
The entire lineage of a taxon was not being checked when tagging it for pathogens, resulting in some species not being properly tagged despite belonging to a pathogenic genus.
image

This change ensures that the entire available lineage of a taxon will be checked for pathogenicity and tag it properly.
image

# Tests

* Verified that species belonging to a pathogenic genus will be tagged as pathogenic.


cdebourcy:
Hi! I want to offer some clarification on point 1:
The pathogen list (https://github.com/chanzuckerberg/idseq-web/blob/fe12d973852b03bd1874e572c7935f4a177c9681/app/models/taxon_lineage.rb#L42) contains some species entries and some genus entries. This is because on the NIAID site some entire genera were listed as emerging priority pathogens, whereas other priority pathogens are restricted to certain species within a genus. So I believe Julie indeed captures the existing/desired behavior here. Of course, as part of the pending update to our pathogen list, we should reevalute the behavior. Curious to hear Katrina's thoughts.

j-x-han:
@gregdingle Regarding your comments:

1. I had no idea! I was comparing the new and old report pages and noticed that in the old version, all species within a pathogenic genus would be tagged accordingly (and digging through the logic behind tagging supports this). I'd love to hear @katrinakalantar 's thoughts and see what the most accurate approach would be.

2. After a quick look through the help center, I don't think the pathogen tags are discussed anywhere. Clicking on a tag in the UI does take you to this link though https://www.niaid.nih.gov/research/emerging-infectious-diseases-pathogens, which is where the list of pathogens is copied from (according to a comment, the version we use in the code was last updated 9/19/2018...).

3. That's a really good point, I'll add some tests! 😄

tfrcarvalho:
Funny that Shigella is one of the few that is list as `Shigella species` with the explicit species in the end. In the end, seems more of a visual/UX decision of what transmits the best idea for these pathogens and what feels redundant. So, maybe @happyimadesignr can chime in too.

katrinakalantar:
1. This is definitely a challenging issue. There are some genera where most or all of the known species within the genus are pathogenic (ie Shigella, Salmonella) - often these are noted in the literature on pathogenicity as i.e. "Salmonella spp". However, there may be new yet-classified species within a genus. In these cases, it's possible that we would "over-highlight" species within a genus. My thoughts are that if the goal of the pathogen tags is to highlight "points of interest" within the UI then I think it would be safe to highlight all species within one of these genera and leave it to the user to look up the relevant information for their particular species. If it is an uncommon or new species then they will have to look this information up anyways. 
2. We should definitely document this! As part of the revamped pathogen list project, it would be great to provide this list linked in docs in such a way that the user can clearly understand which pathogens are highlighted at the genus level and which are highlighted only at the species level.

j-x-han:
Thanks for the input everyone! Does it make sense to merge in this change for now to have the new report page match the old behavior and add documentation to the help center once we re-evaluate how to tag pathogens in the revamped pathogen list project?

spec/services/pipeline_report_service_spec.rb
```diff
@@ -366,4 +366,67 @@
...
+                           pipeline_run_ids: [
+                             create(:pipeline_run,
+                                    sample: create(:sample, project: create(:project))).id,
+                             create(:pipeline_run,
+                                    sample: create(:sample, project: create(:project))).id,
+                           ])
+      @report = PipelineReportService.call(@pipeline_run, @background.id)
+    end
+
+    it "should tag pathogenic genera correctly" do
```
gregdingle: Not to be too picky, but the documentation value of the test is minimal if the documented behavior is "X works correctly". Where are the behaviors we discussed? I can't tell as a reader. For example, all pathogens in a genus should be labeled if the genus is pathogenic.

j-x-han: Good point 😅 Thanks for the continuous push towards readability!

PR #2936, opened 2020-01-09
[Report page] Fix margins so genus and species columns are aligned

j-x-han:
# Description

The numerical columns of the genus and species rows would not be aligned on narrower screens (species row numbers pushed slightly to the right due to not correcting for an extra margin).

**Before:**
image

**Fixed:**
image

# Notes

Consolidated redundant class declarations.

# Tests

* Verified that the columns are now properly aligned.

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/report_table.scss
+++ b/app/assets/src/components/views/SampleViewV2/report_table.scss
@@ -121,6 +121,8 @@
 .speciesRow {
   .nameCell {
     font-weight: $font-weight-regular;
+    margin-left: $space-s;
+    margin-right: $space-xxxs;
   }
 }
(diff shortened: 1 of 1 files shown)


tfrcarvalho:
Thanks!
Interesting that the right margin does not affect the behavior for a wider window. Do you know why?

j-x-han:
@tfrcarvalho Name cells have a min-width of 300px, which doesn't seem to include the margins.

For genus rows, the name cell only has a margin-right: 14px and no left margin. Species rows have a margin-left: 10px to nest the rows underneath genus rows, but also had the margin-right: 14px by default, so those cells were actually 10px wider than the genus name cells.

PR #2937, opened 2020-01-10
[IDSEQ-1984] Change taxon search query type to `match` and remove wildcards

tfrcarvalho:
# Description

* Changes the logic of taxon search to use a query by `match`. This type of query matches multiple strings in a fuzzy way according to the operator specified (`and` in this case). `match` is the query type recommended by ElasticSearch for full-text search.
* Level of fuzziness is automatic computed based on the size of the query string
* Previously, we were using the `query_string` query type. Together with the use of wildcards made the results hard to understand.

# Notes

* Jira: https://jira.czi.team/browse/IDSEQ-1984

# Tests

* Tested search suggestions and for several pathogens and the results seem more logical and easy to understand.
* Query "wuhan mosquito" returns: 
Before:
```
Wuhan tick virus 2
Wuhan ledantevirus
```
Now: 
```
Wuhan Mosquito Virus 9
Wuhan Mosquito Virus 6
Wuhan Mosquito Virus 4
Wuhan mosquito orthophasmavirus 2
Wuhan mosquito orthophasmavirus 1
Wuhan Mosquito Virus 8
```

### DIFF
--- a/app/helpers/elasticsearch_helper.rb
+++ b/app/helpers/elasticsearch_helper.rb
@@ -30,22 +30,25 @@ module ElasticsearchHelper
       search_params = {
         size: ElasticsearchHelper::MAX_SEARCH_RESULTS,
         query: {
-          query_string: {
-            query: "*#{query}*",
-            fields: ["#{level}_name"],
+          match: {
+            "#{level}_name": {
+              query: "*#{query}*",
+              fuzziness: "auto",
+              operator: "and",
+            },
           },
         },
         aggs: {
           distinct_taxa: {
             terms: {
               field: "#{level}_taxid",
+              size: ElasticsearchHelper::MAX_SEARCH_RESULTS,
             },
           },
         },
       }
       search_response = TaxonLineage.search(search_params)
       search_taxon_ids = search_response.aggregations.distinct_taxa.buckets.pluck(:key)
-
       taxon_data = TaxonLineage
                    .where("#{level}_taxid" => search_taxon_ids)
                    .order(id: :desc)
--- a/app/helpers/elasticsearch_helper.rb
+++ b/app/helpers/elasticsearch_helper.rb
@@ -32,7 +32,7 @@ module ElasticsearchHelper
         query: {
           match: {
             "#{level}_name": {
-              query: "*#{query}*",
+              query: query,
               fuzziness: "auto",
               operator: "and",
             },
--- a/app/helpers/elasticsearch_helper.rb
+++ b/app/helpers/elasticsearch_helper.rb
@@ -27,6 +27,9 @@ module ElasticsearchHelper
     matching_taxa = []
     taxon_ids = []
     tax_levels.each do |level|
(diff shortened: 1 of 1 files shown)


MarkAZhang:
For your "wuhan" example, I'm glad that it returns more results, but I'm a little worried that the previous results (Wuhan tick virus 2) seem to have disappeared. Is that an issue?

Also, did you mean to leave the wildcard characters in? I ask because in your title and comment you mention "removing wildcards"

tfrcarvalho:
> For your "wuhan" example, I'm glad that it returns more results, but I'm a little worried that the previous results (Wuhan tick virus 2) seem to have disappeared. Is that an issue?

This was my error describing the example. The final example I used was actually "wuhan mosquito", but forgot to change the description.

> Also, did you mean to leave the wildcard characters in? I ask because in your title and comment you mention "removing wildcards"

No, thanks! Must have lost the change as I was cleaning the code. Updated.

app/helpers/elasticsearch_helper.rb
```diff
@@ -30,15 +30,19 @@ def taxon_search(query, tax_levels = TaxonCount::NAME_2_LEVEL.keys, filters = {}
       search_params = {
         size: ElasticsearchHelper::MAX_SEARCH_RESULTS,
         query: {
-          query_string: {
-            query: "*#{query}*",
-            fields: ["#{level}_name"],
+          match: {
```
gregdingle: I would like to see your comment in the code here:

"query by match. This type of query matches multiple strings in a fuzzy way according to the operator specified (and in this case). match is the query type recommended by ElasticSearch for full-text search."

Plus: "optimized with feedback from users" -- that's the really expensive part!

gregdingle: Also it would be nice to know how `MAX_SEARCH_RESULTS` was determined.

tfrcarvalho:
> Also, as a sidenote, we do have stored all the strings that people have searched by, and the number of results, which we can use to optimize search further.

Do not really see much point of optimizing here... but

app/helpers/elasticsearch_helper.rb
tfrcarvalho: > optimized with feedback from users

I don't think of it as optimized, but as a fix. The previous query was just not adequate for the use case and provided seemingly wrong results.

tfrcarvalho: Added a few extra comments.

app/helpers/elasticsearch_helper.rb
```diff
@@ -1,4 +1,9 @@
 module ElasticsearchHelper
+  # Limit maximum number of results for performance
+  # Since we currently might filter by sample and/or project only after the ElasticSearch,
+  # it should be large enough to avoid filtering everything out
+  # TODO(tiago): Eventually, this filtering should be moved to ElasticSearch, at which point we should reduce the max value
+  # (We can also move it to a parameter that clients of this functions can set)
```
gregdingle: Awesome, learned something new 😄

PR #2939, opened 2020-01-10
[Report page] Check results monitor instead of pipeline monitor for report status

j-x-han:
# Description

Use results monitor's results_finalized instead of pipeline_monitor's finalized?/completed? for report status. This should fix the issue where the report page displays an error rather than a loading state if all the pipeline jobs are checked and completed but the outputs are still waiting to be loaded (see https://chanzuckerbergteam.slack.com/archives/CM5K7MA03/p1578505508000700).

This should now match the check that the old report page did on the frontend to see if the report should display a loading state. https://github.com/chanzuckerberg/idseq-web/blob/476581d1259c0e8eee4b54e38ae26fda58ee9774/app/assets/src/components/views/SampleView/SampleView.jsx#L246

# Notes

The flow of logic for determining the displayed status of a report (found in `renderReport()` and `renderSampleMessage()` of SampleViewV2.jsx) is now as follows:
* If reportReady (taxon_counts output has loaded without errors), then display the report table.
* Otherwise, render a status message.
  * If the report is still loading, then display loading state.
  * If WAITING (results_finalized is false, i.e. results monitor is still waiting for some output state to be completed) and there's no errors, then display waiting/in progress state.
  * Otherwise, display an error/warning.

# Tests

* Verified that report page will display waiting, loading, errors, or the completed report table appropriately.


app/services/pipeline_report_service.rb
```diff
@@ -249,13 +249,15 @@ def get_pipeline_status(pipeline_run)
...
     pipeline_status = "WAITING"
-    # pipeline_run.completed? and pipeline_run.finalized? both indicate if the pipeline run
-    # is no longer being checked by the pipeline monitor (i.e. the run has completed),
-    # but .completed? includes extra logic for an old version of pipeline without run stages.
-    if pipeline_run.completed?
+    if pipeline_run.results_finalized?
       pipeline_status = "COMPLETE"
-    elsif pipeline_run.failed?
+    end
+    if pipeline_run.failed?
```
tfrcarvalho: So, can it be finalized AND failed? 
If so, we should do `if failed elsif finalized` instead. Seems easier to understand.

j-x-han: Yes, it can be both finalized and failed. Changed to `failed/elsif finalized`!

PR #2948, opened 2020-01-13
Ensure collection date metadata for humans returned as month-only.

MarkAZhang:
# Description

Before:
Screen Shot 2020-01-13 at 3 49 05 PM

After:
image

# Notes

* Previously, the metadata was returned as a Date, which was then coerced into a string when returned as JSON. However, when Dates are coerced to strings, the day is always included.
* `Metadatum.by_sample_ids` now returns strings instead of Dates, which can be unexpected. I gated this behavior on a flag `use_raw_date_strings`. Every place where this function is currently used then passes the metadata back to the front-end, but conceivably the function could have other uses in the future.

# Tests

* Updated tests.
* Verified that the front-end receives the expected data.

### DIFF
--- a/app/models/metadatum.rb
+++ b/app/models/metadatum.rb
@@ -325,10 +325,7 @@ class Metadatum < ApplicationRecord
           sample_metadata.map do |m|
             # When fetching metadata for a human sample for displaying on the front-end, we want 2001-01 for dates, not 2001-01-01.
             # date_validated_value is a Date object and will show the day when converted to a string. We use the original raw_value string instead.
-            Rails.logger.info("#{m.metadata_field.base_type} #{MetadataField::DATE_TYPE}")
-
             if m.metadata_field.base_type == MetadataField::DATE_TYPE && use_raw_date_strings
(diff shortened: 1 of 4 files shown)


app/models/metadatum.rb
```diff
@@ -313,12 +313,28 @@ def csv_template_value
...
-        [sample_id, Hash[sample_metadata.map { |m| [m.key.to_sym, m.validated_value] }]]
+        [
+          sample_id,
+          sample_metadata.map do |m|
+            # When fetching metadata for a human sample for displaying on the front-end, we want 2001-01 for dates, not 2001-01-01.
+            # date_validated_value is a Date object and will show the day when converted to a string. We use the original raw_value string instead.
+            Rails.logger.info("#{m.metadata_field.base_type} #{MetadataField::DATE_TYPE}")
+
+            if m.metadata_field.base_type == MetadataField::DATE_TYPE && use_raw_date_strings
+              Rails.logger.info("GOT IT #{m.raw_value}")
```
jshoe: assuming you meant to remove the `GOT IT`

PR #2952, opened 2020-01-14
[Report page] [Bug] Make species names lowercase

j-x-han:
# Description

As per IDSEQ-2074, genus names should be uppercase and species names should be lowercase, but all taxon names were capitalized on the new report page.
image

# Notes
`text-transform: capitalize` is applied on taxonNames by default to ensure genus rows are capitalized properly, which is then overridden with `text-transform: none` for species rows.

# Tests

* Verified that species names are in lowercase.

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/report_table.scss
+++ b/app/assets/src/components/views/SampleViewV2/report_table.scss
@@ -91,7 +91,6 @@
 
   .taxonContainer {
     .taxonName {
-      text-transform: capitalize;
       white-space: normal;
       &:hover {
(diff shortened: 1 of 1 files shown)


j-x-han:
@gregdingle 
It seems in the old report page, the non-specific rows weren't capitalized and there was a JIRA bug filed for that (IDSEQ-1749).
image

After using the method you suggested for capitalization, this is how the row now looks, which hopefully should be more natural casing 😄 
image

gregdingle:
Nice improvements!

PR #2959, opened 2020-01-15
Change sample row click behavior

gregdingle:
# Description

image

The current behavior is seriously annoying me and others because missing a checkbox opens the sample page and resets all checkboxes. The screenshot above shows the new target area. The rest of the row now selects the checkbox, which is easy to undo if unintentional. 

# Notes

* I did not try to fix the current bug where click drag selecting text causes `handleRowClick` to fire
* The meaning of our row click stats will change

# Tests

* Click on new target area, see sample page open
* Click on row, see checkbox

### DIFF
--- a/app/assets/src/components/views/discovery/DiscoveryView.jsx
+++ b/app/assets/src/components/views/discovery/DiscoveryView.jsx
@@ -657,6 +657,7 @@ class DiscoveryView extends React.Component {
     );
   };
 
+  // TODO (gdingle): fix me
   handleSampleSelected = ({ sample, currentEvent }) => {
     openUrl(`/samples/${sample.id}`, currentEvent);
   };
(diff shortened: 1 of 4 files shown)


app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -1036,7 +1036,6 @@ class DiscoveryView extends React.Component {
                 onMapLevelChange={this.handleMapLevelChange}
                 onMapMarkerClick={this.handleMapMarkerClick}
                 onMapTooltipTitleClick={this.handleMapTooltipTitleClick}
-                onSampleSelected={this.handleSampleSelected}
```
gregdingle: `handleSampleSelected` is still used for search

app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -1210,18 +1209,19 @@ class DiscoveryView extends React.Component {
...
-                  dim => dim.values,
-                  keyBy("dimension", dimensions)
-                )}
-                {...filters}
-                domain={domain}
-                onFilterChange={this.handleFilterChange}
-                allowedFeatures={allowedFeatures}
-              />
-            )}
+            {showFilters &&
```
gregdingle: this is just an incidental js prettier change

gregdingle:
That's a good alternative @tfrcarvalho . Gmail also has hover effects (which IDseq lacks) to make it obvious what the effect of the click will be. 

image

tfrcarvalho:
Additionally, we can consider persisting the checkbox selection (although this can be tricky) and if we go with click on row to select i would get rid of the box of the checkbox (just use the shaded checkmark when hovering).

happyimadesignr:
Hey @gregdingle, glad you're working on this! We should just expand to allow the invisible area around the checkbox to be a part of the checkbox click. It's more intuitive for the rest of the row to navigate you to the sample report.

gregdingle:
> allow the invisible area around the checkbox to be a part of the checkbox click

Thanks @happyimadesignr . Could you provide some guidance on how much area around the checkbox and whether there is an existing element that covers it?

happyimadesignr:
@gregdingle sure ill provide a mock

happyimadesignr:
Screen Shot 2020-01-15 at 2 32 31 PM

Here you go! That area around the checkbox, it'll help as they're clicking down that column. I like your suggestion to have a hover state, that will help, this hover state is aligned with the design system work!

jshoe:
is this one still in progress?

gregdingle:
I'm going to close this PR because I don't want to make it a priority right now. I only posted the PR because I hit the issue myself and I saw an easy fix. Anyone is welcome to pick up the JIRA story: https://jira.czi.team/browse/IDSEQ-1700 .

PR #2961, opened 2020-01-15
[Bulk Downloads] Filter out in-progress and failed samples selected by users

lvreynoso:
# Description

This PR changes the bulk downloads modal to automatically filter out any in-progress or failed samples that users have selected in a samples view. Users are given a warning indicating which samples will be left out of the download.

Also added is a backend route to validate sample ids selected by the user.

# Notes

- Since validating the sample ids does not actually create the bulk download, sanitation code is left in place, so any funny business should still be dealt with properly. This can all be consolidated when the extraneous review step is removed (my next priority).
- Parallelized the async requests made by the `ChooseStep` of the bulk downloads modal.
- Wrote a service for extra logic needed by the controller, per @gregdingle 's guide. 

# Tests

Used the `Todd Test` project, selected all samples, generated a bulk download.
Backend rspec tests added to `spec/bulk_downloads_controller_spec.rb`. It implicitly tests everything in the `BulkDownloadsValidationService` but a separate test can be written for it if needed.

Warning notification for samples that are filtered out:
Screen Shot 2020-01-21 at 1 46 01 PM
Screen Shot 2020-01-21 at 1 46 07 PM


app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -484,10 +503,26 @@ class ChooseStep extends React.Component {
...
-    const numSamples = selectedSampleIds.size;
+    return (
+      <div className={cs.notification}>
+        <Notification type="warn" displayStyle="flat">
+          <span className={cs.filteredSamplesWarning}>
+            {filteredSampleNames.length}{" "}
+            {filteredSampleNames.length > 1 ? "samples " : "sample "}
+            won't be included in this bulk download
+          </span>, because they are in progress or failed samples:{" "}
+          {filteredSampleNames.join(", ")}
```
gregdingle: I haven't seen the design but it seems like a bulleted list would be more appropriate here.

app/assets/src/components/views/bulk_download/choose_step.scss
```diff
@@ -166,3 +166,11 @@
     margin-top: 6px;
   }
 }
+
+.notification {
+  margin-bottom: 20px;
```
gregdingle: try to use our standard spacing from app/assets/src/styles/themes/_elements.scss when possible

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -154,6 +154,27 @@ class SamplesView extends React.Component {
...
+  // Queries the samples prop (an ObjectCollectionView) for the job
+  // status of each selected sample id. Returns a Map where the
+  // keys are sampleIds and values are the name and status of the sample.
+  getSelectedSampleStatuses = () => {
+    const { selectedSampleIds, samples } = this.props;
+    const statuses = new Map();
+    selectedSampleIds.forEach(id => {
+      const entry = samples.get(id);
+      // if unloaded samples are somehow selected
+      if (entry == undefined) {
```
gregdingle: prefer `=== undefined`, because... javascript 😉

app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
```diff
@@ -71,6 +74,11 @@ class BulkDownloadModal extends React.Component {
         selectedFields: {},
       });
     }
+    if (
+      prevProps.selectedSampleStatuses !== this.props.selectedSampleStatuses
```
gregdingle: FYI, this will only compare object identity and not values

lvreynoso: That's intentional. I want this to be triggered whenever `selectedSampleStatuses` changes.  `prevProps.selectedSampleStatuses` and `this.props.selectedSampleStatuses` will only point to different objects in React when any of the underlying values in the object change, otherwise they will point to the same object. Saves some work doing a deep comparison, since React has already done so.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
lvreynoso: This is from @happyimadesignr 's design.

happyimadesignr:
Ahh this is a good point once we're seeing it with the real data. The example Mark is showing might be too much though. I'll look across to see what other patterns we have.

happyimadesignr:
What about this pattern, we use it in the "your samples are going public" notification @MarkAZhang @lvreynoso 
Screen Shot 2020-01-16 at 9 45 10 AM

lvreynoso:
I like that pattern. Compact, but you can scroll to see all the names if you really want to. Along with giving the notification a max-height, could we bump up the max-height on the whole modal? Wouldn't make much of a difference on smaller screens but on the screens I've been using it's awfully small.

happyimadesignr:
@lvreynoso I was hoping to bump up the max height of the modal too!

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -154,6 +154,27 @@ class SamplesView extends React.Component {
     }
   }
 
+  // Queries the samples prop (an ObjectCollectionView) for the job
+  // status of each selected sample id. Returns a Map where the
+  // keys are sampleIds and values are the name and status of the sample.
+  getSelectedSampleStatuses = () => {
```
MarkAZhang: Does this work if there are 1000 samples on the page, and `ObjectCollectionView` hasn't loaded all of them yet? In that case, I think everything that's not yet loaded will be marked as `loading`.

Unfortunately, there is a way to select unloaded samples, and this is with the `Select All` checkbox. Select All is an important thing we need to support for ease-of-use. If there isn't an easy way to get the sample statuses here, you might need to make an API call i nBulkDownloadModal after all.

MarkAZhang:
@lvreynoso great idea to bump up the modal height. FYI there's a list of UX burndown items in https://czi.quip.com/srRfArwDfamx/Bulk-Downloads-Burndown (under Form Jenn). The modal height increase is in there. Just to give you a sense of the UX tweaks we've discussed but haven't implemented yet.

app/assets/src/components/views/samples/SamplesView.jsx
lvreynoso: You're right, I had to rewrite things and add a route on the backend.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -479,25 +416,73 @@ class ChooseStep extends React.Component {
...
+          {invalidSampleNames.length} sample
+          {invalidSampleNames.length > 1 ? "s" : ""} won't be included in the
+          bulk download
+        </span>, because they are in progress or failed samples:
+      </div>
+    );
+
+    const content = (
+      <div className={cs.messageContainer}>
+        {invalidSampleNames.map((name, index) => {
```
MarkAZhang: I would recommend passing the list of names (as an array of strings) to CompactListNotification, instead of rendering the list here and then passing the `messageContainer` node to CompactListNotification.

If the purpose of the compact "list" notification is to display a list of items, then I would pass the list and let the component figure out how to render the list (instead of rendering it here).

You can also put the styling for `messageContainer` and `messageLine` in CompactListNotification.

app/assets/src/components/views/bulk_download/choose_step.scss
```diff
@@ -175,10 +175,41 @@
...
+  @include font-body-xxs;
+  margin: $space-s 0;
+  background-color: $warning-bg;
+}
+
+.messageContainer {
+  max-height: 150px;
+  overflow: auto;
+  background-color: $warning-bg;
+  margin-left: 30px; // to align text with icon
```
MarkAZhang: It's a little weird to have this value depend on a css value in another component. You could solve this by moving the rendering of the list of strings into compact list notification as I describe above.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -395,10 +416,74 @@ class ChooseStep extends React.Component {
...
+        displayStyle={"flat"}
+      />
+    );
+  };
+
+  renderValidationError = () => {
+    const { validationError } = this.props;
+
+    const header = (
+      <div className={cs.header}>There was an error fetching sample data:</div>
```
MarkAZhang: I would reword this to "an unexpected error occurred when verifying your selected samples" and not show the `validationError`. I think a phrase like "fetching sample data" would be confusing to the user, who has no idea we're   doing all this validation behind-the-scenes.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -395,10 +416,74 @@ class ChooseStep extends React.Component {
...
+          return (
+            <div key={index} className={cs.messageLine}>
+              {name}
+            </div>
+          );
+        })}
+      </div>
+    );
+
+    return (
```
MarkAZhang: What happens if all selected samples end up being invalid? Perhaps we should show an error and not allow the user to create the bulk download. (perhaps even hide the Continue button)

app/assets/src/components/views/bulk_download/ChooseStep.jsx
lvreynoso: What if I just changed the name the `CompactNotification` or something similar? I thought about doing this, but I decided that having more flexibility over what to display in the toggleable container would be more useful.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
lvreynoso: My thinking on this is that showing the error would be useful, since if `validationError` does have a value it's either a message indicating that the user selected more samples than are allowed in a bulk download - a limit that as far as I can tell they have no other way of knowing about - or something more pernicious has occured, and it asks the user to contact us for more help. Seems pretty valuable to me.

Agreed on the rewording.

app/controllers/bulk_downloads_controller.rb
```diff
@@ -27,6 +27,33 @@ def types
...
+  # query strings longer than a certain amount (1024 * 10 chars), which causes
+  # trouble with projects with a large number of samples.
+  def validate
+    queried_sample_ids = params[:sampleIds]
+    max_samples_allowed = get_app_config(AppConfig::MAX_SAMPLES_BULK_DOWNLOAD)
+
+    validator = BulkDownloadsValidationService.new(queried_sample_ids, current_user, max_samples_allowed)
+
+    begin
+      id_validation_info = validator.validate_sample_ids()
```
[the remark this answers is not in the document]
lvreynoso: You're right, I didn't catch that I had made this redundant.

app/services/bulk_downloads_validation_service.rb
```diff
@@ -0,0 +1,85 @@
...
+  private ### *** PRIVATE METHODS *** ###
+
+  # Returns only the samples viewable by the current user.
+  # Raises an error if more than the max samples allowed are passed.
+  def validate_sample_access(sample_ids, user)
+    # Max samples check.
+    max_samples_allowed = @max_samples
+
+    # Max samples should be string containing an integer, but just in case.
+    if max_samples_allowed.nil?
```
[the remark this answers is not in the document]
lvreynoso: I don't think there's a front-end check - I never encountered one even when I selected over a thousand samples. The front-end wouldn't know what max samples is set to anyways since it's configured in the backend right? But yeah there's no reason to check this twice so I'll get rid of it.

app/services/bulk_downloads_validation_service.rb
```diff
@@ -0,0 +1,85 @@
...
+      LogUtil.log_err_and_airbrake("BulkDownloadsImproperAccessEvent: User made bulk download request for samples they don't have access to: #{private_samples}")
+    end
+
+    return viewable_samples
+  end
+
+  def get_finalized_sample_ids(samples)
+    # Gets the first pipeline runs for multiple samples in an efficient way.
+    created_dates = PipelineRun.select("sample_id, MAX(created_at) as created_at").where(sample_id: samples.pluck(:id)).group(:sample_id)
+    valid_pipeline_runs = PipelineRun
```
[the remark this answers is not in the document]
lvreynoso: Right, I agree. I could modify the code in the `PipelineRunsHelper` into something that meets the requirements for both routes and then call it from the controller, passing the samples from the validation service. That would fit best practices better I think.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
MarkAZhang: That would also work! I would call it `ExpandableNotification` or `AccordionNotification` (since it uses an Accordion). I think you should still move some of the styling into this component, rather than have the caller do the styling. (styling such as the left-margin and the background color, as we discussed)

app/services/bulk_downloads_validation_service.rb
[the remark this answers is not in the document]
MarkAZhang: There is a front-end check but it's disabled for admins. You can look at `handleBulkDownloadModalOpen` in `SamplesView`. The MAX_SAMPLES value is passed to the front-end via the UserContext object. (search for `user_context`)

app/assets/src/components/views/bulk_download/ChooseStep.jsx
lvreynoso: Done

PR #2962, opened 2020-01-15
New host organism info message on review step

gregdingle:
# Description

image

This adds the above message to tell the user how host subtraction works depending on the host. There are 4 states of the message depending on the number of hosts and matching known hosts or explicit ERCC only. I put them all in the playground page. 

image

**This feature is admin only while we iterate on it**

# Notes

* This feature is admin only while we iterate on it
* The unknown hosts will only be relevant when we start to allow plain text in host input

# Tests

* test admin only to show the message
* test all states in playground
* test positioning in review step

### DIFF
--- a/app/assets/src/components/views/SampleUploadFlow/HostOrganismMessage.jsx
+++ b/app/assets/src/components/views/SampleUploadFlow/HostOrganismMessage.jsx
@@ -1,6 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { keys, countBy, groupBy } from "lodash/fp";
+import { keys, countBy, groupBy, mapValues, keyBy, map } from "lodash/fp";
 
 const MATCH = "MATCH";
 const NO_MATCH = "NO_MATCH";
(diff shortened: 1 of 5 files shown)


app/assets/src/components/views/playgrounds/PlaygroundComponents.jsx
```diff
@@ -24,13 +25,49 @@ export default class PlaygroundComponents extends React.Component {
...
   renderComponent = () => {
     const { currentTab } = this.state;
-    if (currentTab === "Geo Search Box") {
+
+    if (currentTab === TABS[0]) {
+      return [
+        // one match
+        <HostOrganismMessage
+          key="1"
+          hostGenomes={[{ name: "Human", id: 1 }]}
```
jshoe: I'm concerned about having the host_genome_ids here. In staging vs. prod I'm pretty sure the host_genome_ids vary already for Human/ERCC/Mosquito. Is there another way?

jshoe: Oh, I see this is just the playground. Doesn't matter then?

gregdingle: That's correct, the values in the playground are arbitrary, though I imitated prod values.

app/assets/src/components/views/SampleUploadFlow/HostOrganismMessage.jsx
```diff
@@ -0,0 +1,84 @@
...
+import { keys, countBy, mapValues, keyBy, map } from "lodash/fp";
+
+import cs from "./host_organism_message.scss";
+
+/**
+ * Shows a message depending on whether the selected host matches a
+ * IDseq-supported host.
+ */
+export default class HostOrganismMessage extends React.Component {
+  static propTypes = {
```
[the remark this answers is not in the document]
gregdingle: I just wanted to see if this would work and if anyone would notice 😉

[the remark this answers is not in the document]
gregdingle: I'll fix this in next PR.

PR #2967, opened 2020-01-16
Add ability to disable pipeline processing after sample upload.

MarkAZhang:
# Description

Add a checkbox during the review step of sample uploading.

Screen Shot 2020-01-16 at 1 51 02 PM

This checkbox is gated by a user setting. The user will only see this option if the setting is set to true.

If the checkbox is selected, the sample will skip processing and show a special blue message to indicate that processing was skipped.

Screen Shot 2020-01-16 at 1 52 28 PM

Screen Shot 2020-01-16 at 1 52 41 PM

# Notes

* Also updated how UserSettings are parsed so that "value" is returned as a boolean instead of a string (but the value is stored in the database as a string)

# Tests

* Verified that the user setting guards the feature properly.
* Verified that for all three upload types, the checkbox prevents the sample from kicking off a pipeline run. Verified that it works for multiple samples.
* Updated tests.

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -36,6 +36,7 @@ class Sample < ApplicationRecord
   UPLOAD_ERROR_S3_UPLOAD_FAILED = "S3_UPLOAD_FAILED".freeze
   UPLOAD_ERROR_LOCAL_UPLOAD_STALLED = "LOCAL_UPLOAD_STALLED".freeze
   UPLOAD_ERROR_LOCAL_UPLOAD_FAILED = "LOCAL_UPLOAD_FAILED".freeze
+  DO_NOT_PROCESS = "DO_NOT_PROCESS".freeze
 
   TOTAL_READS_JSON = "total_reads.json".freeze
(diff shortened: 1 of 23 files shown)


cdebourcy:
This is awesome! Super helpful -- please also check an ongoing discussion on https://jira.czi.team/browse/IDSEQ-1946

app/models/sample.rb
```diff
@@ -717,6 +718,11 @@ def kickoff_pipeline
     # only kickoff pipeline when no active pipeline_run running
     return unless pipeline_runs.in_progress.empty?
 
+    if do_not_process
```
gregdingle: rails tip: you can write...

`if do_not_process?`

rails will add the `?` method

tfrcarvalho:
> This is awesome! Super helpful -- please also check an ongoing discussion on [jira.czi.team/browse/IDSEQ-1946](https://jira.czi.team/browse/IDSEQ-1946)

@cdebourcy this actually solves the QA problem, not the on mentioned on the ticket, i believe.

PR #2968, opened 2020-01-17
Refine host genome messages

gregdingle:
# Description

This changes the messages according to the latest designs at https://chanzuckerberg.invisionapp.com/d/main/#/console/19023327/401090624/preview . 

All combinations are loaded on one page at /playground/components/. See:

image

# Notes

* Some of the styles differ because the default `Notification` component styles differ. I opted to use the defaults whenever available. 
* I discovered an oddity about our global CSS which is that we remove link styling with `!important` which leads to 10+ components such as this one redefining the same style
* I modified `Notification` to make it actually show an icon in the `info` case. I'm not sure why there was one, I don't see the benefit and I checked all callers of the component to be sure. 

# Tests

* Load in playground, compare with designs


app/assets/src/components/views/SampleUploadFlow/HostOrganismMessage.jsx
```diff
@@ -37,30 +56,72 @@ export default class HostOrganismMessage extends React.Component {
...
+            <strong>{host}</strong> Host Organism
+          </span>
+        )},
       </span>
     );
   }
 
   renderManyHosts(uniqHosts) {
+    // Strangely, the " " spaces are not being ending up in the header so
+    // we use &nbsp;
```
lvreynoso: You can also insert a string in curly braces, e.g. `{"  "}` to force React to render an exact amount of spaces or other characters, just fyi

app/assets/src/components/views/SampleUploadFlow/HostOrganismMessage.jsx
```diff
@@ -2,30 +2,49 @@ import React from "react";
...
+        {!this.hasMatch(host) &&
+          " We don't have any hosts matching your selection."}
+        {this.renderTextLine(host, count)} {this.renderLearnMoreLink()}.
+      </Notification>
+    );
+  }
+
+  renderLearnMoreLink() {
+    return (
+      <a
```
tfrcarvalho: Didnt we an element for this - `ExternalLink`?

app/assets/src/components/views/SampleUploadFlow/host_organism_message.scss
```diff
@@ -15,12 +17,25 @@
 .messageLine {
   margin: 0 $space-s;
   padding: $space-s;
-
-  &:nth-child(1) {
-    margin-top: $space-s;
-  }
+  font-size: $font-size;
```
tfrcarvalho: Also font-mixins?

app/assets/src/components/views/SampleUploadFlow/host_organism_message.scss
```diff
@@ -15,12 +17,25 @@
...
+  margin: 0 44px; // align to right of icon
```
tfrcarvalho: Can this be a space variable?

app/assets/src/components/views/SampleUploadFlow/host_organism_message.scss
```diff
@@ -15,12 +17,25 @@
...
+  &:nth-child(even) {
+    background-color: $info-bg;
+  }
+
+  &:last-child {
+    margin-bottom: $space-m;
+  }
+}
+
+// NOTE: we should really have a Link component
```
tfrcarvalho: We do have one... If does not satisfy your needs you should either make it more generic/configurable or add a new one.

app/assets/src/components/views/SampleUploadFlow/host_organism_message.scss
```diff
@@ -15,12 +17,25 @@
...
+.link {
+  // Overwrite <a> inherit rule in header.scss
+  color: $primary-light !important;
```
tfrcarvalho: In any case color is a very specific attribute that is not that common throughout the page for links. Feel free to check with Jenn what is the most common value for color.

app/assets/src/components/views/SampleUploadFlow/HostOrganismMessage.jsx
gregdingle: Yea, that's what I first tried and it strangely didn't work in this component.

app/assets/src/components/ui/notifications/Notification.jsx
```diff
@@ -10,6 +10,7 @@ class Notification extends React.Component {
     switch (type) {
       case "warn":
       case "error":
+      case "info":
```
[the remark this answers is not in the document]
gregdingle: yes, good catch!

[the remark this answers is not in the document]
gregdingle: yes, good catch!

app/assets/src/components/views/SampleUploadFlow/host_organism_message.scss
```diff
@@ -2,11 +2,13 @@
...
 @import "~styles/themes/typography";
 
-.messageContainer {
-  margin: $space-s 0;
-  padding: $space-s;
+.listContainer {
   background-color: $info-bg;
+}
 
+.messageContainer {
```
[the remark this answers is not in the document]
gregdingle: 👍  @happyimadesignr  any guidance on what mixin? `font-body-xs` ?

app/assets/src/components/views/SampleUploadFlow/host_organism_message.scss
gregdingle: No, because it depends on `notification.scss` which itself does not use variables. Would you recommend changing `notification.scss` to do so?

app/assets/src/components/views/SampleUploadFlow/host_organism_message.scss
gregdingle: Which link component are you referring to? `ExternalLink`? 

In any case, the common property that would be valuable is the color, and it's not clear what the generic color should be. There appear to be several shades of blue in use, and there is a global override to force no color on links. 

I'll remove the comment as it may be premature.

app/assets/src/components/views/SampleUploadFlow/host_organism_message.scss
tfrcarvalho: Yes, if simple enough... in general, something we try to do is keep updating the existing components for new methods/rules. But it is up to you to evaluate the effort.

app/assets/src/components/views/SampleUploadFlow/host_organism_message.scss
tfrcarvalho: SGTM, my comment was mostly triggered by the comment.

app/assets/src/components/ui/notifications/Notification.jsx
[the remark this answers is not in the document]
gregdingle: I discovered there were already two existing info icons. In consultation with @happyimadesignr , I renamed one of them and deprecated the other and added a new one.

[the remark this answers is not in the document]
gregdingle: See updated screenshot above. I also tested the renamed icon.

PR #2971, opened 2020-01-18
Add separate sample limit for original input files download type.

MarkAZhang:
# Description

Add a separate sample limit for original input files download type.

Show warning via hover tooltip if user exceeds limit. Same pattern as when user isn't the original uploader of the samples.

Screen Shot 2020-01-17 at 5 37 33 PM

# Tests

* Manually test that upload behaves as expected. Other download types are not affected by new sample limit.
* Update tests.

### DIFF
--- a/app/assets/src/components/common/UserContext.jsx
+++ b/app/assets/src/components/common/UserContext.jsx
@@ -3,6 +3,8 @@ import React from "react";
 export const UserContext = React.createContext({
   admin: false,
   allowedFeatures: new Set(),
+  appConfig: {},
+  userSettings: {},
 });
 // Name to show in DevTools
 UserContext.displayName = "UserContext";
--- a/app/assets/src/components/views/bulk_download/ChooseStep.jsx
+++ b/app/assets/src/components/views/bulk_download/ChooseStep.jsx
@@ -305,7 +305,7 @@ class ChooseStep extends React.Component {
       selectedSampleIds,
     } = this.props;
     const { allSamplesUploadedByCurrentUser } = this.state;
-    const { admin, maxSamplesBulkDownloadOriginalFiles } = this.context || {};
+    const { admin, appConfig } = this.context || {};
 
     const selected = selectedDownloadTypeName === downloadType.type;
     let disabled = false;
@@ -322,12 +322,14 @@ class ChooseStep extends React.Component {
       }, you must be the original uploader of all selected samples.`;
     } else if (
       downloadType.type === "original_input_file" &&
-      maxSamplesBulkDownloadOriginalFiles &&
-      selectedSampleIds.size > maxSamplesBulkDownloadOriginalFiles &&
+      appConfig.maxSamplesBulkDownloadOriginalFiles &&
+      selectedSampleIds.size > appConfig.maxSamplesBulkDownloadOriginalFiles &&
       !admin
     ) {
       disabled = true;
-      disabledMessage = `No more than ${maxSamplesBulkDownloadOriginalFiles} samples
+      disabledMessage = `No more than ${
+        appConfig.maxSamplesBulkDownloadOriginalFiles
+      } samples
         allowed for ${downloadType.display_name} downloads`;
     }
 
--- a/app/assets/src/components/views/bulk_download/ChooseStep.jsx
+++ b/app/assets/src/components/views/bulk_download/ChooseStep.jsx
@@ -299,9 +299,13 @@ class ChooseStep extends React.Component {
   };
 
   renderDownloadType = downloadType => {
-    const { selectedDownloadTypeName, onSelect } = this.props;
+    const {
+      selectedDownloadTypeName,
+      onSelect,
+      selectedSampleIds,
+    } = this.props;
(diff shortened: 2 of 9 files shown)


app/helpers/application_helper.rb
```diff
@@ -34,7 +34,8 @@ def user_context
     {
       admin: current_user ? current_user.role == 1 : false,
       allowedFeatures: current_user && current_user.allowed_feature_list,
-      maxSamplesBulkDownload: get_app_config(AppConfig::MAX_SAMPLES_BULK_DOWNLOAD),
+      maxSamplesBulkDownload: get_app_config(AppConfig::MAX_SAMPLES_BULK_DOWNLOAD).to_i,
```
tfrcarvalho: I dont think these should be parameters of this context, since it is a user specific context.
I believe we should create an AppContext for this... wdyt?

app/assets/src/components/common/UserContext.jsx
```diff
@@ -3,6 +3,8 @@ import React from "react";
 export const UserContext = React.createContext({
   admin: false,
   allowedFeatures: new Set(),
+  appConfig: {},
```
gregdingle: this is still abusing the concept of `UserContext` because it is does not depend on the user... any comment?

app/assets/src/components/views/bulk_download/ChooseStep.jsx
```diff
@@ -316,6 +320,17 @@ class ChooseStep extends React.Component {
       disabledMessage = `To download ${
         downloadType.display_name
       }, you must be the original uploader of all selected samples.`;
+    } else if (
+      downloadType.type === "original_input_file" &&
+      appConfig.maxSamplesBulkDownloadOriginalFiles &&
```
gregdingle: don't think you need this line

app/models/app_config.rb
```diff
@@ -12,6 +12,9 @@ class AppConfig < ApplicationRecord
   USE_AUTH0_FOR_NEW_USERS = 'use_auth0_for_new_users'.freeze
   # The maximum number of samples that can be part of one bulk download.
   MAX_SAMPLES_BULK_DOWNLOAD = 'max_samples_bulk_download'.freeze
+  # The maximum number of samples that can be part of an original input files bulk download.
+  # Original input file downloads are significantly bigger and slower than other downloads, so a separate limit is needed.
+  MAX_SAMPLES_BULK_DOWNLOAD_ORIGINAL_FILES = 'max_samples_bulk_download_original_files'.freeze
```
gregdingle: I'd love to see here what the initial values are and how they were chosen, for both configs.

app/assets/src/components/views/samples/SamplesView.jsx
```diff
@@ -473,18 +473,20 @@ class SamplesView extends React.Component {
...
-    if (!maxSamplesBulkDownload) {
+    const { appConfig, admin } = this.context || {};
+    if (!appConfig.maxSamplesBulkDownload) {
       this.setState({
         bulkDownloadButtonTempTooltip:
           "Unexpected issue. Please contact us for help.",
       });
     } else if (
-      this.props.selectedSampleIds.size > parseInt(maxSamplesBulkDownload) &&
+      this.props.selectedSampleIds.size > appConfig.maxSamplesBulkDownload &&
```
gregdingle: please add a comment here that `maxSamplesBulkDownloadOriginalFiles` is handled later by disabling the option

app/assets/src/components/common/UserContext.jsx
MarkAZhang: Yeah, Tiago and I discussed this offline. We're going to rename `UserContext` to `AppContext` and possibly rename variables like `admin` to be more clear with that renaming (for example, `admin` might become `isUserAdmin`). Didn't want to do in this PR.

app/assets/src/components/views/bulk_download/ChooseStep.jsx
MarkAZhang: My thinking was that there's a chance (however small) that `appConfig. maxSamplesBulkDownloadOriginalFiles` is undefined, and I'd rather make it explicit what the behavior is in that case rather than relying on the code to handle `undefined` correctly. Does that sound reasonable?

app/models/app_config.rb
MarkAZhang: Max samples will be 1000 and max samples original files will be 100. This was based on stress tests which you can find here: https://czi.quip.com/gYUwA4tL0aIr/Bulk-Download-Scalability#faOACABnDFb

I considered how many samples the projects in IDseq contained. I wanted users to be able to bulk download data for all samples in their project for the vast majority of projects, instead of having to split their downloads up. The stress tests verified that the size and download time for the files was roughly reasonable.

Original files is much smaller because based on testing, it tends to be much bigger and slower than other downloads. Also, original files seems to be a bit of a special case. (it's uploader-only, etc.)

MarkAZhang:
Regarding deployment, I planned to add the new app config value to staging and prod when I submitted the PR. The app config value will just sit there unused until the deployment happens.

PR #2972, opened 2020-01-18
Generalize TaxonContigSelect to handle reads as well.

MarkAZhang:
# Description

This finishes a partial refactor that occurred during bulk downloads and removes duplicate logic.

The behavior is unchanged.

The refactored component is this one:

Screen Shot 2020-01-17 at 7 12 17 PM

# Tests
* Verified that the selects behave as expected (and other dropdowns were unaffected).
* Verified that bulk downloads of various types can still be created.
* Updated tests.


app/helpers/samples_helper.rb
```diff
@@ -588,7 +588,7 @@ def self.samples_by_metadata_field(sample_ids, field_name)
 
   # For each taxon, count how many samples have taxon counts for that taxon.
   # Add these counts to the taxon objects.
-  def augment_taxon_list_with_sample_count(taxon_list, samples)
+  def augment_taxon_list_with_sample_count_reads(taxon_list, samples)
```
tfrcarvalho: This is still a sample count correct? Find it a bit confusing name. Seems like you counting the reads.
If so, do we really need to add `reads` in the end. I would either leave the name as it was or do `augment_taxon_list_with_sample_with_reads_count`.

PR #2975, opened 2020-01-21
Fix bug with capitalized emails in 'new user new project' flow

jshoe:
### Description
- There was a bug with capitalized emails in the 'new user new project' flow: https://chanzuckerberg.airbrake.io/projects/158403/groups/2658385090763832215
- The main problem is that Auth0 automatically lowercases (or strips case-sensitivity) emails, so we need to be consistent with them.
- We made a product decision to automatically lowercase as well and have the same behavior. This was favored over generating error messages that would add user friction.

### Tests
- Tested adding a new user on a new project with a capitalized email successfully.
- Tested adding a new user via the admin form successfully.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -464,7 +464,7 @@ class ProjectsController < ApplicationController
   end
 
   def add_user
-    @user = User.find_by(email: params[:user_email_to_add])
+    @user = User.find_by(email: params[:user_email_to_add].downcase)
     if @user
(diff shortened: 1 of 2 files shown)


gregdingle:
Looks good to me. 

FYI, there is a multi html email input which could be useful here: https://www.w3schools.com/tags/att_input_type_email.asp .

jshoe:
- There's already a validator in user.rb for the capitalization 👍 
- For the controller methods, there's a little bit of room for consolidation but the functionality is slightly different for "admin making a new user" vs "user getting added to a new project". But this is good to revisit when we need to touch these flows again

PR #2976, opened 2020-01-21
Add Rails config for new `sandbox` environment

jshoe:
### Description
- This PR adds the config required for the new `sandbox` environment, although there may still be follow-up PRs.
- Adds the mustard-colored sandbox favicon.
- `environments/sandbox.rb` just calls `staging.rb` for zero duplication.

### Tests
- This is deployed to `sandbox.idseq.net` and the site/database/logging settings are behaving as expected.


MarkAZhang:
Right now the cluster name is still `idseq-sandbox-ecs`, not `sandbox`, so if you try to run this script with `deploy.sh sandbox sha-xxx` it'll say `cluster not found`. Is that intentional?

jshoe:
@MarkAZhang ah yes that is intentional because I haven't figured out how to rename the cluster in-place / might rename staging/prod in the other direction. So that's part of the follow-up soon :)

PR #2977, opened 2020-01-21
Specify explicit ordering of queues for Resque workers.

MarkAZhang:
# Description

Verified in sandbox environment that queue of Resque workers is set correctly.

Screen Shot 2020-01-21 at 3 49 07 PM

# Notes

* Prior to this change, the priority of the queues was determined by alphabetical order. (that's the behavior of `*`). With this change, we move bulk download to be lowest priority.

### DIFF
--- a/balances.json
+++ b/balances.json
@@ -3,6 +3,5 @@
   "project": "idseq",
   "env": "staging",
   "name": "web",
-  "region": "us-west-2",
-  "resque_queues": "data_migration,precache_report_info,q03_pipeline_run,transfer_basespace_files,generate_bulk_download,*"
+  "region": "us-west-2"
 }
--- a/balances.json
+++ b/balances.json
@@ -3,5 +3,6 @@
   "project": "idseq",
   "env": "staging",
   "name": "web",
-  "region": "us-west-2"
+  "region": "us-west-2",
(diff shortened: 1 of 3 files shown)


balances.json
```diff
@@ -3,5 +3,6 @@
   "project": "idseq",
   "env": "staging",
   "name": "web",
-  "region": "us-west-2"
+  "region": "us-west-2",
+  "resque_queues": "data_migration,precache_report_info,q03_pipeline_run,transfer_basespace_files,generate_bulk_download,*"
```
gregdingle: what is `balances.json` for and why not put the value in `czecs-resque.json` directly?

MarkAZhang:
Good point, I'll remove the variable from `docker-compose.yml`. There isn't an easy way to share the variable from the json files, but the queue priority shouldn't matter for development anyways.

Re-did the priorities. "data migrations" are rare one-off scripts that we manually run to migrate data, so having that be highest-priority seems fine. Put "precache_report_info" last as you recommended. I agree that q03_pipeline_run is outdated and contains way too many things, but would like to defer that refactor for now.

Good point about `czecs-resque.json`. Will just put the variable directly in there.

PR #2980, opened 2020-01-22
[Report Page] Only show delete button if the sample is deletable

tfrcarvalho:
# Description

* Deletable if user is admin, there are not pipeline runs OR all pipeline runs have failed
* Ensure that the `Sample.deletable?` method return a boolean

# Notes

* This change has no effects on the legacy page

# Tests

* Check that admin sees the delete button for sample (currently only while the report is not ready)
* Check that normal user only see delete button for sample:
	- the user owns AND
	- have no successful pipeline runs


app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -1100,19 +1100,15 @@ export default class SampleViewV2 extends React.Component {
           <div className={cs.sampleViewHeader}>
             <SampleViewHeader
               backgroundId={selectedOptions.background}
+              deletable={sample ? sample.deletable : false}
               editable={sample ? sample.editable : false}
               onDetailsClick={this.toggleSampleDetailsSidebar}
               onPipelineVersionChange={this.handlePipelineVersionSelect}
               pipelineRun={pipelineRun}
+              pipeilneRunStatus={reportMetadata.pipeilneRunStatus}
```
j-x-han: Typo! Should be `pipelineRunStatus={reportMetadata.pipelineRunStatus}`

tfrcarvalho: Good catch... actually I end up not using this variable... removed

PR #2981, opened 2020-01-22
Fix homo sapiens in search

gregdingle:
# Description

We filter out homo sapiens from the reports but the species was still appearing in search suggestions. 

image

# Notes

If it is critical that we never show homo sapiens anywhere in the UI, we should consider some deeper changes, such as
* Not writing to taxon_counts
* Not writing to elasticsearch
* Removing homo sapiens from taxon_lineages
* Changing active record default queries for taxon_counts

# Tests

Check search results before and after, on both report page and DD page.


app/helpers/elasticsearch_helper.rb
```diff
@@ -5,6 +5,7 @@ module ElasticsearchHelper
   # TODO(tiago): Eventually, this filtering should be moved to ElasticSearch, at which point we should reduce the max value
   # (We can also move it to a parameter that clients of this functions can set)
   MAX_SEARCH_RESULTS = 50
+  HOMO_SAPIEN_TAX_ID = 9606
```
tfrcarvalho: Can you create a common place to put this constant? 
I wonder where it makes more sense... `TaxonLineage`?

gregdingle: Yes, good idea. And looking now there is actually a `HOMO_SAPIENS_TAX_ID` already in `TaxonLineage`... which contains two IDs!

Any idea why? And why this constant wasn't used in report helper?

tfrcarvalho: Not sure... probably missed at the time...

app/controllers/phylo_trees_controller.rb
```diff
@@ -118,6 +118,7 @@ def show
     end
   end
 
+  # NOTE: this is used by the report page as well as phylo trees
```
tfrcarvalho: There is a Jira ticket to fix this soon: https://jira.czi.team/browse/IDSEQ-2127

gregdingle: thanks, added a comment

PR #2984, opened 2020-01-23
Rank host genomes options by popularity 

gregdingle:
# Description

image

As part of the host genome change, we want to order by popularity instead of alphabetical. 

# Tests

* Upload a sample, see host genomes in ranked order
* console log counts, see correct counts


db/migrate/20200123002424_add_sample_count_to_host_genome.rb
```diff
@@ -0,0 +1,9 @@
+class AddSampleCountToHostGenome < ActiveRecord::Migration[5.1]
+  def up
```
jshoe: I'd recommend also having the `down` function

PR #2985, opened 2020-01-23
Change host genome default to ERCC only

gregdingle:
# Description

This changes the model definition to make the default files of new host genomes to be ERCC only, in preparation for plain text host genomes. 

# Notes

* I took the opportunity to add some validation and clean up of the model. 

# Tests

* run migration, see all existing records validate
* run new rspec test


app/models/host_genome.rb
```diff
@@ -2,14 +2,39 @@ class HostGenome < ApplicationRecord
...
+
   validates :taxa_category, inclusion: { allow_blank: true, in: [
     # From "Sample Type Groupings" at
     # https://docs.google.com/spreadsheets/d/1_hPkQe5LI0Zw_C0Ls4HVCEDsc_FNNVOaU_aAfoaiZRE/
     "human",
     "insect",
   ], }, if: -> { respond_to?(:taxa_category) } # for migrations
 
-  before_create :add_default_metadata_fields
+  ERCC_PATH_PREFIX = "s3://idseq-database/host_filter/ercc/2017-09-01-utc-1504224000-unixtime__2017-09-01-utc-1504224000-unixtime/".freeze
```
jshoe: can this path come from the "ERCC Only" host genome instead potentially? I could see someone adding a "2020 ERCC" update and forgetting to change this.

jshoe:
I'm out of the loop, but what does 'plain text host genomes' mean?

app/models/host_genome.rb
gregdingle: Good idea.

My question then is: if we upgrade to ERCC 2020, would we update the existing ERCC host genome or create a new one? I guess it depends on the nature of ERCCs.

gregdingle:
> I'm out of the loop, but what does 'plain text host genomes' mean?

This is a short hand way of referring to creating new host genomes on demand, similar to plain text location records.

app/models/host_genome.rb
jshoe: In the past, we have updated host genomes like the mosquito in-place, but I guess it depends

PR #2987, opened 2020-01-23
Show host subtracted in pipeline info sidebar

gregdingle:
# Description

In preparation for user-entered hosts that default to ERCC-only, this adds some info to disambiguate cases where the host subtracted is not the same as the host entered. 

image

# Notes

* On upload, users will be warned about no subtraction if they enter an unknown host
* The `host_subtracted` method depends on internal structure of dag json (similar to pipeline viz) and s3 URLs. To be safe, it always returns `nil` in case of missing data, and no row displayed in the sidebar.

# Tests

* new rspec test
* load sample, see info in sidebar
* load ERCC only sample, see custom string

image


app/models/pipeline_run.rb
```diff
@@ -450,6 +451,30 @@ def unidentified_fasta_s3_path
...
+  # two reasons:
+  #
+  # 1) there are no host genome files so we only subtract ERCCs
+  # 2) the sample host or index files were changed after the run
+  #
+  # Note: This method returns a string identifier extracted from the index
+  # filename, not a HostGenome instance, which could be ambiguous.
+  def host_subtracted
+    pipeline_run_stage = pipeline_run_stages.find { |prs| prs["step_number"] == 1 }
+    dag = pipeline_run_stage && JSON.parse(pipeline_run_stage.dag_json)
```
jshoe: `JSON.parse(... || "{}")` to be safe?

gregdingle: Good point. `dag_json` could be `nil`.

PR #2989, opened 2020-01-23
Add list of ERCC only hosts

gregdingle:
# Description

See comments in code. 

# Tests

run migration
check database with `select name from host_genomes order by name`

### DIFF
--- a/db/migrate/20200123221752_add_ercc_only_host_genomes.rb
+++ b/db/migrate/20200123221752_add_ercc_only_host_genomes.rb
@@ -1,0 +1,90 @@
+# This adds a bunch of hosts which users are known to have taken samples from.
+# They initially default to have ERCCs only subtracted.
+# See https://czi.quip.com/yPgpAPj1dtEC/Hosts-Dropdown-Options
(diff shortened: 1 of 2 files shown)


db/migrate/20200123221752_add_ercc_only_host_genomes.rb
```diff
@@ -0,0 +1,90 @@
...
+    'Seastar',
+    'Sloth',
+    'Snake',
+    'Stork',
+    'Taurine Cattle',
+    'Tick',
+    'Tiger',
+    'Turkey',
+    'Unknown',
+    'Water Buffalo',
```
jshoe: it looks like there are some missing from here / the doc that we have already like `Salpingoeca rosetta` or `Torafugu` https://idseq.net/host_genomes ? Is that OK?

gregdingle: @oliviabholmes ? @katrinakalantar ?

FWIW, the migration should skip over existing. ```    return if HostGenome.find_by(name: name)
```

PR #2991, opened 2020-01-24
[Heatmap] Apply filters on the frontend

j-x-han:
# Description

Previously, heatmap data was recalculated on the backend every time a new filter was applied, which could result in a lot of loading in between each application of a filter.

Categories (and Subcategories), Metric, Read Specificity, and Threshold Filtering (for z-score, rPM, r) will now be done on the frontend rather than refetching from the server, which should cut down on waiting time when applying filters.

# Notes

* This is currently protected by `heatmap_filter_fe` feature, since it's a WIP.

* Sort Taxa, Sort Samples, and Scale were already implemented purely on the frontend. Taxon Level, Background, and Taxa per Sample still require requests from the server.

* ***No backend changes were made in this PR, so these changes result in behavior that is not 100% consistent with the current implementation.*** TODO comments were added wherever additional backend changes are needed to complete the functionality.

  * Some threshold filters aren't functional (%id, L, log(1/e)) because that data is currently not being returned from the backend.

  * The heatmap can now only filter from the data fetched upon the initial load, or until a request to the server is made again, so less data will be shown than before. Backend changes to send more data (ex: the top 10 taxa per sample for each category) are required to fix this.

    * For example, say the heatmap is loaded in with the top 10 taxa per sample, and the top 10 taxa for those samples all happen to be bacteria. If a user then applies the "Viruses" filter, rather than finding the top 10 viruses from each sample, no data will be displayed since it finds no viruses among the top 10 taxa.

    * An example in a heatmap generated from the Medical Detectives Project:

***Top 10 uncategorized taxa per sample (how filters currently work)***
image

***Uncategorized taxa from the top 10 taxa per sample (how filters work in this PR)***
image

# Tests

* Verified that the Taxon Categories, Metric, Threshold Filters (for z-score, rPM, r), and Read Specificity Filters work more quickly (albeit with the caveats listed above).
* Verified that the other filters still work as they did previously.
* Verified that filters still work as they did previously if the user does not have the `heatmap_frontend_fe` feature enabled.


gregdingle:
There is also the backwards compatibility issue for users who expect top N after filtering. For example, would their saved heatmaps still show the same results? Will we tell users who want to the old behavior to download the unfiltered results and perform top N in excel or elsewhere?

gregdingle:
If there is some background I'm missing, please link me to any relevant docs 🙂 I understand that the PR is part of the larger re-write of heatmaps for better performance.

j-x-han:
@gregdingle Really great questions!

I don't believe there are currently any docs for it at the moment, but I'll be working with @tfrcarvalho on how we should load more data from the initial backend request to ensure that the results are comparable to the current heatmap. There was also a discussion with @happyimadesignr on clarifying the meaning of "top N" for the user and possibly adding a button so the user can explicitly request for more data to be fetched

tfrcarvalho:
@gregdingle 
This change will be guarded by a flag, and will be pending on us being able to send enough data from the backend to satisfy the `top n` in most occasions. Until then we have to explore several data sets to see the effects, and probably explore the idea of "load more data" button. Having this option to filter only on client side would help us with further exploration (see it as part of spike too).

Regarding the other requirements:
Agree that the final solution should always include an explanation of how the data is filtered - including filters hidden in constants if any. Not yet worth doing it because we are not set on the final design yet.
Also thank you for noting the need for backward compatibility with saved heatmaps. Those heatmaps should not change for sure and we should discuss our they will behave after the initial load. Actually, the saved heatmaps are great use cases to test coverage of the new data loading.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -370,13 +365,143 @@ class SamplesHeatmapView extends React.Component {
...
+  setTaxonThresholdFilterState() {
+    // Set the state of whether or not a taxon passes the custom threshold filters
+    // for each selected sample.
+    let {
+      sampleDetails,
+      allTaxonDetails,
+      allData,
+      taxonFilterState,
+    } = this.state;
+    Object.values(sampleDetails).forEach(sample => {
```
tfrcarvalho: Should we keep an aggregated state per sample and another per taxa to avoid going through all the options later?

gregdingle:
@tfrcarvalho Thanks for filling in the background. I think this PR makes sense then for the team to try frontend only filters. I'll let you or someone else closer to the project review the code line by line.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -536,20 +661,50 @@ class SamplesHeatmapView extends React.Component {
...
-    );
+    if (this.props.allowedFeatures.includes("heatmap_filter_fe")) {
+      const frontend_filters = [
+        "categories",
+        "subcategories",
+        "thresholdFilters",
+        "readSpecificity",
+        "metric",
+      ];
+      const backend_filters = ["species", "background", "taxonsPerSample"];
```
tfrcarvalho: Use camelCase

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -536,20 +661,50 @@ class SamplesHeatmapView extends React.Component {
...
-    const excluding = ["dataScaleIdx", "sampleSortType", "taxaSortType"];
-    const shouldRefetchData = difference(keys(newOptions), excluding).length;
-    this.setState(
-      {
-        selectedOptions: assign(this.state.selectedOptions, newOptions),
-      },
-      shouldRefetchData ? this.updateHeatmap : null
-    );
+    if (this.props.allowedFeatures.includes("heatmap_filter_fe")) {
+      const frontend_filters = [
```
tfrcarvalho: use camelCase on js code

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
tfrcarvalho: We can definitely do the taxonsPerSample on the client side (should rename `taxons` -> `taxa`)... Eventually we should do the `species` too.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -370,13 +365,143 @@ class SamplesHeatmapView extends React.Component {
...
+      allTaxonIds,
+      allTaxonDetails,
+      taxonFilterState,
+      allData,
+    } = this.state;
+    let taxonDetails = {},
+      taxonIds = [],
+      filteredData = {};
+    if (this.props.allowedFeatures.includes("heatmap_filter_fe")) {
+      for (let taxonId of allTaxonIds) {
```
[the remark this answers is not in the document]
j-x-han: Done

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -370,13 +365,143 @@ class SamplesHeatmapView extends React.Component {
...
+  filterTaxons() {
+    this.setTaxonThresholdFilterState();
+    let {
+      allTaxonIds,
+      allTaxonDetails,
+      taxonFilterState,
+      allData,
+    } = this.state;
+    let taxonDetails = {},
+      taxonIds = [],
```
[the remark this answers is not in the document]
j-x-han: Done, although I have to convert the set back into an array at the end of the function (SamplesHeatmapVis expects `taxonIds` to be an array)

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: Done

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -370,13 +365,143 @@ class SamplesHeatmapView extends React.Component {
...
+      filteredData = {};
+    if (this.props.allowedFeatures.includes("heatmap_filter_fe")) {
+      for (let taxonId of allTaxonIds) {
+        let taxon = allTaxonDetails[taxonId];
+        if (
+          !taxonIds.includes(taxonId) &&
+          this.taxonPassesSelectedFilters(taxon)
+        ) {
+          if (Object.values(taxonFilterState[taxon["index"]]).includes(true)) {
```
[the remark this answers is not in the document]
j-x-han: Changed to have `getTaxonThresholdFilterState` to also return the the aggregated state for each taxon across all samples, so this check will no longer iterate across all samples again.

PR #2992, opened 2020-01-24
Add new host genome on upload if not exists

gregdingle:
# Description

In preparation for the new input, this enables backend creation of a new host genome on upload. This also adds a `belongs_to :user` assoc to `HostGenome` so we can track who created new host genomes. 

# Notes

* name matching is case insensitive

# Tests

* new rspec tests

### DIFF
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -513,9 +513,14 @@ module SamplesHelper
 
       if sample_attributes[:host_genome_name]
         name = sample_attributes.delete(:host_genome_name)
-        # TODO: (gdingle): find or create!!!
-        # TODO: (gdingle): check for both upload form and csv
-        sample_attributes[:host_genome_id] = HostGenome.find_by(name: name).id
+        begin
+          hg = HostGenome.find_or_create_by!(name: name)
+        rescue => e
+          errors << e.message
+          metadata.delete(sample_attributes[:name])
+          next
+        end
+        sample_attributes[:host_genome_id] = hg.id
       end
       sample = Sample.new(sample_attributes)
       sample.input_files.each { |f| f.name ||= File.basename(f.source) }
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -517,7 +517,7 @@ module SamplesHelper
           # TODO: (gdingle): remove admin feature gating.
           # See https://jira.czi.team/browse/IDSEQ-2051
           hg = if user.admin
-                 HostGenome.find_or_create_by!(name: name)
+                 HostGenome.find_or_create_by!(name: name, user: user)
                else
                  HostGenome.find_by(name: name)
                end
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -514,7 +514,13 @@ module SamplesHelper
       if sample_attributes[:host_genome_name]
         name = sample_attributes.delete(:host_genome_name)
         begin
-          hg = HostGenome.find_or_create_by!(name: name)
+          # TODO: (gdingle): remove admin feature gating.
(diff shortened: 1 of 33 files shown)


app/helpers/samples_helper.rb
```diff
@@ -513,7 +513,14 @@ def upload_samples_with_metadata(samples_to_upload, metadata, user)
 
       if sample_attributes[:host_genome_name]
         name = sample_attributes.delete(:host_genome_name)
-        sample_attributes[:host_genome_id] = HostGenome.find_by(name: name).id
+        begin
+          hg = HostGenome.find_or_create_by!(name: name)
```
tfrcarvalho: Shouldnt this be guarded by feature flag?

gregdingle: This functionality will be hidden from users until there is the new input is in place, which I was planning to put behind a feature flag.

tfrcarvalho: I see, so, it assumes that users will not have any choice other than existing host genomes.

ok, up to you, but, imho, seems to me this should not be an isolated change, since a lot more goes into creating host genomes... its a bit risky. 

Is there a design doc and/or product spec to have a better idea of this feature?

gregdingle: The idea is to enable users to create new host genomes at upload time. The new host genomes will default to using ERCC index files, effectively doing what we tell users to do now -- choose the "ercc only" option if their host is not in the list. 

I'd be happy to put this in admin only if you think that's best.

tfrcarvalho: I see. So, it will be ok to create host genomes with name only?

I do not have the full picture of the sequence of tasks and timeline to make the whole feature work, so I will leave it up to you to evaluate the risk - if any - of leaving this open.

gregdingle: I added the `admin` gating 👍

cdebourcy:
Thanks for tagging @oliviabholmes and @katrinakalantar in the channel for their input on this PR! Let's make sure they respond and approve before you merge

katrinakalantar:
I think it may be good to think a bit more about the downstream workflow. Adding the genome to this list of available genomes for that user makes sense to me (since we are making it possible to add genomes that weren't previously available, with ERCCs removed) but is there a mechanism to alert the team when new genomes have been added via this mechanism as just ERCCs so that we can go back and create true host genomes for them in the future? I think it would be great to have a system in place before enabling addition of ERCC-only hosts.

gregdingle:
> is there a mechanism to alert the team when new genomes have been added via this mechanism as just ERCCs so that we can go back and create true host genomes for them in the future

Good question. This came up in initial design discussions with @oliviabholmes. We agreed to manually check for new host genomes from time to time. This would be a very simple SQL query. We could add alerting on top of that easily inside periscope.

oliviabholmes:
@gregdingle I think it's a good plan to periodically check Periscope for these stats and decide if we want to support that as a new host genome. However, I think we should think about automatically adding all of those host organism options as host **genomes** automatically. The intention of adding those host organisms was to provide a dropdown menu value and remove some of the duplicates and messiness of allowing for free text. But, since we don't have a host genome for these, I'm not sure it makes sense to mark them as such on the backend.

gregdingle:
> However, I think we should think about automatically adding all of those host organism options as host genomes automatically.

To be clear, this PR enables backend creation of a new host genome on upload. New host genomes will be defaulted to ERCC only and they will appear in the dropdown. 

The other PR, #2989 , adds the list that were in Quip. Those will appear in the dropdown. 

The new input will continue to be gated to admin only so we can see how we like it.

katrinakalantar:
Is it possible to add the new hosts to the list in the dropdown and run them against the existing ERCC-only "host genome" without creating new host genomes in the backend? It seems beneficial to maintain only high quality vetted host genomes in the back end to avoid confusion.

PR #2993, opened 2020-01-27
Fix water control default not being applied

gregdingle:
# Description

image

Making the host genome default blank caused the water control default apply all to no longer work, because there was logic to depend on a valid host genome in the "apply all" method. That logic would seem to be relevant only when there is a mix of fields available depending on the selected host genomes. It's not clear to me what the consequences would be of setting a value in JS for a field that is not available for a certain sample. I would expect it to be ignored by the backend on submission. 

In any case, the simple workaround here is to skip the logic for required fields which by definition are available for all samples. 

# Notes

* There was a CSS issue I noticed for the first time -- not sure how it could be related -- but I fixed it. 

# Tests

* upload samples, fill out form, except water control
* see progress to review step, no more water control required error

### DIFF
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -249,6 +249,7 @@ class MetadataManualInput extends React.Component {
   isHostGenomeIdValidForField = (hostGenomeId, field) =>
     // Special-case 'Host Genome' (the field that lets you change the Host Genome)
     field === "Host Genome" ||
+    this.props.projectMetadataFields[field].is_required ||
     includes(
       hostGenomeId,
       get([field, "host_genome_ids"], this.props.projectMetadataFields)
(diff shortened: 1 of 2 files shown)


app/assets/src/components/common/MetadataManualInput.jsx
```diff
@@ -249,6 +249,7 @@ class MetadataManualInput extends React.Component {
   isHostGenomeIdValidForField = (hostGenomeId, field) =>
     // Special-case 'Host Genome' (the field that lets you change the Host Genome)
     field === "Host Genome" ||
+    this.props.projectMetadataFields[field].is_required ||
```
tfrcarvalho: Is there any chance of field not being in `this.props.projectMetadataFields`?
If so, you can always use the `get([field, "is_required"]` as below. Sometimes is good to do it just to be on the safe side.

gregdingle: Good point, to be safe.

PR #2994, opened 2020-01-27
[BulkDownloads] Remove Review Step and add Notification when download created (IDSEQ-1979)

lvreynoso:
# Description

We are changing the flow slightly to show a notification when the download is created and keeping the user on the page, instead of taking them to the download page.

Also, removing the second review step, since it's not that useful.

# Notes

* Only front end changes.
* Notification is set to autoclose after 12 seconds.

# Tests

* Used the `Todd Test` project, select all samples, choose a field, and generate a download. Ensure notification pops up and download is created.

Demo:
notification


app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
```diff
@@ -118,102 +98,65 @@ class BulkDownloadModal extends React.Component {
...
-
-  render() {
-    const { open } = this.props;
-
     return (
-      <Modal narrow open={open} tall onClose={this.props.onClose}>
-        {this.renderStep()}
+      <Modal narrow open={open} tall onClose={onClose}>
+        {
+          //loading && "L O A D I N G"
```
tfrcarvalho: Do we need this comment? What does it mean?

app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
```diff
@@ -58,38 +28,44 @@ class BulkDownloadModal extends React.Component {
...
-        selectedDownloadTypeName: null,
-        currentStep: "choose",
-        selectedFields: {},
-      });
-    }
-  }
+  // *** Async requests ***
 
-  async fetchDownloadTypesAndValidationInfo() {
+  async parallelFetchNeededData() {
```
tfrcarvalho: Can we have a more explicity name?
Using `parallel` is less important than trying to say what are we fetching.

app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
```diff
@@ -118,102 +98,65 @@ class BulkDownloadModal extends React.Component {
...
-  handleChooseStepContinue = () => {
-    this.setState({ currentStep: "review" });
-  };
-  handleBackClick = () => {
-    this.setState({ currentStep: "choose" });
-  };
+    return backgroundOptions;
+  }
+
+  // We use the heatmap metrics as the valid metrics for bulk downloads.
```
tfrcarvalho: If we are using heatmap metrics more widely, we should probably move to a more generic endpoint.
I guess this was not introduced on this PR, so might not be in scope...

app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
```diff
@@ -118,102 +98,65 @@ class BulkDownloadModal extends React.Component {
...
-
     return (
-      <Modal narrow open={open} tall onClose={this.props.onClose}>
-        {this.renderStep()}
+      <Modal narrow open={open} tall onClose={onClose}>
+        {
+          //loading && "L O A D I N G"
+        }
+        {
+          <Chooser
```
tfrcarvalho: Can we give it a more specific name? DownloadsOptions? DownloadsMenu?

app/assets/src/components/views/bulk_download/Chooser.jsx
```diff
@@ -183,11 +172,85 @@ class ChooseStep extends React.Component {
     orderBy(["sampleCount", "text"], ["desc", "asc"], options)
   );
 
+  // *** BULK DOWNLOAD GENERATION ***
```
tfrcarvalho: What does this comment indicate? a lot of the functions below are just handlers, right?

app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
```diff
@@ -1,48 +1,18 @@
...
 import {
   getBulkDownloadTypes,
   validateSampleIdsForBulkDownload,
 } from "~/api/bulk_downloads";
 import Modal from "~ui/containers/Modal";
 
-import ChooseStep from "./ChooseStep";
-import ReviewStep from "./ReviewStep";
-
-const assembleSelectedDownload = memoize(
```
tfrcarvalho: I do not have that much context, so it is harder for me to grasp some of the changes and this could be a naive question. 
But what is the reasoning to move this function to the Choose step / Chooser?

Sees to me that the chooser object should be just a way to select options. But the main class would handle the actual download... no?

app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
lvreynoso: That should not have been there, oops.

app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
lvreynoso: I thought it would make for clearer code if all the async requests were done in the `BulkDownloadModal` component and then the resulting data passed down as props to the `ChooseStep`. The purpose of the callback functions I moved was to take data from the `ChooseStep`, assemble a download and pass it to the `ReviewStep`. Since we're eliminating the `ReviewStep` I thought it would be better to put all the logic for handling the selected options and creating a download in the `ChooseStep`.

I agree that I should name it something clearer, I'll do that.

lvreynoso:
Refactored the logic to make it clearer and more sensible, addressed comments, cleaned up stray text.

MarkAZhang:
Screen Shot 2020-01-30 at 5 41 20 PM

- The two notifications have different font sizes (12px and 13px). Would be great to make them consistent.
- On my large monitor, there's a big gap below the Start button. Avoiding the big gap would be best, but at the very least, the button should be bottom-aligned. 
- There should be a bigger margin between the error notification and the start button. See the margin in the mock here: https://chanzuckerberg.invisionapp.com/d/main?redirHash=#/console/18628681/398037533/preview

MarkAZhang:
Also, it looks like there isn't a way to dismiss the notification. Could you add a "DISMISS" button next to "VIEW DOWNLOADS"? `<Notification>` does something similar, although you can't use it easily since you already have the view downloads link.

lvreynoso:
I incorporated your comments, see attached pictures. I just used the built in dismissal that the `Notification` component has for consistency. Limited the modal to 1200px, which looks like the maximum it needs.


Screen Shot 2020-01-30 at 7 08 35 PM
Screen Shot 2020-01-30 at 7 11 07 PM

MarkAZhang:
Using `position: absolute` to position elements can be risky. Items can overlap if the modal is too small (see screenshot)

Screen Shot 2020-01-30 at 8 09 58 PM

The approach I usually use is to add a `<div className={cs.fill}>` element with `flex: 1 1 0; min-height: 0;` to take up any remaining space and force the footer to the bottom. If you search in the code base, you'll see me do this a lot. 

This might not be necessary though. I think if you change `max-height: 1200px; height: calc(85vh - 80px);` to just `max-height: calc(85vh - 80px);` I think the modal will behave the way we want. It'll take up exactly as much space as it needs, without going over `calc(85vh - 80px)` (so we'll never see that big gap).

PR #2997, opened 2020-01-28
New host organism input 

gregdingle:
# Description

image

This changes the host genome input to be like the new sample type input. It is searchable and accepts plain text, which become new host genomes on submit, defaulting to ERCC only subtraction. (See #2968 for messaging to users on review step. And see #2987 for how to distinguish downstream.)

image

I also updated the top description according to the new design by @happyimadesignr .

image

I also added a "ERCC subtraction only" subtext to the dropdown according to design review.

image

# Notes

* Feature is admin only gated
* Not sure if our QA team has admin accounts to test this in staging
* New host genomes simply point to the ERCC index files by new defaults in previous PR
* Invalid names such as "1234" will return an error
* The dropdown order is initially by popularity, then by text match rank, same as sample type
* I added a behavior to the dropdown: exact match eliminates the plain text option
* I will perform a global replace of all UI strings to "host organism" in a later PR when we are ready to launch. 

# Tests

* new and old unit tests
* try upload as admin, see new input, complete steps
* try upload as non admin, see old input, complete steps
* try invalid name, see error message

image


app/helpers/metadata_helper.rb
```diff
@@ -347,4 +352,17 @@ def validate_metadata_csv_for_samples(samples, metadata, new_samples = false)
...
     }
   end
+
+  private
+
+  def find_or_create_host_genomes(metadata, allow_new_host_genomes)
+    host_genome_index = metadata["headers"].find_index("host_genome") || metadata["headers"].find_index("Host Genome")
+    host_genome_names = metadata["rows"].map { |row| row[host_genome_index] }.uniq
+    host_genomes = if allow_new_host_genomes
+                     host_genome_names.map { |name| HostGenome.find_or_initialize_by(name: name).save }
```
gregdingle: Strangely, find_or_create_by returns an unsaved instance on validation failure, so I use find_or_initialize_by().save

app/assets/src/components/ui/controls/HostOrganismSearchBox.jsx
```diff
@@ -0,0 +1,83 @@
...
+    return this.buildResults(this.getMatchesByCategory(query), query);
+  };
+
+  // TODO (gdingle): extract all these to utils if re-used
+  getMatchesByCategory(query) {
+    const matchedHostGenomes = this.props.hostGenomes.filter(
+      hostGenome =>
+        // IMPORTANT NOTE: Only existing, null-user host genomes will be shown as
+        // options for new samples until the team gets a chance to review this
+        // policy in light of the data. See HostGenome.rb.
```
jshoe: should we have a JIRA and date for this note?

app/helpers/error_helper.rb
```diff
@@ -122,9 +122,11 @@ class ErrorAggregator
         headers: ["Row #", "Sample Name"],
         title: ->(num_rows, _) { "#{num_rows} rows are missing host genomes." },
       },
+      # TODO: (gdingle): This behavior will change after removal of admin-only of new host genome input.
```
jshoe: add date?

app/helpers/metadata_helper.rb
```diff
@@ -169,9 +169,29 @@ def metadata_csv_has_duplicate_columns(columns, existing_fields, error_aggregato
...
   end
 
+  # Convenience wrapper
+  def validate_metadata_csv_for_project_samples(samples, metadata)
+    validate_metadata_csv_for_samples(samples, metadata, false, false)
+  end
+
+  # Convenience wrapper
+  def validate_metadata_csv_for_new_samples(samples, metadata)
+    # TODO: (gdingle): remove admin only after launch. See https://jira.czi.team/browse/IDSEQ-2051.
```
jshoe: add date?

app/helpers/metadata_helper.rb
```diff
@@ -270,8 +289,10 @@ def validate_metadata_csv_for_samples(samples, metadata, new_samples = false)
       end
 
       if extract_host_genome_from_metadata
-        host_genome = host_genomes.select { |cur_host_genome| cur_host_genome.name == row[host_genome_index] }.first
+        host_genome = host_genomes.select { |cur_hg| cur_hg && cur_hg.name.casecmp?(row[host_genome_index]) }.first
 
+        # TODO: (gdingle): This behavior will change after removal of admin-only of new host genome input.
```
jshoe: add date?

MarkAZhang:
I notice that LiveSearchPopBox uses a different ui for search versus some of our older dropdowns:

Screen Shot 2020-01-30 at 3 03 53 PM

Screen Shot 2020-01-30 at 3 06 26 PM

Should we try to consolidate this? @happyimadesignr

app/assets/src/components/ui/controls/HostOrganismSearchBox.jsx
gregdingle: Good idea. @oliviabholmes do you want to create a JIRA or should I?

app/helpers/error_helper.rb
gregdingle: I think the link to the JIRA is sufficient.

app/helpers/metadata_helper.rb
gregdingle: I think the link to the JIRA is sufficient.

app/helpers/metadata_helper.rb
gregdingle: I think the link to the JIRA is sufficient.

app/helpers/metadata_helper.rb
```diff
@@ -219,9 +239,8 @@ def validate_metadata_csv_for_samples(samples, metadata, new_samples = false)
     sample_name_index = metadata["headers"].find_index("sample_name") || metadata["headers"].find_index("Sample Name")
 
     if extract_host_genome_from_metadata
-      host_genome_index = metadata["headers"].find_index("host_genome") || metadata["headers"].find_index("Host Genome")
-      host_genomes = HostGenome.where(name: metadata["rows"].map { |row| row[host_genome_index] }.uniq)
-                               .includes(:metadata_fields).to_a
+      host_genomes, host_genome_index, new_host_genomes =
+        find_or_create_host_genomes(metadata, allow_new_host_genomes)
```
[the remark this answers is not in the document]
MarkAZhang: Update: It looks like you need a host_genome_id for each sample in the SampleUploadFlow, so you do need to create the host genomes here. This is unfortunate. (it would be a pretty big refactor to have `upload_samples_with_metadata` accept either a host_genome_id or a custom name.)

I would still create the user-defined host genome objects in a separate function (outside of `validate_metadata_csv_for_samples`). I don't think a validate function should create objects. This seems similar to the situation David was talking about where "saving" a sample shouldn't do something unexpected like kick off a pipeline run. You might even go so far as to have `create_custom_host_genomes` be a completely separate controller endpoint that you call in parallel with `validate_csv_for_new_samples` on the front-end.

app/assets/src/components/views/SampleUploadFlow/SampleUploadFlow.jsx
```diff
@@ -148,20 +157,28 @@ class SampleUploadFlow extends React.Component {
...
+              project={this.state.project}
+              sampleNamesToFiles={this.state.sampleNamesToFiles}
+              hostGenomes={this.state.hostGenomes}
+              visible={this.state.currentStep === "review"}
+              onUploadStatusChange={this.onUploadStatusChange}
+              onStepSelect={this.handleStepSelect}
+              onUploadComplete={this.onUploadComplete}
+            >
+              {this.props.admin && (
+                <HostOrganismMessage
```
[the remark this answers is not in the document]
MarkAZhang: I'm guessing it's because you needed to pass in `this.props.hostGenomes` (the original host genomes). I would just pass in `originalHostGenomes` as an additional prop. It's still a bit awkward, and some comments would probably be helpful.

app/assets/src/components/ui/controls/HostOrganismSearchBox.jsx
```diff
@@ -0,0 +1,83 @@
...
+        results: [{ title: query, name: query }],
+      };
+    }
+    return results;
+  }
+
+  render() {
+    const { className, value, onResultSelect } = this.props;
+    return (
+      <LiveSearchPopBox
```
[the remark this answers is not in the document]
gregdingle: Thanks for the explanation. @oliviabholmes directed me to make it the same as the collection location which is using `LiveSearchPopBox`. I see that `LiveSearchPopBox` uses `BareDropdown` internally. There is considerable stuff in `LiveSearchPopBox` beyond data fetching, for example `handleBlur` behavior. 

To keep things clean in the future, it would be good to write a doc block on top of such common components to clarify the intention beyond names like "live" and "bare".

app/assets/src/components/ui/controls/HostOrganismSearchBox.jsx
```diff
@@ -0,0 +1,83 @@
+import React from "react";
```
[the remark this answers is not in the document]
gregdingle: Ok thanks for clarifying. I'll move it.

app/assets/src/components/views/SampleUploadFlow/utils.js
```diff
@@ -19,3 +21,35 @@ export const openBasespaceOAuthPopup = params => {
     BASESPACE_OAUTH_WINDOW_HEIGHT
   );
 };
+
+// The following three functions were extracted from SampleTypeSearchBox.
+// They aid client-side search of options.
+export const matchType = (type, query) => {
```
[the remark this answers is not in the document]
gregdingle: Good call. I renamed `type` to `result` everywhere.

> I wonder if components/utils/search would be a better place for these functions

I don't see any `util.js` in there and it would see more strange to have it in a source tree where it is not used. For re-use, I expect a developer will first want something "like the upload page" then find the functions. At that point, then should reorganize as they see fit. 

As a general comment, it seems strange to me that there are so many util.js files in our dirs but so few global js utils.

app/assets/src/components/views/SampleUploadFlow/SampleUploadFlow.jsx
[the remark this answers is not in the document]
gregdingle: Passing in as a child node I thought was good (!), on the principle of flatter dependencies and smaller interfaces (props).

In this case the child node could be anything. I can imagine many different messages we may want to go above the continue button.

app/helpers/metadata_helper.rb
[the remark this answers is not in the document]
gregdingle: >  They are only doing validation, and they might change their metadata before the final submission

I agree, this is a weakness of this PR. I did it for the reason you saw, `host_genome_id`. Note there is no user impact of spurious host genomes because we filter them out. It would just be up to use to clean them up later. 

> I would still create the user-defined host genome objects in a separate function (outside of validate_metadata_csv_for_samples).

The reason I put the creation in the validate function is so invalid host genomes will be caught and processed by the existing validation code. What if `validate_metadata_csv_for_samples` initialized the new host genomes but did not save them? Would that be clean enough in your view?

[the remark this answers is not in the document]
gregdingle: Made that change, have a look.

happyimadesignr:
> I notice that LiveSearchPopBox uses a different ui for search versus some of our older dropdowns:
> 
> Screen Shot 2020-01-30 at 3 03 53 PM
> 
> Screen Shot 2020-01-30 at 3 06 26 PM
> 
> Should we try to consolidate this? @happyimadesignr

Hey @MarkAZhang, would be great to consolidate to this newer style in the future!

app/controllers/metadata_controller.rb
```diff
@@ -42,11 +42,19 @@ def validate_csv_for_new_samples
       )
     end
 
-    issues = validate_metadata_csv_for_samples(samples, metadata, true)
+    issues, new_host_genomes = validate_metadata_csv_for_new_samples(
```
MarkAZhang: For the `validate_csv_for_new_samples` method, I would add the comment:


If validation passes and the user provides host genome names that we currently don't support, new custom host genomes will be created and assigned to that user.
This is done at this step because the sample upload flow requires host genome ids to be passed to the create endpoint.

PR #3002, opened 2020-01-30
Fix branch version tagging potential issue

davidrissato:
# Description

Fix a potential issue in `patch_branch_version_tags.sh` where failing to push tags to origin would create an inconsistent state if the command were re-executed.

Also adding some extra release cycle state detection to resume from error conditions.

# Tests

* Manual

### DIFF
--- a/bin/deploy_automation/_shared_functions.sh
+++ b/bin/deploy_automation/_shared_functions.sh
@@ -105,9 +105,9 @@ _git_fetch_and_cleanup() {
   # fetch from remote origin
   git fetch origin
 
-  # remove any dangling version tags not present in the remote origin
+  # remove any dangling version tags not present in remote origin
   # (this could happen if a previous run failed to push tags to remote)
-  git tag -d $(
+  declare dangling_tags; dangling_tags=$(
     grep -vxFf  \
       <(git ls-remote --tags origin | cut -f 2 | sed -E 's/^refs\/tags\///') \
       <(git tag -l \
@@ -115,4 +115,8 @@ _git_fetch_and_cleanup() {
       ) \
       || [[ $? == 1 ]]
   )
+  if [ ! -z "$dangling_tags" ]; then
+    _trace "Found some dangling version tags not present in remote origin. Removing..."
+    git tag -d "$dangling_tags"
+  fi
 }
--- a/bin/deploy_automation/_shared_functions.sh
+++ b/bin/deploy_automation/_shared_functions.sh
@@ -25,8 +25,8 @@ _trace() {
 }
 
 _get_latest_commit() {
-  declare branch_or_tag_name="$1"
-  git log -n 1 --pretty=format:"%h" "$branch_or_tag_name"
+  declare git_rev="$1" # https://git-scm.com/docs/git-rev-parse#_specifying_revisions
+  git log -n 1 --pretty=format:"%h" "$git_rev"
 }
 
 _fetch_current_release_checklist_from_github() {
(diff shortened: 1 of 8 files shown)


bin/deploy_automation/_shared_functions.sh
```diff
@@ -25,8 +25,8 @@ _trace() {
 }
 
 _get_latest_commit() {
-  declare branch_or_tag_name="$1"
-  git log -n 1 --pretty=format:"%h" "$branch_or_tag_name"
+  declare git_rev="$1" # https://git-scm.com/docs/git-rev-parse#_specifying_revisions
+  git log -n 1 --pretty=format:"%h" "$git_rev"
```
davidrissato: Better terminology for this hybrid parameter

bin/deploy_automation/_shared_functions.sh
```diff
@@ -99,4 +99,20 @@ _assert_current_branch_is_not() {
...
+
+  # remove any dangling version tags not present in the remote origin
+  # (this could happen if a previous run failed to push tags to remote)
+  git tag -d $(
+    grep -vxFf  \
+      <(git ls-remote --tags origin | cut -f 2 | sed -E 's/^refs\/tags\///') \
+      <(git tag -l \
+          | grep -E '^v(?:[0-9]+\.){1,2}(?:[0-9]+)_[^_]+_[0-9]{4}-[0-9]{2}-[0-9]{2}$'
+      ) \
+      || [[ $? == 1 ]]
```
davidrissato: `grep` exits with code 1 when no lines are found (and other error codes if something else goes wrong). This condition tells bash exit code 1 is actually fine for this command.

bin/deploy_automation/automated_release_cycle.sh
```diff
@@ -36,15 +35,9 @@ main() {
...
-
-  declare msg;
-  msg="**** READY FOR DEPLOYMENT ****${LF}"
-  msg+="${YELLOW}PLEASE EXECUTE THE FOLLOWING COMMANDS${LF}"
-  msg+="./bin/deploy_automation/deploy_tag.sh staging '${staging_tag}'${LF}"
-  msg+="./bin/deploy_automation/deploy_tag.sh prod '${prod_tag}'"
-  _log "$msg"
+  _log "**** READY FOR DEPLOYMENT ****"
+  _log "New tags: ${YELLOW}" \
+       "$(_get_latest_tag "$PROD_ENV")    $(_get_latest_tag "$STAGING_ENV")"
```
davidrissato: I'm simplifying this output and changing deployment instructions to not depending on copying what was printed here.

bin/deploy_automation/deploy_rev.sh
```diff
@@ -4,25 +4,27 @@ SCRIPT_DIR=$(dirname "$0")
 source "$SCRIPT_DIR/_global_vars.sh"
 source "$SCRIPT_DIR/_shared_functions.sh"
 
-# Deploy a tag to an environment.
-# This command will wait for the commit of this tag
+# Deploy a git rev to an environment.
```
davidrissato: I renamed this script since it does more than just deploying tags.

bin/deploy_automation/deploy_rev.sh
```diff
@@ -4,25 +4,27 @@ SCRIPT_DIR=$(dirname "$0")
...
 
   _log "Checking if docker image $docker_image is available"
-  if ! __retry 5 15 __docker_tag_exists "${DOCKER_REPOSITORY_NAME}" "sha-$sha"; then
+  if ! __retry 10 60 __docker_tag_exists "${DOCKER_REPOSITORY_NAME}" "sha-$sha"; then
     _exit_with_err_msg "Couldn't find image ${docker_image} in docker hub"
   fi
 
   _log "Checking if Github commit ${sha} in a valid state"
-  if ! __retry 5 15 __check_commit_state "${sha}" "$env"; then
+  if ! __retry 10 60 __check_commit_state "${sha}" "$env"; then
```
davidrissato: Increasing the timeout, since travis can take longer than 75 seconds (5 * 15s) to run all tests. Increasing this timeout to 10 minutes (10 * 60s).

jshoe: Hm, based on some recent comments Travis might take up to 15-ish minutes?

davidrissato: Ok, improving to 15 minutes

PR #3004, opened 2020-01-31
Fix bug where water control was showing in review step.

MarkAZhang:
# Description

Fixes bug where "water_control" shows up in the Review Step, even for projects without water_control.
This was being caused by `setDefaultWaterControl` being called even when water_control wasn't in the projectMetadataFields. ReviewStep simply displays whatever is in the metadata object that is passed to it, and assumes that the previous steps only put valid things into the metadata object.

Also fix a (previously undetected) bug where if you GO BACK to step 1 and pick a different project, step 2 (select metadata) will not refresh the project metadata fields. An edge case, but it's now fixed.

# Tests

* Verified that water_control no longer shows up.
* Verified that when you return to step 1 and switch projects, the metadata fields are as expected.

### DIFF
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -13,6 +13,7 @@ import _fp, {
   orderBy,
   find,
   pickBy,
+  has,
 } from "lodash/fp";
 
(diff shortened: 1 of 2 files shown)


gregdingle:
Thanks for the fix @MarkAZhang !!

PR #3006, opened 2020-01-31
[BulkDownloads] Add analytics events

lvreynoso:
# Description

Adds analytics logging to user interactions with the bulk downloads modal:

* when a user selects a download type
* when a user selects a field for a download type
* when a user clicks the "Start Download" button
* when the creation of a bulk download succeeds
* when the creations of a bulk download fails
* when the user clicks the "Downloads page" link of the notification that pops up

# Tests

* Tested with Todd Test project, selected samples to create a bulk download, generated it, clicked on the downloads page link in the notification. All interactions were normal.


app/assets/src/components/views/bulk_download/BulkDownloadModal.jsx
```diff
@@ -193,6 +195,12 @@ class BulkDownloadModal extends React.Component {
 
   handleFieldSelect = (downloadType, fieldType, value, displayName) => {
     this.setState(prevState => {
+      logAnalyticsEvent("BulkDownloadModal_field-for-download-type-selected", {
```
jshoe: you should have it as `_selected` instead of `-` for the verb for each of these (https://czi.quip.com/67RCAIiHN0Qc/IDseq-product-analytics-How-to-log)

lvreynoso: Thanks for point this out!

PR #3007, opened 2020-01-31
Travis + Docker Optimization

morsecodist:
# Description

We discussed a lot of potential changes but these are low hanging fruit and don't change anything risky. This brings the time down to about 3 minutes for ruby changes and about 5 for JavaScript Changes.

- Parallel Builds and tests. This means the docker image gets built and tagged regardless of test passing but it can happen in parallel with tests that don't use docker.
- Switch to installing chamber with curl instead of `ADD`. Adding from a url breaks the cache every time so it leads to double download of dependencies.
- Build assets before copying the rest of the app. This makes changes that change the API only faster to build.
- Use travis services instead of docker-compose. These are pre-installed so they are faster.
- Remove deprecated sudo option from travis file.
- Remove build number tags. I feel these are redundant with the sha tags.

# Tests

The automated tests run in this container test it's functionality


bin/build-docker
```diff
@@ -15,16 +17,16 @@ echo $REPO
 echo $TAG
 echo '--'
 
-# need to pull images for cache_from first
-docker pull $LATEST_TAG || true
-docker pull $COMPOSE_TAG || true
+# Try to pull the latest image for this branch, otherwise pull latest image for caching
+if ! ( docker pull $BRANCH_TAG && docker tag $BRANCH_TAG $COMPOSE_TAG ); then
```
MarkAZhang: Nit: Instead of using COMPOSE_TAG here to specify the image we use for caching, and then using COMPOSE_TAG again later to specify the built docker image, can we use something like CACHE_TAG="$REPO:cached" here? Just don't like re-using names for two different purposes.

.travis.yml
```diff
@@ -1,32 +1,29 @@
...
-  - docker-compose up -d db
-  - docker-compose up -d redis
-  - sleep 10
-  - docker-compose run web rake db:create
-install:
-  - gem install brakeman
+  - bin/build-docker $TRAVIS_BRANCH ${TRAVIS_COMMIT::8}
+  # map db and redis to localhost to seamlessly use travis-ci services instead of docker-compose
+  - sudo bash -c "echo '127.0.0.1 db' >> /etc/hosts"
+  - sudo bash -c "echo '127.0.0.1 redis' >> /etc/hosts"
```
MarkAZhang: Also, is this relevant? https://docs.travis-ci.com/user/hosts/ Possibly a better way.

.travis.yml
```diff
@@ -1,32 +1,106 @@
...
+        - echo foo > app/assets/dist/bundle.min.css
+        - echo foo > app/assets/dist/bundle.min.js
+        - echo foo > app/assets/dist/main.bundle.min.css
+        - echo foo > app/assets/dist/main.bundle.min.css.map
+        - echo foo > app/assets/dist/main.bundle.min.js
+        - echo foo > app/assets/dist/main.bundle.min.js.map
+        - echo foo > app/assets/dist/vendors~main.bundle.min.css
+        - echo foo > app/assets/dist/vendors~main.bundle.min.css.map
+        - echo foo > app/assets/dist/vendors~main.bundle.min.js
+        - echo foo > app/assets/dist/vendors~main.bundle.min.js.map
```
kislyuk: Can this be made less manual? e.g. somehow run bundler here to create an empty bundle?

morsecodist:
@kislyuk Sorry I didn't signal on the PR, I thought people wouldn't look on the weekend. I have been pushing changes on top to measure the effect on the build hence weird stuff like `# foo` an arbitrary change to test the caching. Marked as WIP I will signal when complete.

morsecodist:
As per conversation with @MarkAZhang modifying this PR to add parallel build. Making this WIP until that piece is complete.

.travis.yml
morsecodist: Yes! Thanks much better!

bin/build-docker
morsecodist: Yeah this is a good point. I originally re-used the name because I assumed it would auto-cache if we tagged the same name but for some odd reason this didn't work when I tested it. I will make the change.

.travis.yml
morsecodist: I will research if webpack can support this

.travis.yml
```diff
@@ -1,32 +1,106 @@
...
+        - mysql
+        - redis
+      before_install:
+        # Emulate docker compose routing
+        #   Docker compose maps the host db and redis to the mysql and redis containers
+        #   We are using travis-ci services, which use localhost, for testing instead of docker compose 
+        #   we need to map these hosts to localhost to maintain compatibility between the two.
+        - sudo bash -c "echo '127.0.0.1 db' >> /etc/hosts" 
+        - sudo bash -c "echo '127.0.0.1 redis' >> /etc/hosts" 
+        - mkdir -p ~/.aws && printf "[default]\nregion = $AWS_REGION\naws_access_key_id = $AWS_ACCESS_KEY_ID\naws_secret_access_key = $AWS_SECRET_ACCESS_KEY\n" > ~/.aws/credentials
```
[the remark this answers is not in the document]
morsecodist: We are not running in docker with these changes but I can actually remove this line altogether since they are already environment variables.

PR #3009, opened 2020-02-01
separate queues for initiate_s3_cp and result_monitor_loader jobs

cdebourcy:
# Description
This separates the queue for S3 uploads from the queue for result loader jobs (https://jira.czi.team/browse/IDSEQ-2207). Until we have a better prioritization scheme, this change makes it possible for the on-call operator to stop S3 upload jobs from processing until result loader jobs have finished.


cdebourcy:
> LGTM assuming it worked in your testing.

Thanks. I tested in sandbox that dispatch to those queues works. I will verify the behavior in more detail once this code hits staging

jshoe:
^ btw the mergebot still appears to be down

czecs-resque.json
```diff
@@ -31,7 +31,7 @@
         },
         {
           "name": "QUEUE",
-          "value": "data_migration,q03_pipeline_run,transfer_basespace_files,generate_bulk_download,precache_report_info,*"
+          "value": "data_migration,q03_pipeline_run,result_monitor_loader,transfer_basespace_files,initiate_s3_cp,generate_bulk_download,precache_report_info,*"
```
davidrissato: I'm taking a look on these queue priorities and I noticed this `data_migration`. @MarkAZhang Do we still need that queue? It looks it is only being used for `PopulateContigSpeciesTaxids` (https://github.com/chanzuckerberg/idseq-web/blob/62a010eca46fdffe7e2ffdc215ae18f9095b5c49/app/jobs/data_migrations/populate_contig_species_taxids.rb#L6) and I couldn't find any parts of our code sending messages to that queue.

jshoe:
^ Going to merge this manually given the `bot/merge`

PR #3010, opened 2020-02-02
cache report only if pipeline run is complete and successful

cdebourcy:
# Description
**Problem**: When a user accesses a report before the Experimental stage of the pipeline run is complete, then the report will be cached. This is problematic because now all users will see the incomplete state of the report (e.g. coverage viz disabled, taxon fasta disabled) even after the results of the Experimental stage become available, because the report is retrieved from the cache. 
**Solution**: Cache reports only once all stages have completed and all results have been loaded.

# Tests
Tested locally that the `complete_and_successful?` method behaves as expected. Will verify caching behavior once this code is in staging

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -814,7 +814,7 @@ class SamplesController < ApplicationController
     background_id = get_background_id(@sample, params[:background])
 
     if pipeline_run
-      # Don't cache the response until all results for the pipeline run have been loaded
+      # Don't cache the response until all results for the pipeline run are available
       # so the displayed pipeline run status and report hover actions will be updated correctly.
       skip_cache = !pipeline_run.complete_and_successful? || params[:skip_cache] || false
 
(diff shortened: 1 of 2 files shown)


tfrcarvalho:
This seems a good short term fix.
On the long run, we should revisit what we cache. The bug is just due to `hasByteRanges` flag... seems awfully inefficient to ignore the entire cached report just because of that. We can consider caching only report contents but not the metadata...

cdebourcy:
> This seems a good short term fix.
> On the long run, we should revisit what we cache. The bug is just due to `hasByteRanges` flag... seems awfully inefficient to ignore the entire cached report just because of that. We can consider caching only report contents but not the metadata...

Makes sense. I do think conceptually it is best to cache a report only after it is 100% complete.

gregdingle:
> On the long run, we should revisit what we cache. 
Given the new report page performance, I'd be happy to consider removing the cache altogether, or running some kind of AB test.

PR #3011, opened 2020-02-03
[Heatmap Filtering] Send more metric data to enable all client-side threshold filters

j-x-han:
# Description

Currently, some threshold filters aren't functional (%id, L, log(1/e)) with client-side filtering because that data is currently not being returned from the backend. (IDSEQ-2189) This change sends the necessary data to enable all threshold filters to operate on the client-side.

# Tests

* Verified that NT and NR %id, L, log(1/e) threshold filters work with the `heatmap_filter_fe` flag enabled.


app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -641,7 +638,14 @@ class SamplesHeatmapView extends React.Component {
 
   getControlOptions = () => ({
     // Server side options
-    metrics: this.props.metrics,
+    metrics: [
```
tfrcarvalho: Which should just filter `this.props.metrics` by value. It will guarantee that text is in sync with what is sent from server side. It will also make bugs less intrusive (?), if values change the option will simply no appear on the client side.

PR #3013, opened 2020-02-03
Extend caching pre-condition to pre-cache too

cdebourcy:
# Description
This change ensures reports don't get precached if the last Batch job is not yet complete. This should remove an edge case where report hover actions are disabled even though they could be served.

It introduces a new edge case, which is that certain reports will miss their chance to get precached at all. However, that behavior is better than the previous behavior, where reports would get precached in an incomplete state. In order to assess how often the precache miss actually happens, a log event has been added.


cdebourcy:
@gregdingle you requested explanation of the check that was missing -- see comment in `ready_for_cache?` method. Super exotic edge case

cdebourcy:
> thanks for the follow up!

Actually I just realized this might mean pre-cache never happens. The last output_state may be loaded before the last Batch job completes. That will be the last incarnation of the result monitor for that sample, so it never gets a chance to do the pre-caching. I will need to think about this PR some more

cdebourcy:
Ah, I wasn't ready to merge yet!

PR #3015, opened 2020-02-04
[Heatmap Filtering] Make taxon level a client-side filter

j-x-han:
# Description

Migrate the Taxon Level filter to the client-side for users with the `heatmap_filter_fe` flag enabled (see IDSEQ-2185).

# Tests

* Verified that the Taxon Level filter works on the client-side when the `heatmap_filter_fe` flag is enabled.


app/helpers/heatmap_helper.rb
```diff
@@ -27,7 +27,7 @@ module HeatmapHelper
   ].freeze
 
   # Samples and background are assumed here to be vieweable.
-  def self.sample_taxons_dict(params, samples, background_id)
+  def self.sample_taxons_dict(params, samples, background_id, client_filtering_enabled = false)
```
tfrcarvalho: I started to user keyword arguments in ruby. It's not a pattern we follow for all currently, but i think it is better than just positional with default, because it allows you to have multiple optional parameters without having to set them all. I have been starting to push for it for new method declarations. Wdyt?

`def self.sample_taxons_dict(params, samples, background_id, client_filtering_enabled: false)`

app/helpers/heatmap_helper.rb
```diff
@@ -166,7 +169,7 @@ def self.fetch_top_taxons(
     min_reads = MINIMUM_READ_THRESHOLD,
     sort_by = DEFAULT_TAXON_SORT_PARAM,
     threshold_filters = [],
-    species_selected = true
+    _species_selected = true
```
tfrcarvalho: Can we remove this line and change affected calls?

app/helpers/heatmap_helper.rb
```diff
@@ -493,9 +500,10 @@ def self.top_n_query(query, sort_by, num_results, threshold_filters)
           count_type = '#{sort[:count_type]}' DESC,
           #{sort[:metric]} #{sort[:direction] == 'highest' ? 'DESC' : 'ASC'}
       ) b
-      -- Overfetch by a factor of 4 to allow for a) both count types, and
+      -- Overfetch by a factor of 8 to allow for a) both count types, and
```
tfrcarvalho: We do not want to change the factor unless the feature is enable.
I am not sure about the factor 8 also, because you will probably get a lot more species than genus... 
Any concerns with that?

app/helpers/heatmap_helper.rb
j-x-han: Sounds good to me! I also like how it forces you to be more explicit, I think it makes it more readable/understandable at first glance.

app/helpers/heatmap_helper.rb
j-x-han: Kept this parameter to use the flag to modify the query as we discussed.

app/helpers/heatmap_helper.rb
j-x-han: Reverted back to 4!

app/helpers/heatmap_helper.rb
```diff
@@ -159,6 +160,7 @@ def self.top_taxons_details(
     candidate_taxons.values.sort_by { |taxon| -1.0 * taxon["max_aggregate_score"].to_f }
   end
 
+  # rubocop:disable Metrics/ParameterLists
```
tfrcarvalho: By coincidence, we were talking about this, it has a way to ignore keyword parameters: `https://rubocop.readthedocs.io/en/latest/cops_metrics/#metricsparameterlists`.
Can you try to convert this to keyword parameters and set `CountKeywordArgs` to false on the rubocop configuration?

app/helpers/heatmap_helper.rb
```diff
@@ -187,7 +193,11 @@ def self.fetch_top_taxons(
       phage_clause = " AND is_phage = 1"
     end
 
+    # Select both species and genus level counts.
+    # Taxon counts are later filtered down to the desired taxon level in samples_taxons_details and top_taxons_details
+    # by calling only_species_or_genus_counts!, OR will be filtered on the client-side if a user has "heatmap_filter_fe" enabled.
```
tfrcarvalho: Dont forget to update the comments given the last change.

app/helpers/heatmap_helper.rb
j-x-han: Done

app/helpers/heatmap_helper.rb
j-x-han: Oops, good catch 😅

PR #3021, opened 2020-02-04
Add CZI employees method to feature

gregdingle:
# Description

See comments in code. 

# Tests

* run in localhost and see changes to user objects

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -84,7 +84,7 @@ class User < ApplicationRecord
   # general release. IMPORTANT NOTE: it assumes that only valid personnel have
   # a @chanzuckerberg address. This method is intended to be called from the rails
   # console.
-  def add_allowed_feature_for_all_czi(feature)
+  def self.add_allowed_feature_for_all_czi(feature)
     # covers contractors and regular employees
     users = User.where(
       "email LIKE '%@chanzuckerberg.com' OR email LIKE '%@contractor.chanzuckerberg.com'"
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -79,6 +79,17 @@ class User < ApplicationRecord
     end
   end
 
+  # This is a convenience function for the common case of adding all team
+  # members, test accounts, contractors and beta testers to a new feature before
+  # general release. IMPORTANT NOTE: it assumes that only valid personnel have
+  # a @chanzuckerberg address. This method is intended to be called from the rails
+  # console.
+  def add_allowed_feature_for_all_czi(feature)
+    # covers contractors and regular employees
+    users = User.where("email like '%@%chanzuckerberg.com'")
+    users.each { |u| u.add_allowed_feature(feature) }
+  end
+
   def can_upload(s3_path)
     return true if admin?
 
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -79,6 +79,17 @@ def remove_allowed_feature(feature)
...
   end
 
+  # This is a convenience function for the common case of adding all team
+  # members, test accounts, contractors and beta testers to a new feature before
+  # general release. IMPORTANT NOTE: it assumes that only valid personnel have
+  # a @chanzuckerberg address. This method is intended to be called from the rails
+  # console.
+  def add_allowed_feature_for_all_czi(feature)
+    # covers contractors and regular employees
+    users = User.where("email like '%@%chanzuckerberg.com'")
```
jshoe: could we merge `@contractor.chanzuckerberg.com` and `@chanzuckerberg.com` instead? Test firm might report something like this

gregdingle: sure that would be safer

app/assets/src/components/views/samples/MetadataUploadModal/Instructions.jsx
```diff
@@ -43,6 +43,19 @@ class UploadInstructions extends React.Component {
               >
                 CSV template.
               </a>
+              <p>
+                {`If your entered Host Organism does not match a supported host
+              genome, we will only subtract out ERCCs and the Human genome. You can
+              read more about how to request a new genome to be added to IDseq `}
+                <a
+                  href="https://chanzuckerberg.zendesk.com/hc/en-us/articles/360034790814-How-do-I-request-a-new-host-genome-be-added-to-IDseq-"
```
tfrcarvalho: Can we remove the redundancy? Define the description and url as constants somewhere?

tfrcarvalho: Wait is this part of this PR?

jshoe:
^ I think the last commit went to the wrong branch!

app/assets/src/components/views/samples/MetadataUploadModal/Instructions.jsx
gregdingle: Sorry wrong branch. I have two in progress now. One min please.

app/models/user.rb
```diff
@@ -79,6 +79,19 @@ def remove_allowed_feature(feature)
     end
   end
 
+  # This is a convenience function for the common case of adding all team
+  # members, test accounts, contractors and beta testers to a new feature before
+  # general release. IMPORTANT NOTE: it assumes that only valid personnel have
+  # a @chanzuckerberg address. This method is intended to be called from the rails
+  # console.
+  def self.add_allowed_feature_for_all_czi(feature)
```
tfrcarvalho: This kind of custom actions make more sense to me as a rake task.

gregdingle: true. and would that include the other `add_allowed_feature` method?

tfrcarvalho: No, what i find very specific is a the email parameter... E.g. eventually we might have a rake task for list of users, or biohub emails...

gregdingle:
I don't want to invest the time to make a rake task right now so I'm just going to close this PR. Anyone is welcome to pick up the idea. I only posted this PR because I had just run the exact commands in the console.

PR #3023, opened 2020-02-04
log if report is not precached

cdebourcy:
follow-up to https://github.com/chanzuckerberg/idseq-web/pull/3013, which was merged before I was ready


app/models/pipeline_run.rb
```diff
@@ -876,6 +876,8 @@ def monitor_results
         if ready_for_cache?
           Resque.enqueue(PrecacheReportInfo, id)
           Resque.enqueue(PrecacheReportInfoV2, id)
+        else
+          MetricUtil.log_analytics_event("PipelineReport_not_precached", nil)
```
tfrcarvalho: Could we make the second argument into a default keyword argument?

PR #3024, opened 2020-02-05
Add water control as possible DD column

gregdingle:
# Description

So users can see at a glance the new required field values.

image

image

# Tests

* add and remove column
* check default by deleting local storage


jshoe:
thoughts on placing it next to "Sample Type" in the list?

gregdingle:
> thoughts on placing it next to "Sample Type" in the list?

Good question. It's true that it is next to sample type in upload. On the other hand there is no apparent order in the add columns list. Is there?

jshoe:
I see yeah, seems fine to me

PR #3026, opened 2020-02-05
[Heatmap Filtering] Make top taxa per sample a client-side filter

j-x-han:
# Description

Implement the Taxa per Sample slider on the client rather than making a server call for users with the `heatmap_filter_fe` flag enabled (see IDSEQ-2186).

# Notes

This PR doesn't include any backend changes to load extra data, so increasing the Taxa per Sample value beyond what was initially loaded (default top 10 per sample) doesn't actually do anything.

# Tests

* Verified that the Taxa per Sample slider works on the client-side when the `heatmap_filter_fe` flag is enabled.

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
@@ -333,6 +333,7 @@ class SamplesHeatmapView extends React.Component {
         index: i,
         host_genome_name: sample.host_genome_name,
         metadata: processMetadata(sample.metadata, true),
+        taxa: [],
       };
       if (sample.taxons) {
         for (let j = 0; j < sample.taxons.length; j++) {
@@ -356,6 +357,8 @@ class SamplesHeatmapView extends React.Component {
             };
             allTaxonDetails[taxon.name] = allTaxonDetails[taxon.tax_id];
           }
+
+          sampleDetails[sample.sample_id].taxa.push(taxon.tax_id);
 
           this.props.metrics.forEach(metric => {
             let [metricType, metricName] = metric.value.split(".");
@@ -381,7 +384,7 @@ class SamplesHeatmapView extends React.Component {
     };
   }
 
-  filterTaxons() {
+  filterTaxa() {
     let {
       taxonFilterState,
       taxonPassesThresholdFilters,
@@ -395,17 +398,13 @@ class SamplesHeatmapView extends React.Component {
         let taxon = allTaxonDetails[taxonId];
         if (!taxonIds.has(taxonId) && this.taxonPassesSelectedFilters(taxon)) {
           if (taxonPassesThresholdFilters[taxon["index"]]) {
-            taxonDetails[taxon["id"]] = taxon;
-            taxonDetails[taxon["name"]] = taxon;
             taxonIds.add(taxon["id"]);
-            this.props.metrics.forEach(metric => {
-              filteredData[metric.value] = filteredData[metric.value] || [];
-              filteredData[metric.value][filteredData[metric.value].length] =
-                allData[metric.value][taxon["index"]] || [];
-            });
           }
         }
       });
+      [taxonIds, taxonDetails, filteredData] = this.getTopTaxaPerSample(
+        taxonIds
+      );
       taxonIds = Array.from(taxonIds);
     } else {
       taxonDetails = allTaxonDetails;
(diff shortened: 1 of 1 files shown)


app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -512,6 +511,53 @@ class SamplesHeatmapView extends React.Component {
...
+      allTaxonDetails,
+      selectedOptions,
+    } = this.state;
+    let { metric, taxonsPerSample } = selectedOptions;
+    let topTaxIds = new Set(),
+      topTaxonDetails = {},
+      filteredData = {};
+
+    Object.values(sampleDetails).forEach(sample => {
+      let filteredTaxaInSample = sample.taxa.filter(taxon =>
```
tfrcarvalho: nit: can we call it `taxonId`?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -512,6 +511,53 @@ class SamplesHeatmapView extends React.Component {
...
+      for (let taxId of filteredTaxaInSample) {
+        if (count >= taxonsPerSample) {
+          break;
+        } else if (!topTaxIds.has(taxId)) {
+          let taxon = allTaxonDetails[taxId];
+          topTaxIds.add(taxId);
+          topTaxonDetails[taxId] = allTaxonDetails[taxId];
+          topTaxonDetails[taxon["name"]] = allTaxonDetails[taxId];
+
+          this.props.metrics.forEach(metric => {
```
tfrcarvalho: Please load props on top of the function just like you loaded state, i.e. `const { metrics } = this.props`

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -512,6 +511,53 @@ class SamplesHeatmapView extends React.Component {
...
+    // Fetch the top N taxa from each sample, sorted by the selected metric,
+    // that passed all selected filters.
+    let {
+      sampleDetails,
+      allData,
+      allTaxonDetails,
+      selectedOptions,
+    } = this.state;
+    let { metric, taxonsPerSample } = selectedOptions;
+    let topTaxIds = new Set(),
```
tfrcarvalho: nit: can you add extra line before?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -512,6 +511,53 @@ class SamplesHeatmapView extends React.Component {
...
+          break;
+        } else if (!topTaxIds.has(taxId)) {
+          let taxon = allTaxonDetails[taxId];
+          topTaxIds.add(taxId);
+          topTaxonDetails[taxId] = allTaxonDetails[taxId];
+          topTaxonDetails[taxon["name"]] = allTaxonDetails[taxId];
+
+          this.props.metrics.forEach(metric => {
+            filteredData[metric.value] = filteredData[metric.value] || [];
+            filteredData[metric.value][filteredData[metric.value].length] =
```
tfrcarvalho: Is this not just an array push?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -681,9 +727,9 @@ class SamplesHeatmapView extends React.Component {
         "thresholdFilters",
         "readSpecificity",
         "metric",
+        "taxonsPerSample",
       ];
-      // TODO(julie): taxaPerSample should eventually be a frontend filter
-      const backendFilters = ["background", "taxonsPerSample"];
+      const backendFilters = ["background"];
```
tfrcarvalho: 💯

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -512,6 +511,53 @@ class SamplesHeatmapView extends React.Component {
...
+        } else if (!topTaxIds.has(taxId)) {
+          let taxon = allTaxonDetails[taxId];
+          topTaxIds.add(taxId);
+          topTaxonDetails[taxId] = allTaxonDetails[taxId];
+          topTaxonDetails[taxon["name"]] = allTaxonDetails[taxId];
+
+          this.props.metrics.forEach(metric => {
+            filteredData[metric.value] = filteredData[metric.value] || [];
+            filteredData[metric.value][filteredData[metric.value].length] =
+              allData[metric.value][taxon["index"]] || [];
```
tfrcarvalho: Just to confirm, this loads data for all samples for that particular tax id, right?

j-x-han: Yes, I added a comment by `allData` as you suggested above that will hopefully make the data structure more clear!

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: Oops, good catch.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: Done!

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: Done!

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: Done!

PR #3041, opened 2020-02-06
[Report page] [Bug] Fix csv column names

j-x-han:
# Description

[IDSEQ-2232] The sample report csv would sometimes be missing values because the column names were taken from the first row in the table, which could be missing values (for example, contigs).

This PR makes the column names a constant, which also has the nice property of guaranteeing that the columns are in a constant, predictable order.

An alternate solution of iterating through all rows and getting all the unique keys was also considered, but some initial tests showed that this could slow down csv generation, especially for large samples (3.8s -> 5s for a sample with almost 30k rows).

# Notes

Added a timer log to see how long it takes to convert the report info into csv format.

# Tests

* Verified that downloading a csv file that previously had missing values now contains all values.

### DIFF
--- a/app/services/pipeline_report_service.rb
+++ b/app/services/pipeline_report_service.rb
@@ -58,6 +58,38 @@ class PipelineReportService
     nil => "uncategorized",
   }.freeze
 
+  CSV_COLUMNS = [
+    "tax_id",
+    "tax_level",
+    "genus_tax_id",
+    "name",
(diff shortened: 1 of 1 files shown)


tfrcarvalho:
> This will now show columns even if none of the samples have data for it, but I think that's okay.

I think it's no only ok but also preferable, as we commit to a specific format.

PR #3049, opened 2020-02-07
[Heatmap filtering] Skip filter clauses in query if client-filtering enabled

j-x-han:
# Description

If the user has client-side filtering for the heatmap enabled, don't apply those filters to the query when the server is fetching data.

This allows saved heatmaps to behave as expected when the client-side filtering flag is enabled, rather than applying the filters to an already filtered set of data.

# Tests

* Verified that saved heatmaps behave as expected both when the `heatmap_frontend_fe` flag is enabled and disabled.

### DIFF
--- a/app/helpers/heatmap_helper.rb
+++ b/app/helpers/heatmap_helper.rb
@@ -194,22 +194,20 @@ module HeatmapHelper
   )
     categories_map = ReportHelper::CATEGORIES_TAXID_BY_NAME
     categories_clause = ""
+    read_specificity_clause = ""
+    phage_clause = ""
+
     unless client_filtering_enabled
       if categories.present?
         categories_clause = " AND superkingdom_taxid IN (#{categories.map { |category| categories_map[category] }.compact.join(',')})"
       elsif include_phage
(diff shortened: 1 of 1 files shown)


app/helpers/heatmap_helper.rb
```diff
@@ -192,21 +194,27 @@ def self.fetch_top_taxons(
   )
     categories_map = ReportHelper::CATEGORIES_TAXID_BY_NAME
     categories_clause = ""
-    if categories.present?
-      categories_clause = " AND superkingdom_taxid IN (#{categories.map { |category| categories_map[category] }.compact.join(',')})"
-    elsif include_phage
-      categories_clause = " AND superkingdom_taxid = #{categories_map['Viruses']}"
+    unless client_filtering_enabled
```
tfrcarvalho: Maybe initialize the clause strings above and use just one if?

tfrcarvalho: Add add a comment, why we are bypassing all the filters... (specially since we should probably have changed the request on client)

j-x-han: Condensed the `if` statements and added comment 😄

PR #3051, opened 2020-02-07
Limit number of "additional samples" in phylo tree creation

gregdingle:
# Description

In investigating a pipeline error, I discovered that phylo tree creation was broken for admins in prod. The root cause is a unscalable query. 

# Notes

* I isolated the slow query in the console 
* The SQL query is still really ugly and could be optimized more

```
SELECT `pipeline_runs`.`id`
FROM `pipeline_runs`
WHERE `pipeline_runs`.`sample_id` IN (3412, ...
  AND (id IN
         (SELECT max(id)
          FROM pipeline_runs
          WHERE job_status = 'CHECKED'
            AND sample_id IN
              (SELECT id
               FROM samples)
          GROUP BY sample_id))
  AND `pipeline_runs`.`id` IN (2913, ...
ORDER BY `pipeline_runs`.`id` DESC LIMIT 1000
```

# Tests

* See limit in query log

* Try advancing in localhost, see appropriate data return 

* Confirm 1000 rows in table

image

image

image

### DIFF
--- a/app/controllers/phylo_trees_controller.rb
+++ b/app/controllers/phylo_trees_controller.rb
@@ -32,6 +32,10 @@ class PhyloTreesController < ApplicationController
   before_action :assert_access, only: OTHER_ACTIONS
   before_action :check_access
 
+  # This limit was added because the phylo tree creation was timing out for admins
+  # and otherwise the results will grow without bound per user.
+  ELIGIBLE_PIPELINE_RUNS_LIMIT = 1000
+
   def index
     @project = []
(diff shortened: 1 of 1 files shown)


tfrcarvalho:
I understand we just want to avoid the timeout now, but this needs to be followed by a Jira ticket that paginates this information. We are not setting a limit on the samples to use by on the samples that can be selected.

gregdingle:
@tfrcarvalho Good point. See https://jira.czi.team/browse/IDSEQ-2242 .

PR #3056, opened 2020-02-08
[Report V2] Remove report v2 flags

tfrcarvalho:
# Description

* Removes `report_v2` feature flag
* Removes unreachable code in show action
* Rename v2 functions to drop v2 when the equivalent old function was dropped

# Tests

* Open report page and test all interactions
* WIP: test report precache
* WIP: test report csv download


app/assets/src/api/index.js
```diff
@@ -224,7 +224,7 @@ const getSampleStats = ({ domain, filters, projectId, search }) =>
     },
   });
 
-const getSample = ({ sampleId }) => get(`/samples/${sampleId}/show_v2.json`);
+const getSample = ({ sampleId }) => get(`/samples/${sampleId}.json`);
```
j-x-han: Do you also want to change the `getSampleReportData` and `getSampleReportInfo` methods earlier on in the file?

app/assets/src/components/views/SampleViewV2/SampleViewHeader.jsx
```diff
@@ -105,7 +105,7 @@ export default function SampleViewHeader({
             label: sample.name,
             id: sample.id,
             onClick: () => {
-              openUrl(`/samples/${sample.id}/show_v2`, event);
+              openUrl(`/samples/${sample.id}/show`, event);
```
j-x-han: This causes an error when selecting another sample in the project from the dropdown, I think the route is just ``/samples/${sample.id}``

app/assets/src/api/index.js
tfrcarvalho: You mean deleting the `getSampleReportInfo`? Yes.
What is your suggestion regarding `getSampleReportData` ?

tfrcarvalho: But I will probably delete on a future PR. At the same time I delete report_info on samples controller.

app/assets/src/components/views/SampleViewV2/SampleViewHeader.jsx
tfrcarvalho: Done!

PR #3060, opened 2020-02-10
Update hardcoded references to 2019-09-17 index

jshoe:
### Description
- I found 2 hardcoded references to "2019-09-17". They're not very important but I'll add a note about them to the index update guide.

### Notes
- The docker-compose.yml default gets overwritten as long as the env variables are loaded from Parameter Store.
- DEFAULT_S3_NT_INFO_DB_PATH looks like an edge case based on the comment.


jshoe:
Changed one reference 👍 not sure what to do about the docker-compose one b/c it should never be used in a deployed env but I think it does need a value for local development.

PR #3062, opened 2020-02-11
[Report page] Load less data for report page samples dropdown

j-x-han:
# Description

See IDSEQ-2123.

The current samples index action used to list all the project samples in a sample's dropdown on the report page (see picture below) fetches a lot more extra data (sample's metadata, pipeline run info, etc) when all that's necessary for the dropdown are the samples' names and ids.

image

This change adds param `basic` to `index_v2` which will skip sending all the extra sample details.

# Notes

Some rough comparisons for response size and time (just eyeballing locally by looking at the network tab and refreshing a few times):

* Project with 1 sample:
  * Before: 4.8KB, ~500-600ms
  * After: 3.0KB, ~100-250ms
* Project with 15 samples (Medical Detectives):
  * Before: 7.4KB, ~550-700ms
  * After: 3.2KB, ~100-250ms
* Project with 309 samples:
  * Before: 17.9KB, ~900ms-1.64s
  * After: 4.1KB, ~200-350ms

# Tests

* Loaded various sample reports and verified that the data loaded in for `projectSamples` was reduced and loaded in more quickly.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -180,9 +180,9 @@ class SamplesController < ApplicationController
       methods: [:private_until]
     )
 
+    basic = ActiveModel::Type::Boolean.new.cast(params[:basic])
     # If basic requested, then don't include extra details (ex: metadata) for each sample.
-    # params[:basic] becomes a string, so cast it into a boolean for this check.
-    unless ActiveModel::Type::Boolean.new.cast(params[:basic])
+    unless basic
       samples_visibility = get_visibility(limited_samples)
       # format_samples loads a lot of information about samples
       # There are many ways we can refactor: multiple endpoints for client to ask for the information
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -181,7 +181,8 @@ class SamplesController < ApplicationController
     )
 
     # If basic requested, then don't include extra details (ex: metadata) for each sample.
-    if params[:basic].blank?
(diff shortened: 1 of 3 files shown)


app/controllers/samples_controller.rb
```diff
@@ -179,15 +179,18 @@ def index_v2
...
-    # format_samples loads a lot of information about samples
-    # There are many ways we can refactor: multiple endpoints for client to ask for the information
-    # they actually need or at least a configurable function to get only certain data
-    details_json = format_samples(limited_samples).as_json()
-    limited_samples_json.zip(details_json, samples_visibility).map do |sample, details, visibility|
-      sample[:public] = visibility
-      sample[:details] = details
+
+    # If basic requested, then don't include extra details (ex: metadata) for each sample.
+    if params[:basic].blank?
```
tfrcarvalho: I guess we are ok with the client using anything on `basic` parameter?

j-x-han: I set it as `true` on the client, but it becomes a string `"true"` when passed in as a param, so I would need to do a check like `params[:basic] == "true"`. Wasn't sure what the best workaround for that was so I went with this, but let me know if there's a preferred method to handle it!

tfrcarvalho: Let's use: `basic = ActiveModel::Type::Boolean.new.cast(params[:basic])` as above. A bit more generic... 0 will be false...

j-x-han: Sounds good, made that change and updated the call to have basic default to `false`.

PR #3063, opened 2020-02-11
Expand sample row clickable area (IDSEQ-1700)

lvreynoso:
# Description

Building on the discussion in #2959, this PR dramatically expands the clickable area for each row in the Samples view, making it much easier to mark those checkboxes when selecting multiple samples. ~~I also added some simple drop shadow effects to create a kind of "button" visual indicating when a mouse click will safely select a sample instead of navigate to the sample details page.~~

_Updated, clickable area is invisible:_

invisible_click

# Notes

Moved checkbox `.scss` file to the same folder as the component, per our current convention. Also made rendering of the label in the checkbox dynamic so the css would play nicely with the new setup.

# Tests

* Go to My Data --> Samples, hover over checkbox area and click on it in different locations. Made sure samples were properly selected by creating heatmaps and downloads.
* Searched all files for other pages that used a `Checkbox` or `BaseTable` component and made styling compatible with their function.


app/assets/src/components/ui/controls/checkbox.scss
```diff
@@ -1,37 +1,45 @@
...
   height: 28px;
   align-items: center;
+  justify-content: center;
 
   input[type="checkbox"] {
-    position: absolute;
+    position: relative;
     opacity: 0;
     cursor: pointer;
+    clip: rect(0 0 0 0);
```
tfrcarvalho: Curious: why do we need this clip and the adjustments below?

app/assets/src/components/visualizations/table/base_table.scss
```diff
@@ -134,3 +134,31 @@
...
+    overflow: visible;
+    height: 100%;
+    width: 50px;
+    margin: -11px -10px -10px -10px;
+    padding-left: 0px;
+    justify-content: center;
+    border-radius: 4px;
+
+    &:hover {
+      background-color: $primary-lighter;
```
tfrcarvalho: This still seems to dark... just wait on design input, probably.

app/assets/src/components/visualizations/table/base_table.scss
```diff
@@ -134,3 +134,31 @@
...
+      opacity: 0.7;
+      transition: background-color 0.3s;
```
tfrcarvalho: We do not normally use transitions... I would not see a reason to start now... but again, wait on design input.

app/assets/src/components/visualizations/table/base_table.scss
```diff
@@ -134,3 +134,31 @@
...
+    height: 100%;
+    width: 50px;
+    margin: -11px -10px -10px -10px;
+    padding-left: 0px;
+    justify-content: center;
+    border-radius: 4px;
+
+    &:hover {
+      background-color: $primary-lighter;
+      opacity: 0.7;
```
tfrcarvalho: You should not need this. Should just have the right color.

app/assets/src/components/visualizations/table/base_table.scss
```diff
@@ -134,3 +134,31 @@
...
+  // LUCIA: We pass in the "selectableCell" className to
+  // our checkbox to modify the enclosing div as follows;
+  // this creates a large, clickable "button" the height of
+  // the row, making samples easier to select.
+  .selectableCell {
+    position: absolute;
+    overflow: visible;
+    height: 100%;
+    width: 50px;
+    margin: -11px -10px -10px -10px;
```
tfrcarvalho: Why margin vs. padding?

lvreynoso:
Yeah, after discussion with her I've made the clickable area invisible.

app/assets/src/components/ui/controls/checkbox.scss
lvreynoso: Hmm, turns out we don't. Thanks for catching this. (I had added these lines while working on this PR because the default, browser-styled checkbox had reappeared at some point.)

app/assets/src/components/visualizations/table/base_table.scss
lvreynoso: Margin here increases the size of the `div` here to cover the entire clickable area. No such thing as negative padding.

app/assets/src/components/ui/controls/checkbox.scss
```diff
@@ -1,37 +1,41 @@
...
   }
 
   .checkmark {
-    position: absolute;
-    top: 6px;
-    left: 6px;
+    position: relative;
     height: 16px;
     width: 16px;
+    min-width: 16px;
```
tfrcarvalho: Why do we need `min-width` when we have `width`?

lvreynoso:
I thought about that, but I fiddled with the `Checkbox` just enough that to do something else similar would require just a couple of lines of css, like what's at the bottom of `base_table.scss`. I'm not sure a wrapper component is justified because the styling would be particular to whatever page it's a part of.

app/assets/src/components/ui/controls/checkbox.scss
lvreynoso: Yeah, there checkbox would shrink if the window was small enough - but I decided to use `flex: none` instead.

lvreynoso:
It turns out those use a different component! I tried to merge the two but it turned out to be a *lot* trickier than it seemed. So none of the drop downs are affected by this, I checked :)

tfrcarvalho:
> It turns out those use a different component! I tried to merge the two but it turned out to be a _lot_ trickier than it seemed. So none of the drop downs are affected by this, I checked :)

Thanks for checking!

PR #3064, opened 2020-02-11
Update CLI instructions

gregdingle:
# Description

This updates the CLI instructions according to @oliviabholmes spec. It also updates a few more places to "host organism", and refactors one method. 

image

image

# Tests

* load cli_user_instructions page locally, see updated text

### DIFF
--- a/app/models/host_genome.rb
+++ b/app/models/host_genome.rb
@@ -46,7 +46,7 @@ class HostGenome < ApplicationRecord
   def as_json(options = {})
     hash = super(options)
     hash[:ercc_only] = ercc_only?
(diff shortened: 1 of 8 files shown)


app/models/host_genome.rb
```diff
@@ -67,4 +67,8 @@ def ercc_only?
     s3_star_index_path.starts_with?(HostGenome::ERCC_PATH_PREFIX) &&
       s3_bowtie2_index_path.starts_with?(HostGenome::ERCC_PATH_PREFIX)
   end
+
+  def show_as_option?
+    user.nil?
```
tfrcarvalho: By curiosity, when is user nil?

PR #3065, opened 2020-02-11
[Heatmap filters] [Bug] Fix undefined threshold filters on old saved heatmaps

j-x-han:
# Description

See IDSEQ-2245.

Some older saved visualizations have the `thresholdFilters` saved in a different format (missing `metricDisplay`, which is used for displaying the `thresholdFilterTag`), which results in the threshold filters appearing as `undefined`.

This is actually [handled](https://github.com/chanzuckerberg/idseq-web/blob/4795743af2dc952f13f5cf899db282867cd24f7e/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx#L193) for most URLs, but because saved visualizations have a shortened URL that don't include all the selected options, `savedParams` need to be handled as well.

# Tests

* Loaded an old saved heatmap and verified that the threshold filters showed up as expected.

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
@@ -70,7 +70,7 @@ class SamplesHeatmapView extends React.Component {
     this.urlParams = this.parseUrlParams();
     // URL params have precedence
     this.urlParams = {
-      ...props.savedParamValues,
+      ...this.parseSavedParams(),
       ...this.urlParams,
     };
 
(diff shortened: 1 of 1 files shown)


j-x-han:
> I wonder if it would be possible to change the saved heatmaps to avoid leaving these backward compatibility snippets in the code indefinitely?

Definitely would be doable, the change would essentially be updating the `thresholdFilters` in old heatmap entries in the Visualization database to include `metricDisplay`.

One other thing I noticed while looking into this, some older heatmaps also include aggregate score as an applied filter, which we no longer have (see https://staging.idseq.net/visualizations/heatmap/20), so those also show up as `undefined`. This PR won't fix those tags, is that something we would be concerned about?

PR #3073, opened 2020-02-13
Add to PHAGE_TAXIDS list for index update

jshoe:
### Description
- Add 4 new phage tax_ids: `[2703653, 2704941, 2707030, 2707031]`
- Add new PHAGE_TAXIDS to the list as described in the index update guide.

### Notes
- Generated by the command as described: `TaxonLineage.connection.select_all("SELECT taxid FROM taxon_lineages WHERE family_taxid NOT IN (10472, 10474, 10477, 10656, 10659, 10662, 10699, 10744, 10841, 10860, 10877, 11989, 157897, 292638, 324686, 423358, 573053, 1232737) AND (species_name REGEXP '[[:<:]]phage[[:>:]]' OR genus_name REGEXP '[[:<:]]phage[[:>:]]')").to_a.map  { |entry| entry["taxid"] }.uniq`

### Tests
- Checked on NCBI (https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi) and confirmed they are indeed phages.


lvreynoso:
This array is rather large, would it not save some read time to make it a hash map?

jshoe:
@lvreynoso this actually gets inserted into an SQL query in the next step, so the db handles it :)

PR #3074, opened 2020-02-13
Add validation to all models, part 1 of 2

gregdingle:
# Description

This is a revival of #2478 using a more incremental approach. Apologies for the long description but I feel it's important explain more than usual for such a change.

## Motivation

Historically, idseq-web has used few active record validators and has not consistently restricted the use of `NULL` values in MySQL (which are passed on as `nil` in ruby and `null` in JS). This causes cognitive load on all developers of idseq-web, if not outright bugs. 

For example, recently I added a null host genome to a sample in staging by accident because of a bug. The effect was to crash the data discovery interface for all users who could see the sample. Programmers throughout the app should not have to defend against any value coming from the database being `NULL`. They should be able to safely assume that certain values are always present. 
 
(Note that the uncertainty is compounded by the dynamic languages Ruby and Javascript which means that there are no guarantees anywhere in the stack.)

See also https://guides.rubyonrails.org/active_record_validations.html#why-use-validations-questionmark .

## Background

Six months ago, In an attempt to improve the situation, I found all the columns in our prod database which have all non-NULL values but which will still allow NULLs. See https://czi.quip.com/LWbJAyLuX7fn/2019-08-02-NULL-values-in-database . In #2478, I merely wrote a migration to change columns to `NOT NULL`. I abandoned #2478 because the rollout would have been too risky. Any intermediate state of the database where a `NOT NULL` column would need to hold a `NULL` temporarily would cause a crash. And testing all affected code paths would have been impractical.

## Implementation

This PR leverages active record validators to check for `NULL` values while avoiding any crash in production by making validation depend on the deployment environment. For example:

```
  validates :s3_nt_db_path, presence: true, if: :mass_validation_enabled?

  def mass_validation_enabled?
    Rails.env.development? ||
      Rails.env.test? ||
      AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
  end
```
 The `AppConfig` exists to later enable switching on in staging and production.

Note that the validators in this PR provide considerable value as documentation alone. 

The full change algorithm I followed was:
1. Inspected prod db (as described above)
2. Created list of "no nulls" columns that were not "not null" already
3. Added `validates :attribute, presence: true` to all such columns as model attributes
4. Added extra validation where obvious
* Inclusion in set of constants
* Range of number
5. Added `if: :mass_validation_enabled?` to all new validators
6. Ran all mintiest and rspec
        * Updated fixtures and factories as needed
        * Updated test cases only when no good alternative

# Notes

* To aid my work and future code readers, I standardized the order of lines at the top of models. The order I aimed for is:
  1. Assocs
  2. Validations
  3. Callbacks
  4. Constants
* I discovered some Rails-level associations were missing, most often from some model that `belongs_to` `TaxonLineage` via `taxid`. I added these but commented them out if they caused any unit test failure for lack of time. I feel the associations are important at least as documentation. 
* I found we always use integer in `[0, 1]` instead of Rails and Mysql boolean types. I don't think it is worth the effort to refactor these but I would support changing the pattern for new columns.
* I found some columns which had always the same value in the prod database so it looks like they are not serving any purpose. Similarly, the taxon scoring model model is a singleton after several years in production.
* Annoying mix of `taxid` and `tax_id`. What is preferred?
* Several tables copy data from an associated table without explanation. This violates the normalization of the database (which is nearly complete). What is the point of having `tax_name` and `tax_id` when `tax_name` is derived from `tax_id`?
* A lot of "status" columns that allow `NULL` which seems pretty bad
* A bunch of json columns with different APIs. Why can't we use one of many gems for json attributes? 
* I added some comments on current values for columns that looked like enums but for which it wasn't clear where the possible values are defined. Any pointers would be most welcome to complete the documentation. 
* I got an odd rspec error when running certain files in isolation, such as: `"WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request"`

# Tests

Ran all mintiest and rspec. See point 6. above. 

Created a project and uploaded a sample in local host.

# Rollout plan

After merging to master, my plan is to: 
1. wait two weeks for any validation errors that happen in developers's localhosts. 
2. turn on `ENABLE_MASS_VALIDATION` in staging
3. ask QA team to exercise all features that create or update objects
4. wait two weeks for any new validation errors that happen in staging
5. turn on `ENABLE_MASS_VALIDATION` in prod
6. react to any new prod validation error by turning off `ENABLE_MASS_VALIDATION` then fixing the issue
7. begin on the second PR which will add `NOT NULL`s to the database schema

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -1136,6 +1136,8 @@ class SamplesController < ApplicationController
 
     params[:input_files_attributes].reject! { |f| f["source"] == '' }
     params[:alignment_config_name] = AlignmentConfig::DEFAULT_NAME if params[:alignment_config_name].blank?
+    params[:status] = Sample::STATUS_CREATED
+    params[:user] = current_user
 
     @sample = Sample.new(params)
     @sample.project = project if project
--- a/app/models/amr_count.rb
+++ b/app/models/amr_count.rb
@@ -22,11 +22,15 @@
 #   "Tmt"]
 class AmrCount < ApplicationRecord
   belongs_to :pipeline_run
-
-  validates :gene, presence: true, if: :mass_validation_enabled?
-  validates :allele, presence: true, if: :mass_validation_enabled?
-  validates :drug_family, presence: true, if: :mass_validation_enabled?
-  # AmrCount is designed to handle partial data. See pipeline_run_spec.rb.
+  # AmrCount is designed to handle partial data. See db_load_amr_counts.
+  validates :gene, presence: true, allow_nil: true, if: :mass_validation_enabled?
+  validates :allele, presence: true, allow_nil: true, if: :mass_validation_enabled?
+  validates :drug_family, presence: true, allow_nil: true, if: :mass_validation_enabled?
+  validates :annotation_gene, presence: true, allow_nil: true, if: :mass_validation_enabled?
+  validates :genbank_accession, presence: true, allow_nil: true, if: :mass_validation_enabled?
   validates :coverage, inclusion: 0..100, allow_nil: true, if: :mass_validation_enabled?
   validates :depth, numericality: { greater_than: 0 }, allow_nil: true, if: :mass_validation_enabled?
+  validates :total_reads, numericality: { greater_than: 0, integer_only: true }, allow_nil: true, if: :mass_validation_enabled?
+  validates :dpm, numericality: { greater_than_or_equal_to: 0 }, allow_nil: true, if: :mass_validation_enabled?
+  validates :rpm, numericality: { greater_than_or_equal_to: 0 }, allow_nil: true, if: :mass_validation_enabled?
 end
--- a/app/models/amr_count.rb
+++ b/app/models/amr_count.rb
@@ -1,6 +1,30 @@
 # Model that has data fields necessary for displaying
 # anti microbial resistance (AMR) information for a
 # given pipeline run of a sample.
+#
+# Currently the following values exist for drug_family:
+#   ["AGly",
+#   "Bla",
+#   "Colistin",
+#   "Ecoli_Bla",
+#   "Fcd",
+#   "Fcyn",
+#   "Flq",
+#   "Gly",
+#   "hydrolase_Bla",
+#   "MLS",
+#   "Nitroimidazole_Gene_Ntmdz",
+#   "Oxzln",
+#   "Phe",
+#   "Rif",
+#   "Sul",
+#   "Tet",
+#   "Tmt"]
 class AmrCount < ApplicationRecord
   belongs_to :pipeline_run
+  validates :pipeline_run, presence: true
+
+  # AmrCount is designed to handle partial data. See pipeline_run_spec.rb.
+  validates :coverage, inclusion: 0..100, allow_nil: true
+  validates :depth, numericality: { greater_than: 0 }, allow_nil: true
 end
--- a/app/models/amr_count.rb
+++ b/app/models/amr_count.rb
@@ -22,9 +22,11 @@
 #   "Tmt"]
 class AmrCount < ApplicationRecord
   belongs_to :pipeline_run
-  validates :pipeline_run, presence: true
 
+  validates :gene, presence: true, if: :mass_validation_enabled?
+  validates :allele, presence: true, if: :mass_validation_enabled?
+  validates :drug_family, presence: true, if: :mass_validation_enabled?
   # AmrCount is designed to handle partial data. See pipeline_run_spec.rb.
-  validates :coverage, inclusion: 0..100, allow_nil: true
-  validates :depth, numericality: { greater_than: 0 }, allow_nil: true
+  validates :coverage, inclusion: 0..100, allow_nil: true, if: :mass_validation_enabled?
+  validates :depth, numericality: { greater_than: 0 }, allow_nil: true, if: :mass_validation_enabled?
 end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -8,13 +8,11 @@ class ApplicationRecord < ActiveRecord::Base
   after_destroy { |record| log_analytics record, "destroyed" }
 
   # Condition for rollout of mass addition of validation rules.
-  # Memoizes for performance.
+  # Cached for performance.
   def mass_validation_enabled?
-    if AppConfig.table_exists?
-      if defined?(@enable_mass_validation)
-        return @enable_mass_validation
-      else
-        return @enable_mass_validation = AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
+    if AppConfig.table_exists? # for migrations previous to AppConfig creation
+      return Rails.cache.fetch("ApplicationRecord.mass_validation_enabled?") do
+        AppConfigHelper.get_app_config(AppConfig::ENABLE_MASS_VALIDATION)
       end
     end
   end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -9,7 +9,9 @@ class ApplicationRecord < ActiveRecord::Base
 
   # Condition for rollout of mass addition of validation rules.
   def mass_validation_enabled?
-    @enable_mass_validation ||= AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
+    if AppConfig.table_exists?
+      return @enable_mass_validation ||= AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
+    end
   end
 
   # Set current user and request to global for use in logging.
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -8,9 +8,8 @@ class ApplicationRecord < ActiveRecord::Base
   after_destroy { |record| log_analytics record, "destroyed" }
 
   # Condition for rollout of mass addition of validation rules.
-  self._enable_mass_validation = nil
   def mass_validation_enabled?
-    self._enable_mass_validation ||= AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
+    @enable_mass_validation ||= AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
   end
 
   # Set current user and request to global for use in logging.
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -8,7 +8,7 @@ class ApplicationRecord < ActiveRecord::Base
   after_destroy { |record| log_analytics record, "destroyed" }
 
   # Condition for rollout of mass addition of validation rules.
-  def validate_all?
+  def mass_validation_enabled?
     Rails.env.development? ||
       Rails.env.test? ||
       AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -7,12 +7,9 @@ class ApplicationRecord < ActiveRecord::Base
   after_update { |record| log_analytics record, "updated" }
   after_destroy { |record| log_analytics record, "destroyed" }
 
-  # Load once at boot for performance
-  ENABLE_MASS_VALIDATION = AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
-
   # Condition for rollout of mass addition of validation rules.
   def mass_validation_enabled?
-    ENABLE_MASS_VALIDATION
+    self._enable_mass_validation ||= AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
   end
 
   # Set current user and request to global for use in logging.
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -7,6 +7,13 @@ class ApplicationRecord < ActiveRecord::Base
   after_update { |record| log_analytics record, "updated" }
   after_destroy { |record| log_analytics record, "destroyed" }
 
+  # Condition for rollout of mass addition of validation rules.
+  def validate_all?
+    Rails.env.development? ||
+      Rails.env.test? ||
+      AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
+  end
+
   # Set current user and request to global for use in logging.
   # See https://stackoverflow.com/a/11670283/200312
   class << self
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -8,6 +8,7 @@ class ApplicationRecord < ActiveRecord::Base
   after_destroy { |record| log_analytics record, "destroyed" }
 
   # Condition for rollout of mass addition of validation rules.
+  self._enable_mass_validation = nil
   def mass_validation_enabled?
     self._enable_mass_validation ||= AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
   end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -7,11 +7,12 @@ class ApplicationRecord < ActiveRecord::Base
   after_update { |record| log_analytics record, "updated" }
   after_destroy { |record| log_analytics record, "destroyed" }
 
+  # Load once at boot for performance
+  ENABLE_MASS_VALIDATION = AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
+
   # Condition for rollout of mass addition of validation rules.
   def mass_validation_enabled?
-    Rails.env.development? ||
-      Rails.env.test? ||
-      AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
+    ENABLE_MASS_VALIDATION
   end
 
   # Set current user and request to global for use in logging.
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -8,9 +8,14 @@ class ApplicationRecord < ActiveRecord::Base
   after_destroy { |record| log_analytics record, "destroyed" }
 
   # Condition for rollout of mass addition of validation rules.
+  # Memoizes for performance.
   def mass_validation_enabled?
     if AppConfig.table_exists?
-      return @enable_mass_validation ||= AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
+      if defined?(@enable_mass_validation)
+        return @enable_mass_validation
+      else
+        return @enable_mass_validation = AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
+      end
     end
   end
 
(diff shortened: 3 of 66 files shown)


app/controllers/samples_controller.rb
```diff
@@ -1136,6 +1136,8 @@ def create
 
     params[:input_files_attributes].reject! { |f| f["source"] == '' }
     params[:alignment_config_name] = AlignmentConfig::DEFAULT_NAME if params[:alignment_config_name].blank?
+    params[:status] = Sample::STATUS_CREATED
```
gregdingle: I added this because it seems like a sample should always be created with a status and an owner. I'm not actually sure where this `create` method gets called though.

app/models/amr_count.rb
```diff
@@ -1,6 +1,32 @@
...
+#   "Sul",
+#   "Tet",
+#   "Tmt"]
 class AmrCount < ApplicationRecord
   belongs_to :pipeline_run
+
+  validates :gene, presence: true, if: :mass_validation_enabled?
+  validates :allele, presence: true, if: :mass_validation_enabled?
+  validates :drug_family, presence: true, if: :mass_validation_enabled?
+  # AmrCount is designed to handle partial data. See pipeline_run_spec.rb.
```
gregdingle: @lvreynoso any elaboration welcome 🙂

app/models/archived_background.rb
```diff
@@ -1,2 +1,5 @@
 class ArchivedBackground < ApplicationRecord
```
gregdingle: it would be nice to have some class comments here @cdebourcy . when should we be archiving backgrounds?

app/models/host_genome.rb
```diff
@@ -7,8 +7,7 @@ class HostGenome < ApplicationRecord
   # in light of the data. See showAsOption below.
   # See https://jira.czi.team/browse/IDSEQ-2193.
   belongs_to :user, optional: true
-
-  before_create :add_default_metadata_fields!
+  belongs_to :default_background, optional: true, class_name: "Background", foreign_key: :default_background_id
```
gregdingle: seems equivalent to method removed below... passed all tests

app/models/location.rb
```diff
@@ -5,6 +5,11 @@ class Location < ApplicationRecord
   belongs_to :subdivision, class_name: "Location", optional: true
   belongs_to :city, class_name: "Location", optional: true
 
+  validates :lat, inclusion: -90..90, if: :mass_validation_enabled?
```
gregdingle: cc @jshoe

app/models/metadata_field.rb
```diff
@@ -3,6 +3,11 @@ class MetadataField < ApplicationRecord
   has_and_belongs_to_many :projects
   has_many :metadata, dependent: :destroy
 
+  # Obeying the same rules as in MetadataHelper validate_metadata_csv_for_samples
+  # Metadata fields are shared globally and can be created by users.
```
gregdingle: cc @MarkAZhang to verify

app/models/phylo_tree.rb
```diff
@@ -3,16 +3,34 @@ class PhyloTree < ApplicationRecord
...
   belongs_to :project
+  # NOTE: this conflicts with phylo_trees_controller_spec.rb
+  # belongs_to :taxon_lineage, class_name: "TaxonLineage", foreign_key: :taxid, primary_key: :taxid
+
+  # TODO: (gdingle): should this be scoped to project or user?
   validates :name, presence: true, uniqueness: { case_sensitive: false }
-  after_create :create_visualization
+  validates :taxid, presence: true, if: :mass_validation_enabled?
+  validates :tax_name, presence: true, if: :mass_validation_enabled?
+  validates :tax_level, presence: true, inclusion: { in: [
```
gregdingle: cc @cdebourcy to verify

app/models/pipeline_run.rb
```diff
@@ -108,9 +110,30 @@ class PipelineRun < ApplicationRecord
   STATUS_CHECKED = 'CHECKED'.freeze
   STATUS_FAILED = 'FAILED'.freeze
   STATUS_RUNNING = 'RUNNING'.freeze
-  STATUS_RUNNABLE = 'RUNNABLE'.freeze
+  STATUS_RUNNABLE = 'RUNNABLE'.freeze # TODO: (gdingle): not used anywhere?
```
gregdingle: `STATUS_RUNNABLE` is not referred to anywhere in idseq-web and it does not appear in the prod DB so it's not clear how it supposed to be used

app/models/pipeline_run.rb
```diff
@@ -108,9 +110,30 @@ class PipelineRun < ApplicationRecord
...
+  # | job_status                                |
+  # +-------------------------------------------+
+  # | 1.Host Filtering-FAILED                   |
+  # | 1.Host Filtering-FAILED|READY             |
+  # | 2.GSNAPL/RAPSEARCH alignment-FAILED       |
+  # | 2.GSNAPL/RAPSEARCH alignment-FAILED|READY |
+  # | 2.GSNAPL/RAPSEARCH alignment-RUNNING      |
+  # | 3.Post Processing-FAILED                  |
+  # | 3.Post Processing-FAILED|READY            |
+  # | 4.De-Novo Assembly-FAILED|READY           |
```
gregdingle: this was a new one for me 🙂

app/models/taxon_count.rb
```diff
@@ -8,8 +13,31 @@ class TaxonCount < ApplicationRecord
...
+  ], }, if: :mass_validation_enabled?
+
   COUNT_TYPE_NT = 'NT'.freeze
   COUNT_TYPE_NR = 'NR'.freeze
+  validates :count_type, presence: true, inclusion: { in: [
+    COUNT_TYPE_NT,
+    COUNT_TYPE_NR,
+  ], }, if: :mass_validation_enabled?
+
+  # NOTE: some existing rspec tests assume a value of zero
```
gregdingle: Not sure whether to fix the tests or whether zero is a valid count. No zeros exist in prod db.

app/models/taxon_scoring_model.rb
```diff
@@ -1,3 +1,5 @@
+# NOTE: This is DEPRECATED because there has only ever been one model in use.
```
gregdingle: that's my current opinion

app/models/taxon_summary.rb
```diff
@@ -1,4 +1,24 @@
...
   belongs_to :background
+  # NOTE: this conflicts with pipeline_report_service_spec.rb
+  # belongs_to :taxon_lineage, class_name: "TaxonLineage", foreign_key: :tax_id, primary_key: :taxid
+  validates :tax_id, presence: true, if: :mass_validation_enabled?
+
+  validates :count_type, presence: true, inclusion: { in: [
+    TaxonCount::COUNT_TYPE_NT,
+    TaxonCount::COUNT_TYPE_NR,
+    # Unclear meaning. Exists in database, but not in codebase.
+    "NT+",
```
gregdingle: any comment?

test/controllers/power_controller_test.rb
```diff
@@ -517,18 +518,18 @@ class PowerControllerTest < ActionDispatch::IntegrationTest
   test 'joe can create phylo_tree to joe_project from public samples' do
     sign_in(:joe)
     post "/phylo_trees/create", params: { name: 'new_phylo_tree', projectId: projects(:joe_project).id,
-                                          taxId: 1, pipelineRunIds: [pipeline_runs(:public_project_sampleA_run).id,
-                                                                     pipeline_runs(:public_project_sampleB_run).id,],
-                                          tax_name: 'some species', }
+                                          taxId: 573, pipelineRunIds: [pipeline_runs(:public_project_sampleA_run).id,
```
gregdingle: I changed this to a fixture taxa that had realistic data

gregdingle:
Update: I'm not sure why rspec is failing in travis: https://travis-ci.com/chanzuckerberg/idseq-web/jobs/287082372 .  It is passing on the same code for me locally. I'll investigate more.

test/fixtures/host_genomes.yml
```diff
@@ -1,17 +1,29 @@
...
 
 two:
   name: host genome 2
   metadata_fields: [collection_date, sample_type, sex, age, admission_date, nucleotide_type, water_control]
 
 human:
   name: Human
+  default_background: real_background
   metadata_fields: [collection_date, sample_type, sex, age, admission_date, nucleotide_type, core_field, water_control]
+  s3_bowtie2_index_path: s3://czbiohub-infectious-disease/references/human/bowtie2_genome.tar.gz
```
cdebourcy: these S3 paths are obsolete. We've also made a choice to avoid revealing the current ones in the codebase, could we preserve that?

gregdingle: For sure, I'll change these. How about `s3://bowtie2_genome.tar.gz` ?

Also, for what it's worth, I copied this string from elsewhere in the codebase and appears in many places.

lvreynoso:
If I'm reading this correctly, you're not validating for columns that could be transiently null? I noticed that in the phylo_tree example @cdebourcy pointed out, some of the columns with null values that you wanted to change in the old PR are different from the columns validated in this one. Would this mean there's still a possibility of a null propagating upward? 

I wonder if switching to typed languages might be better in the long run - something like Typescript for React and Rust or Go in the backend. It's true that assuming you're never working with null values makes development faster, but I don't know if that's ever a "safe" assumption as long as the guard against it is a human process.

cdebourcy:
> If I'm reading this correctly, you're not validating for columns that could be transiently null? I noticed that in the phylo_tree example @cdebourcy pointed out, some of the columns with null values that you wanted to change in the old PR are different from the columns validated in this one. Would this mean there's still a possibility of a null propagating upward?
> 
> I wonder if switching to typed languages might be better in the long run - something like Typescript for React and Rust or Go in the backend. It's true that assuming you're never working with null values makes development faster, but I don't know if that's ever a "safe" assumption as long as the guard against it is a human process.

Great thoughtful comment, Lucia, and thanks for reminding us of the transiently-null issue from the previous PR (https://github.com/chanzuckerberg/idseq-web/pull/2478).

Greg, I still struggle with the cost-benefit tradeoff in this PR. To me, this change seems a lot riskier than any risk we might have from occasionally neglecting to handle a null value -- an issue that hasn't occurred with any substantial frequency. I'm unable to sign off on the pieces you tagged me in! I would also be inclined to think there are more light-weight ways to guard against potential issues with null values. Ultimately it's up to what the team prefers for their development experience -- maybe you can schedule a team-wide meeting to discuss.

app/models/location.rb
jshoe: Hm, I think during the design we were allowing for lat and lng to be null for edge cases

gregdingle:
Thanks for the reviews so far. I will schedule a team-wide meeting to discuss the cost-benefit tradeoff @cdebourcy . 

In the meantime, to address your comments @lvreynoso :

> If I'm reading this correctly, you're not validating for columns that could be transiently null?

No, this PR validates columns that appear to be always not null based on existing data. Although I have not found any examples, there is always a possibility that some code is temporarily storing a null value.

> I noticed that in the phylo_tree example @cdebourcy pointed out, some of the columns with null values that you wanted to change in the old PR are different from the columns validated in this one. 

Sorry I can't find the phylo_tree example here. Can you point me to it?

> Would this mean there's still a possibility of a null propagating upward?

While I tried to be complete in my earlier analysis and removal of NULLs, there are probably a few columns which should not contain null values which still do.

> I wonder if switching to typed languages might be better in the long run - something like Typescript for React and Rust or Go in the backend. It's true that assuming you're never working with null values makes development faster, but I don't know if that's ever a "safe" assumption as long as the guard against it is a human process.

Migrating to a typed language would be a much more complete solution and require a much bigger discussion. I've witnessed a couple such migrations transitions that were successful. In any case, adding typing information anywhere in the stack would aid any transition to language-level types.

app/models/location.rb
gregdingle: Ok that's interesting. Currently all values in prod are not null. Could you describe the edge cases so I can add them as comments to the columns?

jshoe: I would write: Lat and lng may be null if the location provider is missing coordinates, as opposed to placing a location at (0,0)

gregdingle: Ok sounds good. And the downstream locations code will handle nulls by filtering them out?

jshoe: Correct

app/models/application_record.rb
```diff
@@ -7,6 +7,13 @@ class ApplicationRecord < ActiveRecord::Base
   after_update { |record| log_analytics record, "updated" }
   after_destroy { |record| log_analytics record, "destroyed" }
 
+  # Condition for rollout of mass addition of validation rules.
+  def mass_validation_enabled?
+    Rails.env.development? ||
+      Rails.env.test? ||
+      AppConfig.find_by(key: AppConfig::ENABLE_MASS_VALIDATION)
```
davidrissato: I'd say we need to optimize this method, it can become slow during run_time. Remember this code will be invoked for every field, and it involves a query to the database.
Maybe we should just have a constant loaded when you start rails, and changes in this field would only require a restart of the container.

gregdingle: @davidrissato  Good point. Will do. 

@tfrcarvalho you had some comment about removing the `Rails.env` dependency and relying only on `AppConfig` is that right?

tfrcarvalho: Yes, and actually for test - in theory at least - you should run all the relevant tests with the flag both enabled and disabled, to test current and future behavior. 
You can do loops where you run the same tests and enable/disable the flag.

app/models/amr_count.rb
lvreynoso: The code that loads amr count results for a pipeline run downloads the amr results file from s3 and then adds values for those keys (among others) only if they are present in the results table; they do not write a default value. It's possible that those attributes will not be present if something is wrong either with our data source (the ARGAnnot v2 fasta file) or the run_srst2 step in the pipeline. They shouldn't be null values when we try to save the AmrCount in rails (knowing what gene we detected is the whole point) but in theory it's possible.

I think it's good to validate the fields, as long as one model failing to save doesn't interfere with the rest, since the `AmrCounts` are saved in a batch. See https://github.com/chanzuckerberg/idseq-web/blob/8ae5dd19f3dee65da920ce6f9ef566d52c10b768/app/models/pipeline_run.rb#L643

app/models/application_record.rb
gregdingle: Ok I made the change and I tested locally where `ENABLE_MASS_VALIDATION` is both true and false.

app/models/amr_count.rb
gregdingle: Thanks for the explanation. I adjusted to `allow_nil` for all columns in `amr_counts`.

app/models/application_record.rb
```diff
@@ -7,6 +7,18 @@ class ApplicationRecord < ActiveRecord::Base
   after_update { |record| log_analytics record, "updated" }
   after_destroy { |record| log_analytics record, "destroyed" }
 
+  # Condition for rollout of mass addition of validation rules.
+  # Memoizes for performance.
+  def mass_validation_enabled?
+    if AppConfig.table_exists?
```
davidrissato: I would suggest using `AppConfigHelper#get_app_config` in this implementation.

app/models/application_record.rb
```diff
@@ -7,6 +7,18 @@ class ApplicationRecord < ActiveRecord::Base
...
+      if defined?(@enable_mass_validation)
```
davidrissato: Your approach is caching that flag per object. That means if you create thousands of model objects, it will fetch the config table for each of them.

Although it is better than the previous implementation, I would suggest improving performance even further by using rails cache. I know that will force us to redeploy every time we change this value, but I don't think it is an issue given the nature of this flag.

My suggestion:
```
  def mass_validation_enabled?
    Rails.cache.fetch("ApplicationRecord.mass_validation_enabled?") do
      AppConfigHelper.get_app_config(AppConfig::ENABLE_MASS_VALIDATION)
    end
  end
```

gregdingle: Thanks, good suggestion! Updated.

app/models/project.rb
```diff
@@ -11,8 +12,12 @@ class Project < ApplicationRecord
   has_many :phylo_trees, -> { order(created_at: :desc) }, dependent: :restrict_with_exception
   has_one :background
   has_and_belongs_to_many :metadata_fields
+
   validates :name, presence: true, uniqueness: { case_sensitive: false }
-  include ReportHelper
+  # NOTE: not sure why these columns were not created as booleans
+  validates :public_access, inclusion: { in: [0, 1] }, if: :mass_validation_enabled?
+  # NOTE: all values of background_flag in prod db are currently zero.
+  validates :background_flag, inclusion: { in: [0, 1] }, allow_nil: true
```
davidrissato: `, if: :mass_validation_enabled?`

app/models/sample.rb
```diff
@@ -25,11 +25,41 @@ class Sample < ApplicationRecord
...
+  validates :status, presence: true, inclusion: { in: [
+    STATUS_CREATED,
+    STATUS_UPLOADED,
+    STATUS_RERUN,
+    STATUS_RETRY_PR,
+    STATUS_CHECKED,
+  ], }, if: :mass_validation_enabled?
+
+  validate :input_files_checks
+  validates :name, presence: true, uniqueness: { scope: :project_id, case_sensitive: false }
```
davidrissato: Should we use `, if: :mass_validation_enabled?` here as well?

gregdingle: This one actually already existed as-is. I just moved it.

app/models/project.rb
gregdingle: good catch! will change

app/models/pipeline_run.rb
```diff
@@ -121,9 +132,31 @@ class PipelineRun < ApplicationRecord
...
+  # | 4.Experimental-FAILED                     |
+  # | 4.Experimental-FAILED|READY               |
+  # | 4.Experimental-SUCCEEDED                  |
+  # | 4.Experimental-SUCCEEDED|READY            |
+  # | CHECKED                                   |
+  # | FAILED                                    |
+  # | LOADED                                    |
+  # +-------------------------------------------+
+  # NOTE: kickoff_pipeline does not set a job_status
+  validates :job_status, presence: true, allow_nil: true, if: :mass_validation_enabled?
```
davidrissato: I'm a bit confused about this line, how can `presence: true` and `allow_nil: true` be in the same place?

gregdingle: Ha, very good question! In doing this PR, I learned that `presence` will also catch whitespace and empty string. I think this is marginally beneficial in this case.

davidrissato: Update: I'm working on PipelineMonitor right now and it seems that domain list for `job_status` field is incomplete. Just confirming, you are not validating that domain yet, correct?

gregdingle: That's correct. I'm only validating `presence`. Those values are a snapshot for docs only.

PR #3075, opened 2020-02-14
[Data Discovery] Fix samples and project counters out of sync issues

tfrcarvalho:
# Description

* Fixes bugs when counter were not updating after some filters were set
* Enforce a behavior where a location in the map is deselected if the user selects a filter with no results while the location is selected
	- We might re-evaluate this behavior later
* Added extra comments to Discovery Data Layer.

# Tests

1. Navigate to the My Data (or other domain) and projects tab
2. Set a few filters
	* Verify that counters update accordingly for both projects and samples on top nav
	* Verify that counters update accordingly for both projects and samples on sidebar
3. Navigate to samples tabs
4. Change/Add/Remove filters
	* Verify that counters update accordingly for both projects and samples on top nav
	* Verify that counters update accordingly for both projects and samples on sidebar
5. Change to map view
6. Click on a particular location
	* Verify that counters remain the same in the top nav 
	* Verify that counters update accordingly for both projects and samples on sidebar
7. Deselect the location by clicking anywhere on the map
	* Verify that counters remain the same in the top nav 
	* Verify that counters go back to the previous values


app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -468,11 +484,11 @@ class DiscoveryView extends React.Component {
...
-        label: renderTab("Projects", filteredProjectStats.count || "-"),
+        label: renderTab("Projects", filteredProjectCount || "-"),
         value: "projects",
       },
       {
-        label: renderTab("Samples", filteredSampleStats.count || "-"),
+        label: renderTab("Samples", filteredSampleCount || "-"),
         value: "samples",
       },
       domain !== DISCOVERY_DOMAIN_PUBLIC &&
```
j-x-han: Seems like the counter for the Visualizations tab updates pretty inconsistently. Not sure if this is completely relevant to this PR since this is more focused on samples/projects counter, but thought I'd mention it here.

Example flow:
1) Go to All Data view. Visualizations tab should have a count.
image

2) Select some filter on the left sidebar. Visualizations tab now has no count.
image

3) Interact with the the view in some way (select a sample, toggle a sidebar, etc.) Visualizations tab now has a count again?
image

Some changes in the PR actually seem to mitigate some of this behavior, since I checked on master/staging/prod and the Visualizations counter behavior is even weirder there.

j-x-han: Oh looks like QA already reported this IDSEQ-1914 😅

PR #3076, opened 2020-02-14
Surface insert size metrics on samples table, pipeline viz and results folder

morsecodist:
# Description

This change does the following with respect to insert size metrics:
- Adds the necessary changes to the dag json for host filtering to allow idseq-dag to generate insert size metrics files
- Updates the results monitor to check for these files, parse them, and save the metrcis to the database
- Serve these metrics to the front end
- Add these metrics to sample csvs
- Add these metrics to the sample details panel
- Add these metrics to the discovery view
- Make these files available for download in the pipeline viz
- Make these files available in the results folder

Some unrelated changes I made through the course of development:
- Update dockerignore + gitignore to ignore the sql data dump files at the root of the project that are left there from following the development instructions

Pictures:

Data Discovery. Ignore bottom sample. Middle sample is for a human host so it has a Mean Insert Size. Top sample was not for a human host so it does not.

Selection_027

Column Selection View. Mean Insert Size in the column selection menu. 

Selection_020

Tooltip. The mean insert size tooltip.

menu

Pipeline Viz file download. Note the bottom two files, these are new for download in the Star step.

Selection_021

Sample Details Tab. The mean insert size in the sample details tab.

Selection_023

Results Folder. The insert size metrics files for download in the results folder. 

Selection_024

# Tests

Testing this was pretty challenging due to the level of integration with various services. In order to test this I spun up an environment locally and hard-coded the use of my idseq-dag branch (https://github.com/chanzuckerberg/idseq-dag/pull/248). I tested with different hosts and samples to cover the various cases of having files and not having files.


app/models/pipeline_run.rb
```diff
@@ -814,9 +883,15 @@ def check_and_enqueue(output_state)
     if output_ready?(output)
       output_state.update(state: STATUS_LOADING_QUEUED)
       Resque.enqueue(ResultMonitorLoader, id, output)
+    # check if job is done more than a minute ago
     elsif finalized? && pipeline_run_stages.order(:step_number).last.updated_at < 1.minute.ago
-      # check if job is done more than a minute ago
-      output_state.update(state: STATUS_FAILED)
+      # HACK: don't fail for missing insert, size metrics, they are optional
```
cdebourcy: Great PR, but would you mind finding a different mechanism for making insert size loading failures non-blocking? I think you could have a rescue block in `db_load_insert_size_metrics` and in that block you should at the very least send an alert about the failure

app/models/pipeline_run.rb
```diff
@@ -1570,16 +1650,27 @@ def outputs_by_step(can_see_stage1_results = false)
...
       targets.each_with_index do |(target_name, output_list), step_idx|
         next if given_targets.keys.include?(target_name)
-        file_info = []
+
+        file_paths = []
         output_list.each do |output|
-          file_info_for_output = filename_to_info["#{output_dir_s3_key}/#{pipeline_version}/#{output}"]
+          file_paths << "#{output_dir_s3_key}/#{pipeline_version}/#{output}"
+        end
+        if prs.name == PipelineRunStage::HOST_FILTERING_STAGE_NAME && target_name == 'star_out'
```
cdebourcy: This is the wrong place for this logic. The whole point of `output_states` is that they encode the outputs that are supposed to get loaded. Please leave `outputs_by_step` untouched and find a way to work within the boundaries of the design philosophy?

app/models/pipeline_run.rb
```diff
@@ -1197,6 +1272,11 @@ def self.download_file(s3_path, destination, dest_is_dir = true)
   end
 
   def file_generated(s3_path)
+    unless s3_path
```
cdebourcy: Would you mind leaving `file_generated` untouched and instead doing the necessary validations in your calling of it?

app/models/pipeline_run.rb
cdebourcy: Is there a way to extend the framework with a way of capturing these "optional" files rather than hardcoding this stuff ad-hoc based on name of the step/stage?

app/services/retrieve_pipeline_viz_graph_data_service.rb
```diff
@@ -176,14 +176,27 @@ def input_output_to_file_paths
...
-  def file_path_to_outputting_step
+  def file_path_to_outputting_step(file_path_to_info)
     file_path_to_outputting_step = {}
     @all_dag_jsons.each_with_index do |stage_dag_json, stage_index|
       stage_dag_json["steps"].each_with_index do |step, step_index|
         stage_dag_json["targets"][step["out"]].each do |file_name|
           file_path = "#{stage_dag_json['output_dir_s3']}/#{@pipeline_run.pipeline_version}/#{file_name}"
           file_path_to_outputting_step[file_path] = { from: { stageIndex: stage_index, stepIndex: step_index } }
         end
+        # Add insert size metrics files to output if available
```
cdebourcy: This seems to break the abstraction inherent in`file_path_to_outputting_step` by adding specific outputs ad-hoc, is there a way to avoid that with a more elegant design?

db/schema.rb
```diff
@@ -144,13 +144,18 @@
   end
 
   create_table "host_genomes", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
-    t.string "name", null: false
-    t.text "s3_star_index_path"
-    t.text "s3_bowtie2_index_path"
+    t.string "name", null: false, comment: "Friendly name of host genome. May be common name or scientific name of species. Must be unique and start with a capital letter."
+    t.string "s3_star_index_path", default: "s3://idseq-database/host_filter/ercc/2017-09-01-utc-1504224000-unixtime__2017-09-01-utc-1504224000-unixtime/STAR_genome.tar", null: false, comment: "The path to the index file to be used in the pipeline by star for host filtering."
```
cdebourcy: How did these schema changes come about? I believe that hardcoding these S3 paths here is no good. They should not be hardcoded in this repo, they should be a configuration that we can tweak as necessary

app/models/pipeline_run.rb
```diff
@@ -411,6 +424,60 @@ def db_load_ercc_counts
...
+    if tsv_lines.length != 2
+      Rails.logger.error("Pipleine run ##{id} has an insert size metrics file but metrics could not be found")
+      return
+    end
+    insert_size_metrics = {}
+    tsv_lines[0].zip(tsv_lines[1]).each do |row|
+      insert_size_metrics[row[0]] = row[1]
+    end
+
+    extract_int = lambda do |metrics, metric_name|
```
davidrissato: To be more Ruby like, I would suggest not using a lambda here. That makes the code harder to use since you are forced to use `call`.
You don't have any closures from the parent method, so I would just declare a private method in the same class.

However, if I understood it right, you are just trying to prevent errors when metric_name is not present, correct?
In this case, you don't actually need this rescue block at all:

```
   # String to_i returns 0
   "abc".to_i
    # 0

    # nil to_i returns 0
    nil.to_i
    # 0

    # non-existent key returns nil, therefore it will return 0
    a = {"b": 1}
    a["c"].to_i
    # 0
```

Which means this rescue block will never get executed, so the result will never be nil. Eventually it will be 0.

app/models/pipeline_run.rb
```diff
@@ -411,6 +424,60 @@ def db_load_ercc_counts
...
+      begin
+        return metrics[metric_name].to_i
+      rescue
+        return nil
+      end
+    end
+
+    insert_size_median = extract_int.call(insert_size_metrics, MEDIAN_INSERT_SIZE_NAME)
```
davidrissato: Do you need nil here if the key is not present or is zero is acceptable in this case?

`insert_size_median = insert_size_metrics[MEDIAN_INSERT_SIZE_NAME].to_i` would assign zero if the key is missing.

Let me know if you need `nil` instead and I can show you some patterns

app/models/pipeline_run.rb
```diff
@@ -814,9 +883,15 @@ def check_and_enqueue(output_state)
...
       output_state.update(state: STATUS_LOADING_QUEUED)
       Resque.enqueue(ResultMonitorLoader, id, output)
+    # check if job is done more than a minute ago
     elsif finalized? && pipeline_run_stages.order(:step_number).last.updated_at < 1.minute.ago
-      # check if job is done more than a minute ago
-      output_state.update(state: STATUS_FAILED)
+      # HACK: don't fail for missing insert, size metrics, they are optional
+      #   Doing this may result in missing some errors
+      if output_state.output == "insert_size_metrics"
+        output_state.update(state: STATUS_LOADED)
```
davidrissato: I didn't get the idea here. I don't like ignoring errors either. I would like to understand better why would we need to do that. I agree with Charles, we should try finding a different mechanism.

app/models/pipeline_run.rb
tfrcarvalho: There are two possibilities: 
1) configuration: have an extra field name output_fields that tells the client what they should look for.
2) output: each stage outputs a status file (e.g. `samples/<project_id>/<sample_id>/results/3.17/host_filter_status.json`) that contains information about the execution. You can include there the list of `additional_files_to_uload` and the `optional_files_to_upload` that the pipeline step generates.

I believe the second option is the best...
In the end you can include those files here. @cdebourcy do you know of any privacy limitations on files that pipeline steps add to these two arrays?

app/services/retrieve_pipeline_viz_graph_data_service.rb
tfrcarvalho: See my comment above of how to report these additional files to the client. We can make these generic.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20_200_207_200_038) do
+ActiveRecord::Schema.define(version: 20_200_213_210_949) do
```
tfrcarvalho: Seems like schema changes are not from this PR... Although we should update the others to match the state of the database (- limitations pointed by @cdebourcy) we should not include them... means manually reverting the remaining changes...

db/schema.rb
tfrcarvalho: The non-default values are also normally added in a migration... which is also not helpful.

app/models/pipeline_run.rb
```diff
@@ -411,6 +424,60 @@ def db_load_ercc_counts
...
   end
 
+  def db_load_insert_size_metrics
+    insert_size_metrics_s3_path = "#{host_filter_output_s3_path}/#{INSERT_SIZE_METRICS_OUTPUT_NAME}"
+    _stdout, _stderr, status = Open3.capture3("aws", "s3", "ls", insert_size_metrics_s3_path)
+    return unless status.exitstatus.zero?
+    insert_size_metrics_lines = Syscall.pipe_with_output(["aws", "s3", "cp", insert_size_metrics_s3_path, "-"])
+    tsv_lines = []
+    tsv_header_line = -1
+    insert_size_metrics_lines.split(/\r?\n/).each_with_index do |line, index|
```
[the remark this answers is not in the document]
morsecodist: It may be possible to do a double parse but that won't be so clean. The file is a strange format. It is some text with two different csv's embedded. I will use `lines` though to clean it up.

app/models/pipeline_run.rb
morsecodist: Since this is an optional value and for our reference values it is 0 I want to distinguish between no value and 0. For our artificial samples the value actually is 0. If these keys weren't found in the file for whatever reason I don't want it to show up as 0. I will fix my check and use a private method.

app/models/pipeline_run.rb
```diff
@@ -411,6 +424,60 @@ def db_load_ercc_counts
...
+      return
+    end
+    insert_size_metrics = {}
+    tsv_lines[0].zip(tsv_lines[1]).each do |row|
+      insert_size_metrics[row[0]] = row[1]
+    end
+
+    extract_int = lambda do |metrics, metric_name|
+      begin
+        return metrics[metric_name].to_i
```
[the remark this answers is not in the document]
morsecodist: My misunderstanding of ruby, fixing so it returns `nil` if the metric isn't found by checking with `if`

app/models/pipeline_run.rb
morsecodist: I do need `nil` sorry, I explained my my earlier comment before I saw this. I have this so far:

```ruby
  private def extract_int_metric(metrics, metric_name)
    if metrics[metric_name]
      return metrics[metric_name].to_i
    end
    return nil
  end
```

Is that better?

app/assets/src/components/views/discovery/discovery_api.js
```diff
@@ -50,6 +51,19 @@ const getDiscoveryStats = async ({ domain, filters, projectId, search }) => {
 };
 
 const processRawSample = sample => {
+  const meanInsertSize = get(
+    "derived_sample_output.pipeline_run.insert_size_mean",
```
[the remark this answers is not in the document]
morsecodist: I updated the utility function to also do the null/number type checking. It is built for any number + error but it simplifies the checking code in the same way. Do you feel like this is enough?

```Javascript
/**
  @function numberWithError
  @param {Number} x
  @param {Number} error
  @return {String|Null} comma separated numbers with error, null if either is not a number
*/
export function numberWithError(x, error) {
  if (typeof x !== "number" || typeof error !== "number") {
    return null;
  }
  return `${numberWithCommas(x)}±${numberWithCommas(error)}`;
}
```

app/models/pipeline_run.rb
gregdingle: Thanks @tfrcarvalho . This is also relevant to my PR https://github.com/chanzuckerberg/idseq-dag/pull/251 .

morsecodist:
# Update

Thanks to help from @tfrcarvalho I implemented a more principled approach to this problem. In idseq-dag I [added](https://github.com/chanzuckerberg/idseq-dag/pull/248) an optional field to the status file called `optional_outputs`. This will contain a list of files that are expected to have been generated.

I updated this PR to use that field to determine what extra files should be included in the results folder and pipeline viz. It also uses that information to determine if the insert size metrics should have been generated but weren't due to a failure.

Can I please get a re review?

morsecodist:
Note: I did test the failure mechanism

app/models/pipeline_run.rb
```diff
@@ -411,6 +426,64 @@ def db_load_ercc_counts
     update(total_ercc_reads: total_ercc_reads)
   end
 
+  def should_have_insert_size_metrics
+    host_filtering_stage = pipeline_run_stages.find { |prs| prs["step_number"] == 1 }
```
gregdingle: maybe make `host_filtering_stage` a method of the class

db/migrate/20200213210949_add_insert_size_metrics_to_pipeline_runs.rb
```diff
@@ -0,0 +1,12 @@
+class AddInsertSizeMetricsToPipelineRuns < ActiveRecord::Migration[5.1]
+  def change
```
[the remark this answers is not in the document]
gregdingle: Then you can add some rails validators as well to describe the expected data.

gregdingle:
> I added an optional field to the status file called optional_outputs

Like Tiago, I also prefer `extra_outputs`. Why not `additional_outputs` which is consistent with terms in idseq-dag?

happyimadesignr:
Hey! The UI updates look good!

**Few comments:**
* Is the tooltip description a bit long, can we get aways without “computed by Picard CollectionInsertSizeMetrics” and have that in the docs when they click learn more? cc @oliviabholmes @katrinakalantar 
* Todd can you share a screenshot of the tooltip in the Sample Details panel?
* Did we do the other tooltips in that section also? It’ll be strange if we only do one tooltip there.

morsecodist:
@happyimadesignr Thanks!

I am cool with removing “computed by Picard CollectionInsertSizeMetrics”

I don't see tooltips on the Sample Details panel on any of the entries. I just added it to the centralized place where we keep tooltips. I don't think this tooltip shows up in that panel. Do we want the tooltip values to show up there as well for everything? I agree that it would be weird to have only one but not the others. We can enable the tooltips to all of those items as a separate story as well.

happyimadesignr:
Hey @morsecodist!

For the tooltips in the panel, I was referring to this one! Yea we want the tooltips to show there for everything. I'm not sure if there was a ticket for this. https://chanzuckerberg.invisionapp.com/share/NWVSED7EACK#/403080860_1-1_Insert_Metrics

morsecodist:
@happyimadesignr cool. Since this is a global change (show the pre-existing tooltips on every field in the details panel) I would prefer to do this in a separate PR. This one's already huge and that change will effect some other features. Is that OK with you?

db/migrate/20200213210949_add_insert_size_metrics_to_pipeline_runs.rb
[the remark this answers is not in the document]
morsecodist: I implemented this with a separate table (see https://github.com/chanzuckerberg/idseq-web/pull/3076/commits/7f082910bb4e070abcfb15aaf0727346b43ebdc1) and changed it back. In my opinion it made the code way more complex. It introduced nesting and I ended up needing to ensure the values were present in multiple contexts. While the ercc counts have their own table there are more than one per sample, which requires the secondary table. The total ercc count is in the `pipeline_runs` table. Regarding the historical nulls, these are optional fields so I don't think it's an issue that the historical runs have null values in them.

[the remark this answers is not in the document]
gregdingle: Thanks for trying out the different structure. Although I still prefer a separate table, I acknowledge that it is mostly a philosophical question of data modeling. I would rather work with a separate table containing 8 not-null columns than 8 columns in `pipeline_runs` which are non-null when the host is human. See this post for a good summary of all the pros and cons: https://dba.stackexchange.com/questions/103855/is-it-better-to-create-another-table-to-possible-null-values-in-relational-datab . 

With that in mind, I would be happy to defer to someone else as a "tie-breaker" for this design question. Having some consensus in the team would be good if nothing else. 

Sidenote: you should estimate the time to migrate pipeline_runs in prod if we add the cols to that table, or confirm that it is non-blocking DDL. We've had problems with long migrations in the past. 

PS: For encapsulation, another possible design is to put all the insert size columns in a JSON object.

oliviabholmes:
Yes, I think the tooltips change can be done separately

db/migrate/20200213210949_add_insert_size_metrics_to_pipeline_runs.rb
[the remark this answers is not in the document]
tfrcarvalho: Most precedents so far are on adding the extra columns on the `pipeline_runs` tables. 
Not sure if ERCCs are an equivalent example since it is a 1:N relation vs 1:1.
Having said that, I agree that this is a good example to flag. This is not typically the way you add statistics to a database and is just not sustainable to add 8 fields of statistic to a core table like pipeline runs. 
(Regarding JSON, we need to be careful when using it. There is not good support for json in our database, in particular for querying.)
Questions about the data, what makes insert size metrics different? Why are we adding 8 columns vs 1? If multiple statistics are required how do we know we not going to need more in the future (in which case we should store the raw values somewhere? Are we fully sure that this are the relevant statistics and they are not going to be added/edited/remove later on?
I tend to agree with @gregdingle that the best solution seems to be to put them on a separate table. Code could be slightly more complex due to the extra join, but i do not think it would affect much (@morsecodist can you give an example of worst case code complexity?). Also, as we migrate the pipeline, we might factor data out, and a separate table would make those tasks easier.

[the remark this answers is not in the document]
morsecodist: You've convinced me to move back to the multi-table. The code complexity was mostly an issue of ensuring we had the sample details on the front end. 

What do you mean what makes them different? I am saving more values than we explicitly use because I feel they add some additional insight and we may want them. The raw files are stored in s3 so we could always access them again. The nice thing about a separate table is I can make them all required since we either have all or none.

PR #3077, opened 2020-02-14
[Report Page] Send caching metrics to DataDog

j-x-han:
# Description

See IDSEQ-2224. The caching metrics for the new report page were being sent to Google Analytics rather than DataDog.

Switch from using `log_analytics_event` to `put_metric_now` for the new report page's caching metrics so they're sent to DataDog for consistency with the old report page.

# Notes

Updated some comments related to `put_metric_now` and `log_analytics_event`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -169,7 +169,7 @@ class ApplicationController < ActionController::Base
   # given cache_key and fills out custom response headers.
   # If the attempt results in a cache miss, then the response is generated normally and
   # will be stored in the cache.
-  def fetch_from_or_store_in_cache(skip_cache, cache_key, httpdate, _event_name)
+  def fetch_from_or_store_in_cache(skip_cache, cache_key, httpdate)
     if skip_cache
       yield
     else
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -169,11 +169,11 @@ class ApplicationController < ActionController::Base
   # given cache_key and fills out custom response headers.
   # If the attempt results in a cache miss, then the response is generated normally and
   # will be stored in the cache.
-  def fetch_from_or_store_in_cache(skip_cache, cache_key, httpdate, event_name)
+  def fetch_from_or_store_in_cache(skip_cache, cache_key, httpdate, _event_name)
     if skip_cache
       yield
     else
-      MetricUtil.log_analytics_event(event_name + "_cache-requested", current_user) unless skip_cache
+      MetricUtil.put_metric_now("samples.cache.requested", 1) unless skip_cache
       # This allows 304 Not Modified to be returned so that the client can use its
       # local cache and avoid the large download.
       response.headers["Last-Modified"] = httpdate
@@ -183,7 +183,7 @@ class ApplicationController < ActionController::Base
       Rails.logger.info("Requesting #{cache_key}")
 
       Rails.cache.fetch(cache_key, expires_in: 30.days) do
-        MetricUtil.log_analytics_event(event_name + "_cache-missed", current_user)
+        MetricUtil.put_metric_now("samples.cache.miss", 1)
         response.headers["X-IDseq-Cache"] = 'missed'
         yield
       end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -169,11 +169,11 @@ class ApplicationController < ActionController::Base
   # given cache_key and fills out custom response headers.
   # If the attempt results in a cache miss, then the response is generated normally and
   # will be stored in the cache.
-  def fetch_from_or_store_in_cache(skip_cache, cache_key, httpdate)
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -169,11 +169,11 @@ def instrument_with_timer
   # given cache_key and fills out custom response headers.
   # If the attempt results in a cache miss, then the response is generated normally and
   # will be stored in the cache.
-  def fetch_from_or_store_in_cache(skip_cache, cache_key, httpdate, event_name)
+  def fetch_from_or_store_in_cache(skip_cache, cache_key, httpdate, _event_name)
```
lvreynoso: If you're not longer using `event_name`, I think it would be better to just eliminate the parameter. According to my editor this function is only called once, in `samples_controller.rb` so it should be a simple modification.

app/controllers/application_controller.rb
```diff
@@ -169,11 +169,11 @@ def instrument_with_timer
...
     if skip_cache
       yield
     else
-      MetricUtil.log_analytics_event(event_name + "_cache-requested", current_user) unless skip_cache
+      MetricUtil.put_metric_now("samples.cache.requested", 1) unless skip_cache
```
lvreynoso: Is there a meaning behind logging the value `1`?

app/lib/metric_util.rb
```diff
@@ -28,7 +28,7 @@ class MetricUtil
     location_geosearched: "location_geosearched",
   }.freeze
 
-  # DEPRECATED. Use log_analytics_event.
+  # Use for system monitoring (e.g. performance) metrics. Sends to DataDog.
```
lvreynoso: 👍

app/controllers/application_controller.rb
j-x-han: Oops, good catch. That's leftover from the previous call to the other logging function

app/controllers/application_controller.rb
j-x-han: `put_metric_now` expects some `value` as its second parameter. Seems like other calls to the function use it to log extra info such as runtime, number of samples created, etc. I'm not actually sure what the history behind the `1` for caching metrics is, since this is carried over from the logging statement from the older version of the report page (I assume it's some placeholder default value?).

@gregdingle Do you have any more context on this?

cdebourcy: Might it mean "1 occurrence" for downstream aggregation/summation later?

app/controllers/application_controller.rb
lvreynoso: Don't forget to change the call's arguments over in the samples controller as well :)
https://github.com/chanzuckerberg/idseq-web/blob/82c24cb7ff91a9311201a4b44f761c434cc64c4a/app/controllers/samples_controller.rb#L773

app/controllers/application_controller.rb
tfrcarvalho: Actually that is a good question but I believe Charles is right. We could eventually report more samples requested and do a sum on datadog side.

app/controllers/application_controller.rb
```diff
@@ -169,11 +169,11 @@ def instrument_with_timer
   # given cache_key and fills out custom response headers.
   # If the attempt results in a cache miss, then the response is generated normally and
   # will be stored in the cache.
-  def fetch_from_or_store_in_cache(skip_cache, cache_key, httpdate, event_name)
+  def fetch_from_or_store_in_cache(skip_cache, cache_key, httpdate)
```
tfrcarvalho: Wait, why did we get rid of the event name? 
We want this to be a generic function for loading from cache. So, we should still use an even t name.

j-x-han: Got it, added it back in.

One thing to note is that the old report page used `samples` as the event name, whereas the refactored version previously used `PipelineReport`. I switched it back over to `samples` in this last commit for the sake of consistency, but if there's a strong preference for moving to `PipelineReport` as the new event name then I'm happy to do that!

tfrcarvalho: Let's choose what makes more sense long term. 
My opinion is that samples is too generic... that could be a problem in the future if we choose to cache other stuff about samples.	 And that the event should indicate that it is the report info that is being cached. Would also be good to have a standard way of naming these. Maybe the controller + action name? For instance, `samples_report` (we should also respect Datadog's naming constraints: https://docs.datadoghq.com/developers/metrics/#naming-custom-metrics).
Wdyt?

j-x-han: Sounds good to me! I looked through a few of the other event names we have and it seems they typically use . instead of _ to separate controller + action, so I'll be updating to `samples.report`.

tfrcarvalho: They have different meaning in Datadog. They actually a sort of hierarchy for the metrics keys.
Having said that, I agree, seems to make sense to have `controller.action`.

PR #3081, opened 2020-02-18
[Public] Remove more download options and buttons

jshoe:
### Description
- Remove these download endpoints `show_taxid_fasta, taxid_contigs, contigs_fasta_by_byteranges, contigs_sequences_by_byteranges` and HoverAction buttons.
- Other changes bundled as commented.

### Tests
- Tested locally.

### DIFF
--- a/app/assets/src/components/common/Header.jsx
+++ b/app/assets/src/components/common/Header.jsx
@@ -160,10 +160,6 @@ class Header extends React.Component {
             // Initialize the toast container - can be done anywhere (has absolute positioning)
           }
           <ToastContainer />
-          <iframe
-            className={cs.backgroundRefreshFrame}
-            src="/auth0/background_refresh"
-          />
         </div>
       )
     );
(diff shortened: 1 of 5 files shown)


app/models/pipeline_run.rb
```diff
@@ -1580,14 +1580,11 @@ def outputs_by_step(can_see_stage1_results = false)
...
         end
-        if file_info.present?
-          result[prs.name]["steps"][target_name] = {
-            "step_description" => STEP_DESCRIPTIONS[prs.name]["steps"][target_name],
-            # Do not send list of files.
-            # "file_list" => file_info,
-            "reads_after" => (job_stats_by_task[target_name] || {})["reads_after"],
-          }
-        end
+
```
jshoe: This is to enable the "reads remaining" table. We don't need the `file_info` check in this case, so it's safer to leave off the unneeded files.

app/assets/src/components/common/Header.jsx
```diff
@@ -160,10 +160,6 @@ class Header extends React.Component {
             // Initialize the toast container - can be done anywhere (has absolute positioning)
           }
           <ToastContainer />
-          <iframe
```
jshoe: Not sure if this would cause an error, but better to not need the background_refresh

jshoe:
@MarkAZhang thanks, I forgot about those! will add that

PR #3084, opened 2020-02-19
Add tooltips to metadata fields on upload or update pages

gregdingle:
# Description

As title. See for example.

image

image

# Notes

* Note these descriptions were shortened from a subset of the current stored descriptions of `MetadataField` by @oliviabholmes .
* I added tooltip for host organism
* Learn more goes to metadata dictionary page

# Tests

* Go to upload screen
* Hover over column headers
* See tooltip text in position
* Add a new column, hover, see text
* Add a column that has no text, hover, see no text
* Do same as above on update page

* Try learn more link


app/assets/src/components/common/MetadataManualInput.jsx
```diff
@@ -25,12 +25,53 @@ import PropTypes from "~/components/utils/propTypes";
...
+  collected_by: "Institution/agency that collected sample.",
+  isolate: "Whether or not sample is an isolate.",
+  antibiotic_administered: "Antibiotics administered to host.",
+  comorbidity: "Other chronic diseases present.",
+  host_age: "Age of host (in years).",
+  host_genus_species: "Genus or species of host.",
+  host_id: "Unique identifier for host.",
+  host_race_ethnicity: "Race and-or ethnicity of host.",
+  host_sex: "Sex of host.",
+  immunocomp: "Immunocompromised Information on if host was immunocompromised.",
```
jshoe: I'm guessing L57 was just meant to start at "Information" ?

MarkAZhang:
I notice that for some reason the tooltip points down instead of up here. I believe they point up everywhere else, and it'd be great to be consistent.

MarkAZhang:
The reason why the tooltip is pointing down is because the "popper.js" library, which is used by semantic ui, is detecting a scroll clipping region and flipping the tooltip.

You can disable that popper.js logic by adding a prop to this tooltip:
```
// Disable the preventOverflow popper.js modifer as it is flipping the tooltip unnecessarily.
popperModifiers={{
  preventOverflow: {
    enabled: false,
  }
}}
```

app/assets/src/components/common/MetadataManualInput.jsx
gregdingle: Good catch! 😄

gregdingle:
> You can disable that popper.js logic by adding a prop to this tooltip:

Thanks @MarkAZhang!

PR #3086, opened 2020-02-19
Show only "all" sample types for unknown hosts

gregdingle:
# Description

Previously, we treated new or unknown hosts as nonhuman animals. This PR adds a new category "unknown" and shows no "suggestions" in the dropdown for them. 

image

# Tests

* try migration
* try each category of host, see appropriate suggestions and "all"

### DIFF
--- a/db/migrate/20200219181808_add_taxa_category_values_to_host_genome.rb
+++ b/db/migrate/20200219181808_add_taxa_category_values_to_host_genome.rb
@@ -10,5 +10,8 @@ class AddTaxaCategoryValuesToHostGenome < ActiveRecord::Migration[5.1]
 
     # By previous migration, these are all non-human
     HostGenome.where(user: nil, taxa_category: nil).find_each { |u| u.update(taxa_category: "non-human-animal") }
+
+    change_column_default :host_genomes, "taxa_category", from: nil, to: "unknown"
+    change_column_null :host_genomes, "taxa_category", false
   end
 end
--- a/db/migrate/20200219181808_add_taxa_category_values_to_host_genome.rb
+++ b/db/migrate/20200219181808_add_taxa_category_values_to_host_genome.rb
@@ -12,6 +12,5 @@ class AddTaxaCategoryValuesToHostGenome < ActiveRecord::Migration[5.1]
     HostGenome.where(user: nil, taxa_category: nil).find_each { |u| u.update(taxa_category: "non-human-animal") }
 
     change_column_default :host_genomes, "taxa_category", from: nil, to: "unknown"
-    change_column_null :host_genomes, "taxa_category", false
   end
 end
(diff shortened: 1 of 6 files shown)


db/migrate/20200219181808_add_taxa_category_values_to_host_genome.rb
```diff
@@ -0,0 +1,16 @@
+# This adds more values to taxa_category. Initial values were from:
+# https://czi.quip.com/yPgpAPj1dtEC/Hosts-Dropdown-Options
+# See also AddTaxaCategoryToHostGenome.
+class AddTaxaCategoryValuesToHostGenome < ActiveRecord::Migration[5.1]
+  def change
+    unknowns = ['Unknown', 'ERCC only']
+    HostGenome.where(name: unknowns).find_each { |u| u.update(taxa_category: "unknown") }
```
jshoe: It looks like `taxa_category` isn't in the schema.rb. could you add it / any others like `samples_count` ?

gregdingle: Sure! I just ignore schema.rb out of habit but I guess any PR that runs a migration should update it.

db/schema.rb
```diff
@@ -69,6 +69,18 @@
     t.index ["name"], name: "index_backgrounds_on_name", unique: true
   end
 
+  create_table "backgrounds__dropped_cols_backup", id: false, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
```
jshoe: these `__dropped_cols_backup` shouldn't be added, right? Maybe you could reset this and just add the new fields manually

gregdingle: Ah good catch. I just deleted all those in staging and I'll delete them locally. FYI, `_table_name` are ignored by `schema.rb` generation.

db/schema.rb
```diff
@@ -144,13 +144,18 @@
   end
 
   create_table "host_genomes", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
-    t.string "name", null: false
-    t.text "s3_star_index_path"
-    t.text "s3_bowtie2_index_path"
+    t.string "name", null: false, comment: "Friendly name of host genome. May be common name or scientific name of species. Must be unique and start with a capital letter."
+    t.string "s3_star_index_path", default: "s3://idseq-database/host_filter/ercc/2017-09-01-utc-1504224000-unixtime__2017-09-01-utc-1504224000-unixtime/STAR_genome.tar", null: false, comment: "The path to the index file to be used in the pipeline by star for host filtering."
```
jshoe: I'll flag that Charles raised concerns about the exposed S3 paths before (but looks like it was about the same lines)

db/schema.rb
```diff
@@ -565,14 +579,10 @@
   end
 
   create_table "users", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
-    t.string "email", default: "", null: false
+    t.string "email"
     t.string "name"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.string "encrypted_password", default: "", null: false
```
jshoe: Thanks for including these!

PR #3088, opened 2020-02-19
IDSEQ-2273: Change wording of "create a collection" to "create a background"

gregdingle:
# Description

"Collections" were an unrealized generalization of the background concept. For the time being, it's better to use "background".

Before:
image

image

After: 
image

image

# Notes

* I did this along with #3087 
* I searched the codebase for other instances of "collection" and found also `app/assets/src/components/Samples.jsx` which appears to be deprecated
* I noticed the tooltip for "save a background" is the only one that is the "verb object". the rest are simply "object" like "heatmap"
* I updated the background explanation to be a little more accurate
* It seems like it would be good to add some text about who can see a background

# Tests

* try dialog
* see text changes

### DIFF
--- a/app/assets/src/components/views/samples/CollectionModal.jsx
+++ b/app/assets/src/components/views/samples/CollectionModal.jsx
@@ -1,15 +1,16 @@
 import React from "react";
 import PropTypes from "prop-types";
 
+import { createBackground } from "~/api";
+import { withAnalytics } from "~/api/analytics";
 import PrimaryButton from "~ui/controls/buttons/PrimaryButton";
 import SecondaryButton from "~ui/controls/buttons/SecondaryButton";
 import Input from "~ui/controls/Input";
 import Textarea from "~ui/controls/Textarea";
-import { createBackground } from "~/api";
-import { withAnalytics } from "~/api/analytics";
 import Modal from "~ui/containers/Modal";
-
+import ExternalLink from "~/components/ui/controls/ExternalLink";
 import Notification from "~ui/notifications/Notification";
+
 import cs from "./collection_modal.scss";
 
 /**
@@ -183,7 +184,13 @@ class CollectionModal extends React.Component {
               A background is a group of samples. You can use a background as a
               statistical model to compare your samples to. When you select a
               background on a report or heatmap, the z-scores will indicate how
-              much a sample deviates from the mean of that background.
+              much a sample deviates from the mean of that background.{" "}
+              <ExternalLink
+                className={cs.link}
+                href="https://chanzuckerberg.zendesk.com/hc/en-us/articles/360035166174-How-do-I-create-and-use-background-models-in-IDseq-"
+              >
+                Learn More
+              </ExternalLink>.
             </div>
             {this.renderForm()}
             {backgroundCreationResponse && this.renderStatus()}
--- a/app/assets/src/components/views/samples/CollectionModal.jsx
+++ b/app/assets/src/components/views/samples/CollectionModal.jsx
@@ -141,12 +141,12 @@ class CollectionModal extends React.Component {
     return (
       <div>
         {backgroundCreationResponse.status === "ok" ? (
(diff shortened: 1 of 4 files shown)


happyimadesignr:
**Comments and questions:** cc @oliviabholmes 
* I think it should be "Background Model" instead of just “background"
* Can we add a link to the help center doc about background models? Now that we have that!
* In the green confirmation, add “Your"? "Your Background Model is being created..”
* Like you pointed out for the tooltip, why don’t we change "save a background" to "Background Model”
* Did you change that on both tool tips? The is also a tool tip that should say “choose 2 samples” when there are less than 2 samples selected.
* We need to plan to update the wording and screenshots in the Help Center together with these product updates.
* Can you make the green message not touch the buttons?

**More minor things:**
* The words in the descriptions text area need padding
* I know this is not a part of your changes, but the alerts have the wrong style, they shouldn’t have that color bar on the left.
* Also, not a part of your changes, but the "large number of samples" alert seem strange to say that with only 5 samples. Also, the treatment is too prominent.

gregdingle:
Thanks for the review @happyimadesignr . My responses: 

> I think it should be "Background Model" instead of just “background"

Will do. 

> Can we add a link to the help center doc about background models? Now that we have that!

Good idea. Where would the link go?

> In the green confirmation, add “Your"? "Your Background Model is being created..”

Will do. 

> Like you pointed out for the tooltip, why don’t we change "save a background" to "Background Model”

Sounds good to me. Will do. 

> Did you change that on both tool tips? The is also a tool tip that should say “choose 2 samples” when there are less than 2 samples selected.

Yes, I changed that too!

> We need to plan to update the wording and screenshots in the Help Center together with these product updates.

Good point!

> Can you make the green message not touch the buttons?

Yes, I'll add some margin. 

> More minor things:

I agree with all the changes but I don't think this PR is the right time to do them. It's not obvious to me which notification style to use or what the logic should be for the "large number" or what the right amount of padding should be.

gregdingle:
This is the modal after the updates per @happyimadesignr 

image

app/assets/src/components/views/samples/CollectionModal.jsx
```diff
@@ -12,6 +12,10 @@ import Modal from "~ui/containers/Modal";
 import Notification from "~ui/notifications/Notification";
 import cs from "./collection_modal.scss";
 
+/**
+ * NOTE: "Collections" were an unrealized generalization of the background concept.
+ * For the time being, a collection is equivalent to a background.
+ */
 class CollectionModal extends React.Component {
```
lvreynoso: I think we should change the name of the React component while we're at it (and all the other references to a "collection modal"), to minimize confusion.

happyimadesignr:
Maybe this could be a link at the end of the description "Learn about Background Models." That whole sentence in blue?

@oliviabholmes Any thoughts?

oliviabholmes:
I think we can just have Learn More. at the end of the description paragraph and have that link out to https://chanzuckerberg.zendesk.com/hc/en-us/articles/360035166174-How-do-I-create-and-use-background-models-in-IDseq-

oliviabholmes:
Otherwise this looks great to me

app/assets/src/components/views/samples/CollectionModal.jsx
gregdingle: Sorry for the late reply. I'm not sure of the status of the "collections" project or else I would be happy to rename as well. This PR is intended to be a minimal change of wording to avoid confusing users about the nonexistent collections feature. @happyimadesignr is there any chance collections will come back?

gregdingle:
@oliviabholmes @happyimadesignr I added the learn more link. See below. Please approve if good to merge. 

image

gregdingle:
Hi could I get an eng team member to review this? I'm assuming the changes requested by @happyimadesignr and @oliviabholmes are good.

PR #3093, opened 2020-02-20
Fix bug with uncontrolled inputs.

MarkAZhang:
# Description

This fixes a bug caused by a previous bug fix #3084.
The previous fix broke all uncontrolled inputs (you can no longer type in them).

This reverts the previous change, so now inputs can be both controlled and uncontrolled again.
It fixes the original bug in a different way, by making sure the empty-string is passed in by the caller of the Input.

# Tests

* Verified that broken inputs (heatmap metadata) now work.
* Verified that the original bug is still fixed. (by going through the Sample Upload process).

### DIFF
--- a/app/assets/src/components/common/MetadataInput.jsx
+++ b/app/assets/src/components/common/MetadataInput.jsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { isUndefined, isArray } from "lodash/fp";
+import { isUndefined, isNull, isArray } from "lodash/fp";
 import cx from "classnames";
 
 import { UserContext } from "~/components/common/UserContext";
@@ -15,6 +15,14 @@ import AlertIcon from "~ui/icons/AlertIcon";
 import Toggle from "~ui/controls/Toggle";
 
 import cs from "./metadata_input.scss";
+
+// If value is undefined or null, an empty string should be displayed.
+// However, if the MetadataInput is re-used for different samples, and the second sample has no value
+// a particular metadata field, undefined will be passed to the MetadataInput for that field
+// and the first sample's metadata value will contain to be shown.
+// To avoid this, we explicitly pass in the empty string whenever the field is undefined or null.
+const ensureDefinedValue = value =>
+  isUndefined(value) || isNull(value) ? "" : value;
 
 class MetadataInput extends React.Component {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


app/assets/src/components/common/MetadataInput.jsx
```diff
@@ -118,7 +118,9 @@ class MetadataInput extends React.Component {
           className={className}
           onChange={val => onChange(metadataType.key, val)}
           onBlur={() => onSave && onSave(metadataType.key)}
-          value={value}
+          // If value is undefined, an empty string should be displayed.
+          // We need to explicitly pass in the empty string. Otherwise, the previous value will be retained.
+          value={isUndefined(value) ? "" : value}
```
tfrcarvalho: Why do we accept null?

gregdingle: I guess the thinking is that `null` should never occur. Either the `value` prop is defined as a string, or it is undefined.

MarkAZhang: Yeah, what Greg said. For this particular usage, we don't use `null`.

gregdingle: Lol I deleted my comment because I started to have doubts. So thanks for the confirmation.

tfrcarvalho: ok, thanks. Just curious, where in the code is the non-null value guaranteed?

tfrcarvalho: Btw, if you are just worried about undefined, one other option is also to use defaultProps to set value to string.

MarkAZhang:
Hey Greg, that's a really good point.
I took a deeper look at your changes, and I think that PhyloTreeCreation.jsx change might be a bit risky. The reason is that you are now controlling the input based on `this.state.treeName`, but that variable actually gets changed in a couple of places (for example, it gets sanitized). So it's possible for the user to be typing stuff, and for the value of the input to suddenly change to a sanitized version of a past value.

I think it's best to at least revert the PhyloTreeCreation change. I'll do that and will verify that it still works with my change. Let me know if that sounds good.

gregdingle:
> I think it's best to at least revert the PhyloTreeCreation change. I'll do that and will verify that it still works with my change. Let me know if that sounds good.

Sounds good to me.

app/assets/src/components/common/MetadataInput.jsx
MarkAZhang: I've changed the code to check for null as well, and also added a more detailed comment.

MarkAZhang:
@gregdingle, I verified that the phylo tree still works.

PR #3098, opened 2020-02-20
Add ability to select new step function pipeline in sample upload flow

lvreynoso:
# Description

This PR adds an option to mark new samples for processing on the step function pipeline infrastructure. It adds a checkbox to the sample upload flow; when selected, the parameter `pipeline_type` for the sample(s) is set to `SFN`, one of the enum values for `pipeline_type` added in the `Sample` and `PipelineRun` model. 

Screen Shot 2020-02-21 at 1 46 53 PM

# Notes
You can't select both "skip sample processing" *and* "use the step function pipeline", since they're contradictory options.

Adds a new accepted attribute `pipeline_type` to both `Sample` and `PipelineRun`. The attribute is an enum defined as
````ruby
enum pipeline_type: {
    directed_acyclic_graph: "DAG",
    step_function: "SFN",
  }
````
and stored as a string in the db, with default value `DAG`.

# Tests

- frontend tests done manually by going through the sample upload flow and uploading a sample.
- rspec tests added in `spec/request/sample_request_spec.rb`

### CODE REVIEW

File: app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx [hunk 1]
```diff
@@ -40,6 +40,7 @@ class ReviewStep extends React.Component {
     showLessDescription: true,
     showUploadModal: false,
     skipSampleProcessing: false,
+    useNewStepFunctionPipeline: false,
```
tfrcarvalho: nit: remove the new

app/assets/src/components/views/SampleUploadFlow/ReviewStep.jsx
```diff
@@ -351,36 +377,37 @@ class ReviewStep extends React.Component {
...
           {allowedFeatures.includes("host_genome_free_text") && (
             <HostOrganismMessage
-              hostGenomes={this.props.originalHostGenomes}
+              hostGenomes={originalHostGenomes}
               samples={samples}
             />
           )}
           {get("show_skip_processing_option", userSettings) &&
             this.renderSkipSampleProcessingOption()}
+          {admin && this.renderUseNewStepFunctionPipelineOption()}
```
tfrcarvalho: Are you planning to add the flag here?

cdebourcy:
Great work! As you complete this integration, are you also adding a pipeline type flag to `pipeline_runs`? In some cases, we will want to rerun the same sample with both the WDL and the idseq-dag pipeline. 

Conceptually, this is similar to the `pipeline_branch` field, which propagates from `sample` to `pipeline_run`. When we have wanted to rerun a sample with a different branch of idseq-dag, we would edit the `pipeline_branch` field and kick off a new pipeline run. So then the sample has two different sets of results we can toggle between for comparison.

lvreynoso:
Yes, I think both are good ideas I could easily add to this PR. Will do!

(Over the long term I think it would be a good idea to change the whole pipeline kickoff process to be more functional, so we have more flexibility and an easier time controlling the process; I'll explore that on my own time, since that's outside the scope of this epic)

davidrissato:
> Talking with @davidrissato , an idea came out (...) indicates somehow what pipeline was run.
> We can call it `pipeline_type` (open to suggestions for a better name) and start with two possible values `DAG` and `SFN`.
Yes, I think it is way better than "old|new" pipeline. However `pipeline_type` may be too generic, no?
What about something like `execution_strategy`?

app/models/pipeline_run.rb
```diff
@@ -194,6 +194,11 @@ class PipelineRun < ApplicationRecord
 
   delegate :status_url, to: :sample
 
+  enum pipeline_type: {
```
tfrcarvalho: As I read this, I think I agree with @davidrissato that pipeline type might be a bit to generic (e.g. pipeline type seems to mean more like phylo tree vs. alignment pipeline ).
I think he proposed `pipeline_execution_strategy`.
Wdyt?

app/models/sample.rb
```diff
@@ -93,6 +93,11 @@ class Sample < ApplicationRecord
   # setter
   attr_writer :bulk_mode
 
+  enum pipeline_type: {
```
tfrcarvalho: We should not duplicate this definition.... should use the same map as pipeline run.

spec/requests/sample_api_spec.rb
```diff
@@ -0,0 +1,100 @@
+require 'rails_helper'
```
tfrcarvalho: why is this called sample api spec? Is it a request type of test kind of thing?

spec/requests/sample_api_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        @client_params = "web"
+
+        # don't actually do any remote work
+        allow_any_instance_of(Sample).to receive(:set_presigned_url_for_local_upload).and_return(true)
+      end
+
+      describe "when we create a sample via sample upload flow" do
+        it "should properly add the pipeline_type flag, directed_acyclic_graph, to the sample and pipeline run" do
```
tfrcarvalho: Is there a way to split these into four different its?

app/models/pipeline_run.rb
lvreynoso: I wasn't aware that there was a phylo tree pipeline, and I figured "alignment" is a stage. But if it makes it easier for us to understand, sure, since it's only going to be seen by devs anyways

spec/requests/sample_api_spec.rb
lvreynoso: Yeah it's a request spec, I suppose `sample_request_spec` is a better name :)

spec/requests/sample_api_spec.rb
lvreynoso: Yep, can do

app/models/pipeline_run.rb
tfrcarvalho: Naming of variables is always fo devs :)
i used `alignment` as general way to refer to the regular pipeline.

spec/requests/sample_api_spec.rb
tfrcarvalho: I dont think you need it: `sample_spec.rb` is fine because it is already inside the requests folder

tfrcarvalho: Unless there is a conflict...

lvreynoso: `sample_spec.rb` is the name of the Model's spec test file, I don't want to cause any confusion.

app/models/sample.rb
```diff
@@ -726,6 +726,7 @@ def kickoff_pipeline
     pr.subsample = subsample || PipelineRun::DEFAULT_SUBSAMPLING
     pr.max_input_fragments = max_input_fragments || PipelineRun::DEFAULT_MAX_INPUT_FRAGMENTS
     pr.pipeline_branch = pipeline_branch.blank? ? "master" : pipeline_branch
+    pr.pipeline_execution_strategy = pipeline_execution_strategy.blank? ? PipelineRun.directed_acyclic_graph : pipeline_execution_strategy
```
cdebourcy: Great :)

PR #3101, opened 2020-02-21
Remove dead code from pipeline monitor

davidrissato:
# Description

It looks like the original intent was restarting the job when the reason for batch termination was a EC2 instance being terminated (supposedly killed by an evicted spot instance, but could be something else). We already have auto-restarts in a different part of the code, so this is redundant.

The code is marked as "not being used" for more than 2 years, and adds unnecessary complexity. So I'm removing it.

# Tests

* None

### DIFF
--- a/app/models/pipeline_run_stage.rb
+++ b/app/models/pipeline_run_stage.rb
@@ -137,12 +137,6 @@ class PipelineRunStage < ApplicationRecord
     end
   end
 
-  def instance_terminated?(job_hash)
-    job_hash['status'] == STATUS_FAILED &&
-      job_hash['statusReason'].start_with?("Host EC2 (instance") &&
-      job_hash['statusReason'].end_with?(") terminated.")
-  end
-
   def add_failed_job
     existing_failed_jobs = failed_jobs ? "#{failed_jobs}, " : ""
     new_failed_job = "[#{job_id}, #{job_log_id}]"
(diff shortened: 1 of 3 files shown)


app/models/pipeline_run_stage.rb
```diff
@@ -168,37 +162,15 @@ def update_job_status
...
-        _job_status, self.job_log_id, _job_hash, self.job_description = job_info(job_id, id)
+        _job_status, self.job_log_id, self.job_description = job_info(job_id, id)
         save
       end
       return
     end
     # The job appears to be in progress.  Check to make sure it hasn't been killed in AWS.   But not too frequently.
     return unless due_for_aegea_check?
-    self.job_status, self.job_log_id, job_hash, self.job_description = job_info(job_id, id)
-    if [STATUS_ERROR, STATUS_FAILED].include?(job_status)
```
davidrissato: Although it looks strange removing this if condition, it is actually redundant with the `unless` statement below. That condition works almost as a "catch all", and does exactly the same thing as this `if` block.

app/models/pipeline_run_stage.rb
```diff
@@ -168,37 +162,15 @@ def update_job_status
...
-    unless instance_terminated?(job_hash)
-      save
-      return
-    end
-
-    # This code path is not being used anymore because we no longer use spot
-    # instances for the Batch jobs. Spot instances can be terminated at any time
-    # if someone bids more, so a retry is necessary. Now we use on-demand
-    # instances. If we decided to go back to spot instances, this code would
-    # ensure that terminated Batch jobs get resubmitted.
```
jshoe: interesting that we are switching back to spot instances, anyway :)

app/models/pipeline_run_stage.rb
```diff
@@ -168,37 +162,15 @@ def update_job_status
     if failed? || succeeded?
       unless job_log_id
         # set log id if not set
-        _job_status, self.job_log_id, _job_hash, self.job_description = job_info(job_id, id)
+        _job_status, self.job_log_id, self.job_description = job_info(job_id, id)
         save
       end
       return
     end
     # The job appears to be in progress.  Check to make sure it hasn't been killed in AWS.   But not too frequently.
```
tfrcarvalho: This comment makes not sense now, right?

app/models/pipeline_run_stage.rb
```diff
@@ -168,37 +162,15 @@ def update_job_status
...
-
-    # note failed attempt and retry
-    add_failed_job
-    unless count_failed_tries <= MAX_RETRIES
-      LogUtil.log_err_and_airbrake("Job #{job_id} for pipeline run #{id} was killed #{MAX_RETRIES} times.")
-      save
-      return
-    end
-    run_job # this saves
+    self.job_status, self.job_log_id, self.job_description = job_info(job_id, id)
```
tfrcarvalho: How can we be sure that the code below that triggered run_job 3x times would not happen ?
If it did, do we get an alert to restart the job?

app/models/pipeline_run_stage.rb
davidrissato: Actually it does, I haven't changed this logic. Maybe the method could have been originally phrased better, since it does more than just that.

app/models/pipeline_run_stage.rb
davidrissato: That code would only run when the ec2 instance was terminated, and we already have auto-restarts in a different part of our code. So this is redundant.

PR #3104, opened 2020-02-25
Append [Datadog] to alerting logs

j-x-han:
# Description

See IDSEQ-2239.

As a temporary stopgap measure as we move away from using Datadog, relevant alerting log lines will have "[Datadog]" appended to the front, and only those logs will be sent via the Lambda. (The Lambda should probably be modified **after** this PR makes it into prod.)

# Notes

This should just be a temporary solution until the Datadog monitors are completely replaced.

### DIFF
--- a/app/jobs/autoscaling.py
+++ b/app/jobs/autoscaling.py
@@ -74,83 +74,90 @@ def autoscaling_update(config):
           You can provision machines for development jobs by manually setting MinSize on the ASGs,
           which is respected by this autoscaler.
     '''
-    my_environment = config['my_environment']
-    if my_environment == "development":
-        return
+    try:
+        my_environment = config['my_environment']
+        if my_environment == "development":
+            return
 
-    asg_list = get_asg_list()
-    tag_list = get_tag_list()
-    gsnap_ASG = ASG("gsnapl", asg_list, tag_list, config)
-    rapsearch_ASG = ASG("rapsearch2", asg_list, tag_list, config)
-    if not (gsnap_ASG.can_scale and rapsearch_ASG.can_scale):
-        print "Scaling by agents of {my_environment} is not permitted.".format(my_environment=my_environment)
-        return
+        asg_list = get_asg_list()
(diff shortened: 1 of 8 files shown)


app/jobs/autoscaling.py
```diff
@@ -74,83 +74,90 @@ def autoscaling_update(config):
...
+                print "No instances are ready to move from 'draining' to 'discarded' state."
+
+        if 'batch_configurations' in config:
+            print "Autoscaling batches:"
+            response_batch_autoscaling = batch_autoscaling.autoscale_compute_environments(config['batch_configurations'])
+            print json.dumps(response_batch_autoscaling, default=_default_json_serializer)
+    except Exception as error:
+        # If any error occurs while performing an autoscaling update, append [Datadog] to ensure that the log is
+        # forwarded to Datadog.
+        # TODO(julie): Note that this is part of a temporary stopgap measure as we move away from Datadog and should be
```
jshoe: could you date this comment and also include the JIRA number?

jshoe: could you also `raise error` again at the end of this? in case someone ever adds more handling around this

jshoe:
Thanks for taking this on and also volunteering to help follow through to remove DD logs entirely.

jshoe:
There's also one more in idseq-dag for `AmrAlleleMismatchError` (rare) that you can change! Or maybe @MarkAZhang can comment on if that one is still necessary.

app/jobs/autoscaling.py
j-x-han: Done!

MarkAZhang:
If it's not too much trouble to keep it, I think it would be good. While it shouldn't get triggered under normal circumstances, if we were to ever update our AMR database, it might catch an error or two. I can still see it having use in the future.

j-x-han:
I'll submit a PR to idseq-dag to update the `AmrAlleleMistmatchError` as well 😄

PR #3105, opened 2020-02-25
[Public] Add Basic Auth for preview and testing purposes

jshoe:
### Description
- We agreed to add Basic Auth in the interim (as opposed to no auth whatsoever) to facilitate previewing the site. We can take this away later.

### Tests
- You can test in the browser normally at `http://localhost:3000/` or with a command like `curl -v http://user:password@localhost:3000`

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -177,6 +177,7 @@ Rails.application.routes.draw do
     get :sample_locations, on: :collection
   end
 
+  # Replaced Warden 'authenticate' with BasicAuth in ApplicationController.
   # authenticate :auth0_user, ->(u) { u.admin? } do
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -177,9 +177,9 @@
     get :sample_locations, on: :collection
   end
 
-  authenticate :auth0_user, ->(u) { u.admin? } do
-    mount RESQUE_SERVER, at: "/resque"
-  end
+  # authenticate :auth0_user, ->(u) { u.admin? } do
```
MarkAZhang: Nit: Could you add a comment here as well? To explain why this is commented out.

jshoe: 👍

PR #3108, opened 2020-02-26
Migrate from Dockerhub to Amazon ECR.

MarkAZhang:
# Description

* Change the travis-ci build/push scripts and deployment scripts to use ECR for the docker registry.

# Notes
* Created new idseq-web ECR repo.
* Also needed to add appropriate permissions to travis-ci IAM role. These permissions are scoped to only the newly created idseq-web ECR repo.
* AWS authentication on Travis CI relies on Travis CI repository variables already set on the idseq-web repo in Travis CI. The only variable we needed to add was the aws account id.

# Tests

* Verified that Travis CI builds successfully. Verified that Travis CI properly uses cached image on subsequent builds (and speeds up from 8m to 3m).
* Verified that deployment to sandbox environment succeeds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,13 +82,14 @@ jobs:
         - ./node_modules/.bin/eslint app/assets/src --ext .js,.jsx
         - npm test
     - name: "Docker Build"
-      language: python
+      language: shell
       services:
         - docker
       cache: pip
       before_script:
         # Need to install the AWS command line to authenticate to ECR.
-        - pip install awscli
+        # Need to use user to avoid permission error on Travis CI when using shell language.
+        - pip install --user awscli
         # Need to set the region to do ECR authentication.
         - aws configure set default.region ${AWS_REGION}
         - $(aws ecr get-login --no-include-email)
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
 addons:
   # Emulate docker compose routing
   #   Docker compose maps the host db and redis to the mysql and redis containers
-  #   We are using travis-ci services, which use localhost, for testing instead of docker compose 
+  #   We are using travis-ci services, which use localhost, for testing instead of docker compose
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -82,9 +82,16 @@ jobs:
         - ./node_modules/.bin/eslint app/assets/src --ext .js,.jsx
         - npm test
     - name: "Docker Build"
-      language: shell
+      language: python
```
jshoe: was this supposed to change?

bin/push-docker
```diff
@@ -11,25 +11,6 @@ echo $REPO
...
-        echo "Error: Exceeded retries with Docker."
-        exit 1
-    fi
-    count=$((count+1))
-    sleep 15
-done
-
 # Retry up to 5 times total if failed Docker push
 run=1
 count=1
```
jshoe: While you're at it, could change these comments below to say something like "Trying ECR push"/"Exceeded retries with ECR"

bin/deploy
```diff
@@ -63,6 +63,7 @@ end
 
 def post_slack_message(payload, slack_hook)
   puts `curl -X POST --data-urlencode payload=#{Shellwords.escape(JSON.dump(payload))} #{slack_hook}`
+
```
jshoe: nit: extra blank line

.travis.yml
MarkAZhang: I had changed it thinking we needed python for "pip install". It turns out that travis CI includes python with the "minimal/shell" build anyways. I will change it back and see what happens.

bin/push-docker
MarkAZhang: Good catch!

MarkAZhang:
Also fixed an issue brought up by Tiago where I was storing `AWS_ACCOUNT_ID` in an environment variable in the Terminal and not removing it after the script was run.

Changed to using backticks, which allows you to get the output of a shell command in Ruby.

PR #3113, opened 2020-02-27
Refactor dag json generation and stage batch job trigger

tfrcarvalho:
# Description

* Split json generation from pipeline dispatch, so that we can use the json for reusability by the SFN pipeline dispatch
* Split upload json s3 from generate batch command for reusability

# Tests

QA: Test by submitting samples through the pipeline and verify that they process correctly.

Thanks to @lvreynoso:
Tested by running 4 different samples through both `master` and `tiago/pipeline/dispatch`. The sample dags were collected and compared using `diff -r`. The output is viewable at https://gist.github.com/lvreynoso/e444dc5ba1f1313e3ec591e9aaff28b3 and the only differences are due to user input or expected (e.g. different sample ids). For the "echo" sample I accidentally uploaded fastqs for one and fastas for the other, which explains the differences visible there.

The generated JSON is the same.


tfrcarvalho:
> Tested by running 4 different samples through both `master` and `tiago/pipeline/dispatch`. The sample dags were collected and compared using `diff -r`. The output is viewable at https://gist.github.com/lvreynoso/e444dc5ba1f1313e3ec591e9aaff28b3 and the only differences are due to user input or expected (e.g. different sample ids). For the "echo" sample I accidentally uploaded fastqs for one and fastas for the other, which explains the differences visible there.
> 
> The generated JSON is the same. 👍

Thank you so much @lvreynoso !

PR #3116, opened 2020-02-28
IDSEQ-2278 - Update PipelineRunStage job_status for step function executions

davidrissato:
# Description

Update `PipelineRunStage#job_status` from step function execution history for PipelinesRuns using `pipeline_execution_strategy = "step_function"`

# Notes
IDSEQ-2278 / IDSEQ-2257

# Tests

* Unit tests only

### DIFF
--- a/app/models/pipeline_run_stage.rb
+++ b/app/models/pipeline_run_stage.rb
@@ -159,6 +159,18 @@ class PipelineRunStage < ApplicationRecord
   end
 
   def update_job_status
+    if pipeline_run.step_function?
+      # this logic will be replaced soon by step functions async notifications (IDSEQ-2310)
+
+      if !id || !pipeline_run.sfn_execution_arn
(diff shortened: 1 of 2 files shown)


app/models/pipeline_run_stage.rb
```diff
@@ -159,6 +159,18 @@ def due_for_aegea_check?
   end
 
   def update_job_status
+    if pipeline_run.step_function?
+      # this logic will be replaced soon by step functions async notifications (IDSEQ-2310)
+
+      if !id || !pipeline_run.sfn_execution_arn
```
tfrcarvalho: I see it was there before but what does `!id` do? Does this mean we call this function even before creating/saving the PipelineRunStage?

davidrissato: No idea. That is my understanding as well, the only way it could happen would be if you invoked this method before saving. I haven't invested time to understand if it would even be possible, since I'm remove this entire block in the near future, so I just copied it.

PR #3117, opened 2020-02-28
[SFN Pipeline] Dispatch new pipeline 

tfrcarvalho:
### Description

* This PR dispatches the new SFN pipeline
* It tries to minimize changes to the current workflow. Because of that some less obvious decisions were made:
  - For the DAG pipeline would initialize each stage For SFN, it starts SFN execution when the first stage starts
  - It saves the SFN arn in job_command field of pipeline_run_stage for compatibility of the rest of the code
* Copies `idd2wdl` script from `chanzuckerberg/idseq-web` for converting dag json to wdl sfl pipeline configuration
* Adds utility function to s3 for uploading a string

### Notes

* The pipeline currently fails on insert metrics step. Probably will need to adapt `idd2wdl` script in future PRs.

### Tests

* Check that a regular user can upload a sample and it processes correctly (regression test) 
* Unit tests


app/services/sfn_pipeline_service.rb
```diff
@@ -0,0 +1,133 @@
...
+    # Not using SDK because aws-sdk-sfn is currently on version 1.0.0.rc4,
+    # which in terms requires aws-sdk-core (= 3.0.0.rc1)
+    #
+    # sfn_client = Aws::SFN::Client.new
+    # resp = sfn_client.start_execution({
+    #   state_machine_arn: AppConfigHelper.get_app_config(AppConfig::SFN_ARN),
+    #   name: "idseq-#{Time.now.to_i}",
+    #   input: JSON.dump(sfn_input_json)
+    # })
+    # return resp[:execution_arn]
```
lvreynoso: Why not add `aws-sdk-core` and use the SDK? I think that's more desirable over running a shell command.

app/models/pipeline_run_stage.rb
```diff
@@ -118,8 +122,17 @@ def redacted_dag_json
   def run_job
     # Check output for the run and decide if we should run this stage
     return if started? && !failed? # job has been started successfully
-    self.job_command = send(job_command_func)
-    self.command_stdout, self.command_stderr, status = Open3.capture3(job_command)
+    if step_function?
```
lvreynoso: The `step_function` attribute is part of the `PipelineRun` though, not the `PipelineRunStage`. I don't think this will work.

app/services/sfn_pipeline_service.rb
```diff
@@ -0,0 +1,133 @@
...
+    return {
+      wdl_input_s3_path: "s3://#{SAMPLES_BUCKET_NAME}/#{wdl_input_s3_key}",
+      wdl_output_s3_path: "s3://#{SAMPLES_BUCKET_NAME}/#{wdl_output_s3_key}",
+    }
+  end
+
+  def dispatch(sfn_input_json)
+    sfn_name = "idseq-#{Time.now.to_i}"
+    sfn_input = JSON.dump(sfn_input_json)
+    sfn_arn = AppConfigHelper.get_app_config(AppConfig::SFN_ARN)
```
lvreynoso: We should be getting this from the `sfn_execution_arn` attribute of the `PipelineRun`, no?

app/models/pipeline_run_stage.rb
tfrcarvalho: Indeed! :)

app/services/sfn_pipeline_service.rb
tfrcarvalho: Because it is a release candidate, **because it pushes the version we have currently down** and will impact also other code. 
I'll add that to the comment...

tfrcarvalho: Also, unless there is no other resort, i tend to not install rc versions.

lvreynoso: I see v3.4.8 in Gemfile.lock, it won't work with any higher version? :(

tfrcarvalho: No,  the requirements are strictly what I put in the comment: `= 3.0.0.rc1`.
We should definitely revisit this but is not something I want to invest that much time right now.

app/services/sfn_pipeline_service.rb
```diff
@@ -0,0 +1,133 @@
+class SfnPipelineService
+  # This service is responsible for dispatching a request to the
+  # StepFunctions-based pipeline.
+  # It generates jsons for all the pipeline run stages, converts them to WDL,
+  # creates the Step Function's input JSON and start SFN execution
```
lvreynoso: Would you mind changing the name to `SfnPipelineDispatchService` or something similar? I can imagine creating more services in the future that interact with the sfn pipeline - I might have to for the Pipeline Viz

tfrcarvalho: Done.

kislyuk:
Thanks @tfrcarvalho, could you update the idd2wdl.py script with the version from https://github.com/chanzuckerberg/idseq/blob/akislyuk-forward-error-metadata/workflows/idd2wdl.py?

app/jobs/idd2wdl.py
```diff
@@ -0,0 +1,143 @@
...
+    if name.startswith("PipelineStep"):
+        name = name[len("PipelineStep"):]
+    return name
+
+
+def target_filename(target, index):
+    if target == "fastqs":
+        return f"{target}_{index}"
+    else:
+        return f"{target}_{file_path_to_name(dag['targets'][target][index])}"
```
lvreynoso: So it appears that the Docker images we use have Debian 9.0 _stretch_, and the latest python3 version for that distro is `3.5.3`. Unfortunately f-strings require python3 v`3.6` or greater. We have to either rewrite this script without 3.6+ features or upgrade our Docker image to use the latest stable Debian 10.0 _buster_, which has python3 v`3.7`. Otherwise, as currently built trying to kickoff the sfn pipeline fails with

```
SfnPipelineDispatchService::Idd2WdlError (Command to convert dag to wdl failed ('idd2wdl.py'). Error:   File "app/jobs/idd2wdl.py", line 43)
    return f"{target}_{index}"
                             ^
SyntaxError: invalid syntax
```

As an aside, this is a good reason why running Docker locally is valuable for development, since we can reproduce the production environment and catch things like this. :)

tfrcarvalho: Good catch. Thanks! 
Point taken. Indeed, we are trading better performance for not fully recreating the prod environment. Nothing replaces the exact docker configuration that will be used in prod.

I still prefer to be on the side of performance for most vs. the occasional error. And there are other ways to provide "similar" guarantees. We can definitely improve our CI configuration, make sure we freeze versions when installing any packages and now that @kislyuk added a script to run it locally we can keep improving it to make sure it is synced with docker config.

kislyuk: @lvreynoso unlike Docker on Linux, Docker on MacOS has performance issues that have their own negative impacts on development productivity. The reason is that Docker on MacOS uses full-blown virtualization, while Docker on Linux uses lightweight pseudo-virtualization that is much less resource intensive.

I agree that we should test our code in Docker to accurately represent production conditions. I think this should be done in a CI/CD pipeline that runs on GitHub Actions or in another CI environment, not on our laptops.

kislyuk: I would add here that I believe Ubuntu LTS (not Debian) should be our preferred OS baseline. However, for this specific issue, I'm happy to rewrite the script to avoid the use of f-strings and be otherwise python3.5-compatible. I'll push a branch and link here when done.

kislyuk: The Python 3.5 compatible copy of idd2wdl.py is now ready at https://github.com/chanzuckerberg/idseq/blob/akislyuk-idd2wdl-py35-compat/workflows/idd2wdl.py. It's definitely less readable and maintainable, but that shouldn't be a problem in the long term since our plan is to move away from generating DAG JSON and converting it to WDL, and toward using static WDL-defined workflows directly (which will remove the need for this script).

lvreynoso: I'm also a fan of Ubuntu! I did not know about the differences in Docker performance between Linux and MacOS, thanks for the information.

tfrcarvalho: @kislyuk what the factors in favor of Ubuntu LTS? Better support?

app/jobs/idd2wdl.py
```diff
@@ -0,0 +1,168 @@
...
+task {task_name} {{
+  runtime {{
+    docker: "{AWS_ACCOUNT_ID}.dkr.ecr.us-west-2.amazonaws.com/idseq-workflows:{STAGE}"
+  }}
+  input {{
+{wdl_step_input}
+  }}
+  command<<<
+  python3 <<CODE
+  import os, sys, json, contextlib, importlib, threading, logging, subprocess, traceback
```
davidrissato: I have a suggestion for this code: could we create a static class in idseq_dag with all this logic below, and instead of adding this huge block here, just a static method sending a JSON entry. That would not only make the code more readable, but also would prevent potential command injection.

Ex:
- Create a `idseq-dag/wdl_task_wrapper.py`
```python3
def wdl_execute_task(inputJson):
   # move all the logic to this block
```

And here in the wdl template, just use:
```python3
task {task_name} {{
    ...
    command<<<
       python3 <<CODE
            from idseq_dag.wdl_task_wrapper import wdl_wrap_task
            wdl_execute_task({input_json})
    ...
    }}""".format(
        JSON.dumps({
             task_name=task_name(step),
             ...
        })
```

tfrcarvalho: For me, just depends how long this file is going to be here. @kislyuk:  wdyt ?

app/jobs/idd2wdl.py
kislyuk: Yes - Ubuntu LTS is better supported than Debian, has access to a wider variety of packages, and is used more widely in bioinformatics (making it more likely that we can standardize on it across all our environments).

app/jobs/idd2wdl.py
kislyuk: The transition plan after we ship the initial version of this code is to stop using jbuilder templates and start using static WDL files representing each of the stages. In other words, this code (idd2wdl.py) will be removed from idseq-web as soon as we can get rid of the jbuilder templates. Any parameters that parameterize the jbuilder templates will be moved into the static WDL interface.

The suggestion to keep calling idseq-dag to translate the dynamically generated DAG (using facilities internal to idseq-dag) breaks this interface migration, and takes us further away from a clean separation between the front-end and the pipeline. So I think we should stick with the transition plan as we have defined it so far.

kislyuk: Actually, reading further into David's proposal, I can see that it concerns just the python code in the WDL command block. I initially misread it as concerning the whole idd2wdl.py layer.

I think it would be reasonable to put this code in a `idseq-dag/wdl_task_wrapper.py` file instead of autogenerating it here. I can go ahead and implement that. I'll post here when I get a PR merged in idseq-dag, release it, and rebuild a container image for the sfn-wdl pipeline with the new idseq-dag version.

kislyuk: Because the task of moving code to idseq-dag will necessitate additional testing and validation, I would like to postpone it and break it out into a separate sub-task, without blocking this PR on it.

kislyuk: As an aside, I consider the command injection concern to be unwarranted, since the inputs are extracted from DAG JSON.

kislyuk: Filed https://jira.czi.team/browse/IDSEQ-2390 to track the additional task.

PR #3120, opened 2020-03-02
Always make additional output available for download

morsecodist:
# Description

Makes additional output for download in the results folder.

Opposite of https://github.com/chanzuckerberg/idseq-web/pull/3121


morsecodist:
Going with the opposite approach for now

PR #3121, opened 2020-03-02
Always hide additional output from host filtering

morsecodist:
# Description

Prevents downloading additional outputs in the pipeline viz.

Opposite of https://github.com/chanzuckerberg/idseq-web/pull/3120.

# Testing

I spun up my local environment with the pipeline runs from my previous tests. I hard-coded hiding host filtering files by setting that modifier to true (so it will treat me as if I don't own the samples). I checked all of the cases (errored case, having files, not having files, steps with additional outputs, normal steps). I checked in both the pipeline viz and the results folder.


morsecodist:
Pending testing

morsecodist:
replaced with https://github.com/chanzuckerberg/idseq-web/pull/3122

PR #3122, opened 2020-03-02
[Release fix]  Always hide additional output from host filtering

morsecodist:
# Description

Prevents downloading additional outputs in the pipeline viz.

# Testing

I spun up my local environment with the pipeline runs from my previous tests. I hard-coded hiding host filtering files by setting that modifier to true (so it will treat me as if I don't own the samples). I checked all of the cases (errored case, having files, not having files, steps with additional outputs, normal steps). I checked in both the pipeline viz and the results folder.


app/services/retrieve_pipeline_viz_graph_data_service.rb
```diff
@@ -241,7 +241,10 @@ def modify_step_name(step_name)
 
   def remove_host_filtering_urls(edges)
```
gregdingle: could you add a comment why it is important here to remove host filtering urls? is it for privacy? 

and would those files then be a good candidate for `additional_output_files_hidden` ?

morsecodist: Yes. This is because the data may contain PGI. 

No, I know this is confusing but those files are hidden in all contexts. These files are available to the pipeline runner, and their existence is visible to everyone. We probably could expose them but we wanted to err on the side of caution when it comes to making any files available from this stage. Will explain in a comment.

PR #3125, opened 2020-03-02
Update default alignment config date for local dev

jshoe:
### Description
- Update hardcoded '2020-02-03' reference that is used in local dev.

### Notes
- Thanks Lucia for reminding me to update this.
- The docker-compose.yml default gets overwritten as long as the env variables are loaded from Parameter Store, but in local development that's not usually the case. I don't see an obvious way to change this besides always loading from Parameter Store (?).
- See: https://github.com/chanzuckerberg/idseq-web/pull/3060

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 # Shared variables using Docker x- extension and YAML merging syntax.
 
 x-web-variables: &web-variables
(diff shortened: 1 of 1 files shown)


lvreynoso:
Should we also update the rake tasks `create_alignment_config` and `update_lineagedb` to use `ENV["ALIGNMENT_CONFIG_DEFAULT_NAME"]` instead of `ENV["NCBI_DATE"]` as well?

PR #3127, opened 2020-03-02
[Release fix] Fix bulk downloads

morsecodist:
# Description

Fixes bulk downloading of insert size metrics

# Tests
Locally tested downloading summary csv

### DIFF
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -24,7 +24,8 @@ module SamplesHelper
                     total_reads nonhost_reads nonhost_reads_percent total_ercc_reads subsampled_fraction
                     quality_control compression_ratio reads_after_star reads_after_trimmomatic reads_after_priceseq reads_after_cdhitdup
                     sample_type nucleotide_type collection_location
-                    host_genome notes]
+                    host_genome notes
+                    insert_size_median insert_size_mode insert_size_median_absolute_deviation insert_size_min insert_size_max insert_size_mean insert_size_standard_deviation insert_size_read_pairs]
     CSVSafe.generate(headers: true) do |csv|
       csv << attributes
       formatted_samples.each do |sample_info|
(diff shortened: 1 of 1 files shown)


app/helpers/samples_helper.rb
```diff
@@ -60,14 +61,14 @@ def generate_sample_list_csv(formatted_samples)
...
-                        insert_size_median: pipeline_run ? pipeline_run.dig(:insert_size_metric_set, :median) : '',
-                        insert_size_mode: pipeline_run ? pipeline_run.dig(:insert_size_metric_set, :mode) : '',
-                        insert_size_median_absolute_deviation: pipeline_run ? pipeline_run.dig(:insert_size_metric_set, :median_absolute_deviation) : '',
-                        insert_size_min: pipeline_run ? pipeline_run.dig(:insert_size_metric_set, :min) : '',
-                        insert_size_max: pipeline_run ? pipeline_run.dig(:insert_size_metric_set, :max) : '',
-                        insert_size_mean: pipeline_run ? pipeline_run.dig(:insert_size_metric_set, :mean) : '',
-                        insert_size_standard_deviation: pipeline_run ? pipeline_run.dig(:insert_size_metric_set, :standard_deviation) : '',
-                        insert_size_read_pairs: pipeline_run ? pipeline_run.dig(:insert_size_metric_set, :read_pairs) : '', }
+                        insert_size_median: pipeline_run && pipeline_run.insert_size_metric_set && pipeline_run.insert_size_metric_set.median ? pipeline_run.insert_size_metric_set.median : '',
```
gregdingle: I think the `dig` method would shorten these lines but no worries if re-testing is annoying

See https://tiagoamaro.com.br/2016/08/27/ruby-2-3-dig/.

morsecodist: The dig method caused the issue, was I using it incorrectly? I think it is only for hashes and not class members.

gregdingle: nevermind! I didn't even read the red lines.

PR #3136, opened 2020-03-03
Clarify ALIGNMENT_CONFIG_DEFAULT_NAME

morsecodist:
# Description

Removes the setting of `ALIGNMENT_CONFIG_DEFAULT_NAME` from `docker-compose.yml` as it will always be overridden by chamber.

Added a comment above it's access to specify that this configuration is loaded from ssm/chamber.

# Tests

This effects local only.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,6 @@
 # Shared variables using Docker x- extension and YAML merging syntax.
 
 x-web-variables: &web-variables
-  ? ALIGNMENT_CONFIG_DEFAULT_NAME=2020-02-10
   ? SAMPLES_BUCKET_NAME=idseq-samples-development
   ? ES_ADDRESS=http://elasticsearch:9200
   ? AIRBRAKE_PROJECT_ID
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,7 +1,6 @@
 # Shared variables using Docker x- extension and YAML merging syntax.
 
 x-web-variables: &web-variables
-  ? ALIGNMENT_CONFIG_DEFAULT_NAME=2020-02-10
```
jshoe: I think you actually still need the `? ALIGNMENT_CONFIG_DEFAULT_NAME` part in order for the variable to get into the container

morsecodist: You need it to import the environment variable from the outside environment but this one is loaded in the image itself. The command calls `chamber exec` so they are loaded by the image at startup. See:

`Dockerfile:61`

```Dockerfile
ENTRYPOINT ["bin/entrypoint.sh"]
```

`bin/entrypoint:7`

```bash
if [ "$OFFLINE" = "1" ]
then
    exec bundle exec "$@"
else
    # Use Chamber to inject secrets via environment variables.
    exec chamber exec idseq-$ENVIRONMENT-web -- bundle exec "$@"
fi
```

jshoe: Oh I see yeah

PR #3145, opened 2020-03-04
[Public] Update site links + clarify the use of 'index'

jshoe:
### Description
- Use the referrer link Greg recommended and update to include that (all IDseq controlled pages).
- Change cases of 'index' to not reference to the NCBI index, to avoid confusion.

### Testing
- Tested locally.


jshoe:
That's a great point.. "index" only really makes sense to us as a technical detail, actually..

PR #3147, opened 2020-03-05
Add banner on landing page linking to idseq public site

lvreynoso:
# Description

Adds a banner at the top of the page linking to https://public.idseq.net/ to redirect people who may be looking for our response to the nCov-2019 virus.

Screen Shot 2020-03-05 at 10 07 30 AM


jshoe:
I'll point out that the current video bulletin is still active, so it'll look a little different

app/assets/src/components/views/Landing.jsx
```diff
@@ -125,6 +131,39 @@ class Landing extends React.Component {
...
+                  Learn how researchers in Cambodia used IDseq to sequence
+                  SARS-CoV-2
+                </ExternalLink>
+              </span>
+            }
+          />
+        </div>
+      );
+    } else {
+      publicSiteBanner = <Divider />;
```
jshoe: The `else` Divider isn't really needed, right?

app/controllers/home_controller.rb
```diff
@@ -25,6 +25,19 @@ def landing
           @show_bulletin = true
         end
       end
+
+      @show_public_site = false
+      if get_app_config(AppConfig::SHOW_LANDING_PUBLIC_SITE_BANNER) == "1"
+        @show_public_site = true
+      else
+        time_zone = ActiveSupport::TimeZone.new("Pacific Time (US & Canada)")
```
jshoe: I don't think you really need the time check for this one ;)

lvreynoso:
Going to do this as a "release fix" instead

PR #3148, opened 2020-03-05
IDSEQ-1398: Ensure required metadata fields (and host defaults) are associated with all host genomes

gregdingle:
# Description

In IDSEQ-1398, a user was unable to upload because their host (bat) did not have required field `collection_location_v2`. This adds model validation and a callback to ensure this can't happen again.

In addition, I will run the following script in prod to update existing hosts.

```
  MetadataField.where(is_required: 1).each do |mf|  
    mf.update_host_genomes!
    mf.save!
  end

  MetadataField.where(default_for_new_host_genome: 1).each do |mf|  
    mf.update_host_genomes!
    mf.save!
  end
```

# Notes

* This also raises the question of whether we should update existing projects on adding a required field

# Tests

* try creating a required field with no default_for_new_host_genome
* try updating to a required field with no default_for_new_host_genome
* try creating a new field that is required, see hosts update

### DIFF
--- a/app/models/host_genome.rb
+++ b/app/models/host_genome.rb
@@ -46,9 +46,9 @@ class HostGenome < ApplicationRecord
     HostGenome::ERCC_PATH_PREFIX + HostGenome::S3_BOWTIE2_INDEX_FILE
   end
 
-  # This is designed to be called only during migrations and console sessions.
   def self.all_without_metadata_field(name)
-    joins(:metadata_fields).where.not(metadata_fields: { name: [name] }).distinct
+    exists = joins(:metadata_fields).where(metadata_fields: { name: [name] })
+    where.not(id: [exists.pluck(:id).uniq])
   end
 
   def as_json(options = {})
--- a/app/models/host_genome.rb
+++ b/app/models/host_genome.rb
@@ -47,8 +47,8 @@ class HostGenome < ApplicationRecord
   end
 
   def self.all_without_metadata_field(name)
-    exists = joins(:metadata_fields).where(metadata_fields: { name: [name] })
-    where.not(id: [exists.pluck(:id).uniq])
+    exists = joins(:metadata_fields).where(metadata_fields: { name: [name] }).distinct
+    where.not(id: exists)
   end
 
   def as_json(options = {})
--- a/app/models/host_genome.rb
+++ b/app/models/host_genome.rb
@@ -46,6 +46,11 @@ class HostGenome < ApplicationRecord
     HostGenome::ERCC_PATH_PREFIX + HostGenome::S3_BOWTIE2_INDEX_FILE
   end
 
+  # This is designed to be called only during migrations and console sessions.
+  def self.all_without_metadata_field(name)
+    joins(:metadata_fields).where.not(metadata_fields: { name: [name] }).distinct
+  end
+
   def as_json(options = {})
     hash = super(options)
     hash[:ercc_only] = ercc_only?
--- a/app/models/metadata_field.rb
+++ b/app/models/metadata_field.rb
@@ -88,7 +88,7 @@ class MetadataField < ApplicationRecord
   validates :name, presence: true, uniqueness: true, if: :mass_validation_enabled?
   validates :display_name, presence: true, uniqueness: true, if: :mass_validation_enabled?
 
-  validate :metadata_field_subsets
+  validate :metadata_field_validations
 
   before_save :update_host_genomes!
 
@@ -125,12 +125,13 @@ class MetadataField < ApplicationRecord
   validates :is_required, inclusion: { in: [0, 1] }
   validates :default_for_new_host_genome, inclusion: { in: [0, 1] }, if: -> { respond_to?(:default_for_new_host_genome) && mass_validation_enabled? } # respond_to? for migrations
 
+  # Note: default_for_host_genome is no longer only for "new" hosts
   def default_for_host_genome?
     # for migration compatibility
     respond_to?(:default_for_new_host_genome) && default_for_new_host_genome == 1
   end
 
-  def metadata_field_subsets
+  def metadata_field_validations
     if is_default == 1 && is_core == 0
       errors[:name] << 'Default field must also be core field'
     end
@@ -142,8 +143,6 @@ class MetadataField < ApplicationRecord
     end
   end
 
-  # This method is designed to be called when a field is made required or
-  # default by an admin to update existing host genomes.
   def update_host_genomes!
     if is_required == 1 || default_for_host_genome?
       host_genomes << HostGenome.all_without_metadata_field(name)
--- a/app/models/metadata_field.rb
+++ b/app/models/metadata_field.rb
@@ -138,7 +138,9 @@ class MetadataField < ApplicationRecord
   end
 
   def update_host_genomes!
-    host_genomes << HostGenome.all_without_metadata_field(name)
+    if is_required
+      host_genomes << HostGenome.all_without_metadata_field(name)
+    end
   end
 
   # Important attributes for the frontend
--- a/app/models/metadata_field.rb
+++ b/app/models/metadata_field.rb
@@ -125,6 +125,11 @@ class MetadataField < ApplicationRecord
   validates :is_required, inclusion: { in: [0, 1] }
   validates :default_for_new_host_genome, inclusion: { in: [0, 1] }, if: -> { respond_to?(:default_for_new_host_genome) && mass_validation_enabled? } # respond_to? for migrations
 
+  def default_for_host_genome?
+    # for migration compatibility
+    respond_to?(:default_for_new_host_genome) && default_for_new_host_genome == 1
+  end
+
   def metadata_field_subsets
     if is_default == 1 && is_core == 0
       errors[:name] << 'Default field must also be core field'
@@ -132,16 +137,15 @@ class MetadataField < ApplicationRecord
     if is_required == 1 && is_default == 0
       errors[:name] << 'Required field must also be default field'
     end
-    if respond_to?(:default_for_new_host_genome) && # for migration compatibility
-       is_required == 1 && default_for_new_host_genome == 0 &&
-       errors[:name] << 'Required field must also be default_for_new_host_genome field'
+    if is_required == 1 && !default_for_host_genome?
+      errors[:name] << 'Required field must also be default_for_new_host_genome field'
     end
   end
 
   # This method is designed to be called when a field is made required or
   # default by an admin to update existing host genomes.
   def update_host_genomes!
-    if is_required == 1 || default_for_new_host_genome == 1
+    if is_required == 1 || default_for_host_genome?
       host_genomes << HostGenome.all_without_metadata_field(name)
     end
   end
--- a/app/models/metadata_field.rb
+++ b/app/models/metadata_field.rb
@@ -138,7 +138,7 @@ class MetadataField < ApplicationRecord
   end
 
   def update_host_genomes!
-    if is_required
+    if is_required == 1
       host_genomes << HostGenome.all_without_metadata_field(name)
     end
   end
--- a/app/models/metadata_field.rb
+++ b/app/models/metadata_field.rb
@@ -89,7 +89,7 @@ class MetadataField < ApplicationRecord
 
   validate :metadata_field_subsets
 
-  before_create :update_host_genomes!
+  before_save :update_host_genomes!
 
   # As of Feb 2020, we renamed "Host Genome" in the UI to better reflect its
   # purpose as "Host Organism". For backwards compatibility, especially for CSV
--- a/app/models/metadata_field.rb
+++ b/app/models/metadata_field.rb
@@ -1,3 +1,82 @@
+# ActiveRecord documentation summary
+
+# Name/key ex: "sample_type"
+# t.string :name
+
+# User-friendly display name ex: "Sample Type"
+# t.string :display_name
+
+# Human-readable description e.g. "Type of sample or tissue such as plasma, whole blood, etc."
+# t.string :description
+
+# Base data type. 0 for string, 1 for number, 2 for dates. Used for figuring out which column
+# to use in "metadata-data" ex: string_validated_value, number_validated_value,
+# date_validated_value.
+# t.integer :base_type
+
+# Name of a validation type corresponding to a 'hardcoded' validation function.
+# Ex: "positive_number" here would call a positive_number validator in code dynamically (with
+# a restricted set of functions we make).
+
+# An array of options when applicable. Ex for nucleotide_type: ["DNA", "RNA"]. Unfortunately our
+# RDS flavor/version doesn't support json types so this is string.
+# t.string :options
+
+# If true then only allow users to use the options (i.e. no freetext).
+# t.integer :force_options
+
+# A hash that maps host genome ids to an array of examples for that metadata field and host genome.
+# Stored as JSON.
+# If the key is "all", the examples apply to all host genomes.
+# We separate by host genome because, for example, valid sample_types are different for human vs mosquito,
+# and the examples should reflect this.
+# t.string :examples
+
+# +----------------------+
+# |      All fields      |
+# | +------------------+ |
+# | |      Core        | |
+# | | +--------------+ | |
+# | | |   Default    | | |
+# | | | +----------+ | | |
+# | | | | Required | | | |
+# | | | +__________+ | | |
+# | | +______________+ | |
+# | +__________________+ |
+# +______________________+
+
+# The core fields are basically the set of fields that we think are important/interesting/have
+# curated ourselves. All user-created custom types will not be core (unless we make them).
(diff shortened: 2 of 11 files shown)


tfrcarvalho:
I still think that if we *add a new metadata field that is default for new host genomes*, it should be added for old ones too (not just for required ones).

gregdingle:
I'd like to get @MarkAZhang 's input on that question. The same point applies to projects also, by the way.

tfrcarvalho:
> The same point applies to projects also, by the way.

Not exactly the same. The association of metadata fields to projects is supposed to be fluid. You can add custom fields and remove custom ones.
On the other hand, the default fields for host genomes are to indicate what fields you immediately add to a project when you first create it. That's why you must add any new default field created to existing host genomes.

app/models/metadata_field.rb
```diff
@@ -1,10 +1,91 @@
...
+# See add_default_metadata_fields in host_genome.rb.
+# t.integer :default_for_new_host_genome, limit: 1, default: 0
 class MetadataField < ApplicationRecord
   has_and_belongs_to_many :host_genomes
   has_and_belongs_to_many :projects
   has_many :metadata, dependent: :destroy
 
   validate :metadata_field_subsets
 
+  before_save :update_host_genomes!
```
cdebourcy: It seems like `update_host_genomes!` will run every time a metadata_field is saved. Where does  it check that metadata_field is a required field before adding it to all host_genomes?

I'm asking because custom metadata fields are added on the fly when samples are created (via https://github.com/chanzuckerberg/idseq-web/blob/d0b690549c105e2db0d2a9892551c91601e43ea4/app/helpers/metadata_helper.rb#L16) and we wouldn't want to trigger the callback for those.

cdebourcy:
Thank you for tackling this. Unless I'm missing something, I think you need to reduce the number of scenarios in which `update_host_genomes` is triggered, as per my in-line comment. To match your description, it should only be triggered when a _required_ field is added, which can only be caused through manual action of an admin member of the development team.

I also echo Tiago's comment about default metadata fields for old host genomes. His request is in fact the definition & acceptance criterion for task https://jira.czi.team/browse/IDSEQ-1398

app/models/metadata_field.rb
gregdingle: Yes, that's correct. The unit tests failure also showed this missing logic. I'll fix it now.

gregdingle:
> His request is in fact the definition & acceptance criterion for task https://jira.czi.team/browse/IDSEQ-1398

Sorry, I'm not seeing that in the acceptance criteria of the task. I was going by 

> Jonathan Sheu Assigned to you to get clarity on what we should do here.
> My opinion: default fields should be loaded in real time, i.e. when we need the metadata fields for a particular host genome we load all associated fields + all fields marked as default (as opposed to having to create/delete associations when we add/remove a default field).

If the question has been already been decided before I arrived, I'm happy to go with that.

MarkAZhang:
> On the other hand, the default fields for host genomes are to indicate what fields you immediately add to a project when you first create it. That's why you must add any new default field created to existing host genomes.

When a project is created, all fields marked as `is_default` are added to the project. It shouldn't depend on any host genome. Projects and host genomes shouldn't rely on each other for any defaults related to metadata fields. A sample takes the intersection of its project and host genome's metadata fields (which are independent) when determining what fields are valid for itself.

I do think it makes sense for us to update past projects as well as host genomes when new default fields are added. There is a slight downside that we are modifying past projects (example of a possible downside: the user was very diligent in uploading complete metadata based on the schema at the time, but now we are adding additional fields so they are missing metadata), but the intention was that required and default fields should change infrequently.
I think retroactively adding default fields will reduce the likelihood of a certain class of bugs. For example, we make a new default metadata field and we make some UI assuming that this field is present. A bug might occur if the user views a past sample  that doesn't have this default field. 

So yes, I would support doing this not just for required fields, but for default fields as well.

Note that when a host genome is created, it adds all metadata fields that are `default_for_new_host_genome`, where when a project is created, it adds all metadata fields that are `is_default`.

The reason why there is this distinction is that there are some metadata fields, such as `is_blood_fed` which are default fields, but specific to certain host genomes (mosquitoes). We do want `is_blood_fed` added to new projects, so that mosquito samples within these projects get the field, but we don't want `is_blood_fed` added to new host genomes, so `default_for_new_host_genome` is false. `default_for_new_host_genome` is true for fields like `collection_location` that should apply to all new host genomes.

MarkAZhang:
To summarize the proposed changes, when a new metadata field is created, if it has `default_for_new_host_genome`, add it to all past host genomes. If it has `is_default`, add it to all past projects. 

Separately, there is an implicit assumption that if it `is_required`, then it will also be both `default_for_new_host_genome` and `is_default`. I don't know if there is validation for this, and it would be good to add it in metadata field.

MarkAZhang:
> My opinion: default fields should be loaded in real time, i.e. when we need the metadata fields for a particular host genome we load all associated fields + all fields marked as default (as opposed to having to create/delete associations when we add/remove a default field).

I think Tiago is talking about not storing default fields on the host genome via the many-to-many relationship, but instead just loading them ad-hoc whenever we need them. (i.e. changing the current implementation)

I'm a little against this because 1) if we remove `default_for_new_host_genome` from a metadata field, then it gets implicitly deleted from all the host genomes, causing metadata to possibly stop being shown in the UI and 2) now you have to remember to also load the default fields when you're writing code, which can increase the chance of developer error (especially as we've already invested in the existing solution).

app/models/metadata_field.rb
MarkAZhang: Agree! This seems like it might be a bug.

gregdingle: Fixed now with test case. See https://github.com/chanzuckerberg/idseq-web/pull/3148/files#diff-154aca7aaaf3a253a78f2691590336c9R59 .

gregdingle:
Thanks for all the thoughts @MarkAZhang ! 

I'm still working on making the unit tests pass for now. It looks like duplicate fields are getting added somehow because of the callback.

gregdingle:
@MarkAZhang, about this:

> Separately, there is an implicit assumption that if it is_required, then it will also be both default_for_new_host_genome and is_default. I don't know if there is validation for this, and it would be good to add it in metadata field.

`is_required` and `default_for_new_host_genome` assumption should be covered in https://github.com/chanzuckerberg/idseq-web/pull/3148/files#diff-30474228c4745c95ef6519536e46e8d6R129 . 

`default_for_new_host_genome` and `is_default` seems questionable. Should every default field for a a host also be a default a field for all projects?

cdebourcy:
> Sorry, I'm not seeing that in the acceptance criteria of the task. I was going by
> 

Based on the title of the task, "Default metadata fields should be automatically added to old host genomes" :)

gregdingle:
Hi reviewers, here is a linear summary up of the PR discussions up to this point with follow up questions.

There is consensus that fields that are default for new hosts should also be added to old hosts.

I will add that logic and corresponding tests now.

A follow-up set of questions is
* whether we should change the name of the field default_for_new_host_genome to default_for_host_genome?
* where to update the internal documentation? I know there is a quip doc and the code docs.
* is there any update needed to external documentation? Anything about metadata fields being frozen that would no longer be true?

Further, Mark supports the change to make fields that are default for new projects to also be default for old projects.

> I think retroactively adding default fields will reduce the likelihood of a certain class of bugs. For example, we make a new default metadata field and we make some UI assuming that this field is present

> ...possible downside: the user was very diligent in uploading complete metadata based on the schema at the time, but now we are adding additional fields so they are missing metadata

While I think these are both good points, I think on balance it's not worth the time and risk of disruption to make this change now.

Any thoughts @tfrcarvalho , @cdebourcy ?

Further, Tiago recommends not storing default fields on the host genome via the many-to-many relationship which removes the need to synchronize via callbacks.

> default fields should be loaded in real time, i.e. when we need the metadata fields for a particular host genome we load all associated fields + all fields marked as default (as opposed to having to create/delete associations when we add/remove a default field).

Mark is slightly against it:

> 1) if we remove default_for_new_host_genome from a metadata field, then it gets implicitly deleted from all the host genomes, causing metadata to possibly stop being shown in the UI and
> 2) now you have to remember to also load the default fields when you're writing code, which can increase the chance of developer error (especially as we've already invested in the existing solution).

I'm also against it for those reasons. Any thoughts @tfrcarvalho , @cdebourcy ?

cdebourcy:
> 
> I'm also against it for those reasons. Any thoughts @tfrcarvalho , @cdebourcy ?

Thank you for this great summary and analysis. I vote for doing the minimal change that fixes the bug reported in IDSEQ-1398 (i.e. for _not_ expanding the scope to a refactor).

tfrcarvalho:
Definitely, great way to summarize @gregdingle ! 

Regarding @MarkAZhang points, they are valid from the sense of changing existing code. 

How i would envision it, would be changing the current `metadata_fields` into a `custom_metadata_fields` and adding model method `metadata_fields that merges custom metadata_fields with the default ones. The clients of current metadata_fields attribute would not have to change.

Having said that, it is true that it is more work (not much more but more). So @gregdingle, I will let you decide how to proceed.

gregdingle:
I prefer to keep things as-is, although I readily admit that I prefer we had started with the system that @tfrcarvalho described, given where we have ended up.

gregdingle:
Ping. @MarkAZhang @tfrcarvalho . Could I get a final review here?

app/models/host_genome.rb
```diff
@@ -46,6 +46,11 @@ def self.s3_bowtie2_index_path_default
     HostGenome::ERCC_PATH_PREFIX + HostGenome::S3_BOWTIE2_INDEX_FILE
   end
 
+  def self.all_without_metadata_field(name)
+    exists = joins(:metadata_fields).where(metadata_fields: { name: [name] })
+    where.not(id: [exists.pluck(:id).uniq])
```
tfrcarvalho: DO you nee pluck? What about - for a single query: 
```
exists = HostGenome.joins(:metadata_fields).where(metadata_fields: { name: [name] }).distinct
where.not(id: exists)
```
Not event sure you need the distinct... but maybe for perf.

spec/models/metadata_field_spec.rb
```diff
@@ -26,4 +30,55 @@
     expect { create(:metadata_field, name: 'sample_name') }.to raise_error(ActiveRecord::RecordInvalid)
     expect { create(:metadata_field, name: 'my_name') }.to_not raise_error
   end
+
+  it 'will not create a required field that is not default_for_new_host_genome' do
```
tfrcarvalho: Sometimes is good to create contexts to group tests:
```
context 'for required fields' do
	context 'if is not default for new hosts' do
		it 'will not create ' do
		it 'will not update' do
```
I feel it is easier to read as opposed to the long sentences.

spec/models/metadata_field_spec.rb
```diff
@@ -26,4 +30,55 @@
...
+
+  it 'will update to required a field that is default_for_new_host_genome' do
+    metadata_field = create(:metadata_field, name: 'test', is_default: 1, is_core: 1, default_for_new_host_genome: 1)
+    expect { metadata_field.update(is_required: 1) }.to_not raise_error
+  end
+
+  it 'will update all hosts needed when a required field is created' do
+    host_genome = create(:host_genome)
+    metadata_field = find_test_field(host_genome)
+    expect(metadata_field).to be_nil
```
tfrcarvalho: What is this testing? Is it testing the test setup ?

spec/models/metadata_field_spec.rb
```diff
@@ -26,4 +30,55 @@
...
+    expect { create(:metadata_field, name: 'test', is_default: 1, is_core: 1, default_for_new_host_genome: 1, is_required: 1) }.to_not raise_error
+  end
+
+  it 'will update to required a field that is default_for_new_host_genome' do
+    metadata_field = create(:metadata_field, name: 'test', is_default: 1, is_core: 1, default_for_new_host_genome: 1)
+    expect { metadata_field.update(is_required: 1) }.to_not raise_error
+  end
+
+  it 'will update all hosts needed when a required field is created' do
+    host_genome = create(:host_genome)
```
tfrcarvalho: You can put this host genome on a let block above, right?

app/models/metadata_field.rb
```diff
@@ -43,85 +125,10 @@ class MetadataField < ApplicationRecord
...
-
-  # Whether this metadata field should be added automatically to new host genomes.
-  # See add_default_metadata_fields in host_genome.rb.
-  # t.integer :default_for_new_host_genome, limit: 1, default: 0
+  def default_for_host_genome?
+    # for migration compatibility
+    respond_to?(:default_for_new_host_genome) && default_for_new_host_genome == 1
+  end
 
   def metadata_field_subsets
```
MarkAZhang: Nit: Could we name this `metadata_field_validations`? I believe the Rails convention is to describe the property you are validating, but it's a little hard to put into words. (is_default is a "metadata-field field", so it would be `metadata_field_field_dependencies`). `subsets` feels confusing to me though and causes me to think the function is about something else.

app/models/metadata_field.rb
```diff
@@ -43,85 +125,10 @@ class MetadataField < ApplicationRecord
...
-  # User-defined meta-fields will belong to each (and potentially multiple) projects. So
-  # meta-fields have many projects and projects have many meta-fields.
-  # When a user creates a new project, they'll basically get a list of all the meta-fields marked
-  # "default". Then they can add and subtract from their set of meta-fields from there.
-  # create_join_table :projects, :metadata_fields
-
-  # Whether this metadata field should be added automatically to new host genomes.
-  # See add_default_metadata_fields in host_genome.rb.
-  # t.integer :default_for_new_host_genome, limit: 1, default: 0
+  def default_for_host_genome?
```
MarkAZhang: Nit: Can we rename this function to `default_for_new_host_genome`? To keep it consistent with the underlying field name.

app/models/metadata_field.rb
```diff
@@ -130,6 +137,17 @@ def metadata_field_subsets
     if is_required == 1 && is_default == 0
       errors[:name] << 'Required field must also be default field'
     end
+    if is_required == 1 && !default_for_host_genome?
+      errors[:name] << 'Required field must also be default_for_new_host_genome field'
+    end
+  end
+
+  # This method is designed to be called when a field is made required or
```
MarkAZhang: Nit: This comment looks out of date, as this method is currently being called whenever a field is updated.

app/models/metadata_field.rb
```diff
@@ -130,6 +137,17 @@ def metadata_field_subsets
...
+  # default by an admin to update existing host genomes.
+  def update_host_genomes!
+    if is_required == 1 || default_for_host_genome?
```
MarkAZhang: Nit: Would it make sense to just have `if default_for_host_genome?`. Since if it's required, it will be default_for_host_genome.

spec/factories/host_genomes.rb
```diff
@@ -8,15 +8,12 @@
...
-    after :create do |host_genome, options|
+    after :create do |_host_genome, options|
       options.metadata_fields.each do |metadata_field_name|
         metadata_field = MetadataField.find_by(name: metadata_field_name)
         unless metadata_field
-          metadata_field = create(:metadata_field, name: metadata_field_name)
+          create(:metadata_field, name: metadata_field_name)
         end
-
-        # Add the metadata field to the host genome if it doesn't already exist.
```
MarkAZhang: Should this still be removed? It seems like the metadata field wouldn't be added to the host genome unless it was required or default for host genome.

spec/factories/samples.rb
```diff
@@ -26,16 +26,12 @@
 
     # metadata fields
     # ensure the metadata field is added to the host genome
-    before(:create) do |sample, options|
+    before(:create) do |_sample, options|
       options.metadata_fields.each_key do |metadata_field_name|
         metadata_field = MetadataField.find_by(name: metadata_field_name)
         unless metadata_field
-          metadata_field = create(:metadata_field, name: metadata_field_name)
+          create(:metadata_field, name: metadata_field_name)
```
MarkAZhang: I'm glad we got rid of the `default_for_new_host_genome` hack! But don't we still need the `sample.host_genome.metadata_fields << metadata_field` line?

gregdingle:
@MarkAZhang ready for review again.

app/models/metadata_field.rb
MarkAZhang: Okay that makes sense. Would it be too much trouble then to rename the column itself with `rename_column`? It would remove confusion in the future, now that we've permanently changed the meaning of this column.

gregdingle: `rename_column` would be easy but we would need to change all ruby references as well and test them which I think is too much trouble. 

I guess the middleground is to rename the column, add an attribute `default_for_new_host_genome` for legacy and leave a deprecated comment. 

However, I'm not sure that is even enough. I'd want to research "renaming activerecord attribute and backward compatibility" .

PR #3149, opened 2020-03-05
Special case reads remaining display for cdhitdup

gregdingle:
# Description

This change communicates the new meaning of reads after the cdhitdup changes. NOTE: it should only be committed after https://github.com/chanzuckerberg/idseq-dag/pull/252/files . 

image

# Tests

* try in localhost, see updated display

### DIFF
--- a/app/helpers/pipeline_runs_helper.rb
+++ b/app/helpers/pipeline_runs_helper.rb
@@ -10,7 +10,7 @@ module PipelineRunsHelper
         "star_out" => "Remove host reads using STAR.",
         "trimmomatic_out" => "Trim Illumina adapters using trimmomatic.",
         "priceseq_out" => "Remove low-quality reads using PriceSeqFilter.",
-        "cdhitdup_out" => "Remove duplicate reads using CD-HIT-DUP.",
+        "cdhitdup_out" => "Identify duplicate reads using CD-HIT-DUP. All reads at this step remain in downstream analysis.",
         "lzw_out" => "Remove low-complexity reads using LZW compression filter.",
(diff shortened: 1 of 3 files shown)


cdebourcy:
Nice!!

app/helpers/pipeline_runs_helper.rb
```diff
@@ -10,7 +10,7 @@ module PipelineRunsHelper
         "star_out" => "Remove host reads using STAR.",
         "trimmomatic_out" => "Trim Illumina adapters using trimmomatic.",
         "priceseq_out" => "Remove low-quality reads using PriceSeqFilter.",
-        "cdhitdup_out" => "Remove duplicate reads using CD-HIT-DUP.",
+        "cdhitdup_out" => "Identify duplicate reads using CD-HIT-DUP. All reads at this step remain in downstream analysis.",
```
cdebourcy: I think this is not quite right. If I'm not mistaken, only unique reads will feed into the assembly step, for example, which could make a real difference. I think what you mean is "Downstream read **counts will be reported without deduplication**" or equivalent. Apologies if other team members have already reviewed this copy and I'm weighing in late

gregdingle: This was @katrinakalantar 's language. I think the intention was to communicate that the number does not change at the cdhitdup step. Happy to add the additional `Downstream read counts will be reported without deduplication` or re-work it further with help.

cdebourcy: Ok then let's go with what Katrina said and keep what you have :) Thank you for explaining

PR #3152, opened 2020-03-05
[IDSEQ-2192 ] [Heatmap] Append sample ids to samples with duplicate names

j-x-han:
# Description

When users selected samples with the same name (from multiple projects), some of the sample labels would not appear and the corresponding metadata would also be blank. This change appends numbers the duplicate sample names to differentiate between them, and also makes the label blue and adds a tooltip to notify the user.
**Before:**
image
**After (mouse hovering over left label):**
image

# Tests

* Created a heatmap using samples with the same name from two different projects and verified that the sample names and metadata rendered as expected.


happyimadesignr:
Hey, one of the dups comes from a project called RR038 GCE South Africa Test Nextseq. It seems to be a test project? I asked in the biohub slack if it was if the project was a test and if the data is actually duplicated or if it's just a duplicate name. 

The duplicates are both from RR038. It seems strange to have the same name, especially both being from RR038.

- For dups, I think it should just be a number added to the end of the dups like (1) and (2). People can also see from the details panel which project it's from.
- Maybe the 1 and 2 can be blue links, with a tooltip that points out that it's a dup. I think people shouldn't accidentally think that there is 2 samples with data when it's actually the same data.

j-x-han:
@happyimadesignr Boris pointed out that the two samples from that project were actually slightly different, so it wasn't the exact same sample!

I've made the changes you requested and updated the screenshots accordingly. Let me know what you think 😄

happyimadesignr:
@j-x-han Yay looks good to me. We just chatted about removing the blue box and making the links black by default.

PR #3156, opened 2020-03-06
Add script for running local server without Docker

kislyuk:


bin/simple-local-server.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+brew services start elasticsearch
+brew services start redis
+brew services start mysql
+
+yq -yi '.development.host="127.0.0.1" | .test.host="127.0.0.1"' config/database.yml
+sed -i -e 's/redis:6379/localhost:6379/' config/initializers/redis.rb
```
tfrcarvalho: Maybe define env variable `REDISCLOUD_URL='redis://localhost:6379'` and not change code?

bin/simple-local-server.sh
```diff
@@ -0,0 +1,25 @@
...
+sed -i -e '/TaxonLineage/ d' lib/tasks/create_elasticsearch_indices.rake
```
tfrcarvalho: Are you just trying saving time/space?
This creates a `.rake-e`... was that on purpose?

bin/simple-local-server.sh
tfrcarvalho: If cache is needed we need to set redis also on `config/environments/development.rb`

bin/simple-local-server.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+brew services start elasticsearch
+brew services start redis
+brew services start mysql
+
+yq -yi '.development.host="127.0.0.1" | .test.host="127.0.0.1"' config/database.yml
```
tfrcarvalho: You should be able to use env variables too, no?
```
development:
  host: <%= ENV['DATABASE_HOST'] %>
```

bin/simple-local-server.sh
```diff
@@ -0,0 +1,25 @@
...
+
+trap cleanup INT EXIT
+
+cleanup() {
+    kill -TERM 0
+    wait
+}
+
+"$(dirname $0)/entrypoint.sh" "rake db:migrate"
+"$(dirname $0)/entrypoint.sh" "rake create_elasticsearch_indices"
```
tfrcarvalho: No need to run it all the time...

gregdingle:
Exciting! I'm going to try this out 😄

bin/simple-local-server.sh
kislyuk: Yep - this is to avoid the very expensive TaxonLineage indexing. I'll see if I can get rid of a .rake-e file.

bin/simple-local-server.sh
kislyuk: Did not think of that, will try.

bin/simple-local-server.sh
tfrcarvalho: It seems like it is taking -e as the optional extension to -i option.

PR #3157, opened 2020-03-06
IDSEQ-2311: Remove Min Contigs Reads filter from report page

gregdingle:
# Description

From the ticket: The min contig_r filter on the report page UI does not follow the same behavior as the other filters. It is also rarely used < 5 clicks / 1300 total filters applied in the past 3 months.

In addition, the original behavior of filtering on unique reads would be at odds with the new definition of reads as non-unique with the launch of the cdhitdup change. 

## Before

image

## After 

image

# Notes

* is it still possible to see the v1 report?
* is it worth to also clean up any related server side code?
* there are some screenshots here to change as well: https://chanzuckerberg.zendesk.com/hc/en-us/articles/360034790574-Single-Sample-Report-Table . Also note that the screen is already out of date because of the change from "size" to "reads". I'm not sure exactly how to change the screenshot. @happyimadesignr ?

# Tests

* load the report page v2, see no filter


MarkAZhang:
There are other pieces of code you can remove (such as the callbacks and the back-end logic), but I'm guessing this is a WIP to get feedback.

Removing this feature sounds good to me. Users will want to do quality control on contigs, but I don't think this feature is particularly helpful for quality control (there are other factors to look at besides the contig reads, so it's not a complete solution). As long as Katrina is happy, SGTM.

tfrcarvalho:
@gregdingle If this filter is no longer used we can optimize the server significantly... Currently we are sending all the info necessary to apply the filter on the client side - which is a lot for some samples - so we can narrow that down per taxon.

gregdingle:
@tfrcarvalho That sounds great! Can you point me to a good place to start removing in the server code?

tfrcarvalho:
> Can you point me to a good place to start removing in the server code?

If you want to take a shot at removing the extra `read_count` dimension (and compute the total sum instead) at https://github.com/chanzuckerberg/idseq-web/blob/master/app/models/pipeline_run.rb#L1622.
This is called in the `PipelineReportService` at https://github.com/chanzuckerberg/idseq-web/blob/master/app/services/pipeline_report_service.rb#L125 (and a few lines below).

You probably will need to adapt both the functions that merges the contigs to the rest of the results (https://github.com/chanzuckerberg/idseq-web/blob/master/app/services/pipeline_report_service.rb#L413) and the frontend code that displays it (https://github.com/chanzuckerberg/idseq-web/blob/master/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx#L423). For the latter, you can probably delete most of it (and `computeRowContigStats` above)

PR #3161, opened 2020-03-09
Require description when creating a new project

lvreynoso:
# Description

Disables the create button if a project description is not entered (minimum 8 characters). Adds a popup explaining the requirements.

# Tests

No server side changes.

For frontend: go to sample upload page, click create project. Test that the "create project" buttons is disabled, and on hover has a popup explaining the requirements. Test that the popup changes when each requirement is fulfilled. The popup should disappear completely and the "create project" button enabled when all requirements are fulfilled.

Screen Shot 2020-03-09 at 3 48 18 PM


happyimadesignr:
Hey @lvreynoso! Thanks!

**Few questions/comments:**
- I was wondering: why is there an 8 character limit on the project description?
- I think a shorter description would be sufficient like "Complete all fields to create a project."
- Can the tooltip be centered on the "CREATE PROJECT" button?

lvreynoso:
Thanks for taking a look Jenn! The 8 characters isn't a limit, it's a minimum; but it's also an arbitrary number, I can just check for at least 1 character in the field. I can also make the tooltip have a shorter description and centered on the `Create Project` button.

lvreynoso:
Updated popup:

Screen Shot 2020-03-11 at 11 31 24 AM

happyimadesignr:
Thank you @lvreynoso! Can you make the tooltip width smaller? So it's more balanced with more words on the bottom =]=] Thank you!!

lvreynoso:
More aesthetic popup display:

Screen Shot 2020-03-11 at 12 07 43 PM

app/models/project.rb
```diff
@@ -18,6 +18,7 @@ class Project < ApplicationRecord
   validates :public_access, inclusion: { in: [0, 1] }, if: :mass_validation_enabled?
   # NOTE: all values of background_flag in prod db are currently zero.
   validates :background_flag, inclusion: { in: [0, 1] }, allow_nil: true, if: :mass_validation_enabled?
+  validates :description, presence: true
```
gregdingle: please add a comment that this requirement was added 2020-03-11

PR #3162, opened 2020-03-10
Avoid rails cache in mass validation

gregdingle:
# Description

 #3074 is causing a bunch of log noise in pipeline monitor. Such as 

```
[2020-03-09T17:28:48] DEBUG ActiveSupport::Cache::RedisStore : Cache read: ApplicationRecord.mass_validation_enabled? ({:expires_in=>30 days})
[2020-03-09T17:28:48] DEBUG ActiveSupport::Cache::RedisStore : Cache fetch_hit: ApplicationRecord.mass_validation_enabled? ({:expires_in=>30 days})
[2020-03-09T17:28:48] DEBUG ActiveSupport::Cache::RedisStore : Cache read: ApplicationRecord.mass_validation_enabled? ({:expires_in=>30 days})
```

# Tests

use byebug to ensure `@@mass_validation_enabled` is set and only once

try pipeline monitor
see no more log noise


app/models/application_record.rb
```diff
@@ -19,8 +22,13 @@ def log_errors
   # Cached for performance.
   def mass_validation_enabled?
     if AppConfig.table_exists? # for migrations previous to AppConfig creation
-      return Rails.cache.fetch("ApplicationRecord.mass_validation_enabled?") do
-        AppConfigHelper.get_app_config(AppConfig::ENABLE_MASS_VALIDATION)
+      if !@@mass_validation_enabled.nil?
```
tfrcarvalho: I really don't think we need cache at all. 

And here `@@mass_validation_enabled ||= AppConfigHelper.get_app_config(AppConfig::ENABLE_MASS_VALIDATION)` should be enough...

@davidrissato Why would you say we need redis cache?

davidrissato: We actually don’t need redis cache, a constant is enough. We just need to make sure this line you suggested produces the effect we want, because that is an abstract class, and I believe the behavior of @@ is different in this case.

gregdingle: I updated to use a class var only, and I retested that it is only set once.

> I believe the behavior of @@ is different in this case

I couldn't find any references to that. Any pointers? My understanding of ruby class vars is that they act as global vars.

PR #3163, opened 2020-03-10
[IDSEQ-2171] [Heatmap] Show notification for failed/waiting samples

j-x-han:
# Description
* When a user creates a heatmap with a failed or waiting sample, the sample would appear but have no information.
Screen Shot 2020-01-28 at 11 11 28 AM

* This change ensures that failed/waiting samples do not appear in a heatmap, and also adds a notification describing to the user what happened.

image
image

# Notes

* Added `font-caps` mixins!

* This PR uses the `validateSampleIdsForBulkDownload` API call to validate samples, so I renamed the call to the more generic `validateSampleIds`.

* This PR made changes to the `AccordionNotification` component used by the bulk download modal, so I'm including screenshots of that as well to ensure that the style hasn't changed too much. @happyimadesignr 
image
image

* Changed long sample names to wrap rather than horizontal scroll.
image
image

# Tests

* Selected failed and waiting samples while creating a heatmap and verified that those samples did not appear in the heatmap, but were listed in a warning notification.
* Selected failed and waiting samples for bulk download and verified that the warning notification still displays as expected.


happyimadesignr:
OMG amazing! @j-x-han 

- Thanks for adding the font-cap- mixins!
- Can we move the accordion arrow up and more to the right? If I could inspect in I could give exact values. Move up to center with the circle icon.
- The wrap looks good! Better than horizontal scrolling! The line looks a bit long still, can you make it wrap a bit earlier?

j-x-han:
Updated screenshots after moving the accordion arrow and wrapping long sample names earlier (checked with Jenn!):
image
image

app/assets/src/api/bulk_downloads.js
```diff
@@ -17,7 +17,7 @@ export const getBulkDownload = bulkDownloadId =>
 export const getPresignedOutputUrl = bulkDownloadId =>
   get(`/bulk_downloads/${bulkDownloadId}/presigned_output_url`);
 
-export const validateSampleIdsForBulkDownload = sampleIds =>
+export const validateSampleIds = sampleIds =>
   postWithCSRF("/bulk_downloads/validate_sample_ids", {
     sampleIds: sampleIds,
   });
```
lvreynoso: I'm glad you're also finding a use for this function :) We should probably refactor the backend part to be more generalized, like changing the route and function to live on the samples controller. But I think that can wait for an API review.

app/assets/src/components/views/bulk_download/bulk_download_modal_footer.scss
```diff
@@ -31,12 +31,8 @@
...
 .messageLine {
   @include font-body-xxs;
-  margin: 0 $space-s;
-  padding: $space-xxxs;
-
-  &:last-child {
-    margin-bottom: $space-s;
-  }
+  padding: $space-xxxs 0;
+  margin-right: $space-l;
```
lvreynoso: Thanks for taking care to update this!

app/assets/src/api/bulk_downloads.js
tfrcarvalho: This is a good point! I am actually in favor of doing this small low risk refactors as we find them, before they grow to be used in multiple places.

PR #3166, opened 2020-03-11
Fix logic of mass_validation_enabled class var

gregdingle:
# Description

The original `||=` allowed the db query to run. 

# Tests

* unit tests
* tested by @tfrcarvalho in pipeline monitor

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -21,11 +21,12 @@ class ApplicationRecord < ActiveRecord::Base
   # Condition for rollout of mass addition of validation rules.
   # Cached for performance.
   def mass_validation_enabled?
-    if AppConfig.table_exists? # for migrations previous to AppConfig creation
-      @@mass_validation_enabled ||= AppConfigHelper.get_app_config(AppConfig::ENABLE_MASS_VALIDATION) || false # rubocop:disable Style/ClassVars
+    # for migrations previous to AppConfig creation
+    if AppConfig.table_exists?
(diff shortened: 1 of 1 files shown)


app/models/application_record.rb
```diff
@@ -21,11 +21,12 @@ def log_errors
   # Condition for rollout of mass addition of validation rules.
   # Cached for performance.
   def mass_validation_enabled?
-    if AppConfig.table_exists? # for migrations previous to AppConfig creation
-      @@mass_validation_enabled ||= AppConfigHelper.get_app_config(AppConfig::ENABLE_MASS_VALIDATION) || false # rubocop:disable Style/ClassVars
+    # for migrations previous to AppConfig creation
+    if AppConfig.table_exists?
+      @@mass_validation_enabled = !!AppConfigHelper.get_app_config(AppConfig::ENABLE_MASS_VALIDATION) if @@mass_validation_enabled.nil? # rubocop:disable Style/ClassVars,Style/DoubleNegation
```
tfrcarvalho: Actually you can get rid of double negation and go back to `|| false`...

gregdingle: agree, but I'd prefer not to re-test

tfrcarvalho: just hate disabling alert inline

gregdingle: in general, do you think `!!` is a bad pattern? maybe we should review all the "skips" we have in the codebase in the next eng lunch.

tfrcarvalho: No, i like it ...

PR #3168, opened 2020-03-11
Add font-caps mixins to playground

j-x-han:
# Description

Adding the font-caps mixins created in https://github.com/chanzuckerberg/idseq-web/pull/3163 to the fonts playground.


tfrcarvalho:
Thank you for doing this!

PR #3169, opened 2020-03-11
Clean up unneeded contig code

gregdingle:
# Description

Following #3157 , we no longer need to compute and send per count taxa info to the report page. 

## Before

image

## After

image

# Tests

* try http://localhost:3000/samples/13435 
* see no change in contig rows before and after

### DIFF
--- a/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
+++ b/app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
@@ -1,11 +1,9 @@
 import React from "react";
 import {
   compact,
-  entries,
   every,
   find,
   flatten,
-  flow,
   get,
   keys,
   map,
@@ -160,7 +158,6 @@ export default class SampleViewV2 extends React.Component {
     return {
       categories: {},
       metric: TREE_METRICS[0].value,
-      minContigReads: 4,
       nameType: "Scientific name",
       readSpecificity: 0,
       thresholds: [],
@@ -251,7 +248,6 @@ export default class SampleViewV2 extends React.Component {
     }
 
     this.setDisplayName({ reportData, ...selectedOptions });
-    this.setContigStats({ reportData, ...selectedOptions });
     const filteredReportData = this.filterReportData({
       reportData,
       filters: selectedOptions,
@@ -397,25 +393,6 @@ export default class SampleViewV2 extends React.Component {
       !readSpecificity ||
       (row.taxLevel === "genus" ? row.taxId > 0 : row.genus_tax_id > 0)
     );
-  };
-
-  setRowContigStats = ({ row, minContigReads }) => {
-    ["nr", "nt"].forEach(dbType => {
-      const dbTypeRow = row[dbType];
-      if (dbTypeRow) {
-        dbTypeRow.contigCount = get([dbType, "contigs"], row);
-        dbTypeRow.readsCount = get([dbType, "contig_r"], row);
-      }
-    });
-  };
-
-  setContigStats = ({ reportData, minContigReads }) => {
-    reportData.forEach(genus => {
-      this.setRowContigStats({ row: genus, minContigReads });
-      genus.species.forEach(species => {
-        this.setRowContigStats({ row: species, minContigReads });
-      });
-    });
   };
 
   setDisplayName = ({ reportData, nameType }) => {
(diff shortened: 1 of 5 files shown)


app/services/pipeline_report_service.rb
```diff
@@ -420,18 +422,14 @@ def merge_contigs(contigs, counts_by_tax_level, csv)
         end
 
         if counts_per_db_type
-          if csv
```
gregdingle: after this change, the non-CSV mode is the same as CSV mode

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -399,32 +399,21 @@ export default class SampleViewV2 extends React.Component {
     );
   };
 
-  computeRowContigStats = ({ row, minContigReads }) => {
+  setRowContigStats = ({ row, minContigReads }) => {
```
tfrcarvalho: You might not need this function.
You probably just need to adjust the key names in the ReportTable: https://github.com/chanzuckerberg/idseq-web/blob/master/app/assets/src/components/views/SampleViewV2/ReportTable.jsx#L156

tfrcarvalho: In any case, `minContigReads` its not used

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -399,32 +399,21 @@ export default class SampleViewV2 extends React.Component {
...
+      const dbTypeRow = row[dbType];
+      if (dbTypeRow) {
+        dbTypeRow.contigCount = get([dbType, "contigs"], row);
+        dbTypeRow.readsCount = get([dbType, "contig_r"], row);
       }
     });
   };
 
-  computeContigStats = ({ reportData, minContigReads }) => {
+  setContigStats = ({ reportData, minContigReads }) => {
```
tfrcarvalho: Should be able to remove this too

gregdingle:
> Ideally we would change get_summary_contig_counts_v2 to aggregate the results in the database, if possible... that function is the true bottleneck...

I don't see an easy way to do this. The current SQL is 
```
SELECT read_count,
       species_taxid_nt,
       species_taxid_nr,
       genus_taxid_nt,
       genus_taxid_nr
FROM `contigs`
WHERE `contigs`.`pipeline_run_id` = 20155
  AND (read_count >= 0)
  AND (lineage_json IS NOT NULL)
```

To aggregate we need a query like this **per count type**
```
SELECT species_taxid_nt,
      "nt",
      SUM(read_count) AS contig_r,
      COUNT(1) AS contigs 
FROM `contigs`
WHERE `contigs`.`pipeline_run_id` = 20155
  AND (read_count >= 0)
  AND (species_taxid_nt IS NOT NULL)
GROUP BY 1
```
I'm not sure if UNIONing 4 such queries would be faster. Do you see another way?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -673,11 +639,6 @@ export default class SampleViewV2 extends React.Component {
 
     // different behavior given type of option
     switch (key) {
-      // - min contig size: recompute contig statistics with new size and refresh display
-      case "minContigReads":
```
gregdingle: please confirm this is safe to remove

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -1106,7 +1067,6 @@ export default class SampleViewV2 extends React.Component {
               reportPresent={!!reportMetadata.reportReady}
               sample={sample}
               view={view}
-              minContigReads={selectedOptions.minContigReads}
```
gregdingle: and this

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
gregdingle: Thanks for pointing that out. Done.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: Yep, sounds safe.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: yes. Did you clean up the header on the previous PR?

gregdingle:
Opened https://jira.czi.team/browse/IDSEQ-2420 .

> Did you clean up the header on the previous PR?

I cleaned it up in this PR here: https://github.com/chanzuckerberg/idseq-web/pull/3169/files#diff-6f3d1dcdabd1005da96220e3de664384 . It looks min contigs was only being used there for the CSV download link.

PR #3170, opened 2020-03-11
Add support for a custom taxon whitelist processing mode, behind feature flag

jshoe:
### Description
- This adds a feature flag that adds a checkbox that adds an additional_attribute to enable a custom taxon whitelist processing mode, for a special use case.
- This goes along with this PR: https://github.com/chanzuckerberg/idseq-dag/pull/263

### Notes
- Similar addition to the flag in #3098.
- I wanted to keep it an explicit flag with `use_taxon_whitelist` for simplicity. The whitelist S3 path doesn't have to live in idseq-web at all, and we don't need that level of customizability yet. This is similar to the `do_not_process` attribute.
- I noticed we still had dag_vars around but it looks like having the explicit attributes is preferred now (with the v2 upload page).

### Testing
- I've tested this end-to-end locally, with a dag branch, so it should be good to go.

Checkbox:
Screen Shot 2020-03-11 at 4 04 18 PM

Dag example:
Screen Shot 2020-03-11 at 3 45 56 PM

Report example:
Screen Shot 2020-03-11 at 3 02 49 PM

Steps for QA:
- Ask me to add the "taxon_whitelist" feature flag for you.
- Go through the upload process normally. On the Review Step, check "Use respiratory pathogen whitelist mode".
- Sample should process successfully. Samples without the flag should also process successfully.

### DIFF
--- a/db/migrate/20200311210521_add_use_taxon_whitelist_to_samples.rb
+++ b/db/migrate/20200311210521_add_use_taxon_whitelist_to_samples.rb
@@ -1,0 +1,5 @@
+class AddUseTaxonWhitelistToSamples < ActiveRecord::Migration[5.1]
+  def change
+    add_column :samples, :use_taxon_whitelist, :boolean, default: false, null: false, comment: "If true, sample processing will filter for a whitelist of taxons."
+  end
+end
--- a/db/migrate/20200311210521_add_use_taxon_whitelist_to_samples.rb
+++ b/db/migrate/20200311210521_add_use_taxon_whitelist_to_samples.rb
@@ -1,5 +1,6 @@
 class AddUseTaxonWhitelistToSamples < ActiveRecord::Migration[5.1]
   def change
     add_column :samples, :use_taxon_whitelist, :boolean, default: false, null: false, comment: "If true, sample processing will filter for a whitelist of taxons."
+    add_column :pipeline_runs, :use_taxon_whitelist, :boolean, default: false, null: false, comment: "If true, pipeline processing will filter for a whitelist of taxons."
   end
 end
--- a/app/assets/src/api/upload.js
+++ b/app/assets/src/api/upload.js
@@ -44,6 +44,7 @@ export const bulkUploadLocalWithMetadata = async ({
       "project_id",
       "do_not_process",
       "pipeline_execution_strategy",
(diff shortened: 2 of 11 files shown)


db/migrate/20200311210521_add_use_taxon_whitelist_to_samples.rb
```diff
@@ -0,0 +1,5 @@
+class AddUseTaxonWhitelistToSamples < ActiveRecord::Migration[5.1]
+  def change
+    add_column :samples, :use_taxon_whitelist, :boolean, default: false, null: false, comment: "If true, sample processing will filter for a whitelist of taxons."
```
gregdingle: yay for comments and not null!

jshoe:
@tfrcarvalho For the use case described, I think it is safer to keep it on the sample level so that other results are not surfaced in potential reruns.

This whitelist is specifically for respiratory pathogens. I could go either way for a more specific name, but I think leaving it open to expansion (other whitelists) with this name is appropriate. Wdyt?

tfrcarvalho:
> @tfrcarvalho For the use case described, I think it is safer to keep it on the sample level so that other results are not surfaced in potential reruns.
>

Thats not the pattern we used for some other flags like this, but might make sense in this case. Up to you. Could be useful to get the same sample to run through other pipelines.

> This whitelist is specifically for respiratory pathogens. I could go either way for a more specific name, but I think leaving it open to expansion (other whitelists) with this name is appropriate. Wdyt?

I think we should do it based on the features we have. How set are we on this list being respiratory pathogens? If very set, i would called it that. If not we can keep it generic.
Eventually to expand to other list, this would probably be a field into a string defining the white list. 

Actually, I think we need to surface the name on the report page somehow... and for that the string (or a key to a whitelist table would be good).

jshoe:
1) Added the field to the pipeline_run as well, so it matches the other fields like it, as you mentioned.

2) As discussed, I'm going to defer adding more generalizability to the whitelist customization. It wouldn't be such a great solution for the other custom references use cases b/c in the pipeline this still processes regularly and then just skips hit accumulation. We can add more generalizability by list names when there's a product plan for it.

3) Added a little message on the report page as suggested 👍 

Screen Shot 2020-03-11 at 6 29 23 PM

PR #3172, opened 2020-03-12
Remove table_exists call

gregdingle:
# Description

This call was actually executing a SQL query which hurt performance in staging. The fix is to catch the exception produced during migrations. 

# Tests

try RAILS_ENV=test bin/rails db:drop db:create db:migrate db:seed
see no exception

check after deploy to staging

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -23,7 +23,7 @@ class ApplicationRecord < ActiveRecord::Base
   def mass_validation_enabled?
     @@mass_validation_enabled ||= AppConfigHelper.get_app_config(AppConfig::ENABLE_MASS_VALIDATION) || false # rubocop:disable Style/ClassVars
   rescue StandardError
-    # This will occur during some migrations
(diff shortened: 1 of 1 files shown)


app/models/application_record.rb
```diff
@@ -21,12 +21,14 @@ def log_errors
   # Condition for rollout of mass addition of validation rules.
   # Cached for performance.
   def mass_validation_enabled?
-    # for migrations previous to AppConfig creation
-    if AppConfig.table_exists?
-      @@mass_validation_enabled = !!AppConfigHelper.get_app_config(AppConfig::ENABLE_MASS_VALIDATION) if @@mass_validation_enabled.nil? # rubocop:disable Style/ClassVars,Style/DoubleNegation
+    if AppConfig.table_exists? # for migrations previous to AppConfig creation
```
tfrcarvalho: You're sill not removing the call?

gregdingle: sorry didn't push all changes yet

PR #3175, opened 2020-03-12
Handle duplicate sample names in AMR Heatmap

j-x-han:
# Description

Related to https://github.com/chanzuckerberg/idseq-web/pull/3152.

Differentiate between samples with duplicate names in the AMR heatmap by adding (#) at the end.
image

# Tests

* Created an AMR heatmap using samples with the same name from two different projects and verified that the sample names and metadata rendered as expected.

### DIFF
--- a/app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
+++ b/app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
@@ -104,16 +104,19 @@ export default class AMRHeatmapView extends React.Component {
   }
 
   correctSampleAndGeneNames(filteredSamples) {
-    let sampleNames = {};
+    let sampleNames = new Map();
     let duplicateSampleNames = new Set();
     filteredSamples.forEach(sample => {
       // Keep track of samples with the same name, which may occur if
       // a user selects samples from multiple projects.
-      if (Object.keys(sampleNames).includes(sample.sampleName)) {
-        sampleNames[sample.sampleName].push(sample);
+      if (sampleNames.has(sample.sampleName)) {
+        let duplicateSamples = sampleNames
+          .get(sample.sampleName)
+          .concat([sample]);
+        sampleNames.set(sample.sampleName, duplicateSamples);
         duplicateSampleNames.add(sample.sampleName);
       } else {
-        sampleNames[sample.sampleName] = [sample];
+        sampleNames.set(sample.sampleName, [sample]);
       }
 
       sample.amrCounts.forEach(amrCount => {
@@ -129,8 +132,8 @@ export default class AMRHeatmapView extends React.Component {
 
     // Append a number to a sample's name to differentiate between samples with the same name.
     duplicateSampleNames.forEach(sampleName => {
-      for (let i = 0; i < sampleNames[sampleName].length; i++) {
-        let sample = sampleNames[sampleName][i];
+      for (let i = 0; i < sampleNames.get(sampleName).length; i++) {
+        let sample = sampleNames.get(sampleName)[i];
         sample.sampleName = `${sampleName} (${i + 1})`;
       }
     });
--- a/app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
+++ b/app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
@@ -104,19 +104,17 @@ export default class AMRHeatmapView extends React.Component {
   }
 
   correctSampleAndGeneNames(filteredSamples) {
-    let sampleNames = new Map();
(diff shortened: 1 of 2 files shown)


app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -103,8 +103,19 @@ export default class AMRHeatmapView extends React.Component {
...
+    let duplicateSampleNames = new Set();
     filteredSamples.forEach(sample => {
+      // Keep track of samples with the same name, which may occur if
+      // a user selects samples from multiple projects.
+      if (Object.keys(sampleNames).includes(sample.sampleName)) {
+        sampleNames[sample.sampleName].push(sample);
+        duplicateSampleNames.add(sample.sampleName);
+      } else {
+        sampleNames[sample.sampleName] = [sample];
+      }
```
lvreynoso: You could use a JS [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead of a separate Set and object - it allows you to store arbitrary keys while retaining the ability to iterate over it (with the `forEach` method) without having to dump them to an array.

j-x-han: Switched to using a map instead of an object!

I kept the set since I want to use that to track only the duplicate samples so I can update only those names rather than iterating over the entire map of sample names again later.

app/assets/src/components/views/amr_heatmap/AMRHeatmapView.jsx
```diff
@@ -116,6 +130,14 @@ export default class AMRHeatmapView extends React.Component {
       });
     });
 
+    // Append a number to a sample's name to differentiate between samples with the same name.
+    duplicateSampleNames.forEach(sampleName => {
+      for (let i = 0; i < sampleNames.get(sampleName).length; i++) {
+        let sample = sampleNames.get(sampleName)[i];
```
lvreynoso: I just realized that you need to get the sample from `filteredSamples` in order to modify the name used by the AMR Heatmap. That's the only thing returned by the function, `sampleNames` gets discarded. You only need to modify the sample names that are duplicated, so there's no need to track every single sample name.

j-x-han: I'm not sure what you're suggesting. Wouldn't I need to iterate through `filteredSamples` to find the duplicated sample names, then do a second pass to update them?

I figured that keeping a set of `duplicatedSampleNames` and a map of {sampleName: [samples with sampleName]}, then just iterating through the `duplicatedSampleNames` would be the quickest way to update the names to prevent a second pass through all of `filteredSamples`. Sorry if I misunderstood your suggestion! 😅

tfrcarvalho: You could just keep track of the number of duplicates and immediately set the the suffix given the current count.
You would only need to do one cycle.

j-x-han: I would need to do another pass afterwards to add (1) to the first sample found with that name, since I wouldn't know it was a duplicate during the first cycle.

lvreynoso: AFAICT what happens in the `sampleNames` object isn’t returned or otherwise used in the component’s state, so once the function is done executing and the object goes out of scope it’s deleted.

lvreynoso: If you look at the `requestSampleData` function you can see the processing pipeline and where `filteredSamples` is used.

lvreynoso: You’re right that you have to track every name for dupes, my bad

tfrcarvalho: Ah, they all get a suffix ?

j-x-han: Yeah, that's the pattern I went with for the taxon heatmap as well. Looping in @happyimadesignr to double check!

lvreynoso: Ohh you’re storing a reference to the sample in the Map. I see that now

j-x-han:
Checked with Jenn and she was fine with not all duplicates having the suffix so now the duplicate names will look like this:
image

Updated the implementation for duplicate sample names in the Taxon Heatmap accordingly as well.

PR #3178, opened 2020-03-13
Make pipeline monitor save only when a run has changed

gregdingle:
# Description

#3074 caused a perf regression in staging because of 100s of validation calls per second. #3177 fixed the issue by caching.

This is an attempt to fix the root cause: the simple fact that pipeline monitor class `save!` on every iteration of every pipeline run in progress, which then calls `mass_validation_enabled` once per model attribute where it exists. 

Same for phylo tree, but I left that alone because I couldn't test it locally after an hour of debugging. 

This change makes `save` conditional on `has_changes_to_save?`. 

(Update: the root cause I now believe was the interaction of repeated saves and `validates_associated` which was since been removed in #3180)

# Notes

* See https://stackoverflow.com/questions/24314584/run-a-callback-only-if-an-attribute-has-changed-in-rails
* We could additionally reduce potential callback explosion by validating associated only on create. For example `validates_associated :job_stats` to `validates_associated :job_stats, on: :create` (note: I already did this in #3180 )
* Without this change, any type of callback added to PipelineRun could easily cause a perf issue

# Tests

* ran pipeline monitor and did not observe any flood of validation calls
* ran four samples to completion


gregdingle:
@tfrcarvalho Still need to test but if you want to, please be my guest!

tfrcarvalho:
Apparently, just calling `save` it does not update the record but still runs validations?
https://stackoverflow.com/questions/24412634/rails-save-method-if-no-changes

davidrissato:
I don't see why this would be related to that perf problem. PipelineMonitor runs in a separate container. It shouldn't affect web performance at all.

davidrissato:
Additionally, if these validations are affecting perfomance at that level, even after #3176, then I would recommend rolling back the entire change from #3074 . Because if performance is being affected this level in PipelineMonitor in this method, it is certainly affecting many other methods around.

gregdingle:
> Apparently, just calling save it does not update the record but still runs validations?

Yes, that is what I learned.

> I don't see why this would be related to that perf problem. PipelineMonitor runs in a separate container. It shouldn't affect web performance at all.

My working assumption is that the requests from PipelineMonitor overwhelmed the database. My main data points are that Tiago observed 100s of queries from PM while working with it locally and and I did not observe 100s of queries from using web locally. After #3176, the total volume of queries in staging went down roughly 10x. I'd like to confirm the cause was PipelineMonitor in some way tomorrow.

> Additionally, if these validations are affecting perfomance at that level, even after #3176, then I would recommend rolling back the entire change from #3074 . Because if we have a performance issue in PipelineMonitor for this reason, we certainly have it in a lot of other places.

To be clear, the perf issue in staging has been fixed by proper caching. This PR is an attempt to address the root cause.

tfrcarvalho:
@gregdingle maybe we should hold off on this. I think i understand @davidrissato's concern.
Since our application abuses of callbacks associated with save (which would no longer run), this change can have unintended consequences. Since you seem to have fixed the caching issue, and that seemed to have a positive impact, let's just observe the performance over the next few days.

Hopefully, as we refactor the pipeline we will be able to remove a lot of the intricate code.

gregdingle:
I agree, it makes sense to hold off and try to understand the consequences better.

tfrcarvalho:
@gregdingle Can we close this? At this point, I don't think the potential gains outweigh the potential risks.

gregdingle:
@tfrcarvalho While I admit the change is low-pri, and we plan to refactor the pipeline monitor, I still believe the risk of **not** acting is higher than that of **acting**. Right now, we are:

1) Calling all callbacks of all active runs on every iteration of PM

2) Relying on a low level Rails behavior to not run mutating SQL queries for all active runs on every iteration of PM. This behavior is not documented at https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-save . 

@davidrissato any opinion on this change?

davidrissato:
> @davidrissato any opinion on this change?

My opinion is that this change won't bring the benefits you expect. It just adds complexity.

My point is that if Pipeline Monitor is still taking too long to validate fields (which I don't believe it is happening anymore), there are tons of other places in idseq-web that we would need to make a similar change, more important than this one. And if that was the case, I would rather roll back the original change.

gregdingle:
Ok, I'll take the comments to mean that our interest in stability is greater than our interest in removing possibly dangerous side effects. I'll close the PR. Good discussion 👍

davidrissato:
The only benefit that I see in this change is eventually reducing potential page locks in our database, which is not exactly bad.

If that is the intention, I would rewrite the description of the entire PR, because the current description seems to target a performance issue that probably is not there.

PR #3179, opened 2020-03-13
Fix unknown firstpos: NilClass bug in development.

MarkAZhang:
# Description

q1Ac2

Fix bug that causes this screen in development when changing Rails code.
Fix taken from: # See: https://stackoverflow.com/questions/44465118/rails-5-1-unknown-firstpos-nilclass-issue-reloading-application

Enabling this fix only in development (since it doesn't affect other environments).

# Tests

* Verified that bug no longer appears in development.
* Verified that code changes are correctly picked up when the page is reloaded once.
* Since the bug appears to be related to routing, visited some different pages and verified the routing worked properly.


tfrcarvalho:
Not for QA testing, right? Please add QA flags!

gregdingle:
OMG! No more red screen? We will need to tell the entire team.

PR #3185, opened 2020-03-16
Upgrade eslint and its plugins to latest versions

kislyuk:
# Description

Upgrade eslint and its plugins to latest versions

# Notes

Need help updating the js/jsx code to pass new lint tests.

# Tests


tfrcarvalho:
I will be taking over this issue.

PR #3191, opened 2020-03-17
Correct output states validation

gregdingle:
# Description

Some of the needed states were missing from validation. 

# Tests

sql query against prod database for all actual values

### DIFF
--- a/app/models/output_state.rb
+++ b/app/models/output_state.rb
@@ -16,11 +16,28 @@ class OutputState < ApplicationRecord
   # +----------------------+
   validates :output, presence: true, if: :mass_validation_enabled?
 
-  validates :state, inclusion: { in: [
+  # Current values of state are:
+  # +----------------+
+  # | state          |
+  # +----------------+
+  # | CHECKED        |
+  # | FAILED         |
+  # | LOADED         |
+  # | LOADING        |
+  # | LOADING_ERROR  |
+  # | LOADING_QUEUED |
+  # | NULL           |
+  # | RUNNING        |
+  # | UNKNOWN        |
+  # +----------------+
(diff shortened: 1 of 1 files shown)


app/models/output_state.rb
```diff
@@ -16,11 +16,28 @@ class OutputState < ApplicationRecord
...
+    PipelineRun::STATUS_CHECKED,
+    PipelineRun::STATUS_FAILED,
     PipelineRun::STATUS_LOADED,
-    PipelineRun::STATUS_UNKNOWN,
     PipelineRun::STATUS_LOADING,
-    PipelineRun::STATUS_LOADING_QUEUED,
     PipelineRun::STATUS_LOADING_ERROR,
+    PipelineRun::STATUS_LOADING_QUEUED,
+    PipelineRun::STATUS_RUNNING,
+    PipelineRun::STATUS_UNKNOWN,
```
MarkAZhang: Hey Greg, I couldn't find any evidence that STATUS_RUNNING and STATUS_CHECKED are valid output states.

Comment from pipeline_run.rb:
The output_states indicate the state of each target output, the progression being as follows:
UNKNOWN -> LOADING_QUEUED -> LOADING -> LOADED / FAILED 

(the docs are missing LOADING_ERROR)

MarkAZhang: I think the only one that needs to be added is STATUS_FAILED.

gregdingle: > I couldn't find any evidence that STATUS_RUNNING and STATUS_CHECKED are valid output states

The states I guess are legacy. They are only found older than 2019. Also `NULL`. 

Good investigation!

gregdingle:
@MarkAZhang ready for review again.

PR #3192, opened 2020-03-17
Handle error in kickoff_pipeline

gregdingle:
# Description

In trying to run samples locally, I found a silent error when I didn't have the correct alignment config. This handles any error in `kickoff_pipeline` and makes it a upload error so a "sample failed" message is shown instead of "waiting". 

image

# Notes

* In general we should favor `save!` over `save`. 

# Tests

* force an error, see message on sample page
* save no error, see pipeline run kickoff

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -66,6 +66,7 @@ class Sample < ApplicationRecord
   UPLOAD_ERROR_S3_UPLOAD_FAILED = "S3_UPLOAD_FAILED".freeze
   UPLOAD_ERROR_LOCAL_UPLOAD_STALLED = "LOCAL_UPLOAD_STALLED".freeze
   UPLOAD_ERROR_LOCAL_UPLOAD_FAILED = "LOCAL_UPLOAD_FAILED".freeze
+  UPLOAD_ERROR_PIPELINE_KICKOFF = "PIPELINE_KICKOFF_FAILED".freeze
   DO_NOT_PROCESS = "DO_NOT_PROCESS".freeze
 
   TOTAL_READS_JSON = "total_reads.json".freeze
@@ -747,7 +748,13 @@ class Sample < ApplicationRecord
 
     pr.alignment_config = AlignmentConfig.find_by(name: alignment_config_name) if alignment_config_name
     pr.alignment_config ||= AlignmentConfig.find_by(name: AlignmentConfig::DEFAULT_NAME)
-    pr.save
+    pr.save!
+  rescue StandardError => err
+    LogUtil.log_err_and_airbrake("Error saving pipeline run: #{err.inspect}")
+    # This will cause a message a to be shown to the user on the sample page.
+    # See app/assets/src/components/utils/sample.js
+    self.upload_error = Sample::UPLOAD_ERROR_PIPELINE_KICKOFF
+    save!
   end
 
   def get_existing_metadatum(key)
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -61,7 +61,10 @@ class Sample < ApplicationRecord
   before_save :check_host_genome, :concatenate_input_parts, :check_status
   after_save :set_presigned_url_for_local_upload
 
-  # Constants for upload errors.
+  # Constants for upload_error field. NOTE: the name of this field is
+  # inaccurate. The intention of the field is to carry a message (a constant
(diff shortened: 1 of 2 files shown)


app/models/sample.rb
```diff
@@ -747,7 +748,13 @@ def kickoff_pipeline
 
     pr.alignment_config = AlignmentConfig.find_by(name: alignment_config_name) if alignment_config_name
     pr.alignment_config ||= AlignmentConfig.find_by(name: AlignmentConfig::DEFAULT_NAME)
-    pr.save
+    pr.save!
+  rescue StandardError => err
+    LogUtil.log_err_and_airbrake("Error saving pipeline run: #{err.inspect}")
+    # This will cause a message to be shown to the user on the sample page.
+    # See app/assets/src/components/utils/sample.js
```
tfrcarvalho: nit: Ok, to leave it... but normally  Ido not comment on behavior of the client based on the backend behavior since client can change behavior later.

app/models/sample.rb
```diff
@@ -747,7 +748,13 @@ def kickoff_pipeline
...
+    self.upload_error = Sample::UPLOAD_ERROR_PIPELINE_KICKOFF
+    save!
```
tfrcarvalho: This code is being run on a `before_save` callback and you call save again here. Seems like a dangerous pattern. What is the status of the sample at this point?

app/models/sample.rb
```diff
@@ -747,7 +748,13 @@ def kickoff_pipeline
 
     pr.alignment_config = AlignmentConfig.find_by(name: alignment_config_name) if alignment_config_name
     pr.alignment_config ||= AlignmentConfig.find_by(name: AlignmentConfig::DEFAULT_NAME)
-    pr.save
+    pr.save!
+  rescue StandardError => err
```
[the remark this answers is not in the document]
gregdingle: It could be for `ActiveRecord::RecordInvalid` only but I was thinking we would want to handle any error the same way. I don't see cases where we would want to fail silently.

app/models/sample.rb
gregdingle: That's a fair point. I will change to `This may cause a message`.

app/models/sample.rb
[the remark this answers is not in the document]
tfrcarvalho: You want fail silently anymore since you are using `!`. It is just how much noise you make... Typically, you use the more narrower to provide a better error message. In this case, I agree that it might not be very important as long as err.inspect provides all info.

app/models/sample.rb
gregdingle: Good catch! I will remove the save. It should still work.

app/models/sample.rb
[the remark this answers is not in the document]
gregdingle: I agree. By fail silently, I was referring to the lack of user facing message.

app/models/sample.rb
tfrcarvalho: As I think about it, this is not an upload error, right? We should find a better field and status for this.

gregdingle: True, this is not an upload error in the sense that the files have been copied successfully. The user message is "issue processing". 

FWIW, we already overload `upload_error` with `DO_NOT_PROCESS`. 

I don't think it is worth creating a new field for this type of error and I'm not sure where else it would go. 

I think it's a big improvement over showing "waiting for files", the current user facing message.

tfrcarvalho: I guess at this point it is more the name of the field that is inaccurate. And is true that it will be unlikely for this to happen in staging and prod.
Can you add a comment with argument on why we use this field (e.g. the do not process precedent)?

app/models/sample.rb
```diff
@@ -747,7 +748,14 @@ def kickoff_pipeline
...
+    LogUtil.log_err_and_airbrake("Error saving pipeline run: #{err.inspect}")
+    # This may cause a message to be shown to the user on the sample page.
+    # See app/assets/src/components/utils/sample.js
+    # Use low-level update_columns to avoid callbacks, because kickoff_pipeline
+    # may be running in a callback already.
+    update_columns(upload_error: Sample::UPLOAD_ERROR_PIPELINE_KICKOFF) # rubocop:disable SkipsModelValidations
```
tfrcarvalho: Can you add some sort of "do not try this at home" kind of message? Do not want this to be used as a precedent.

PR #3193, opened 2020-03-17
Upgrades ESLint, configures rules and adds fixes for default rules

tfrcarvalho:
# Description

* This PR fixes 2 security vulnerabilities on ESLint dependencies by upgrading ESLint to the last version
* Last version introduces multiple default rules for which we were not being consistent on our codebase.
* Some rules where configured to adapt to our practices (see `.eslintrc`.

# Notes

* Will do a follow up PR to update other packages that depend on the package versions reported in the alerts (the alert only mention the eslint package as being affected). 
* This PR makes sense even if it does fully remove it.

# Tests

* Changes are very simple for each component (mostly style changes), but since it affects a very high number of components most of the checks will be done during a *full regression test*.


tfrcarvalho:
> Looks like there are a lot of equality comparisons that are being replaced with the stricter version

Indeed, I definitely want to run this by QA and for a full regression testing. Having said, i was also considering breaking it apart.

PR #3194, opened 2020-03-17
[IDSEQ-2414] [Heatmap filtering] Add explanatory text for heatmap data

j-x-han:
# Description

Adding explanatory text of how many taxons passing selected filters are being displayed on the heatmap.
image

The tooltip displayed when the info icon is hovered over.
image

The text will be displayed beneath selected filter tags.
image

# Notes

Added a new info icon at @happyimadesignr 's request!

# Tests

* Loaded a heatmap and verified that the text would update with the appropriate number of shown taxa as filters were changed.
* Verified that the text does not display if the user does not have client side filtering enabled.

### DIFF
--- a/app/assets/src/components/ui/containers/ColumnHeaderTooltip.jsx
+++ b/app/assets/src/components/ui/containers/ColumnHeaderTooltip.jsx
@@ -14,6 +14,10 @@ class ColumnHeaderTooltip extends React.Component {
               <span className={cs.title}>{this.props.title}:</span>
             )}
             {this.props.content}
+            {this.props.list &&
+              this.props.list.map(item => {
+                return <li key={item}>{item}</li>;
+              })}
             {this.props.link && (
               <a
                 className={cs.link}
@@ -35,6 +39,7 @@ ColumnHeaderTooltip.propTypes = {
   size: PropTypes.string,
   title: PropTypes.string,
   link: PropTypes.string,
+  list: PropTypes.arrayOf(PropTypes.string),
 };
 
 ColumnHeaderTooltip.defaultProps = {
--- a/app/assets/src/components/ui/containers/ColumnHeaderTooltip.jsx
+++ b/app/assets/src/components/ui/containers/ColumnHeaderTooltip.jsx
@@ -5,25 +5,24 @@ import cs from "./column_header_tooltip.scss";
 
 class ColumnHeaderTooltip extends React.Component {
   render() {
+    const { title, content, list, link } = this.props;
     return (
       <BasicPopup
         {...this.props}
         content={
           <div className={cs.tooltip}>
-            {this.props.title && (
-              <span className={cs.title}>{this.props.title}:</span>
-            )}
-            {this.props.content}
-            {this.props.list &&
-              this.props.list.map(item => {
+            {title && <span className={cs.title}>{title}:</span>}
+            {content}
+            {list &&
+              list.map(item => {
                 return <li key={item}>{item}</li>;
               })}
-            {this.props.link && (
+            {link && (
               <a
                 className={cs.link}
                 target="_blank"
                 rel="noopener noreferrer"
-                href={this.props.link}
+                href={link}
               >
                 Learn more.
               </a>
--- a/app/assets/src/components/ui/icons/InfoSIcon.jsx
+++ b/app/assets/src/components/ui/icons/InfoSIcon.jsx
@@ -1,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const InfoSIcon = props => {
+  return (
+    <svg
+      className={props.className}
+      width="14px"
+      height="14px"
+      viewBox="0 0 14 14"
+      version="1.1"
+      xmlns="http://www.w3.org/2000/svg"
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+    >
+      <g id="Icons" stroke="none" strokeWidth="1" fillRule="evenodd">
+        <g id="icon-/-small-/-icon-s-info">
+          <path
+            d="M7,0 C10.8661,0 14,3.1339 14,7 C14,10.8661 10.8661,14 7,14 C3.1339,14 0,10.8661 0,7 C0,3.1339 3.1339,0 7,0 Z M7,6 C6.44771525,6 6,6.44771525 6,7 L6,7 L6,10.75 C6,11.3022847 6.44771525,11.75 7,11.75 C7.55228475,11.75 8,11.3022847 8,10.75 L8,10.75 L8,7 C8,6.44771525 7.55228475,6 7,6 Z M7,2.5 C6.67076544,2.5 6.34807913,2.6351598 6.11379505,2.87068219 C5.88605935,3.10034399 5.75,3.4318335 5.75,3.75819499 C5.75,4.09224851 5.88242143,4.41714485 6.11379505,4.65010322 C6.57617876,5.11601997 7.42418517,5.11601997 7.87711002,4.6537661 C8.11648719,4.41494714 8.25,4.09151593 8.25,3.75819499 C8.25,3.42817062 8.11685098,3.10583829 7.88438599,2.87397877 C7.64682778,2.6351598 7.32596043,2.5 7,2.5 Z"
+            id="Combined-Shape"
+          />
+        </g>
+      </g>
+    </svg>
+  );
+};
+
+InfoSIcon.propTypes = {
+  className: PropTypes.string,
+};
+
+export default InfoSIcon;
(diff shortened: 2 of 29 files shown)


app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -414,7 +416,38 @@ export default class SamplesHeatmapControls extends React.Component {
...
+        <ColumnHeaderTooltip
+          trigger={
+            <span>
+              <InfoSIcon className={cs.infoIcon} />
+            </span>
+          }
+          content="The data included in this heatmap 
+            was preselected based on the following conditions:"
+          list={[
+            "The top 1,000 unique taxa per sample",
```
gregdingle: how is the top 1,000 computed? I would assume as a user that it is ranked by the current selected metric, which I don't believe is our intention.

app/assets/src/components/ui/icons/InfoSIcon.jsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const InfoSIcon = props => {
```
gregdingle: what does the 'S' in InfoSIcon standfor?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -414,7 +416,38 @@ export default class SamplesHeatmapControls extends React.Component {
...
+      </span>
+    );
+  };
+
   render() {
+    const { selectedOptions, loading, displayFilterStats } = this.props;
+    const { thresholdFilters, categories, subcategories } = selectedOptions;
+    // Only display the filter tag row if relevant filters are selected,
+    // otherwise an empty row will be rendered.
+    let displayFilterTags =
```
gregdingle: prefer `const` over `let`

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -414,7 +416,38 @@ export default class SamplesHeatmapControls extends React.Component {
     return filterTags;
   };
 
+  renderFilterStatsInfo = () => {
+    let { filteredTaxaCount, totalTaxaCount } = this.props;
```
gregdingle: as above prefer `const`

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -414,7 +416,38 @@ export default class SamplesHeatmapControls extends React.Component {
...
+          trigger={
+            <span>
+              <InfoSIcon className={cs.infoIcon} />
+            </span>
+          }
+          content="The data included in this heatmap 
+            was preselected based on the following conditions:"
+          list={[
+            "The top 1,000 unique taxa per sample",
+            "Only taxa with at least 5 reads",
```
gregdingle: I recommend putting in a code comment in here that cites the constant in the code, so that in the future if we change the constant value and search for instances, we will find this text that needs to change as well.

Of course, sharing constants with rails would be even better, though outside the scope of this PR, and not something that is easy in our web code infrastructure.

app/assets/src/components/ui/containers/ColumnHeaderTooltip.jsx
```diff
@@ -14,6 +14,10 @@ class ColumnHeaderTooltip extends React.Component {
               <span className={cs.title}>{this.props.title}:</span>
             )}
             {this.props.content}
+            {this.props.list &&
```
tfrcarvalho: nit: Can we move loading the props taking advantage and refactor the other properties/

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -24,6 +24,8 @@ import Slider from "~ui/controls/Slider";
 import SequentialLegendVis from "~/components/visualizations/legends/SequentialLegendVis.jsx";
 import ThresholdFilterTag from "~/components/common/ThresholdFilterTag";
 import FilterTag from "~ui/controls/FilterTag";
+import ColumnHeaderTooltip from "~/components/ui/containers/ColumnHeaderTooltip";
+import InfoSIcon from "~/components/ui/icons/InfoSIcon";
```
tfrcarvalho: Dont import it direclty like this. User the index in `ui/icons/index.js`.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
tfrcarvalho: This 1000 value should probably come from the server, otherwise we need to keep updating this message.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
tfrcarvalho: We will review this value (and probably remove it) as part of the current Epic

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -414,7 +416,38 @@ export default class SamplesHeatmapControls extends React.Component {
...
+    );
+  };
+
   render() {
+    const { selectedOptions, loading, displayFilterStats } = this.props;
+    const { thresholdFilters, categories, subcategories } = selectedOptions;
+    // Only display the filter tag row if relevant filters are selected,
+    // otherwise an empty row will be rendered.
+    let displayFilterTags =
+      (thresholdFilters && thresholdFilters.length > 0) ||
```
tfrcarvalho: nit: `(thresholdFilters || []).length` would be shorter...

app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
```diff
@@ -32,6 +32,19 @@
         margin-right: 0;
       }
     }
+
+    .statsRow {
+      @include font-body-xs;
+      color: $medium-grey;
+
+      .infoIcon {
+        margin-left: 4px;
```
tfrcarvalho: Can we spaces variables?

app/assets/src/components/ui/icons/InfoSIcon.jsx
j-x-han: I believe it stands for "solid," since we have a similar icon that isn't filled in. @happyimadesignr ?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
j-x-han: It's based off the rPM! I updated the tooltip's copy, here's the new screenshot.
image

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
j-x-han: Updated the code so that both constants are passed in from the server, so we should only need to update it on the rails side now.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
j-x-han: Done

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
j-x-han: Done

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
j-x-han: Done!

app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
j-x-han: Good catch! Done

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
gregdingle: Nice! I'd shorten it to "top 1000 unique taxa per sample by rPM" but I will leave the final copy to you and @happyimadesignr .

app/assets/src/components/ui/icons/InfoSIcon.jsx
happyimadesignr: Oh hey, it's for small. Sorry it's not clear now b/c it's a part of the new design system work. We're creating different sized icons, the smaller one will be solid and larger ones will be line icons!

gregdingle: I'd prefer then to use the whole word "InfoSolidIcon"

j-x-han: Renamed it to InfoSmallIcon!

PR #3195, opened 2020-03-18
Log and notify if more than 100 samples uploaded at once

gregdingle:
# Description

After reading the announcement in biohub slack, I had the thought that it would be easy to add an airbrake alert when >100 samples are uploaded, as a second layer of warning.

```
"[Datadog] LargeBulkUploadEvent: 1 samples by non-admin user. See: http://localhost:3000/projects/855"
```

After merging this, I intend to setup a datadog alert similar to the LongRunningSample alert to raise the warning in our slack. 

# Notes

* The code assumes that all samples are in the same project and from the same user. Not sure how safe an assumption this is. @MarkAZhang ?
* Not sure whether `project_url` or DD project page is better to  link to

# Tests

`bundle exec rspec spec/requests/sample_request_spec.rb`

use byebug to inspect error message

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -1434,8 +1434,10 @@ class SamplesController < ApplicationController
     if samples.length < 100
       return
     end
-    msg = "[Datadog] LargeBulkUploadEvent: #{samples.length} samples by #{samples.first.user.role_name} in project #{samples.first.project.name}." \
-      "See: #{project_url(samples.first.project)}"
+    project = samples.first.project
+    uploader = samples.first.user
+    msg = "[Datadog] LargeBulkUploadEvent: #{samples.length} samples by #{uploader.role_name} in project #{project.name}." \
+      " See: #{project.status_url}"
     LogUtil.log_err_and_airbrake(msg)
   end
 end
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -1431,9 +1431,11 @@ class SamplesController < ApplicationController
 
   def warn_if_large_bulk_upload(samples)
     # 100 is the limit that was chosen during covid19 surge for biohub projects.
+    # At the time, we had successfully handled up to ~1000 uploads per day.
     if samples.length < 100
       return
     end
+    # assume that all samples are in the same project and from the same user
     project = samples.first.project
     uploader = samples.first.user
     msg = "[Datadog] LargeBulkUploadEvent: #{samples.length} samples by #{uploader.role_name} in project #{project.name}." \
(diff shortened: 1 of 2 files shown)


app/controllers/samples_controller.rb
```diff
@@ -1426,4 +1428,21 @@ def check_owner
...
+  def warn_if_large_bulk_upload(samples)
+    # 100 is the limit that was chosen during covid19 surge for biohub projects.
+    # At the time, we had successfully handled up to ~1000 uploads per day.
+    if samples.length < 100
+      return
+    end
+    # assume that all samples are in the same project and from the same user
+    project = samples.first.project
+    uploader = samples.first.user
+    msg = "[Datadog] LargeBulkUploadEvent: #{samples.length} samples by #{uploader.role_name} in project #{project.name}." \
```
gregdingle: @j-x-han is this sufficient for whitelisting in datadog?

app/controllers/samples_controller.rb
```diff
@@ -1426,4 +1428,19 @@ def check_owner
...
   end
+
+  def warn_if_large_bulk_upload(samples)
+    # 100 is the limit that was chosen during covid19 surge for biohub projects.
+    if samples.length < 100
+      return
+    end
+    project = samples.first.project
+    uploader = samples.first.user
+    msg = "[Datadog] LargeBulkUploadEvent: #{samples.length} samples by #{uploader.role_name} in project #{project.name}." \
```
tfrcarvalho: Lets not put project name here. The status should be enough.

gregdingle: For privacy? The project name I feel reveals a lot of information about the intent and priority of the uploads.

tfrcarvalho: Yes, for privacy. That field is not controlled by use, so it should be exposed to logs. 
Does not matter how useful it might be and more info is just a click away

gregdingle: Ok I'll remove it. FWIW, we have been logging the project name elsewhere.

tfrcarvalho: OK, I'll approve a PR that changes that :P

app/controllers/samples_controller.rb
```diff
@@ -1426,4 +1428,21 @@ def check_owner
...
+    end
+    # assume that all samples are in the same project and from the same user
+    project = samples.first.project
+    uploader = samples.first.user
+    msg = "[Datadog] LargeBulkUploadEvent: #{samples.length} samples by #{uploader.role_name}." \
+      " See: #{project.status_url}"
+    LogUtil.log_err_and_airbrake(msg)
+  rescue StandardError => e
+    # catch all errors because we don't want to ever block uploads
+    LogUtil.log_err_and_airbrake("warn_if_large_bulk_upload: #{e}")
```
tfrcarvalho: i think i forgot this on the other PR. I think you need to use `log_backtrace` too if you want the full backtrace of the exception.
We should definitely merge both functions.

PR #3196, opened 2020-03-19
[Release Fix] Disable project creation trigger properly (IDSEQ-2454)

lvreynoso:
# Description

A logic error in React allowed a user to trigger the project creation function and make an api call to the backend even when the button was supposed to be disabled, although the backend properly rejected it. This fix prevents the project creation function from being triggered unless all the requirements are met: a name, description, and either public or private access.

# Notes

I also cleaned up some of the code to help prevent future programmers from making similar mistakes.

# Tests

Same as in #3161

### DIFF
--- a/app/assets/src/components/common/ProjectCreationForm.jsx
+++ b/app/assets/src/components/common/ProjectCreationForm.jsx
@@ -30,7 +30,7 @@ class ProjectCreationForm extends React.Component {
     disableCreateButton: true,
   };
 
-  areCreationReqsUnmet(changed = {}) {
+  areCreationReqsMet(changed = {}) {
     const reqs = {
       name: this.state.name,
       accessLevel: this.state.accessLevel,
@@ -46,14 +46,14 @@ class ProjectCreationForm extends React.Component {
       reqs.accessLevel === ACCESS_LEVEL.noSelection ||
       reqs.description.length < 1
     ) {
-      return true;
+      return false;
     }
 
-    return false;
+    return true;
   }
 
   handleNameChange = name => {
-    const disableCreateButton = this.areCreationReqsUnmet({ name });
+    const disableCreateButton = !this.areCreationReqsMet({ name });
     this.setState({
       name,
       disableCreateButton,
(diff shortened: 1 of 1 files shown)


app/assets/src/components/common/ProjectCreationForm.jsx
```diff
@@ -97,10 +118,12 @@ class ProjectCreationForm extends React.Component {
...
           <div
             className={cs.sharingOption}
-            onClick={() => this.setState({ publicAccess: 1 })}
+            onClick={() =>
+              this.handleAccessLevelChange(ACCESS_LEVEL.publicAccess)
+            }
           >
             <RadioButton
-              selected={publicAccess === 1}
+              selected={accessLevel === 1}
```
tfrcarvalho: You should use the enum you defined above for the comparisons too.

app/assets/src/components/common/ProjectCreationForm.jsx
```diff
@@ -14,42 +14,61 @@ import { logAnalyticsEvent } from "~/api/analytics";
...
   handleNameChange = name => {
+    const disableCreateButton = name === "";
     this.setState({
       name,
+      disableCreateButton,
+    });
+  };
+
+  handleAccessLevelChange = accessLevel => {
+    const disableCreateButton = accessLevel === -1;
```
tfrcarvalho: You should use the enum you defined above for the comparisons too.

tfrcarvalho: Doesn't `disableCreateButton` always depend on the all conditions: accessLevel and disableCreateButton (not name length)? You cannot check them just one at a time, right?

lvreynoso: Yes, thank you for catching.

app/assets/src/components/common/ProjectCreationForm.jsx
```diff
@@ -14,42 +14,83 @@ import { logAnalyticsEvent } from "~/api/analytics";
...
     name: "",
-    publicAccess: -1, // No selection by default
+    accessLevel: ACCESS_LEVEL.noSelection, // No selection by default
     error: "",
     description: "",
     showInfo: false,
+    disableCreateButton: true,
   };
 
+  areCreationReqsUnmet(changed = {}) {
```
tfrcarvalho: nit: it is normally easier to process when we name it on the positive and then do `!creationReqsMet`...

PR #3197, opened 2020-03-19
IDSEQ-2364 - Async pipeline notification service

davidrissato:
# Description

IDSEQ-2364
This PR creates a service that will be invoked from pipeline step function via AWS -> Rails async interface (IDSEQ-2374/IDSEQ-2375/IDSEQ-2363) to notify successes or failures in pipeline stages.

This code is created in a way to support messages arriving out of order, and it will try to mimic the current {{PipelineRun#job_status}} behavior.

# Notes

We should consider refactoring and normalize `PipelineRun#job_status` field. It is a string field with an unclear domain, and there are multiple places in our source code using this field.
Examples:
https://github.com/chanzuckerberg/idseq-web/blob/b4bed182db116a77ad3a93898b18f983620fec54/app/models/pipeline_run.rb#L263-L274
We probably have more places using logic that parses this field content.
I created a new ticket for that task IDSEQ-2482.

# Tests

* Unit tested

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -1110,7 +1110,7 @@ class PipelineRun < ApplicationRecord
         check_and_log_long_run
       end
       self.job_status = "#{prs.step_number}.#{prs.name}-#{prs.job_status}"
-      self.job_status += "|#{STATUS_READY}" if report_ready?
+      self.job_status = "|#{STATUS_READY}" if report_ready?
     end
     save!
     enqueue_new_pipeline_run if automatic_restart
@@ -1154,9 +1154,9 @@ class PipelineRun < ApplicationRecord
   end
 
   def job_status_display
-    return "Pipeline Initializing" unless self.job_status
-    stage = self.job_status.to_s.split("-")[0].split(".")[1]
-    stage ? "Running #{stage}" : self.job_status
+    return "Pipeline Initializing" unless job_status
+    stage = job_status.to_s.split("-")[0].split(".")[1]
+    stage ? "Running #{stage}" : job_status
   end
 
   def run_time
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -1109,11 +1109,36 @@ class PipelineRun < ApplicationRecord
         # Check for long-running pipeline run and log/alert if needed
         check_and_log_long_run
       end
-      self.job_status = "#{prs.step_number}.#{prs.name}-#{prs.job_status}"
-      self.job_status = "|#{STATUS_READY}" if report_ready?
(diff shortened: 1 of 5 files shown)


app/models/pipeline_run.rb
```diff
@@ -1109,13 +1109,38 @@ def update_job_status
         # Check for long-running pipeline run and log/alert if needed
         check_and_log_long_run
```
tfrcarvalho: What happened to checks for long runs?

app/models/pipeline_run.rb
```diff
@@ -1109,13 +1109,38 @@ def update_job_status
...
       end
-      self.job_status = "#{prs.step_number}.#{prs.name}-#{prs.job_status}"
-      self.job_status += "|#{STATUS_READY}" if report_ready?
+      self.job_status = format_job_status_text(prs.step_number, prs.name, prs.job_status, report_ready?)
     end
     save!
     enqueue_new_pipeline_run if automatic_restart
```
tfrcarvalho: What happens to the automatic restart?

app/models/pipeline_run.rb
```diff
@@ -1109,13 +1109,38 @@ def update_job_status
...
+    if prs.nil?
+      # all stages succeeded
+      self.finalized = 1
+      self.job_status = STATUS_CHECKED
+    else
+      if prs.failed?
+        self.job_status = STATUS_FAILED
+        self.finalized = 1
+        self.known_user_error, self.error_message = check_for_user_error(prs)
+        automatic_restart = false
```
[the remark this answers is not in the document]
davidrissato: No, auto restarts should be handled by the step function. I kept this variable over there to to let it clear it should be false, but I just realized that it is better to change the invoked method signature and default `automatic_restart` to `false`, and remove it from here.

app/models/pipeline_run.rb
davidrissato: Removed on intent. It doesn't make sense to check for long runs here, since this method will be invoked asynchronously, therefore it means we had some progress. Checking for long runs need to be done through a polling mechanism, timeouts in the step function, or both. It is more related to ticket IDSEQ-2378.

app/models/pipeline_run.rb
davidrissato: Removed. This logic should be part of the step function.

app/models/pipeline_run.rb
tfrcarvalho: We could trigger timeout events and keep the logic asynchronous instead of having a separate poll mechanism.
However, since it seems we can poll from the DB just runs that have not been updated by using updated_at, i am not too worried.

PR #3199, opened 2020-03-19
Add admin upload options.

MarkAZhang:
# Description

Migrate admin options from legacy sample flow to new one.

These admin options are the last feature that needed to be ported over. Notably, in the legacy flow, you can only allow these admin options in the SINGLE sample upload flow, whereas in this updated flow, you can apply the options to multiple samples if needed.

This unblocks deleting the legacy sample flow as everything is ported over.

Screen Shot 2020-03-19 at 1 55 42 PM

Screen Shot 2020-03-19 at 1 55 46 PM

Screen Shot 2020-03-19 at 1 55 48 PM

Also changed the help icon to the SmallInfoIcon (Jenn's request). Verified that icon in coverage viz is still behaving correctly.

# Tests

* Verify that when the options are filled out, the corresponding fields on the sample are set correctly on all uploaded samples.
* Verify that when options are not filled out, the fields are set to nil.
* Verify that non-admin users don't see the options, and the fields are set to nil for their samples.

# QA Impact
* Normal users should not see admin options.

### DIFF
--- a/app/assets/src/api/upload.js
+++ b/app/assets/src/api/upload.js
@@ -45,6 +45,12 @@ export const bulkUploadLocalWithMetadata = async ({
       "do_not_process",
       "pipeline_execution_strategy",
       "use_taxon_whitelist",
+      "max_input_fragments",
+      "subsample",
+      "pipeline_branch",
+      "dag_vars",
+      "alignment_config_name",
+      "s3_preload_result_path",
     ]),
     samples
   );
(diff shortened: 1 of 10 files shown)


MarkAZhang:
@happyimadesignr I ended up not including the other checkbox options for now, as they are not technically "admin" options (for example, they are granted to QA testers).

happyimadesignr:
> @happyimadesignr I ended up not including the other checkbox options for now, as they are not technically "admin" options (for example, they are granted to QA testers).

happyimadesignr:
@MarkAZhang Sorry clicked the wrong button!!!

happyimadesignr:
@MarkAZhang sonuds good. Small things, the Admin Options title looks good, but could use a little space above to separate from the table. Also the question mark icon, can we use the info icon? Thanks =D

PR #3201, opened 2020-03-19
IDSEQ-2490: Fix new host bug on csv upload

gregdingle:
# Description

There was a subtle bug where the location check on CSV upload reset a bunch of state in the parent component. As a result, any new host that was created would not be recognized on hitting continue.

image

The bug does not reproduce locally because the location check errors out without the location API KEY. 

# Tests

* add location api key
* repro bug
* apply fix
* no more bug


app/assets/src/components/common/MetadataUpload.jsx
```diff
@@ -150,7 +150,13 @@ class MetadataUpload extends React.Component {
         metadata,
         this.getRequiredLocationMetadataType()
       );
-      onMetadataChange({ metadata: newMetadata });
+      // Here we set issues to null on the assumption that getCSVLocationMatches
+      // is called only when warnings and errors have been dealt with.
```
gregdingle: this is just making explicit what was implicit before

PR #3208, opened 2020-03-20
Add customizable subsampling parameters for specified projects.

MarkAZhang:
# Description

In order to serve some special projects, this adds the ability to configure the subsample and max_input_fragments options for samples uploaded to particular configurable projects. This will allow uploaded samples to run with modified subsampling parameters.

Example usage:
AppConfigHelper.set_app_config(AppConfig::SUBSAMPLE_WHITELIST_DEFAULT_SUBSAMPLE, 200000000)
          AppConfigHelper.set_app_config(AppConfig::SUBSAMPLE_WHITELIST_DEFAULT_MAX_INPUT_FRAGMENTS, 10000000)
          AppConfigHelper.set_json_app_config(AppConfig::SUBSAMPLE_WHITELIST_PROJECT_IDS, [123, 124])

# Tests

* Added tests.
* Verified manually that only samples uploaded to the specified projects are affected.
* Verified that app config if values are nil, samples still upload properly.

### DIFF
--- a/app/models/app_config.rb
+++ b/app/models/app_config.rb
@@ -30,4 +30,12 @@ class AppConfig < ApplicationRecord
   # List of launched features still guarded by a flag.
   # Use the features rake tasks for editing this key
   LAUNCHED_FEATURES = 'launched_features'.freeze
+  # The projects to apply the following two defaults to.
+  # Stored as a JSON array of project_ids.
+  # This is intended to be a temporary short-term mechanism to service critical projects with our partners.
+  SUBSAMPLE_WHITELIST_PROJECT_IDS = 'subsample_whitelist_project_ids'.freeze
+  # Default subsample for special biohub samples.
+  SUBSAMPLE_WHITELIST_DEFAULT_SUBSAMPLE = 'subsample_whitelist_default_subsample'.freeze
+  # Default max_input_fragments for special biohub samples.
+  SUBSAMPLE_WHITELIST_DEFAULT_MAX_INPUT_FRAGMENTS = 'subsample_whitelist_default_max_input_fragments'.freeze
 end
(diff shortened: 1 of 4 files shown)


app/models/app_config.rb
```diff
@@ -30,4 +30,11 @@ class AppConfig < ApplicationRecord
   # List of launched features still guarded by a flag.
   # Use the features rake tasks for editing this key
   LAUNCHED_FEATURES = 'launched_features'.freeze
+  # Default subsample for special biohub samples.
+  BIOHUB_DEFAULT_SUBSAMPLE = 'biohub_default_subsample'.freeze
+  # Default max_input_fragments for special biohub samples.
+  BIOHUB_DEFAULT_MAX_INPUT_FRAGMENTS = 'biohub_default_max_input_fragments'.freeze
+  # The special biohub projects to apply the above two defaults to.
+  # Stored as a JSON array of project_ids.
+  BIOHUB_PROJECT_IDS = 'biohub_project_ids'.freeze
```
gregdingle: We could generalize this concept to `WHITELIST_PROJECT_IDS`. It doesn't look like there is any biohub specific logic, only the name.

MarkAZhang: Good point. I've generalized it to SUBSAMPLE_WHITELIST_PROJECT_IDS.

MarkAZhang:
I had to remove a test (backgrounds_controller_test) because it suddenly started failing in Travis CI.

The endpoint that is being tested is not in use (and there is a bug with the endpoint). It's unknown why the test started failing but I was unable after several attempts to get the test to pass. The test appears to behave differently in Travis CI vs locally. (it returns a different http status, even when I force `format: html`). 

Will attempt to address in a subsequent PR.

PR #3213, opened 2020-03-24
[IDSEQ-2411] [Heatmap filtering] Refactor background filter to load less data

j-x-han:
# Description

Refactor the background filter when client-side filtering is enabled to simplify the SQL query so only background-relevant metrics are refetched and less data is sent from the client.

# Notes

Generated a few heatmaps to get some very rough time/space estimates:

With 12 samples (3394 taxa):
* Initial request takes approximately 15-20s, response size is 2.7MB
* Background filter change previously takes approximately 15-20s, response size is 1.5MB
* New background filter change takes approximately 7-10s, response size is 753KB

With 73 samples (12597 taxa):
* Initial request takes approximately 50-60s, response size is 5.2MB
* Background filter change previously takes approximately 35-40s, response size is 2.7MB
* New background filter  change takes approximately 30-37s, 2.6MB

Also made a change to `SamplesHeatmapControls` to ensure that the filters are disabled while data is being fetched from the server, since prematurely changing the filters will cancel the request and result in no data being displayed.

# Tests

* Loaded a heatmap and switched to several different backgrounds to check that the NT/NR zscores updated as expected.


app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -50,7 +50,7 @@ export default class SamplesHeatmapControls extends React.Component {
         value={this.props.selectedOptions.species}
         onChange={this.onTaxonLevelChange}
         label="Taxon Level"
-        disabled={!this.props.data}
+        disabled={this.props.loading}
```
tfrcarvalho: Is checking loading by itself better than checking if data is empty? Should we use both?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -456,6 +460,64 @@ class SamplesHeatmapView extends React.Component {
...
+    let newState = this.extractBackgroundMetrics(heatmapData);
+
+    this.updateHistoryState();
+    this.setState(newState, this.updateFilters);
+  }
+
+  extractBackgroundMetrics(rawData) {
+    let { sampleDetails, allTaxonDetails, allData } = this.state;
+
+    const backgroundMetrics = [
```
tfrcarvalho: Does this need to be defined every time? Should be a constant no?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -456,6 +460,64 @@ class SamplesHeatmapView extends React.Component {
...
+        background: this.state.selectedOptions.background,
+        heatmapTs: this.props.heatmapTs,
+      },
+      this.lastRequestToken.token
+    );
+  }
+
+  async fetchBackground() {
+    this.setState({ loading: true });
+    let heatmapData = await this.fetchBackgroundData();
```
tfrcarvalho: Should we call the data backgroundData? Or is it `fetchBackgroundData` that is deceiving?

app/helpers/heatmap_helper.rb
```diff
@@ -397,42 +417,67 @@ def self.fetch_samples_taxons_counts(samples, taxon_ids, parent_ids, background_
...
-        taxon_counts.tax_level  = taxon_summaries.tax_level       AND
-        taxon_counts.tax_id     = taxon_summaries.tax_id
-      WHERE
-        pipeline_run_id IN (#{pr_id_to_sample_id.keys.join(',')})
-        AND taxon_counts.genus_taxid != #{TaxonLineage::BLACKLIST_GENUS_ID}
-        AND taxon_counts.count_type IN ('NT', 'NR')
-        AND (taxon_counts.tax_id IN (#{taxon_ids.join(',')})
-         #{parent_ids_clause}
-         OR taxon_counts.genus_taxid IN (#{taxon_ids.join(',')}))").to_hash
+    sql_results =
```
tfrcarvalho: Can you put these queries into their own functions?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
j-x-han: I've added back `data` so that we check both since that seems to be a pattern elsewhere in the code as well.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: Moved to a constant!

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: Oops, renamed to `backgroundData`. `heatmapData` is carried over from an old variable I forgot to rename.

app/helpers/heatmap_helper.rb
j-x-han: Done!

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
tfrcarvalho: We should not just follow a pattern, but ask ourselves does it make sense for these filter to be enabled.
Data makes sense because they are useless without data, and we would have to consider that edge case, loading could make sense because we also do not want people changing filters while we are about to change data, which avoids plenty of other edge cases.

j-x-han: I think having both makes sense in order to cover the most number of edge cases!

I believe we **need** to disable filters when the heatmap is in a loading state, otherwise filters can be applied when the data is in the middle of being fetched or in an incomplete state. And as you mentioned, having the filters disabled when there's no data makes sense.

PR #3215, opened 2020-03-24
Move taxon whitelisting to taxon count method of idseq-dag

cdebourcy:
# Description
Change taxon whitelist mode so that it applies only to the compilation of taxon counts, not to candidate accessions for BLAST after GSNAP/RAPSearch2. This is necessary because we first need to find the global best match, and then whitelist only once the best match has been found, rather than restricting the search space upfront.

This change needs to apply in both stage 2 and stage 3 of the pipeline, both of which use the taxon counting method.

# Notes
Goes with https://github.com/chanzuckerberg/idseq-dag/pull/270

# Tests


cdebourcy:
> My only thought is that it would be nice, while you're touching this code, to add the `taxon_whitelist` and `taxon_blacklist` variables to the json templates. Right now, these variables are accepted in idseq_dag but there is no way to pass them in.
> 
> You might use the `dag_vars` variables. Something like `attr[:dag_vars_dict]["taxon_whitelist"]`. This would allow admins to pass in these values via the `dag_vars` admin option.

Thank you for the review and for pointing this out. Design-wise we're moving away from letting pipeline implementation details creep into the web app code, so I'll keep as is. I don't expect the whitelist needing to change much on a per-sample basis. If we do need to make updates to it down the line, uploading a new file to DEFAULT_WHITELIST_S3 or making a change to idseq-dag are, in my mind, easier ways of deploying the change :)

PR #3220, opened 2020-03-24
Remove legacy upload page.

MarkAZhang:
# Description

Now that the admin options are ported over, we can remove this legacy upload page.

# Notes

* Also removed numerous files that are no longer needed, including the old host genome icons and a js dependency for tooltips (react-tipsy).

# Tests

* Verified locally that local, remote, and basespace uploads still work.
* Verified that uploading via CLI still works for both single and bulk upload (by using an ngrok tunnel to upload to local server) 

# Tests for QA

* Verify that sample uploads using local files still works properly.


app/assets/src/api/index.js
```diff
@@ -4,7 +4,6 @@
 import axios from "axios";
 import axiosRetry from "axios-retry";
 
-import { cleanFilePath } from "~utils/sample";
```
tfrcarvalho: Should we delete cleanFilePath?

MarkAZhang: This function isn't actually deleted, just removed from this file.

tfrcarvalho: I know, but it is not used anymore! So we should delete it.

PR #3222, opened 2020-03-25
IDSEQ-2515 Disable report caching

cdebourcy:
# Description

# Notes

# Tests

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -347,6 +347,8 @@ class PipelineRun < ApplicationRecord
   end
 
   def ready_for_cache?
+    # TEMPORARILY DISABLE CACHING COMPLETELY - REMOVE EARLY RETURN TO REINSTATE
+    return false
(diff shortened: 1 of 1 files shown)


app/models/pipeline_run.rb
```diff
@@ -347,6 +347,8 @@ def results_finalized?
   end
 
   def ready_for_cache?
+    # TEMPORARILY DISABLE CACHING COMPLETELY - REMOVE EARLY RETURN TO REINSTATE
+    return false
```
gregdingle: I'd prefer to see this as an AppConfig so we can still debug in other envs

gregdingle: I'd be happy to take over this PR and do that.

gregdingle: See https://github.com/chanzuckerberg/idseq-web/pull/3223

cdebourcy: Ok thanks, please feel free to do that

PR #3229, opened 2020-03-27
[IDSEQ-2425] [IDSEQ-2413] Allow users to add taxon to heatmap

j-x-han:
# Description

When client-side filtering is enabled on the heatmap, allow users to manually select taxa to display on the heatmap. If the taxon they have selected does not pass the currently selected filters, notify the user that it will not be displayed.

This PR does not include the changes to automatically highlight and scroll to the position of a newly added taxon (IDSEQ-2426).

**New heatmap look**
* The line next to the `Add Metadata` trigger has been removed, and the metadata cells are now moved below the trigger rather than above.
* Added `Add Taxon` trigger.
image

**Opening the `Add Taxon` dropdown**
* All taxa that are currently being displayed will automatically be selected.
  * The user can deselect a taxon as an alternative method of removing it from the heatmap (as opposed to clicking on the `x` by the taxon label).
* The dropdown also indicates how many samples the taxon is found in.
image

**Notification when trying to add a taxon that doesn't pass filters**
* In this case, I tried to add a genus when the taxon level filter was set to species.
* `SamplesHeatmapView` will still keep track of the taxon. If the user changes filters, the taxon will be checked against the new set of filters: if it passes, then it will be displayed; if not, then the notification will be displayed again.
image

**Successfully adding a taxon that passes filters into the heatmap**
* Adding `Aminobacter aminovorans`
image
* Manually highlighting the newly added taxon 😄 
* Not pictured in this screenshot, but the `Showing {x} of {y} preselected taxa.` line below the filters will also be updated accordingly.
image

# Notes

Some style changes were made to the base heatmap component, notably the placement of the metadata cells, so I verified that these changes would not negatively impact other heatmaps.

**Taxon Heatmap without client-side filtering**
* No `Add Taxon` option
* `Add Metadata` line removed, metadata cells below the trigger.
image

**AMR Heatmap**
* `Add Metadata` line removed, metadata cells below the trigger.
image

# Tests
* Searched for and added taxon to the heatmap.
* Verified that added taxon that did not pass filters triggered a notification.
  * Changed filters and verified that the added taxon is now being displayed on the heatmap.
* Deselected a taxon and verified that it is no longer being displayed on the heatmap.


app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -119,6 +120,18 @@ class SamplesHeatmapView extends React.Component {
...
+      ),
+      // notifiedFilteredOutTaxa keeps track of the taxon ids for which
+      // we have already notified the user that they have manually added
+      // but did not pass filters.
+      // This is to ensure that we do not notify the user of ALL
+      // manually added taxa that failed filters every time they
+      // make a selection in the Add Taxon dropdown.
+      // This will be reset whenever filters change, so the user will be
+      // notified of which manually added taxa do not pass the new filters.
+      notifiedFilteredOutTaxa: new Set(),
```
tfrcarvalho: s/notifiedFilteredOutTaxa/notifiedFilteredOutTaxonIds/ ?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -362,7 +380,7 @@ class SamplesHeatmapView extends React.Component {
     // If there are failed/waiting samples selected, display a warning
     // to the user that they won't appear in the heatmap.
     if (sampleValidationInfo.invalidSampleNames.length > 0) {
-      this.showNotification();
+      this.showNotification("invalid samples");
```
tfrcarvalho: Define a constant for these strings... it easy to make typos on these.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -663,6 +705,9 @@ class SamplesHeatmapView extends React.Component {
     let topTaxIds = new Set();
     let topTaxonDetails = {};
     let filteredData = {};
+    // skippedTaxa includes taxa that passed filters but were explicitly
+    // removed by the user, which will still be counted toward the top taxa count.
```
tfrcarvalho: Why should we still count these towards the top?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -712,13 +783,52 @@ class SamplesHeatmapView extends React.Component {
...
   };
 
+  handleAddedTaxonChange = selectedTaxonIds => {
+    // selectedTaxonIds includes taxa that pass filters
+    // and the taxa manually added by the user.
+    let { taxonIds, addedTaxonIds, notifiedFilteredOutTaxa } = this.state;
+
+    // currentAddedTaxa is all the taxa manually added by the user.
+    let currentAddedTaxa = new Set([
+      ...[...selectedTaxonIds].filter(taxId => !taxonIds.includes(taxId)),
```
tfrcarvalho: Why do we need the `[... ]` here? 
I see that you repeat this pattern below...

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: This is a user experience/design choice so I'd ultimately like to check with @happyimadesignr  and Maria, but my logic was:

If I am a user and I have a heatmap that says its displaying 100 taxa and I remove one to clean up what I"m seeing, I expect it to now display 99 taxa. However, if we don't count the removed taxon towards the "top 100 taxa" count, the heatmap logic will just look for the next best taxon and silently add it to the heatmap so that it's still displaying the 100 taxa, which could be confusing to the user.

I could also see the other perspective where the user wants to see the top N taxa but explicitly not including specific taxa. Not sure which is more intuitive, so would love to discuss more!

tfrcarvalho: For me the latter makes more sense. Select the top 10 given all the filters (including tax id removed).
Imagine the case that the user is not interested in the top 10 and removes them all... It will display an empty heatmap.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: The [spread operator](https://exploringjs.com/impatient-js/ch_sets.html#missing-set-operations) is used in set operations in Javascript. (I just learned this, so if there's another way to do it please let me know! 😅 )

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: Done!

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: Done!

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: That makes sense to me! I'll change it.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
tfrcarvalho: I see... this the most compact way. You probably do not need the leading `...`. So, it would be `new Set([...selectedTaxonIds].filter(taxId => !taxonIds.includes(taxId))`.
But also, have in mind that for each includes you have a N operation. So it could be better to loop through the taxonIds array instead and remove them from a set.

tfrcarvalho: We could rewrite this using functions named merge, difference and intersection!

j-x-han: Good point, rewrote to use those functions 😄

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -712,13 +778,54 @@ class SamplesHeatmapView extends React.Component {
     window.history.replaceState("", "", this.getUrlForCurrentParams());
   };
 
+  handleAddedTaxonChange = selectedTaxonIds => {
+    // selectedTaxonIds includes taxa that pass filters
+    // and the taxa manually added by the user.
+    let { taxonIds, addedTaxonIds, notifiedFilteredOutTaxonIds } = this.state;
```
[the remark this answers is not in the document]
j-x-han: Done

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -712,13 +778,54 @@ class SamplesHeatmapView extends React.Component {
...
+
+    // currentAddedTaxa is all the taxa manually added by the user.
+    let currentAddedTaxa = new Set([
```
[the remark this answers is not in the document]
j-x-han: Done

PR #3231, opened 2020-03-30
IDSEQ-2514: Update DCR tooltip

gregdingle:
# Description

After cdhitdup change, we no longer remove dupes.

## Before
image

## After

image

# Tests

hover over tooltip, see changed text


happyimadesignr:
Hey thanks for tagging, I think the last sentence is helpful, but we don't provide that explanation in any of the other metrics. The other tooltips explain what the metric is and you click through to learn more. Would be nice to be consistent, either all have it or all don't. So in this case, maybe follow the current pattern? What do you think?

katrinakalantar:
Good catch @happyimadesignr - given that logic, it seems sufficient to remove the last sentence for consistency.

tfrcarvalho:
LGTM from eng perspective. @katrinakalantar or @happyimadesignr can you give the final approval?

PR #3238, opened 2020-04-01
[IDSEQ-2426] Trigger hover effect when taxon is added to heatmap

j-x-han:
# Description

When a user adds a taxon to the heatmap, scroll to and temporarily highlight the newly added row.

ezgif com-video-to-gif

Since a user can potentially add multiple taxa, the effect is only triggered for the most recently added taxon when the user closes the `Add Taxon` dropdown menu.

# Tests

* Added taxa and verified that once the dropdown menu was closed, the heatmap would automatically scroll to and highlight the newest added taxon.

### DIFF
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
@@ -557,6 +557,7 @@ class SamplesHeatmapView extends React.Component {
       allData,
       addedTaxonIds,
       notifiedFilteredOutTaxonIds,
+      newestTaxonId,
     } = this.state;
     let taxonDetails = {};
     let taxonIds = new Set();
@@ -581,6 +582,7 @@ class SamplesHeatmapView extends React.Component {
           ) {
             this.showNotification(NOTIFICATION_TYPES.taxaFilteredOut, taxon);
             notifiedFilteredOutTaxonIds.add(taxon["id"]);
+            newestTaxonId = null;
           }
         }
       });
@@ -604,6 +606,7 @@ class SamplesHeatmapView extends React.Component {
       loading: false,
       data: filteredData,
       notifiedFilteredOutTaxonIds,
+      newestTaxonId,
     });
   }
 
@@ -790,6 +793,7 @@ class SamplesHeatmapView extends React.Component {
       ...selectedTaxonIds,
     ]);
     const currentAddedTaxa = new Set(newlyAddedTaxa, previouslyAddedTaxa);
+    const newestTaxonId = newlyAddedTaxa[newlyAddedTaxa.length - 1];
 
     // Update notifiedFilteredOutTaxonIds to remove taxa that were unselected.
     const currentFilteredOutTaxonIds = new Set(
@@ -808,6 +812,7 @@ class SamplesHeatmapView extends React.Component {
       {
         addedTaxonIds: currentAddedTaxa,
         notifiedFilteredOutTaxonIds: currentFilteredOutTaxonIds,
+        newestTaxonId: newestTaxonId,
       },
       this.updateFilters
     );
@@ -1081,6 +1086,7 @@ class SamplesHeatmapView extends React.Component {
           sampleDetails={this.state.sampleDetails}
           scale={SCALE_OPTIONS[scaleIndex][1]}
           selectedTaxa={this.state.addedTaxonIds}
+          newestTaxonId={this.state.newestTaxonId}
           allTaxonIds={this.state.allTaxonIds}
           taxonIds={Array.from(
             new Set(this.state.taxonIds, this.state.addedTaxonIds)
--- a/app/assets/src/components/visualizations/heatmap/Heatmap.js
+++ b/app/assets/src/components/visualizations/heatmap/Heatmap.js
@@ -198,9 +198,6 @@ export default class Heatmap {
     for (let i = 0; i < this.rowLabels.length; i++) {
       let label = this.rowLabels[i].label;
       this.rowLabelsWidth = Math.max(this.rowLabelsWidth, labelWidth(label));
-      if (this.rowLabels[i].focus) {
-        this.focus = this.rowLabels[i];
-      }
     }
 
     for (let i = 0; i < this.options.columnMetadata.length; i++) {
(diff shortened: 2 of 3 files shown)


tfrcarvalho:
Look nice, should we highlight the label too?

j-x-han:
@tfrcarvalho Good catch, just updated to highlight the label as well!

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -557,6 +557,7 @@ class SamplesHeatmapView extends React.Component {
       allData,
       addedTaxonIds,
       notifiedFilteredOutTaxonIds,
+      newestTaxonId,
```
tfrcarvalho, line 560: What happens if the user adds more than one taxa? Isn't that possible?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -526,7 +538,39 @@ export default class Heatmap {
     d3.event.stopPropagation();
   }
 
-  pan(deltaX, deltaY) {
+  scrollToFocus() {
```
tfrcarvalho, line 541: If you accept my suggestion above this would be called `scrollToRow` or something.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -511,6 +517,12 @@ export default class Heatmap {
 
     // Set up scrolling behavior.
     this.g.on("wheel.zoom", this.scroll.bind(this));
+
+    // If there's a specific row we want to focus on, auto-scroll the heatmap
+    // so the row is centered on screen.
+    if (this.focus) {
+      this.scrollToFocus();
```
tfrcarvalho, line 524: Could this be called from the SamplesHeatmapViz instead?
This feature seems very specific, and i would prefer to just enable the mechanism to scroll to a particular row (by index or by name).
That way heatmap would not have to worry about triggering/disabling this.
SamplesHeatmapViz would be responsible for keeping track of the focus row, choosing which one to scroll to, and asking that change from the heatmap.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -597,30 +647,65 @@ export default class Heatmap {
...
-      "y",
-      y - this.columnLabelsHeight - this.options.marginTop
-    );
-    // Placing the white rectangle to hide column labels in the top left corner.
-    this.metadataLabelsBackground.attr(
-      "y",
-      y - this.columnLabelsHeight - this.options.marginTop
-    );
+  placeColumnLabelAndMetadataContainers(y, transition = false) {
+    if (transition) {
```
tfrcarvalho, line 651: Sorry, i am having trouble understanding what changes in column label and metadata positioning.  Can you explain the difference?

j-x-han: There were no changes in the column label and metadata positioning, I just added a transition so that the heatmap will smoothly scroll to the row instead of jumping to it. This normally isn't necessary when the user manually scrolls since the the delta from scrolling is much more incremental.

app/assets/src/components/visualizations/heatmap/Heatmap.js
j-x-han: Done!

app/assets/src/components/visualizations/heatmap/Heatmap.js
j-x-han: That makes a lot of sense, thanks for the suggestion! Just made the change.

app/assets/src/components/visualizations/heatmap/Heatmap.js
tfrcarvalho: Ok, would just set duration to 0 in case of no transition work?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: The behavior I currently have implemented is that it will only scroll to the taxon once the Add Taxon dropdown is closed, so it will scroll to the most recently added one (the last one selected before the dropdown was closed).

Alternatively, we could have the heatmap scroll to the taxon as soon as it's selected, but I feel it could be a little jarring if the heatmap is constantly scrolling if the user is adding multiple taxa. This is something I'll check with @happyimadesignr !

tfrcarvalho: Yeah, makes sense. The first one given the order on the heatmap might make more sense. But definitely check with Jenn.

app/assets/src/components/visualizations/heatmap/Heatmap.js
j-x-han: Ah yes that works as well. Just changed!

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -548,7 +584,10 @@ export default class Heatmap {
       this.options.marginTop,
       Math.max(deltaY + gCurrentTranslate[1], yScrollMax)
     );
-    this.g.attr("transform", `translate(${[dx, dy]})`);
+    this.g
+      .transition()
+      .duration(transition ? 750 : 0)
```
tfrcarvalho, line 589: nit: Move 750 to option above.

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -597,30 +635,39 @@ export default class Heatmap {
...
-  placeColumnLabelAndMetadataContainers(y) {
-    this.gColumnLabels.attr(
-      "transform",
-      `translate(${this.rowLabelsWidth},${y})`
-    );
-    this.gColumnMetadata.attr("transform", `translate(0, ${y})`);
+  placeColumnLabelAndMetadataContainers(y, transition = false) {
+    this.gColumnLabels
+      .transition()
+      .duration(transition ? 750 : 0)
```
tfrcarvalho, line 641: Replace number with option.

app/assets/src/components/visualizations/heatmap/Heatmap.js
j-x-han: I realized there's an existing `transitionDuration` option, so I'll use that instead.

app/assets/src/components/visualizations/heatmap/Heatmap.js
j-x-han: Done!

PR #3240, opened 2020-04-01
IDSEQ-2517: Fix contig count bug in CSV download of report

gregdingle:
# Description

This fixes an issue that was revealed by #3169. If there is no `params[:min_contig_reads]`, the contigs counts would be stripped. 

# Notes

* This is where it would be great to have static typing or at least use asserts. `min_contig_reads` should never be `nil`. 

# Tests

* new unit test
* download csv locally and see new values in contig cols

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -591,7 +591,7 @@ class SamplesController < ApplicationController
   def report_csv
     pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
     background_id = get_background_id(@sample, params[:background])
-    min_contig_reads = params[:min_contig_reads]
+    min_contig_reads = params[:min_contig_reads] || PipelineRun::MIN_CONTIG_READS
     @report_csv = PipelineReportService.call(pipeline_run, background_id, csv: true, min_contig_reads: min_contig_reads)
     send_data @report_csv, filename: @sample.name + '_report.csv'
   end
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -431,12 +431,6 @@ class PipelineRun < ApplicationRecord
   end
 
   def align_viz_available?
-    if Rails.env == "test"
-      # Otherwise, tests will raise a WebMock::NetConnectNotAllowedError.
-      # TODO: (gdingle): see IDSEQ-2602
-      return true
-    end
-
     # TODO(tiago): we should not have to access aws. see IDSEQ-2602.
     align_summary_file = "#{alignment_viz_output_s3_path}.summary"
     return align_summary_file && get_s3_file(align_summary_file) ? true : false
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -431,7 +431,13 @@ class PipelineRun < ApplicationRecord
   end
 
   def align_viz_available?
-    # TODO(tiago): we should not have to access aws
+    if Rails.env == "test"
+      # Otherwise, tests will raise a WebMock::NetConnectNotAllowedError.
+      # TODO: (gdingle): see IDSEQ-2602
+      return true
+    end
+
+    # TODO(tiago): we should not have to access aws. see IDSEQ-2602.
     align_summary_file = "#{alignment_viz_output_s3_path}.summary"
     return align_summary_file && get_s3_file(align_summary_file) ? true : false
   end
--- a/app/services/pipeline_report_service.rb
+++ b/app/services/pipeline_report_service.rb
@@ -47,8 +47,6 @@ class PipelineReportService
   Z_SCORE_WHEN_ABSENT_FROM_SAMPLE = -100
 
   DEFAULT_SORT_PARAM = :agg_score
-  # See also PipelineRun::MIN_CONTIG_READS
-  DEFAULT_MIN_CONTIG_READS = 0
 
   CATEGORIES = {
     2 => "bacteria",
@@ -91,11 +89,11 @@ class PipelineReportService
     "species_tax_ids",
   ].freeze
 
-  def initialize(pipeline_run, background_id, csv: false, min_contig_reads: DEFAULT_MIN_CONTIG_READS, parallel: true)
+  def initialize(pipeline_run, background_id, csv: false, min_contig_reads: PipelineRun::MIN_CONTIG_READS, parallel: true)
     @pipeline_run = pipeline_run
     @background_id = background_id
     @csv = csv
-    @min_contig_reads = min_contig_reads || DEFAULT_MIN_CONTIG_READS
+    @min_contig_reads = min_contig_reads || PipelineRun::MIN_CONTIG_READS
     @parallel = parallel
   end
 
--- a/app/services/pipeline_report_service.rb
+++ b/app/services/pipeline_report_service.rb
@@ -95,8 +95,7 @@ class PipelineReportService
     @pipeline_run = pipeline_run
     @background_id = background_id
     @csv = csv
-    raise 'bad min_contig_reads' unless min_contig_reads
-    @min_contig_reads = min_contig_reads
+    @min_contig_reads = min_contig_reads || DEFAULT_MIN_CONTIG_READS
     @parallel = parallel
   end
 
--- a/app/services/pipeline_report_service.rb
+++ b/app/services/pipeline_report_service.rb
@@ -95,6 +95,7 @@ class PipelineReportService
     @pipeline_run = pipeline_run
     @background_id = background_id
     @csv = csv
+    raise 'bad min_contig_reads' unless min_contig_reads
     @min_contig_reads = min_contig_reads
     @parallel = parallel
   end
--- a/spec/services/pipeline_report_service_spec.rb
+++ b/spec/services/pipeline_report_service_spec.rb
@@ -3,6 +3,8 @@ require 'json'
 
 RSpec.describe PipelineReportService, type: :service do
   context "converted report test for species taxid 573" do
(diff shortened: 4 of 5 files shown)


app/models/pipeline_run.rb
```diff
@@ -431,7 +431,13 @@ def taxon_byte_ranges_available?
   end
 
   def align_viz_available?
-    # TODO(tiago): we should not have to access aws
+    if Rails.env == "test"
```
gregdingle, line 434: not sure if the medicine is better than the disease here

spec/services/pipeline_report_service_spec.rb
```diff
@@ -137,6 +161,11 @@
 
       expect(JSON.parse(@report)["counts"]["2"]["570"]).to include_json(genus_result)
     end
+
+    it "should show correct values in CSV" do
+      csv_report = PipelineReportService.call(@pipeline_run, @background.id, csv: true)
+      expect(csv_report).to eq("tax_id,tax_level,genus_tax_id,name,common_name,category,agg_score,max_z_score,nt_z_score,nt_rpm,nt_count,nt_contigs,nt_contig_r,nt_percent_identity,nt_alignment_length,nt_e_value,nt_bg_mean,nt_bg_stdev,nr_z_score,nr_rpm,nr_count,nr_contigs,nr_contig_r,nr_percent_identity,nr_alignment_length,nr_e_value,nr_bg_mean,nr_bg_stdev,species_tax_ids\n570,2,570,Klebsiella,,bacteria,2428411764.7058825,99,99,193404.63458110517,217,6,594,99.7014,149.424,89.5822,18.3311,64.2056,99,77540.10695187165,87,6,594,97.9598,46.4253,16.9874,35.0207,238.639,[573]\n573,1,570,Klebsiella pneumoniae,,bacteria,2428411764.7058825,99,99,186274.50980392157,209,2,198,99.6995,149.402,89.5641,9.35068,26.4471,99,61497.326203208555,69,2,198,97.8565,46.3623,16.9101,29.9171,236.332,\n")
```
gregdingle, line 167: not sure how better to show long string

app/services/pipeline_report_service.rb
```diff
@@ -95,6 +95,7 @@ def initialize(pipeline_run, background_id, csv: false, min_contig_reads: DEFAUL
     @pipeline_run = pipeline_run
     @background_id = background_id
     @csv = csv
+    raise 'bad min_contig_reads' unless min_contig_reads
```
gregdingle, line 98: I use `assert` commonly when I code in python... not sure the convention in ruby, or what we what to follow

app/models/pipeline_run.rb
tfrcarvalho: Do not do this. You should just mock the S3 response... YOu can use Webmock or the AWS SDK stubs.

app/services/pipeline_report_service.rb
tfrcarvalho: Fine to raise an exception, but I prefer to define a custom error whenever possible so you can catch later on.
However, in this case, you might want to consider just enforcing `DEFAULT_MIN_CONTIG_READS`, with `@min_contig_reads = min_contig_reads || DEFAULT_MIN_CONTIG_READS`.

In any case, make sure the code outside properly handles the exception if you raise (my preference is just to make it work by enforcing the default value).

spec/services/pipeline_report_service_spec.rb
tfrcarvalho: A slight improvement is to define the a constant above.
Side note: A better (also more code) test would be to split into multiple check. Parse the csv and expect correct CSV, expect containing each header (loop), expect contain correct values, etc.

app/models/pipeline_run.rb
```diff
@@ -431,7 +431,13 @@ def taxon_byte_ranges_available?
...
+      # Otherwise, tests will raise a WebMock::NetConnectNotAllowedError.
+      # TODO: (gdingle): see IDSEQ-2602
+      return true
+    end
+
+    # TODO(tiago): we should not have to access aws. see IDSEQ-2602.
```
tfrcarvalho, line 440: Good catch! Thank you for opening the ticket. Definitely, we need to remove this... I wonder how much time is adding... but should be non-negligible.

app/controllers/samples_controller.rb
```diff
@@ -591,7 +591,7 @@ def bulk_upload_with_metadata
   def report_csv
     pipeline_run = select_pipeline_run(@sample, params[:pipeline_version])
     background_id = get_background_id(@sample, params[:background])
-    min_contig_reads = params[:min_contig_reads]
+    min_contig_reads = params[:min_contig_reads] || PipelineRun::MIN_CONTIG_READS
```
MarkAZhang, line 594: Hey Greg, I notice that the `PipelineRunService.call` in `models/bulk_downloads.rb` to generate the Sample Taxon Report doesn't pass in anything for min_contig_reads. So that call would use a value of 0 while this call would use a value of 4.

Since we only load contigs into our database that have at least four contig reads (see `db_load_contigs` which uses `PipelineRun::MIN_CONTIG_READS`) any value from 0 to 4 would produce the same result. However, it still seems good to keep it consistent.

Perhaps we can make the default in PipelineReportService 4 instead of 0? And add a comment explaining that contigs less than 4 are filtered out anyways so people don't get confused. Curious what @tfrcarvalho thinks about changing the default.

app/models/pipeline_run.rb
gregdingle: I'll just remove it. 

The issue is only that existing mocking somehow does not work when a single case is called
```
rspec spec/services/pipeline_report_service_spec.rb:165
# does not work
spec/services/pipeline_report_service_spec.rb
# works
```

The error is 
```
     WebMock::NetConnectNotAllowedError:
       Real HTTP connections are disabled. Unregistered request: GET https://idseq-samples-development.s3.us-west-2.amazonaws.com/samples/43/43/postprocess/1.0/subsample_1000000/align_viz.summary with headers ...

       You can stub this request with the following snippet:
     
       stub_request(:get, "https://idseq-samples-development.s3.us-west-2.amazonaws.com/samples/43/43/postprocess/1.0/subsample_1000000/align_viz.summary").
         with(
```

However the suggested stub did not work for me either. Any insight?

By the way, the pattern of returning early for test env exists a couple other places. 

```
  def taxon_search(query, tax_levels = TaxonCount::NAME_2_LEVEL.keys, filters = {})
...

  def list_outputs(s3_path, display_prefix = 1, delimiter = "/")
    return TEST_RESULT_FOLDER if Rails.env == "test"
```

app/models/pipeline_run.rb
gregdingle: I don't know much about this feature or else I would take a stab at fixing it here.

app/services/pipeline_report_service.rb
gregdingle: > However, in this case, you might want to consider just enforcing DEFAULT_MIN_CONTIG_READS

Good idea. It's too late to try to do things "properly" . Better to be safe with existing code.

app/controllers/samples_controller.rb
gregdingle: Thanks for pointing that out Mark. We definitely don't want bulk downloads to return different results than single reports.

I propose going further and eliminating the constant from pipeline report service. It does not appear to add any value. 

> Since we only load contigs into our database that have at least four contig reads

As a matter of fact, we filter out contigs less than 4 in the pipeline also. See

```
                    if unique_count < MIN_CONTIG_SIZE:
                        del contig_counts[contig]
```
And in another place in the pipeline, the value appears to come from idseq-web. 
```
        min_contig_size = self.additional_attributes.get("min_contig_size", MIN_CONTIG_SIZE)
```

As a side note, when we add constants that should be in sync across codebases, it is good to add a comment to that effect in the code.

tfrcarvalho: * I don't think web needs to be in sync with pipeline, in this case. More and more we should detach web from pipeline. If the pipeline makes a decision of not publishing contigs smaller than 4, so be it. Exceptions *could* be any parameter that is configurable through the web -> pipeline interface.

* The default should however be in sync throughout the web. In that case, i would prefer `0` which would publish any values that the pipeline decides 
publish (even if the value changes later).

app/models/pipeline_run.rb
tfrcarvalho: Yeah, I would have to see where it is used, too. I think the ticket makes sense and we schedule it on the next sprint.
One question on the ticket: would you call this a bug?

app/models/pipeline_run.rb
tfrcarvalho: Yeah, it is being mocked somewhere else... which is not great, since it is test pollution. We should eventually debug these cases, because it can have unexpected event on tests.
I prefer stubbing the services directly when possible (vs. using generic webmock) like in https://github.com/chanzuckerberg/idseq-web/blob/master/spec/services/sfn_pipeline_dispatch_service_spec.rb#L19

app/controllers/samples_controller.rb
gregdingle: I agree, I don't think these constants should be in sync. I should have written "constants that should be **informed by** constants in other codebases".

So how we the values in idseq-web? PipelineReportService:: DEFAULT_MIN_CONTIG_READS and PipelineRun::MIN_CONTIG_READS?

Note that the method endpoint `summary_contig_counts` already has the expression `min_contig_reads = params[:min_contig_reads] || PipelineRun::MIN_CONTIG_READS` which is why I added it to `report_csv`.

app/models/pipeline_run.rb
gregdingle: I would call this tech debt but not a bug because it does not have a known user impact. Tech debt I think we are categorizing as stories.

app/controllers/samples_controller.rb
gregdingle: I replaced `DEFAULT_MIN_CONTIG_READS` with `PipelineRun::MIN_CONTIG_READS`.

gregdingle: `PipelineRun::DEFAULT_MIN_CONTIG_READS` doesn't make sense because that constant is also used in `PipelineRun` in loading the contigs. While it may be good to remove that, that would be another PR.

PR #3242, opened 2020-04-02
IDSEQ-2603: Fix contig threshold filters

gregdingle:
# Description

The #3157 and #3169 did not account for contig filters. This fixes the broken filters. 

image

# Tests

* test contig_r, contig filter both nt and nr


app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -345,24 +345,6 @@ export default class SampleViewV2 extends React.Component {
   getTaxonMetricValue = (row, metric) => {
     let parsedMetric = metric.split(":");
     let parsedValue = get(parsedMetric, row);
-
-    // Contigs/contig_reads are stored in the format {length of contig: count},
-    // so some extra processing is needed to extract the value.
-    if (parsedMetric.includes("contigs") && parsedValue) {
```
gregdingle, line 351: I'm not sure whether there is more special casing of contigs elsewhere or how to search for it. Please advise @tfrcarvalho .

tfrcarvalho: Not sure either. But I cannot find any other reference.

PR #3246, opened 2020-04-03
Add to pipeline_runs a link to batch job page and batch job exit reason

gregdingle:
# Description

Makes it easier to debug failures, especially OOM errors.

image

# Notes

* I think AWS keeps the batch job status only for a limited time

# Tests

* load page, see status, try link


app/models/pipeline_run_stage.rb
```diff
@@ -390,4 +390,26 @@ def experimental_command
     # Dispatch job
     aegea_batch_submit_command(batch_command, stage_name: DAG_NAME_EXPERIMENTAL, sample_id: sample.id)
   end
+
+  # Gets the URL to the AWS console page of the batch job for display on
+  # admin-only pages.
+  def batch_job_status_url
```
tfrcarvalho, line 396: This is definitely a perfect case to use view helpers (the true purpose of view). Should not be model methods.

gregdingle: Ha, I don't even know what a view helper is. I'll look it up. 

But I still think there is some value for the console or other future callers other than the web page.

tfrcarvalho: Sure, lets do it.

PR #3252, opened 2020-04-08
Add data discovery empty states

tfrcarvalho:
# Description

* Add states for first time user and user with no data (projects, samples and visualizations)

# Notes

Screen Shot 2020-04-07 at 9 34 44 PM
Screen Shot 2020-04-07 at 9 35 03 PM
Screen Shot 2020-04-07 at 9 35 11 PM
Screen Shot 2020-04-07 at 9 35 16 PM

# Tests

* Make sure to start with an empty account and delete key `DiscoveryViewSeenBefore` key from you local storage(in `Developer Tools > Application > LocalStorage`)
* Open `My Data` page and see the modal
* Test:
	* Upload data link
	* Public data link
	* Close button
* Should not see the modal anymore (even on refresh)
* In `My Data`:
	* Navigate to projects tab and see the new no data banner for project an test the link
	* Navigate to samples tab and see the new no data banner for samples, test link
	* Navigate to visualizations tab and see the new no data banner for visualizations, test link (currently redirects to info on saving heatmaps)


app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -1012,22 +1038,121 @@ class DiscoveryView extends React.Component {
...
+
+    if (emptyStateModalOpen) {
+      localStorage.setItem("DiscoveryViewSeenBefore", "1");
+    }
+
+    switch (currentTab) {
+      case "projects":
+        if (userDataCounts.projectCount === 0) {
+          return emptyStateModalOpen ? (
+            <EmptyStateModal onClose={this.handleEmptyStateModalClose} />
```
gregdingle, line 1061: I'd prefer `EmptyStateModal` be named `WelcomeFirstTimeUserModal` to be more clear about when it shows up but not a big deal.

app/assets/src/components/ui/controls/ExternalLink.jsx
```diff
@@ -1,45 +1,8 @@
...
-  // We intentionally don't have an onClick prop, because we don't want to encourage arbitrary onClick handlers,
-  // since there is already an on-click behavior (following the link href). logAnalyticsEvent is an exception.
-  analyticsEventName: PropTypes.string,
-  analyticsEventData: PropTypes.object,
-};
-
-ExternalLink.defaultProps = {
-  analyticsEventData: {},
+const ExternalLink = props => {
+  return <Link external {...props} />;
```
gregdingle, line 5: nice refactor! finally our own `Link` component. this is something @happyimadesignr and I discussed a while ago. maybe it should go in the design system?

app/assets/src/components/ui/icons/BacteriaCultureIcon.jsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const BacteriaCultureIcon = ({ className }) => {
```
gregdingle, line 4: should we be naming these now `IconCultureBacteria` by the new naming schema?

app/assets/src/components/views/discovery/DiscoverySidebar/DiscoverySidebar.jsx
```diff
@@ -254,12 +254,20 @@ export default class DiscoverySidebar extends React.Component {
...
+      noDataAvailable,
+      project,
+    } = this.props;
     if (!loading && !this.hasData()) {
       return (
         <div className={cx(className, cs.sidebar)}>
           <div className={cs.noData}>
-            Try another search to see summary info.
+            {noDataAvailable
+              ? "Add data to view summary info."
```
gregdingle, line 269: 👍

app/assets/src/components/views/discovery/InfoBanner.jsx
```diff
@@ -0,0 +1,71 @@
...
+  iconClassName,
+  link,
+  message,
+  messageClassName,
+  suggestion,
+  title,
+  titleClassName,
+  type,
+}) => {
+  // This is a hack to associate the event with the parent component, DiscoveryView
```
gregdingle, line 23: 👍 also it's hack because we are logging a non-interaction event

app/assets/src/components/views/discovery/EmptyStateModal.jsx
```diff
@@ -0,0 +1,49 @@
...
+      onClose={withAnalytics(onClose, "EmptyStatesModal_close-modal_clicked")}
+    >
+      <InfoBanner
+        className={cs.emptyStateBanner}
+        icon={<BacteriaCultureIcon />}
+        iconClassName={cs.iconClassName}
+        message={
+          <React.Fragment>
+            To get started, you can{" "}
+            <Link className={cs.link} href={"/samples/upload"}>
```
gregdingle, line 28: why not put analytics tracking on all links? 

also I'm wondering whether now is a good time to put the `cs.link` color into `Link` as a default. it seems a little crazy to define the link color in every instance of `Link`.

app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -1012,22 +1038,121 @@ class DiscoveryView extends React.Component {
...
+        if (!visualizations.isLoading() && visualizations.length === 0) {
+          return emptyStateModalOpen ? (
+            <EmptyStateModal onClose={this.handleCloseEmptyStateModal} />
+          ) : (
+            <div className={cs.noDataBannerFlexContainer}>
+              <InfoBanner
+                className={cs.noDataBannerContainer}
+                icon={<BannerVisualizations />}
+                link={{
+                  href: VISUALIZATIONS_DOC_LINK,
```
gregdingle, line 1113: typically we make links to the help center open in a new tab but not a big deal

app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -130,11 +136,15 @@ class DiscoveryView extends React.Component {
...
         showFilters: true,
         showStats: true,
+        userDataCounts: null,
       },
       localState,
       sessionState,
       urlState
     );
+
+    this.loadUserDataStats();
```
gregdingle, line 146: curious why we are not waiting on the result here... for performance in the common case? could this lead to any race condition where something shows before the empty state? maybe worth explaining in a comment

app/assets/src/components/views/discovery/DiscoveryView.jsx
```diff
@@ -1095,12 +1216,8 @@ class DiscoveryView extends React.Component {
...
             {!samples.length &&
               !samples.isLoading() &&
-              currentDisplay === "table" && (
-                <NoResultsBanner
-                  className={cs.noResultsContainer}
-                  type="samples"
-                />
-              )}
+              currentDisplay === "table" &&
+              this.renderNoSearchResultsBanner("samples")}
```
gregdingle, line 1220: good refactoring 👍

app/assets/src/components/views/discovery/empty_state_modal.scss
```diff
@@ -0,0 +1,20 @@
...
+  @include font-header-xl;
+}
+
+.link {
+  color: $primary-light !important;
+}
+
+.iconClassName {
+  svg {
+    fill: $primary-light;
```
gregdingle, line 18: all using CSS vars 🎉

app/assets/src/components/views/discovery/discovery_view.scss
```diff
@@ -86,3 +86,24 @@
...
+  &::before {
+    content: "";
+    flex: 1 0 auto;
+    max-height: 250px;
+  }
+
+  &::after {
+    content: "";
+    flex: 1 0 auto;
+  }
```
gregdingle, line 104: does this vertically center the banner with a min top of 250px?  not that obvious to me, so maybe comment in the code

app/assets/src/components/views/discovery/InfoBanner.jsx
gregdingle: It looks like this event log call was copied from different components a copy times now... I'm glad you added the type ... an alternative would be to rename the event because it has lost its original meaning.

app/assets/src/components/views/discovery/DiscoveryView.jsx
tfrcarvalho: Makes sense. The best to capture it right now would be both... because it only shows up if the user has no data (could have been invited to a project).
@happyimadesignr and @oliviabholmes, we should probably show it anyway to all first time users, no? Maybe we can adapt the message to include users that start the adventure in IDseq with data already. Wdyt?

app/assets/src/components/ui/controls/ExternalLink.jsx
tfrcarvalho: Not sure, but I think it is already...

tfrcarvalho: Not sure, but I think it is already...

app/assets/src/components/ui/icons/BacteriaCultureIcon.jsx
tfrcarvalho: I would say `IconBacteriaCulture` would match the naming schema better. Specially since we already have `BacteriaIcon`. 
But now that you mentioned it, not sure if we should consider this an icon. I might need to review the whole placement of this component.
@happyimadesignr What would this component be in the design system?

app/assets/src/components/views/discovery/InfoBanner.jsx
tfrcarvalho: We can log them all on the clients... probably the best since it is now generic.

As you say, this is also a different type of a event. It is an impression vs. an interaction. We should have that distinction as part of the event at some point.

app/assets/src/components/views/discovery/discovery_view.scss
tfrcarvalho: yes

app/assets/src/components/views/discovery/EmptyStateModal.jsx
tfrcarvalho: > why not put analytics tracking on all links?

Yes, i thought of that but end up forgetting, will put it on the component.

> is a good time to put the cs.link color into Link as a default

Seems like we should do it but it is a bit risky... Will take a look

app/assets/src/components/views/discovery/DiscoveryView.jsx
tfrcarvalho: We do not wait for any data loading. When we use await it is inside an async function. No rendering should assume that the data is loaded.

app/assets/src/components/ui/controls/ExternalLink.jsx
happyimadesignr: Woohoo! 

@tfrcarvalho Is that a component for the blue link?

app/assets/src/components/ui/icons/BacteriaCultureIcon.jsx
happyimadesignr: @tfrcarvalho @gregdingle We are calling these illustrations!  ImgBacteriaXlarge

The illustrations are on the bottom of this page in the design system with names: https://chanzuckerberg.invisionapp.com/share/UWVLU4OH37F#/401265432_1-7_Bases-_Icons_-_Illustrations

app/assets/src/components/views/discovery/DiscoveryView.jsx
happyimadesignr: @tfrcarvalho We should show it to all first-time users. I can't think of a way to word it for both cases other than having 2 different messages. I think if we show the current message to first time users even if they have a project, I think the message still makes enough sense.

With new naming, would it be ModalFirstTimeUser?

app/assets/src/components/views/discovery/discovery_view.scss
tfrcarvalho: Done

app/assets/src/components/views/discovery/DiscoveryView.jsx
tfrcarvalho: Good point. Added option for external link.

app/assets/src/components/views/discovery/DiscoveryView.jsx
tfrcarvalho: Indeed. Renamed.

app/assets/src/components/ui/controls/ExternalLink.jsx
tfrcarvalho: yes, just added ble as default color.

PR #3253, opened 2020-04-08
[IDSEQ-2532] [Heatmap] Address design changes

j-x-han:
# Description

Addressing all the design changes requested in this doc: https://czi.quip.com/Hg21ADRWnnmf/Heatmap-Filtering-Feedback

A summary of the main changes:

**Description**
* Spacing, color, and copy changes for the text and tooltip
* Replaced info icon
* The total preselected taxa count is now restricted to the selected taxon level
image

**`Add Taxa` and `Add Metadata`**
* Font and icon styling updated
* Add Taxa dropdown has enforced minwidth and extra spacing between taxa names and sample counts
* Taxa shown in the dropdown is now restricted to the selected taxon level
* **The heatmap will now immediately scroll to a newly added taxon upon selection**
image

**Manually removing rows with the X**
* X icon has been updated
* Bug where taxa added via the dropdown could not be removed with the X has been fixed (IDSEQ-2611)

**Notification of filtering out added taxon**
* Copy has been simplified
* Notification now only appears when the taxon is initially added. If the user subsequently adds additional filters that filter out the taxon, they will not be notified.
image

# Tests

* Verified that a taxon manually added via the dropdown can be removed through the dropdown and by manually clicking the X in the row.
* Verified that a removed taxon can be added back via the dropdown.
* Verified that the heatmap will **immediately** scroll to the position of a taxon added through the dropdown.


app/assets/src/components/ui/containers/ColumnHeaderTooltip.jsx
```diff
@@ -13,10 +13,18 @@ class ColumnHeaderTooltip extends React.Component {
           <div className={cs.tooltip}>
             {title && <span className={cs.title}>{title}:</span>}
             {content}
-            {list &&
-              list.map(item => {
-                return <li key={item}>{item}</li>;
-              })}
+            {list && (
```
tfrcarvalho, line 16: What does this list contain?

app/assets/src/components/ui/icons/InfoIconSmall.jsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const SmallInfoIcon = props => {
```
tfrcarvalho, line 4: Thank you for taking the opportunity to start updating names according to our new naming guidelines! 💯

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -991,9 +1029,8 @@ class SamplesHeatmapView extends React.Component {
         {
           selectedOptions: assign(this.state.selectedOptions, newOptions),
           loading: shouldRefilterData,
-          // Reset notifiedFilteredOutTaxonIds so the user will be newly
-          // notified if their manually selected taxa do not pass the new filters.
-          notifiedFilteredOutTaxonIds: new Set(),
+          // Don't re-notify the user if their manually selected taxa do not pass the new filters.
```
tfrcarvalho, line 1032: very nit: s/Don't/Do not/
I personally like to avoid single quotes on comments...

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -530,36 +530,40 @@ export default class Heatmap {
...
-    );
-    this.updateCellHighlights();
-
-    for (let i = 0; i < this.rowLabels.length; i++) {
-      this.rowLabels[i].shaded = false;
-    }
-    setTimeout(() => {
-      this.rowLabels[rowIndex].highlighted = false;
-      this.updateCellHighlights();
+    if (row) {
```
tfrcarvalho, line 533: When is row null?

app/assets/src/components/visualizations/heatmap/Heatmap.js
```diff
@@ -530,36 +530,40 @@ export default class Heatmap {
...
+        this.rowLabels[i].shaded = false;
+      }
+      setTimeout(() => {
+        this.rowLabels[rowIndex].highlighted = false;
+        this.updateCellHighlights();
+        this.updateLabelHighlights(
+          this.gRowLabels.selectAll(`.${cs.rowLabel}`),
+          this.rowLabels
+        );
+      }, 2750);
```
tfrcarvalho, line 565: Can we move it to the options structure above?

app/assets/src/components/ui/containers/ColumnHeaderTooltip.jsx
j-x-han: This is an unordered list of text for the tooltip to display, for example: 
[
   `The top ${topN} unique taxa per sample, based off relative abundance (rPM)`,
   `Only taxa with at least ${minReads} reads`,
]
image

The component specifies that the proptype should be PropTypes.arrayOf(PropTypes.string)

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
j-x-han: Makes a lot of sense, I'll keep that in mind for the future!

app/assets/src/components/visualizations/heatmap/Heatmap.js
j-x-han: A user could potentially select and deselect a taxon quickly, in which case scrollToRow would be called with the label but the heatmap would have no longer have the corresponding row information to access, causing the heatmap to crash.

app/assets/src/components/ui/containers/ColumnHeaderTooltip.jsx
tfrcarvalho: Thanks for clarifying... I was looking to see if there was a more specific name that would make it clearer... but i guess it is a list (that it is a bullet list comes from the code).

app/assets/src/components/visualizations/heatmap/Heatmap.js
j-x-han: Done!

PR #3254, opened 2020-04-08
Add new info for nonhost fastq step

gregdingle:
# Description

This goes with https://github.com/chanzuckerberg/idseq-dag/pull/282 . 

# Tests

* run  bundle exec 'rake export_dags[15149,1]'
* see `use_taxon_whitelist` and both cdhitdup files
* tested locally with branch of https://github.com/chanzuckerberg/idseq-dag/pull/282 .


lib/tasks/export_dags.rake
```diff
@@ -25,7 +25,7 @@ task :export_dags, [:sample_id, :print, :step] => :environment do |_, args|
     Sample.find_by(id: args.sample_id).pipeline_runs.each do |pipeline_run|
       PipelineRunStage::STAGE_INFO.each do |step_number, stage_info|
         if args.step && step_number != args.step.to_i
-          break
```
cdebourcy, line 28: Do you mind explaining this change a bit more?

gregdingle: This was an pre-existing bug which I fixed. The `step` are is intended to limit the output to the given step but it was only working for step 1 because then it would exit. That's the difference between `break` and `next`.

cdebourcy: I see, so this is an unrelated change and only affects manual DAG exports, correct?

gregdingle: Yes this script is for running adhoc only AFAIK. I made the change to test the DAG change more easily.

PR #3255, opened 2020-04-09
IDSEQ-2600: Add bulk download type for betacoronavirus

gregdingle:
# Description

This adds a new type for biohub COMET group. Goes with pipeline PRs https://github.com/chanzuckerberg/idseq-dag/pull/282 and https://github.com/chanzuckerberg/idseq-dag/pull/283.

image

# Notes

* Still waiting on https://github.com/chanzuckerberg/idseq-dag/pull/282 and https://github.com/chanzuckerberg/idseq-dag/pull/283 for real files to download
* Not sure whether it is better to allow all admins access or only those with the `allowed_feature`
* It seems a little dangerous to have s3 names that contain spaces like `test_project_79/Test Sample One_131_betacoronavirus_reads_R1.fastq`

# Tests

* unit test
* open bulk download dialog, see new choice
* tested with allowed features

### DIFF
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -570,6 +570,23 @@ class PipelineRun < ApplicationRecord
     files
   end
 
+  def betacoronavirus_fastq_s3_paths
+    # must be in sync with naming in idseq_dag/steps/nonhost_fastq.py
+    s3_path = lambda do |name|
+      "#{postprocess_output_s3_path}/betacoronavirus__#{name.gsub(/.gz$/, '')}"
+    end
+
+    s3_paths = [
+      s3_path.call(sample.input_files[0].name),
+    ]
+
+    if sample.input_files.length == 2
+      s3_paths << s3_path.call(sample.input_files[1].name)
+    end
+
+    s3_paths
+  end
+
   # Unidentified is also referred to as "unmapped"
   def unidentified_fasta_s3_path
     return "#{assembly_s3_path}/#{ASSEMBLY_PREFIX}#{DAG_UNIDENTIFIED_FASTA_BASENAME}" if supports_assembly?
--- a/app/models/pipeline_run.rb
+++ b/app/models/pipeline_run.rb
@@ -573,15 +573,15 @@ class PipelineRun < ApplicationRecord
   def betacoronavirus_fastq_s3_paths
     # must be in sync with naming in idseq_dag/steps/nonhost_fastq.py
     s3_path = lambda do |name|
-      "#{postprocess_output_s3_path}/betacoronavirus__#{name.gsub(/.gz$/, '')}"
+      "#{postprocess_output_s3_path}/betacoronavirus__#{name}"
     end
 
     s3_paths = [
-      s3_path.call(sample.input_files[0].name),
+      s3_path.call(sample.input_files[0].without_gz),
     ]
 
     if sample.input_files.length == 2
-      s3_paths << s3_path.call(sample.input_files[1].name)
+      s3_paths << s3_path.call(sample.input_files[1].without_gz)
     end
 
     s3_paths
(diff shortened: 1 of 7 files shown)


gregdingle:
> if a user picks samples to download that don't have the betacoronavirus files.

Good point. We could show a custom error message if any of the runs do not have `use_taxon_whitelist`. But that maybe overkill for a feature we only intend to whitelist to a few users.  Thoughts?

MarkAZhang:
If it's only for a few Biohub users, should be okay.

app/models/pipeline_run.rb
```diff
@@ -570,6 +570,23 @@ def nonhost_fastq_s3_paths
...
+    s3_path = lambda do |name|
+      "#{postprocess_output_s3_path}/betacoronavirus__#{name}"
+    end
+
+    s3_paths = [
+      s3_path.call(sample.input_files[0].without_gz),
+    ]
+
+    if sample.input_files.length == 2
+      s3_paths << s3_path.call(sample.input_files[1].without_gz)
```
cdebourcy, line 584: Is it necessary to name this file after the user-provided input file? It seems like a complexity we could get rid of. We've also noted that for security purposes it would be better to use our own file names in our backend rather than user-provided ones and this would be a good opportunity to get into that habit. cc @davidrissato ?

cdebourcy:
> LGTM. The only thing I'm a little concerned about is what happens if a user picks samples to download that don't have the betacoronavirus files. The expected behavior is that the download should complete with "complete with issue" but would be good to make sure. Prior to this change, it was very rare that a bulk download would be missing a file, but now this makes it more likely.

That's a good point. We always expect a number of samples in these projects to have zero betacoronavirus reads, namely the negative controls they make a point of including with each batch. So we definitely need to make sure that case is handled properly.

cdebourcy:
> > if a user picks samples to download that don't have the betacoronavirus files.
> 
> Good point. We could show a custom error message if any of the runs do not have `use_taxon_whitelist`. But that maybe overkill for a feature we only intend to whitelist to a few users. Thoughts?

Would be good to check to protect against future breakage!

app/models/pipeline_run.rb
davidrissato: Although it would be a good idea to use our own file names, I believe in this case he might have done that on purpose, since bulk downloads can download multiple samples at once, and it would be hard for users to identify based on the sample id. Am I correct?

MarkAZhang: Bulk downloads has a step to rename the files, and includes the sample name and sample id in the file name. So it's okay to give these files a static name like betacoronavirus_R1.fastq if we want.

gregdingle: It's not necessary at all. I just followed the example of the previous PR which I presumed was done to convey more context to the downloader. 

It's good to learn that we prefer to use our own file names. The bulk download code renames the files according to the sample name which is derived from the original uploaded file name. So is that  a security risk as well?

davidrissato: > @MarkAZhang said:
> Bulk downloads has a step to rename the files, and includes the sample name and sample id in the file name
By "sample name" do you mean `Sample#name` or `InputFile#name`? Could you point us to this code?

MarkAZhang: Yeah, it's in get_output_file_prefix. It's the sample name.

gregdingle:
I changed the name scheme and changed the beta label color to grey per @happyimadesignr . 

If there are no objections, I'd like to merge this now to ease end to end testing with the other PRs that are currently in progress.

app/models/pipeline_run.rb
davidrissato: Considering we want using the sample name, I recommend being very aggressive on name cleanups. My suggestion is only allowing `/[0-9a-zA-Z_]/`. Anything else should be replaced by an underscore.

PR #3256, opened 2020-04-10
Alignment batch

morsecodist:
# Description

This change removes the old ASG management stuff and adds some stuff that will be needed for the batch refactor so we can remove compatibility hacks.

# Tests

Ran various pipeline tests locally

Also tested on a branch of idseq-dag with no compatibility hacks


app/helpers/pipeline_runs_helper.rb
```diff
@@ -65,10 +65,15 @@ def aegea_batch_submit_command(base_command,
                                  docker_image: "idseq_dag",
                                  sample_id: 0,
                                  stage_name: "misc")
+    # TODO: environment values once we use the new pipeline infra
+    #   that new infra uses dev rather than development
+    #   once we switch to that for the pipeline these values won't be needed
+    deployment_environment = Rails.env == "development" ? "dev" : Rails.env
```
tfrcarvalho, line 71: There is a mapping for in https://github.com/chanzuckerberg/idseq-web/blob/master/app/services/sfn_pipeline_dispatch_service.rb#L13. Please use the mapping but feel free to move it a more adequate place if necessary.

db/schema.rb
```diff
@@ -502,8 +502,8 @@
     t.integer "taxid", null: false
     t.bigint "wikipedia_id"
     t.string "title"
-    t.text "summary"
-    t.text "description"
+    t.text "summary", limit: 16_777_215
```
cdebourcy, line 505: Are these unrelated schema changes intended? Was the schema file out of date before?

morsecodist: Sorry another weird mangling I will manually bring this back into line.

app/helpers/pipeline_runs_helper.rb
```diff
@@ -65,10 +65,15 @@ def aegea_batch_submit_command(base_command,
                                  docker_image: "idseq_dag",
                                  sample_id: 0,
                                  stage_name: "misc")
+    # TODO: environment values once we use the new pipeline infra
```
davidrissato, line 68: Should we refer to a Jira ticket here?

app/models/pipeline_run.rb
```diff
@@ -1246,9 +1170,9 @@ def load_chunk_stats
     return unless stdout
     outputs = stdout.split("\n").map { |line| line.split.last }
     gsnap_outputs = outputs.select { |file_name| file_name.start_with?("multihit-gsnap-out") && file_name.end_with?(".m8") }
-    rapsearch_outputs = outputs.select { |file_name| file_name.start_with?("multihit-rapsearch2-out") && file_name.end_with?(".m8") }
+    rapsearch2_outputs = outputs.select { |file_name| file_name.start_with?("multihit-rapsearch2-out") && file_name.end_with?(".m8") }
     self.completed_gsnap_chunks = gsnap_outputs.length
-    self.completed_rapsearch_chunks = rapsearch_outputs.length
+    self.completed_rapsearch2_chunks = rapsearch2_outputs.length
```
tfrcarvalho, line 1175: Do we still need this?

app/helpers/pipeline_runs_helper.rb
morsecodist: I am not sure which one it is. This will eventually need to be removed completely. @kislyuk do we have a ticket to capture the decoupling work?

morsecodist: actually maybe I should just remove the TODO since instead of fixing it this section will just be deleted

tfrcarvalho: Should have a ticket anyway... If you find an existing one, great. Otherwise create a new one.

app/helpers/pipeline_runs_helper.rb
tfrcarvalho: Since it is temporary, it is fine to use that location...

app/models/pipeline_run.rb
```diff
@@ -126,25 +120,25 @@ class PipelineRun < ApplicationRecord
   STATUS_RUNNABLE = 'RUNNABLE'.freeze # TODO: (gdingle): not used anywhere?
   STATUS_READY = 'READY'.freeze
   # NOTE: The current stored job_status are...
-  # +-------------------------------------------+
-  # | job_status                                |
-  # +-------------------------------------------+
-  # | 1.Host Filtering-FAILED                   |
-  # | 1.Host Filtering-FAILED|READY             |
-  # | 2.GSNAPL/RAPSEARCH alignment-FAILED       |
```
gregdingle, line 134: this list is supposed to represent historical values, not current ones––if those are even changing

db/migrate/20200403194501_rapsearch_to_rapsearch2.rb
```diff
@@ -0,0 +1,5 @@
+class RapsearchToRapsearch2 < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :pipeline_runs, :completed_rapsearch_chunks, :completed_rapsearch2_chunks
```
gregdingle, line 3: wow this is a pretty thorough rename!

in general, I'd be cautious about renaming any db col. you need to understand the downstream dependents really well, which could even include analytics reports! in this case, I think its safe

gregdingle:
Red diffs are the best! 😄

db/migrate/20200403194501_rapsearch_to_rapsearch2.rb
morsecodist: `completed_rapsearch_chunks` has no usages in the code base. I assume it was to do with the autoscaling but I am hesitant to drop it for historical reasons. I do want to ensure we rename it everywhere though.

app/models/pipeline_run.rb
morsecodist: Nope, thanks!

morsecodist:
Doing another test before I merge

PR #3260, opened 2020-04-10
IDSEQ-2635 IDSEQ-2641 Fix bugs with location metadata auto-adjusting.

MarkAZhang:
# Description

This PR fixes the following bugs:
- MetadataCSVLocationMenu wasn't properly adjusting city-level locations for human samples. This was because the wrong prop was being passed (isHuman versus taxaCategory)
- Was allowing users to save city-level locations if the city_name wasn't present (but the geo_level was city). This does happen for some LocationIQ locations (possible data quality issue). This is addressed with the changes to `models/location.rb`.
- Was creating duplicate Location entries under specific conditions (due to LocationIQ passing back inconsistent results for the same osm_id). See changes in find_or_new_by_fields. See https://jira.czi.team/browse/IDSEQ-2641.

It makes the following changes:
- Now re-validates the locations after the user adjusts them in MetadataCSVLocationMenu. Shows any errors.
Screen Shot 2020-04-10 at 10 13 15 AM
Note: there should never BE any errors, because we now properly auto-adjust the user's locations. The error message is confusing (it refers to a location in the original CSV) but it would be tricky to fix so deferring that for now since the error message should never actually show up.
- Displays a "Verifying collection locations..." loading message when we are running getCSVLocationMatches (a tiny UX improvement)
- Return a more specific error for invalid locations for human samples.
- Add the restriction that an "adjusted" location (which is shown in the UI as "Changed to county/district for privacy reasons") needs to have "refetchAdjustedLocation: true" set. Otherwise, we will assume that there was a bug and return an error. This will require a small change in idseq-cli. (See https://github.com/chanzuckerberg/idseq-cli/pull/67) 
- When auto-adjusting the location for privacy reasons, remove the city_name (no need to keep it around) and also remove the subdivision_name if it is the same as the city_name (as it creates confusion).

# Notes

https://jira.czi.team/browse/IDSEQ-2635 is partially fixed, in that if you try to perform the bug, the back-end will now catch it and throw a validation error. However, the front-end is also mis-behaving there (it should auto-adjust when Apply to All is clicked but it doesn't) This will be addressed in a subsequent PR.

Note that MetadataCSVLocationMenu also has an Apply to All button, and this one IS now working correctly.

Also, the existing CLI will not always work with these changes. We will need to submit https://github.com/chanzuckerberg/idseq-cli/pull/67.

# Tests

* Added rspec tests.
* Verified that location validation error is raised if you perform the steps in https://jira.czi.team/browse/IDSEQ-2635.
* Verified that sample upload succeeds with valid locations for both manual and CSV upload.
* Verified that duplicate locations are no longer being created in situations where they were being created before.
* Verified that MetadataCSVLocationMenu now correctly auto-adjusts the locations.
* Verifies that auto-adjusting logic behaves as expected.

# Tests for QA

* Verify that uploading a sample via the upload flow succeeds as before.
* Verify that for a human sample, there is no way to select a "city-specific" location (like "San Francisco"). The UI should always auto-correct.

### DIFF
--- a/app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
+++ b/app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { uniq, find, chunk, random } from "lodash/fp";
+import { uniq, find, chunk, random, get } from "lodash/fp";
 
 import { withAnalytics, logAnalyticsEvent } from "~/api/analytics";
 import { getGeoSearchSuggestions } from "~/api/locations";
@@ -139,7 +139,7 @@ class MetadataCSVLocationsMenu extends React.Component {
           metadata: newMetadata,
         });
         this.setState({ applyToAllSample: sampleName });
-        logAnalyticsEvent("MetadataManualInput_input_changed", {
+        logAnalyticsEvent("MetadataCSVLocationsMenu_input_changed", {
           key,
           value,
           sampleName,
@@ -156,7 +156,7 @@ class MetadataCSVLocationsMenu extends React.Component {
             metadataType={locationMetadataType}
             onChange={onChange}
             withinModal={true}
-            isHuman={isRowHuman(row)}
+            taxaCategory={get("taxa_category", this.getHostGenomeForRow(row))}
             warning={CSVLocationWarnings[sampleName]}
           />
           {applyToAllSample === sampleName &&
@@ -166,6 +166,12 @@ class MetadataCSVLocationsMenu extends React.Component {
       ];
     });
   };
+
+  getHostGenomeForRow = row =>
+    find(
+      ["name", row["Host Organism"] || row["Host Genome"]],
+      this.props.hostGenomes
+    );
 
   render() {
     const { metadata, locationMetadataType } = this.props;
@@ -202,6 +208,7 @@ MetadataCSVLocationsMenu.propTypes = {
   }),
   onCSVLocationWarningsChange: PropTypes.func.isRequired,
   onMetadataChange: PropTypes.func.isRequired,
+  hostGenomes: PropTypes.array,
 };
 
 export default MetadataCSVLocationsMenu;
--- a/app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
+++ b/app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
@@ -15,11 +15,22 @@ export const LOCATION_UNRESOLVED_WARNING =
 // Process location selections and add warnings.
 export const processLocationSelection = (result, isHuman) => {
   let warning = "";
+  // For human samples, drop the city part of the name and show a warning.
+  // NOTE: The backend will redo the geosearch for confirmation and re-apply
+  // this restriction.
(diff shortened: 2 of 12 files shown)


app/helpers/error_helper.rb
```diff
@@ -47,14 +47,19 @@ def self.upload_from_basespace_failed(sample_name, sample_id, file_name, basespa
     end
   end
 
+  LOCATION_INVALID_ERROR_HUMAN = "Please input a location at the county/district level of higher. (for human samples)".freeze
```
jshoe, line 50: nit: "or higher"?

app/assets/src/components/ui/controls/GeoSearchInputBox.jsx
```diff
@@ -15,11 +15,22 @@ export const LOCATION_UNRESOLVED_WARNING =
...
+  // NOTE: The backend will redo the geosearch for confirmation and re-apply
+  // this restriction.
   if (isHuman && get("geo_level", result) === "city") {
     result = Object.assign({}, result); // make a copy to avoid side effects
-    // For human samples, drop the city part of the name and show a warning.
-    // NOTE: The backend will redo the geosearch for confirmation and re-apply
-    // this restriction.
+
+    // If the subdivision name is identical to the city name, remove the subdivision name as well to be safe.
+    if (result.subdivision_name === result.city_name) {
```
jshoe, line 25: did you have some specific cases in mind for this? did "San Francisco" get returned like that?

app/models/location.rb
```diff
@@ -131,9 +131,18 @@ def self.find_or_new_by_fields(loc_info)
       success, resp = geosearch_by_osm_id(loc_info[:osm_id], loc_info[:osm_type])
       raise "Couldn't fetch OSM ID #{loc_info[:osm_id]} (#{loc_info[:osm_type]})" unless success
 
-      resp = LocationHelper.adapt_location_iq_response(resp)
-      # 'New' without saving so make sure caller saves.
-      new_from_params(resp)
+      location_fields = LocationHelper.adapt_location_iq_response(resp)
+
+      # Surprisingly, it is possible that adapt_location_iq_response returns location_fields
+      # that are different from loc_info above (where loc_info is passed in from the front-end).
```
jshoe, line 137: did you find a reason for this surprise? or, did you find what fields differed unexpectedly?

app/assets/src/components/common/MetadataCSVLocationsMenu.jsx
```diff
@@ -156,7 +156,7 @@ class MetadataCSVLocationsMenu extends React.Component {
             metadataType={locationMetadataType}
             onChange={onChange}
             withinModal={true}
-            isHuman={isRowHuman(row)}
+            taxaCategory={get("taxa_category", this.getHostGenomeForRow(row))}
```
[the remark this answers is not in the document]
MarkAZhang: That's a great question about the linting. Let me check.

app/models/location.rb
MarkAZhang: I broke down the API inconsistencies I discovered here: https://czi.quip.com/OZw6ATNbMyJi/LocationIQ-Gotchas

PR #3264, opened 2020-04-13
SfnPipelineDispatchService: Pass pipeline_version, sample_output_s3_path

kislyuk:


app/services/sfn_pipeline_dispatch_service.rb
```diff
@@ -113,6 +113,8 @@ def convert_dag_json_to_wdl(dag_json)
       },
       "app/jobs/idd2wdl.py",
       "--name", dag_json['name'].to_s,
+      "--dag-branch", @sample.pipeline_version,
```
tfrcarvalho, line 116: Why do we need dag branch here? Isn't it optional? 
I already load the dag branch from the sfn.

I will hook up a custom dag branch later.

kislyuk: Sorry, I used the wrong argument here, and I suspect that caused the confusion. Please see the udpated version using `@pipeline_run.pipeline_commit`.

This parameter is required when `pipeline_branch` is set to a custom value (not `master`) to induce the WDL script to install the custom version of idseq-dag at runtime. It replaces the logic in https://github.com/chanzuckerberg/idseq-web/blob/master/app/helpers/pipeline_runs_helper.rb#L225-L232.

PR #3266, opened 2020-04-14
[Release Fix] Added staging deployment environment to map

morsecodist:
# Description

Adds `staging` to the map of rails environments to deployment environments.


app/services/sfn_pipeline_dispatch_service.rb
```diff
@@ -9,6 +9,7 @@ class SfnPipelineDispatchService
   # Constains SFN deployment stage names that differ from Rails.env
   ENV_TO_DEPLOYMENT_STAGE_NAMES = {
     "development" => "dev",
+    "staging" => "staging",
```
tfrcarvalho, line 12: The idea here was to just use the environment name if key does not exist.
But it does not hurt.

PR #3268, opened 2020-04-14
[Release fix] Stage rename

morsecodist:
# Description

Renames the stage in the db with a migration

# Tests

Ran locally and looked at my local runs from before the migration and their visualizations worked.


app/services/sfn_pipeline_dispatch_service.rb
```diff
@@ -12,6 +12,7 @@ class SfnPipelineDispatchService
   # Constains SFN deployment stage names that differ from Rails.env
   ENV_TO_DEPLOYMENT_STAGE_NAMES = {
     "development" => "dev",
+    "staging" => "staging",
```
gregdingle, line 15: how is this related?

morsecodist: sorry, forgot to rebase

PR #3272, opened 2020-04-14
[Release fix] Raw sql migration

morsecodist:
# Description

Due to lingering dependency issues with the migration I switched this to raw sql as per discussion with the team.

This issue started with the pipeline viz breaking https://jira.czi.team/browse/IDSEQ-2660. This breakage occurred because I refactored consistent usage of rapsearch2 vs rapsearch which changed the name of one of our stages. This worked for jobs going forward but historical jobs had the old value still which is used to construct the viz. I tried to migrate using the class but that caused issues because of some dependency issues with loading constants. To sidestep these issues I switched to raw sql.

# Tests

Though I have had difficulties with local testing in the past, I have more confidence this will work because it introduces no dependencies and it ran locally.

### DIFF
--- a/db/migrate/20200414220332_rename_alignment.rb
+++ b/db/migrate/20200414220332_rename_alignment.rb
@@ -1,5 +1,10 @@
 class RenameAlignment < ActiveRecord::Migration[5.1]
-  def change
-    ActiveRecord::Base.connection.execute("UPDATE pipeline_run_stages SET name='GSNAPL/RAPSEARCH alignment' WHERE name='GSNAPL/RAPSEARCH alignment'")
+  # Using raw sql here because PipelineRunStage introduces dependencies on constants that may not be initialized
+  def up
+    ActiveRecord::Base.connection.execute("UPDATE pipeline_run_stages SET name='GSNAPL/RAPSEARCH alignment' WHERE name='GSNAPL/RAPSEARCH2 alignment'")
+  end
+
+  def down
+    ActiveRecord::Base.connection.execute("UPDATE pipeline_run_stages SET name='GSNAPL/RAPSEARCH2 alignment' WHERE name='GSNAPL/RAPSEARCH alignment'")
   end
 end
--- a/db/migrate/20200414220332_rename_alignment.rb
+++ b/db/migrate/20200414220332_rename_alignment.rb
@@ -1,8 +1,5 @@
-# This is needed to load the constants in PipelineRunStage
-require 'pipeline_run_stage'
-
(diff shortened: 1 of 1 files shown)


db/migrate/20200414220332_rename_alignment.rb
```diff
@@ -1,8 +1,5 @@
-# This is needed to load the constants in PipelineRunStage
-require 'pipeline_run_stage'
-
 class RenameAlignment < ActiveRecord::Migration[5.1]
   def change
-    PipelineRunStage.where(name: "GSNAPL/RAPSEARCH alignment").find_each { |u| u.update(name: "GSNAPL/RAPSEARCH2 alignment") }
+    ActiveRecord::Base.connection.execute("UPDATE pipeline_run_stages SET name='GSNAPL/RAPSEARCH alignment' WHERE name='GSNAPL/RAPSEARCH alignment'")
```
davidrissato, line 3: I recommend using reversible migrations:
```ruby
class RenameAlignment < ActiveRecord::Migration[5.1]
   def up
       # update
   end
   def down
       # revert
   end
end
```

davidrissato: Also, it looks like both strings are the same: `'GSNAPL/RAPSEARCH alignment'`

morsecodist: Thanks for catching!

davidrissato:
Could you add some extra info on what this PR is doing in the description

db/migrate/20200414220332_rename_alignment.rb
```diff
@@ -1,8 +1,10 @@
-# This is needed to load the constants in PipelineRunStage
-require 'pipeline_run_stage'
-
 class RenameAlignment < ActiveRecord::Migration[5.1]
-  def change
-    PipelineRunStage.where(name: "GSNAPL/RAPSEARCH alignment").find_each { |u| u.update(name: "GSNAPL/RAPSEARCH2 alignment") }
+  # Using raw sql here because PipelineRunStage introduces dependencies on constants that may not be initialized
+  def up
+    ActiveRecord::Base.connection.execute("UPDATE pipeline_run_stages SET name='GSNAPL/RAPSEARCH alignment' WHERE name='GSNAPL/RAPSEARCH2 alignment'")
```
davidrissato, line 4: Shouldn't it be:
`ActiveRecord::Base.connection.execute("UPDATE pipeline_run_stages SET name='GSNAPL/RAPSEARCH2 alignment' WHERE name='GSNAPL/RAPSEARCH alignment'")`

PR #3276, opened 2020-04-15
IDSEQ-2569 Cache location validation keyed on raw_value.

MarkAZhang:
# Description

I ended up caching not the location API calls, but the validation of location metadata.
This is because validation of location metadata also includes additional calls to Location table for each metadata value, and this is also a contributing factor to the slowness of validation.

Benchmarks for validation of 24 samples:
With text string locations (no location validation at all) - 300ms
With location objects prior to PR - 5000ms
With caching of just location API calls - 3000ms
With caching of location metadata validation (this PR's approach) - 500ms

We make a couple of assumptions in order to cache the validation. The most important is if the raw location value (e.g. the JSON) that is being passed in is exactly the same (including the locationiq_id, the osm_id, etc.), then the location validation will always produce the same result, and return the same Location object in our database. This allows us to use the raw_value as the key for the cache. It is quite common that the raw_value is exactly the same across all location objects in a batch of sample uploads (when the user clicks Apply All), so this strict strategy for caching gets a lot of hits.

Also some benchmarks for the actual uploading of 24 samples:
With text string locations (no location validation at all) - 4.3s
With location objects prior to PR - 9.1s
With caching of location metadata validation (this PR's approach) - 4.3s

We may want to investigate other causes for the slowness of uploads, though that is outside the scope of this PR.

# Notes

See various comments in the code. 
Things to note: choice of very short cache expiration time.

# Tests

* Verified for both CSV and manual metadata upload that upload still works.
* Verified for when Redis cache is enabled locally (by adding tmp/caching-dev.txt)
* Verified for both raw text, the same location value for all samples, and differing location values for samples in one upload.
* Updated rspec tests and added a test for the cache.

# Tests for QA
* Upload a batch of multiple samples with the same Collection Location metadata value and another with different Collection Location metadata values, and ensure that the Collection Locations are what you expect after the upload is complete.

### DIFF
--- a/app/models/metadatum.rb
+++ b/app/models/metadatum.rb
@@ -102,26 +102,35 @@ class Metadatum < ApplicationRecord
     errors.add(:raw_value, MetadataValidationErrors::INVALID_DATE)
   end
 
-  def check_and_set_location_type
-    # Skip if location was already resolved
-    return if location_id && !raw_value
-
+  # Helper function of check_and_set_location_type.
+  # Returns the {location_id, string_validated_value, raw_value} values that should be assigned to the metadata object.
+  # Note: May create Location objects as a side effect.
+  # Important assumptions:
+  # The same raw_value will always cause the same response to be returned.
+  # Location objects in our database may need to be created the first time a new raw_value is processed,
+  # but only the first time.
+  def _determine_location_params(raw_value)
     # Based on our metadata structure, the location details selected by the user will end up in
     # raw_value.
     begin
       loc = JSON.parse(raw_value, symbolize_names: true)
     rescue JSON::ParserError
       # CSV uploads will be unwrapped strings
-      self.string_validated_value = raw_value
-      self.location_id = nil
-      return
+      return {
+        location_id: nil,
+        string_validated_value: raw_value,
+        raw_value: nil,
+      }
     end
 
     unless loc[:locationiq_id]
       # Unresolved plain text selection (wrapped 'name')
       if loc[:name].present?
-        self.string_validated_value = loc[:name]
-        self.location_id = nil
+        return {
+          location_id: nil,
+          string_validated_value: loc[:name],
+          raw_value: nil,
+        }
       end
       return
     end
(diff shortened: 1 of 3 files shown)


spec/models/metadata_spec.rb
```diff
@@ -123,12 +139,61 @@
       expect(created_location.osm_id).to eq(200)
     end
 
+    # This tests the cache by saving the same raw_value twice.
+    it "should return correct response if called twice with the same raw_value" do
+      # This needs to be different from all other locations in this file, so as not to trigger the cache.
```
gregdingle, line 144: what guarantees do we have that the cache will always be on during unit tests?

app/models/metadatum.rb
```diff
@@ -154,10 +163,36 @@ def check_and_set_location_type
...
+    cache_key = "#{raw_value}&_v=#{cache_version_key}"
+
+    # Use a cache for this function so that, if the exact same raw location value is passed for multiple samples
+    # in one request, we only validate it once, which saves time.
+    # Use the raw value as the cache key.
+    # Use a very short cache expiration time because our current performance requirements only require this cache
+    # to persist for the duration of one request.
+    # A short cache expiration time mitigates the risk in case someone forgets to
+    # increment the cache_version_key above when making a change.
+    location_params = Rails.cache.fetch(cache_key, expires_in: 1.minute) do
```
gregdingle, line 188: Will one minute be enough for uploads of up to 1000 samples? it might be a good idea to increase it to ensure the expectation that all samples will be covered in one "request"

gregdingle:
Good fix, and nice to see all the benchmarks!

spec/models/metadata_spec.rb
MarkAZhang: Hey @gregdingle. It looks like we already testing the cache elsewhere so I assumed it would be on. Also, it looks like in config/environments/test.rb, we set the cache_store to Redis.

I remember you worked on the cache though, so please let me know if you still have concerns!

app/models/metadatum.rb
MarkAZhang: That's a good point. Right now it wouldn't be enough. I'll set it to 10 minutes which should be enough.

spec/models/metadata_spec.rb
gregdingle: Ok that's good to know. Caching is something I would expect could be turned on or off in tests depending on the application.

PR #3280, opened 2020-04-17
Adapt pipeline viz and admin pages to SFN pipeline

tfrcarvalho:
# Description

* Adapt pipeline viz to show live updates to the new SFN pipeline
* Add link to step function on the pipeline runs admin page
* Side updates:
	- Add formatter to display dag event logs in WDL execution logs
	- Fix environment mapping issue

# Tests

* Run pipeline runs with and without the step function enabled and verify:
	- pipeline run succeeds
	- pipeline viz is displayed correctly (including during intermediary stages)


app/jobs/idd2wdl.py
```diff
@@ -118,10 +118,21 @@ def target_filename(target, index):
...
+  )
   {install_custom_idseq_dag_version}
   import idseq_dag, idseq_dag.util.s3, idseq_dag.util.count
-  logging.basicConfig(level=logging.INFO)
+
+  root_logger = logging.getLogger()
+  root_logger.setLevel(level=logging.INFO)
+  stream_handler = logging.StreamHandler()
+  stream_handler.setFormatter(idseq_dag.util.log.JsonFormatter())
+  root_logger.addHandler(stream_handler)
```
kislyuk, line 133: I'm guessing you had to do this to format logs correctly for cloudwatch?

If we need to do this, can we just call `idseq_dag.util.configure_logger()` instead?

app/jobs/idd2wdl.py
```diff
@@ -138,16 +149,29 @@ def target_filename(target, index):
...
     json.dump(dict(), status_file)
+
   try:
     {count_input_reads}
     step_instance.update_status_json_file("running")
     step_instance.run()
     step_instance.count_reads()
     step_instance.save_counts()
-    step_instance.update_status_json_file("finished_running")
+    # temoorary until we instrument miniwdl - not yet uploaded, but this is the final status
```
kislyuk, line 162: Typo in "temporary"

app/jobs/idd2wdl.py
```diff
@@ -138,16 +149,29 @@ def target_filename(target, index):
...
+    step_instance.update_status_json_file("uploaded")
+  except idseq_dag.engine.pipeline_step.InvalidInputFileError as e:
+    try:
+        self.update_status_json_file("user_errored")
+    except:
+        logging.error("Failed to update status to 'user_errored'")
+    logging.error("Failure due to user error: %%s", traceback.format_exc())
```
kislyuk, line 169: Shouldn't this (the branch under line 164) exit with an error? It looks like it falls through to a successful exit from the process.

app/jobs/idd2wdl.py
tfrcarvalho: This, is just to output all the `log_event` from pipeline step. They coming out as `None`. True that the JSON format is great for automatic parsing by cloudwatch too.
I think that `idseq_dag.util.configure_logger()` does not work since it uses a named logger from where it is called instead of receiving a logger... but willing to try again.
Maybe we can move on with this version for now given time constraints?

app/jobs/idd2wdl.py
tfrcarvalho: I don't think so. Errors due to malformed inputs from user are supposed to be pipeline successes. It means that the checks were in place and worked.
Having said we might want some kind of warning on these... so, if you feel strongly i can leave the unsuccessful exit for now - until we have a specific way to handle these somewhere eles.

app/jobs/idd2wdl.py
tfrcarvalho: Fixed.

app/jobs/idd2wdl.py
kislyuk: Let's continue with this version, yes.

app/jobs/idd2wdl.py
kislyuk: I would like us to move away from the convention that a user error results in a successful pipeline exit. A successful pipeline exit should mean that the pipeline results are available. Instead of signaling user error out-of-band, we should move to an error reporting system where the reason for the exit is communicated in the last event of the step function execution history, as described here: https://github.com/chanzuckerberg/idseq/blob/master/workflows/Interface.md#reporting-errors. The caller should then parse the reason, and if it matches a dictionary of errors attributable to user input, display a message specific to that cause.

tfrcarvalho: OK, kept the behavior. The final code is a bit convoluted but i think of for a temporary conversion.

app/jobs/idd2wdl.py
```diff
@@ -138,16 +149,32 @@ def target_filename(target, index):
...
+        self.update_status_json_file("user_errored")
+      except:
+        logging.error("Failed to update status to 'user_errored'")
+    except:
+      try:
+        self.update_status_json_file("pipeline_errored")
+      except:
+        logging.error("Failed to update status to 'pipeline_errored'")
     traceback.print_exc()
     exit(json.dumps(dict(wdl_error_message=True, error=type(e).__name__, cause=str(e))))
```
kislyuk, line 179: With the change above, `e` will now end up unbound in the "pipeline_errored" branch above. If you just keep L164 as ` except Exception as e:` this will work, or consider rewriting it like this instead:
```python
except Exception as e:
    status = "user_errored" if isinstance(e, idseq_dag.engine.pipeline_step.InvalidInputFileError) else "pipeline_errored"
    try:
        self.update_status_json_file(status)
    except:
        logging.error("Failed to update status to '%s'", status)
    traceback.print_exc()
    exit(json.dumps(dict(wdl_error_message=True, error=type(e).__name__, cause=str(e))))
```

tfrcarvalho: Sure, makes sense and it is more compact.

PR #3290, opened 2020-04-21
Change improper access alert level to warning and refactor sample validation api

lvreynoso:
# Description

Currently when a user attempts to make a bulk download request for samples they don't have access to, an error is logged and Airbrake is tripped. Since we don't need to be notified in Slack every time this happens (it means the code is working fine!) this PR removes Airbrake from the code path and downgrades the logging level to a warning.

# Notes

The `BulkDownloadsSampleValidationService` has been renamed to `SampleAccessValidationService` to clarify the logic of the service, so that it may be reused in a more generic manner. It's also been made into a simple, easy to use callable.

The `validate_sample_ids` method has been moved to the Samples controller, since it no longer has any code specific to bulk downloads. The `validateSampleIds` method in React has been moved to `access_control.js` as well, for the same reason.

# Tests

* Rspec tests updated and passed.
* Frontend tests: make sure Bulk Downloads modal and Taxon heatmap still work and are still properly filtering out samples that failed or are in progress.


app/services/sample_access_validation_service.rb
```diff
@@ -1,12 +1,14 @@
-class BulkDownloadsSampleValidationService
+class SampleAccessValidationService
```
tfrcarvalho, line 1: 👍 for the name change!

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20_200_403_194_501) do
+ActiveRecord::Schema.define(version: 20_200_414_220_332) do
```
tfrcarvalho, line 13: No need to push this.

app/lib/log_util.rb
```diff
@@ -9,4 +9,8 @@ def self.log_err_and_airbrake(msg)
   def self.log_backtrace(exception)
     Rails.logger.error("Backtrace:\n\t#{exception.backtrace.join("\n\t")}")
   end
+
+  def self.log_warning(msg)
```
tfrcarvalho, line 13: nit: i dont think this justifies a method... Why not just use `Rails.logger.warn` directly? 
Is it because of the test?

lvreynoso: Nah, just following pattern. I agree, I'll change it

app/services/sample_access_validation_service.rb
MarkAZhang, line 1: Could you add a comment summarizing what this service does to the top of the file?

PR #3291, opened 2020-04-22
idd2wdl.py: import changes from idseq

kislyuk:
# Description

# Notes

# Tests


kislyuk:
How can I do a quick test pipeline run with this code?

kislyuk:
Added the tests section back in to the description.

tfrcarvalho:
> Added the tests section back in to the description.

The test section is to add how you tested the code? In this case, i see at least unit tests.

tfrcarvalho:
> How can I do a quick test pipeline run with this code?

In this case, I would say that just an upload of a quick sample through the frontend should suffice and make sure it triggers the step function. Probably you can cancel the step function after that.

kislyuk:
OK, I added a note about unit tests to the description.

I was unable to test this on the frontend, because the frontend has stopped working on my computer due to an elasticsearch configuration issue that I haven't been able to track down yet.

PR #3292, opened 2020-04-22
Upgrade Rails to 5.2.4

lvreynoso:
# Description

Upgrades Rails to 5.2.4 to fix a security vulnerability in `actionview`. Forces an `npm` resolution for `kind-of` to `^6.0.3` in order to avoid a security vulnerability. The other updates involved (including a problem with font rendering) led me to update some `npm` packages as well. The new Rails version installed `elasticsearch-model (7.1.0)` and was incompatible with the elasticsearch image, which meant a new Docker image had to be pulled.

# Notes

Had to refresh `package-lock.json` due to some issues. I also had to follow the instructions found at https://github.com/nodejs/node-gyp/blob/master/macOS_Catalina.md to fix a node-gyp issue since I have Catalina on my computer.

NPM and snyk will now only have to audit ~260k packages instead of ~1.2m, which is a win.

# Tests

* Automated tests pass
* Smoke testing by playing around with the website
* Sample uploads work
* Pipelines are kicked off and resque monitors work
* Will do more integration testing


Gemfile
```diff
@@ -39,7 +39,7 @@ gem 'puma', '~> 3.12'
 # Use Redis adapter to run Action Cable in production
 # gem 'redis', '~> 3.0'
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '~> 5.1.5'
+gem 'rails', '~> 5.2.4.2'
```
tfrcarvalho, line 42: We should avoid freezing very minor versions. They should not change interfaces and allows to always install the latest security fixes. Ideally we should be making it a bit less specific `major.minor`...
Any reason why we need so much specificity?

app/assets/src/components/ui/containers/toast_container.scss
```diff
@@ -9,7 +9,7 @@
       margin-bottom: 5px;
       background: transparent;
 
-      &-body {
+      & .body {
```
tfrcarvalho, line 12: curious, did this change underneath?

app/views/layouts/application.html.erb
```diff
@@ -8,6 +8,7 @@
     <%= csrf_meta_tags %>
     <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet"/>
+    <link href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,400i,600,600i,700,700i,800" rel="stylesheet">
```
tfrcarvalho, line 11: Why did we need this? Shouldn't this be coming from the css?

package.json
```diff
@@ -53,21 +54,21 @@
     "eslint-plugin-promise": "^3.7.0",
     "eslint-plugin-react": "^7.10.0",
     "eslint-plugin-standard": "^3.0.1",
-    "extract-text-webpack-plugin": "^3.0.2",
+    "fibers": "^4.0.2",
```
tfrcarvalho, line 57: what is this one?

Gemfile
lvreynoso: It's not needed, I just inserted the latest version. Good point

app/assets/src/components/ui/containers/toast_container.scss
lvreynoso: Sass started complaining about this after I updated it - `Invalid parent selector ":global(.Toastify__toast)"`.

tfrcarvalho: Wouldn't that be an error on the selector above? Was the error not specific?

app/views/layouts/application.html.erb
lvreynoso: It should...but it wasn't. The `@import url(...)` declaration showed up in the compiled CSS, but for some reason the browser was not making the network requests to Google fonts that it should. I went through the development tools for both my local instance and staging side by side to see why and the applied styles were identical, yet Chrome was rendering Helvetica Neue on the local instance. This was a real headache to debug, I tried updating Sass, tried to migrate modules forward, to no effect. Adding this line finally fixed it, but more investigation needs to be done about our Sass system (some parts of it are ancient) and webpack config.

tfrcarvalho: Yeah, this was my blocker. Let's add a comment and a Jira tasks to understand this. This seems like a side fix that is going to come back to haunt us.

package.json
lvreynoso: A support library for `sass` (the dart version works synchronously) to play nice with node (which does stuff asynchronously). It was recommended by the sass-loader documentation.

Also it turns out we were no longer using `extract-text-webpack-plugin`.

app/assets/src/components/ui/containers/toast_container.scss
lvreynoso: It said the above and specified this line (12). Perhaps I misread it

tfrcarvalho: Probably not if this fixed it! Just found it curious how the error gets reported.

webpack.config.common.js
```diff
@@ -75,7 +75,6 @@ const config = {
           {
             loader: "sass-loader",
             options: {
-              minimize: true,
```
tfrcarvalho, line 78: What happened here? Is it not minimizing?

lvreynoso: Sass-loader 8 doesn't do minimizing, so it's not a valid option. It wasn't doing anything before I believe; the `cssnano` plugin for `postcss-loader` does the bulk of the work.

tfrcarvalho: ok, interesting

lvreynoso: Yeah, the css is the same. `dist/main.bundle.min.js` is 50 kb bigger.

app/models/metadatum.rb
```diff
@@ -234,7 +234,7 @@ def self.bulk_load_import(to_create)
       # The unique key is on sample and metadata.key, so the value fields will
       # be updated if the key exists.
       update_keys = [:raw_value, :string_validated_value, :number_validated_value, :date_validated_value, :location_id]
-      results = Metadatum.import to_create, on_duplicate_key_update: update_keys
+      results = Metadatum.bulk_import to_create, on_duplicate_key_update: update_keys
```
tfrcarvalho, line 237: There are more imports. Found these two:
https://github.com/chanzuckerberg/idseq-web/blob/master/app/models/pipeline_run.rb#L758
https://github.com/chanzuckerberg/idseq-web/blob/master/app/jobs/data_migrations/populate_contig_species_taxids.rb#L37

lvreynoso:
Can log into sandbox now :) still need to fix a few things, it appears

PR #3293, opened 2020-04-22
Remove alignment chunk columns

morsecodist:
# Description

These columns currently have no usages. I kept them in to ease the transition to the new alignment implementation but now that that is running stably we can drop these columns.

I know the last time I migrated it didn't go well so here are some reasons why I feel this will go better:
- The timeout happened as a result of renaming while this is dropping, which is a faster operation
- I did a global search for usages of the column names (I did that last time but just including for thoroughness)
- The string mismatch issue that happened last time was a result of changing values within a column not the column name itself. It would be much more unlikely that we have a dependency on the column name that doesn't show up anywhere in the code.
- The uninitialized constants issue only happens when you use class methods but this migration does not.
- The issue of the pipeline monitor and results monitor spitting out errors won't happen this time because the pipeline monitor no longer uses these values at all.

# Tests

Ran the migration locally and tested a sample.


morsecodist:
@gregdingle to go back now would be a huge undertaking. Unfortunately since this touches so many pieces it is difficult to change in a way that is reversible without leaving a total mess.

PR #3298, opened 2020-04-23
Fix bug with admin-only metadata auto-populate.

MarkAZhang:
# Description

The feature previously relied on a metadata object already existing
for each sample, which only occurred if the water control field was
present in the project and was getting set to a default value.
This fix removes that unintended constraint.

# Tests

* Verified that feature still works as intended (samples can still be submitted successfully) and the bug is fixed.

### DIFF
--- a/app/assets/src/components/common/MetadataManualInput.jsx
+++ b/app/assets/src/components/common/MetadataManualInput.jsx
@@ -160,13 +160,21 @@ class MetadataManualInput extends React.Component {
   };
 
   autoPopulateMetadata = () => {
+    const { samples } = this.props;
     const { metadataFieldsToEdit, headersToEdit } = this.state;
 
+    let newMetadataFieldsToEdit = {};
+
+    // For each sample, merge auto-populate fields into existing fields (which may be empty).
+    // Existing fields take precedence.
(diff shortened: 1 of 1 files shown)


app/assets/src/components/common/MetadataManualInput.jsx
```diff
@@ -160,13 +160,21 @@ class MetadataManualInput extends React.Component {
...
+    // Existing fields take precedence.
+    samples.forEach(sample => {
+      newMetadataFieldsToEdit[sample.name] = merge(
+        AUTO_POPULATE_FIELDS,
+        metadataFieldsToEdit[sample.name] || {}
+      );
+    });
+
     const newHeadersToEdit = union(headersToEdit, keys(AUTO_POPULATE_FIELDS));
-    const newMetadataFieldsToEdit = mapValues(
```
MarkAZhang, line 166: The previous code applied a change to values that were already in `metadataFieldsToEdit`. But unless something makes a change to metadata (such as setting default fields for the water control field), there will be no values in `metadataFieldsToEdit`, so the previous code would do nothing.

The new code iterates through the names of the samples that are being uploaded.

PR #3299, opened 2020-04-24
Index generation tweaks

morsecodist:
# Description

- Modifies the `create_alignent_config` rake task to get `nr` and `nt` from `ncbi-sources` instead of `alignment-data` since the new index generation no longer redundantly copies these. This copying was only ever needed because idseq-dag was used for the generation and it needed the input in the same directory.
- Modifies the `create_alignent_config` rake task to use the `.db` file extension for `nt_info`. I am creating a `.db` version of all of these now, this is the merge blocker. This file never had a shelve (`.db`) version which is why the copying is needed. It is the last one using `.sqlite3` and we can retire a lot of code by making this switch. Also the new index generation no longer makes a `.sqlite3` version at all so this is needed to use the new index generation.
- Modifies the default `nt_info` path for the same reasons as above
- Migrates the alignment configs historically to use the new `.db` copies. This is in case we ever want to use an old config.

# Notes

Waiting on creating copies the .sqlite3 indexes to the shelve format to merge this PR.

Migrations notes:

This is yet another migration, so here's why we won't get issues while we did that one time.
- There are MANY fewer records in the `alignment_config` table so timeouts won't be an issue.
- The `AlignmentConfig` class was used in [another migration](https://github.com/chanzuckerberg/idseq-web/blob/master/db/migrate/20190911200352_pipeline_config_change_sqlite_to_shelve.rb) without issue so that dependency constants issue likely won't happen again. 
- Both values before and after the conversion are supported equally so we won't get any rogue failures during the transition phase.

# Tests

Ran Migration locally. Tested a sample using the master branch of idseq-dag. Tested a sample using a branch of idseq-dag that does not support sqlite.


lvreynoso:
Looks good! You can also do "[Create Draft PR](https://github.blog/wp-content/uploads/2019/02/draft-pull-requests.png?w=1354)" in github to prevent accidental merging.

morsecodist:
> I can't think of when an old config would be used in normal use. Re-runs would use the new config.

Interesting, I guess so. I think it would be clearer to remove references to file types that are no longer supported.

PR #3300, opened 2020-04-24
Surfaces pipeline version on DD Samples Table and as a warning on the heatmap

tfrcarvalho:
# Description

* Add a column for pipeline version on Data Discovery's Samples Table.
* Adds a warning on the heatmap when samples use different pipeline versions.

Screen Shot 2020-04-24 at 3 26 02 PM

Screen Shot 2020-04-24 at 3 24 56 PM

# Tests

* Load a samples table and verify that:
	* You can select pipeline version column
	* Pipeline version shown is correct
	* The header as a proper tooltip
* Create a heatmap with samples with different pipeline version for the last run and verify that:
	* It shows a notification about the different pipeline versions
	* The notification mentions the correct pipeline versions
	* The notification can be dismissed
	* The notification disappears automatically after ~12s


gregdingle:
Looks good based on description except for one point that I think needs discussion. When the heatmap warning was originally conceived, it was for **major version** differences only. I expect minor version warnings will be >10x more common than major. (We could easily quantify this with a SQL query, by the way). 

See https://github.com/chanzuckerberg/idseq-dag/

> We increase X for a paradigm shift in how the pipeline is conceived. Example: adding a de-novo assembly step and then reassigning hits based on the assembled contigs. 

> We increase Y for a change that adds results, changes results, or has the potential to change results.

tfrcarvalho:
> Looks good based on description except for one point that I think needs discussion. When the heatmap warning was originally conceived, it was for **major version** differences only. I expect minor version warnings will be >10x more common than major. (We could easily quantify this with a SQL query, by the way).
> 
> See [chanzuckerberg/idseq-dag](https://github.com/chanzuckerberg/idseq-dag/)
> 
> > We increase X for a paradigm shift in how the pipeline is conceived. Example: adding a de-novo assembly step and then reassigning hits based on the assembled contigs.
> 
> > We increase Y for a change that adds results, changes results, or has the potential to change results.

I don't think how common it is is important. If it "changes results" then it definitely needs a warning (maybe not as obvious though).

tfrcarvalho:
To avoid alarmist warnings we will not be reporting minor version discrepancies using the notification warning.
Will still be important to somehow surface those differences in a future PR.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -77,6 +77,7 @@ const BACKGROUND_METRICS = [
 const NOTIFICATION_TYPES = {
   invalidSamples: "invalid samples",
```
j-x-han, line 78: nit: if using underscores as the new convention, would you mind refactoring the other notification types to be the same format?

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -415,6 +416,20 @@ class SamplesHeatmapView extends React.Component {
       this.fetchMetadataFieldsBySampleIds(),
     ]);
 
+    const pipelineMajorVersionsSet = new Set(
+      compact(
+        map(data => {
+          if (!data.pipeline_version) return null;
```
j-x-han, line 422: When are pipeline versions null? I noticed some older samples that displayed v2.11 on the DD table were missing the pipeline version here, so the warning didn't display on the heatmap.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -1199,21 +1214,54 @@ class SamplesHeatmapView extends React.Component {
     );
   }
 
+  renderFilteredMultiplePipelineVersionsWarning(onClose, versions) {
+    return (
+      <Notification type={"warn"} displayStyle={"elevated"} onClose={onClose}>
+        <div>
+          <span className={cs.highlight}>
+            The selected samples come from multiple major pipeline versions:{" "}
+            {versions.join(", ")}.
```
j-x-han, line 1223: nit: I think there's a space missing after this sentence

MarkAZhang: I would add `Contact us for help` with a link at the end of this message, because I don't believe users can currently re-run samples on their own.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapView.jsx
```diff
@@ -415,6 +416,20 @@ class SamplesHeatmapView extends React.Component {
...
 
+    const pipelineMajorVersionsSet = new Set(
+      compact(
+        map(data => {
+          if (!data.pipeline_version) return null;
+          return `${data.pipeline_version.split(".")[0]}.x`;
+        }, heatmapData)
+      )
+    );
+    if (pipelineMajorVersionsSet.size > 1) {
```
MarkAZhang, line 427: It looks like this notification is shown over and over every time the user changes their filters, even if they are still viewing the same samples. Perhaps we should only show the notification once per component lifecycle?

Also, a nit that perhaps this logic should be separated out into a function `checkPipelineVersions()`.

PR #3304, opened 2020-04-28
[Release Fix] Fix link styling and name collision

lvreynoso:
# Description

Fixes a problem where the link styling on the front page banner for the public website defaulted to blue, making the text invisible.

# Notes

Avoided a collision with materialize.

# Tests

Head to landing page; make sure covid banner text is visible. Log in and browse through IDSeq, make sure other link styles are not changed.

### DIFF
--- a/app/assets/src/components/ui/controls/Link.jsx
+++ b/app/assets/src/components/ui/controls/Link.jsx
@@ -25,11 +25,11 @@ class Link extends React.Component {
   };
 
   render() {
-    const { href, className, children, textColor, external } = this.props;
+    const { href, className, children, external } = this.props;
     return (
       <a
         href={href}
-        className={cx(cs.link, className, textColor)}
+        className={cx(cs.link, className)}
         target={external ? "_blank" : null}
         rel="noopener noreferrer"
         onClick={this.onClick}
@@ -45,7 +45,6 @@ Link.propTypes = {
   className: PropTypes.string,
   external: PropTypes.bool,
   href: PropTypes.string,
-  textColor: PropTypes.string,
 
   // We intentionally don't have an onClick prop, because we don't want to encourage arbitrary onClick handlers,
   // since there is already an on-click behavior (following the link href). logAnalyticsEvent is an exception.
@@ -56,7 +55,6 @@ Link.propTypes = {
 Link.defaultProps = {
   analyticsEventData: {},
   externalLink: false,
-  textColor: "primary-light",
 };
 
 export default Link;
--- a/app/assets/src/components/ui/controls/Link.jsx
+++ b/app/assets/src/components/ui/controls/Link.jsx
@@ -25,11 +25,11 @@ class Link extends React.Component {
   };
 
   render() {
-    const { href, className, children, external } = this.props;
+    const { href, className, children, textColor, external } = this.props;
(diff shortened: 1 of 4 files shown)


MarkAZhang:
I'm a bit confused about this PR and am still looking into it. Just wanted to comment that I'm looking into it so it doesn't get merged!

lvreynoso:
The link on the landing page is an `ExternalLink`, which is a wrapper around the `Link` component. In a previous PR [Tiago made sure this component was styled blue](https://github.com/chanzuckerberg/idseq-web/pull/3252/files#diff-76037c2b9e28d0ab348eec1680714734R4) and unfortunately this overrode the styling specified in the landing page. I decided to create a prop to specify the link color.

lvreynoso:
Handy visual of the issue being fixed:

Screen Shot 2020-04-29 at 12 09 51 PM

MarkAZhang:
Hey @lvreynoso, there are a couple of issues with this PR that suggest it isn't doing what it looks like it's doing:
* the `color` prop in `Landing.jsx` seems like it should be `textColor` which is the prop you added in `Link.jsx`.
* If you pass in `textColor="primary-light"` and then do `className={textColor}` the literal string `primary-light` will be added as a CSS class. However, the actual classname in `link.scss` is something like `primary-light-abc123` because of the hash added by CSS modules. So adding the literal string `primary-light` as a class would not apply those CSS values. The correct way to apply the hashed classname is to use the `cs` object with `className={cs[textColor]}`. When you do `import cs from "./landing.scss"`, `cs` is an object that maps from the string `primary-light` to the hashed classname `primary-light-abc123`.

The fact that the fix works in spite of these issues suggests that you inadvertently fixed the issue with some portion of the PR. What I think happened is that in `link.scss`, you are no longer ever applying the `color: $primary-light` or `color: $white` rules, and the link is coincidentally turning the correct color due to the color it's inheriting from some parent element. I'm a little concerned that not applying `color: $primary-light` may have affected other links on the site that Tiago was originally trying to fix. (it's possible you may have accidentally undid his change)

I think the best approach here is to add CSS to the `.link` class in `Landing.jsx` to make that particular link white and avoid adding any new props. You mentioned there is an issue with `!important` in materialize but I believe there is a way around this that I can show you if you think that's the best way forward.

lvreynoso:
Thanks Mark, for some reason I thought the link color was already specified in the CSS. Adding that one line fixed it, and I've removed the other changes.

MarkAZhang:
@lvreynoso I'm glad! I remember when I was looking at it earlier that `!important` was needed. I just checked, and for my machine at least, `color: white` doesn't work without `!important`. Is that the case for you as well?

lvreynoso:
Hmm I thought I was seeing it without using `!important` but it looks like you're right. Apologies, my setup is a little wonky.

I was really hoping that I didn't have to do so, because I don't want to introduce more tech debt since reworking our CSS is not exactly the highest priority; but at the same time trying to do a better fix seems outside the scope of a release fix. What do you think? Should we just got with `!important` here?

MarkAZhang:
Yeah, `!important` is fine for this case. I would just add a comment `override materialize.css` or `override styling in <Link>` or whatever the case may be.

PR #3317, opened 2020-04-30
Allow operator to deposit custom file on user Downloads page

cdebourcy:
# Description
This is intended as a mechanism for an operator to securely expose a file to a user for download via a rake task. The file appears as a Bulk Download of type "Customer Support Request".

image

# Tests
- Verify that running the rake task allows successful addition of a file to the Downloads page for only the intended user.
- Verify that the description of "Customer Support Request" type downloads appears in the Details sidebar on the Downloads page.
- Verify that all previously existing Bulk Download types work exactly the same as before.

# Tests for QA
- Verify that all previously existing Bulk Download types work exactly the same as before.


MarkAZhang:
FYI currently reviewing this. This expands the definition of what a BulkDownload is, so thinking carefully about the implications.

app/helpers/bulk_download_types_helper.rb
```diff
@@ -14,12 +15,18 @@ module BulkDownloadTypesHelper
   RESQUE_EXECUTION_TYPE = "resque".freeze
   VARIABLE_EXECUTION_TYPE = "variable".freeze
   ECS_EXECUTION_TYPE = "ecs".freeze
+  MANUAL_EXECUTION_TYPE = "manual".freeze
```
MarkAZhang, line 18: Can we call this `MANUAL_UPLOAD_TYPE` since the admin is just uploading a file, and nothing is being executed at all?
Also add a comment explaining that "Bulk downloads with this type are not auto-generated and the kickoff function will do nothing. An admin operator will need to manually upload a file to the s3 location designated by download_output_key before calling mark_success."

app/helpers/bulk_downloads_helper.rb
```diff
@@ -43,7 +43,7 @@ def get_valid_pipeline_run_ids_for_samples(samples)
 
   def format_bulk_download(bulk_download, detailed: false, admin: false)
     formatted_bulk_download = bulk_download.as_json(except: [:access_token])
-    formatted_bulk_download[:num_samples] = bulk_download.pipeline_runs.length
+    formatted_bulk_download[:num_samples] = bulk_download.pipeline_runs.empty? ? "NA" : bulk_download.pipeline_runs.length
```
MarkAZhang, line 46: We should hide the "Number Samples" and "Samples in this Download" sections on the bulk download details sidebar UI if there are no pipeline runs. Instead of showing "NA".

lib/tasks/create_customer_support_download.rake
```diff
@@ -0,0 +1,50 @@
...
+  puts "\n(Note the quotes around the destination path to take care of spaces in the destination file name.)"
+  is_file_uploaded = await_confirmation("Have you uploaded your file? Please confirm once you're ready",
+                                        "Process aborted, deleting the draft BulkDownload")
+  unless is_file_uploaded
+    bulk_download.destroy
+    exit 2
+  end
+
+  begin
+    bulk_download.mark_success
```
MarkAZhang, line 34: I would rename `mark_success` to `verify_and_mark_success` and add a comment in that function that `fetch_output_file_size` will raise an error if the file isn't found. Otherwise, the code is quite confusing to read.

If you're open to paying down a bit of technical debt, it would be great to move these lines from `generate_download_file` into `mark_success`:

```
  rescue
    update(status: STATUS_ERROR)
    raise
  end
```

and renaming `mark_success` to `finalize`. It's a little weird that `mark_success` can raise an error and it expects the caller to handle it.

app/controllers/bulk_downloads_controller.rb
```diff
@@ -22,6 +22,9 @@ def types
       end
     end
 
+    # Filter out types that should not be part of the creation UI because they are created offline by admins
+    download_types = download_types.reject { |type| type[:execution_type] == MANUAL_EXECUTION_TYPE }
```
MarkAZhang, line 26: I would add a new property to the types in `BulkDownloadTypesHelper` called `hide_in_creation_modal` and setting `hide_in_creation_modal: true` for CUSTOMER_SUPPORT_BULK_DOWNLOAD_TYPE, instead of attaching this filtering behavior to the MANUAL_EXECUTION_TYPE execution type.

Ideally, each property should have a single clear purpose.

lib/tasks/create_customer_support_download.rake
```diff
@@ -0,0 +1,50 @@
...
+
+  puts "Please enter a user-facing description of the file you wish to transfer:"
+  description = ActionController::Base.helpers.strip_tags(STDIN.gets.strip)
+
+  bulk_download = BulkDownload.create(download_type: "customer_support_request",
+                                      status: BulkDownload::STATUS_WAITING,
+                                      description: description,
+                                      user_id: user.id)
+  puts "BulkDownload record of type customer_support_request has been created."
+  s3_uri = "s3://#{ENV['SAMPLES_BUCKET_NAME']}/#{bulk_download.download_output_key}"
```
MarkAZhang, line 22: Because of how `download_output_key` is currently implemented, the admin operator will need to name their file `Customer Support Request.tar.gz`, regardless of what the file is. This is because BulkDownloads originally assumed that all files would be like this.

Since this is an admin-only feature, maybe this is okay, but I have ideas for how to make this better. Let me know if you want to discuss.

cdebourcy:
Thanks for the great review @MarkAZhang! I worked offline with @happyimadesignr to arrive at the current design. Will take a look at your other comments now

app/helpers/bulk_download_types_helper.rb
cdebourcy: Good point, I made the change and added the comment

app/helpers/bulk_downloads_helper.rb
cdebourcy: Great point, I made the change to hide the relevant sections. I do still need to handle the edge here in this line to prevent the "Samples" column from showing "0" in the main Downloads table (not Details sidebar). However, I changed my mind from "NA" to just leaving the field empty, how does that sound? (cc @happyimadesignr)

lib/tasks/create_customer_support_download.rake
cdebourcy: This is a great point. I have renamed the method to `verify_and_mark_success` and added the comment next to `fetch_output_file_size`.

I have not made the change of moving the rescue block into the `mark_success` method because this would also mean that errors in the main block of `generate_download_file` are no longer caught, with potential side effects I cannot adequately assess. Prefer to leave a change of that logic to someone who is more intimately familiar with that code!

app/controllers/bulk_downloads_controller.rb
cdebourcy: Great point, I've made that change

lib/tasks/create_customer_support_download.rake
cdebourcy: Very true. I'd love to hear your ideas on this! I see it as becoming especially problematic when the operator seeks to transfer a file that's not actually a tar.gz -- in this case, the extension could be very confusing for the user

app/helpers/bulk_downloads_helper.rb
cdebourcy: image

image

MarkAZhang:
One solution to `download_output_key` is to add a new field on `BulkDownload` called `custom_file_name`. In the `download_output_key` function, you would then use this file name in place of the default one if it is set.

If you did this, you might allow the user to set the custom_file_name as part of `create_customer_support_download.rake`. You could even go as far as to make uploading the file to S3 part of the Rake task script, i.e. "please supply the path to your file" and then the Rake uploads the file to the correct location in s3 on the user's behalf, and also sets the `custom_file_name`.

Note that the "Advanced Download" option that creates a terminal command for the user to use assumes that the file is a `tar.gz` so you will have to modify that if you decided to take this approach (perhaps actually check if the file has `tar.gz` extension).

PR #3319, opened 2020-05-01
Update kind-of to ^6.0.3 and jQuery to ^3.5.0 to resolve security vulns

lvreynoso:
# Description

Updates kind-of (not used directly, but by dependencies) to ^6.0.3 and jQuery to ^3.5.0 to resolve security vulnerabilities. Refreshed package-lock.json.

# Notes

All dependencies updated minimist to at least ^1.2.5, and eslint-scope to at least 3.7.1, obviating the need for a dependency resolution entry.

extract-text-webpack-plugin was removed since it is no longer in use.

# Tests

* Webpack builds succeed
* Smoke testing the website revealed nothing different.
* Deployed to sandbox environment, no differences detected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,6 +133,7 @@
     "webpack-node-externals": "^1.7.2"
   },
   "resolutions": {
+    "//": "Avoids using insecure versions of kind-of as of 2020/05/01",
     "kind-of": "^6.0.3"
   },
   "snyk": true
--- a/package.json
+++ b/package.json
@@ -1,5 +1,7 @@
 {
   "name": "idseq-web",
+  "license": "MIT",
+  "repository": "https://github.com/chanzuckerberg/idseq-web",
   "scripts": {
     "preinstall": "npx npm-force-resolutions",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -132,9 +133,7 @@
     "webpack-node-externals": "^1.7.2"
   },
   "resolutions": {
-    "eslint-scope": "3.7.1",
-    "//": "Avoids use of insecure version of minimist as of 2020/03/30. Revisit.",
```
tfrcarvalho, line 136: Hey, lets keep a similar comment for the new resolution you added.

lvreynoso: Good idea :)

lvreynoso:
I could try - it's only used in three components. I'll take a stab at it and see what happens :)

lvreynoso:
Looking at it more closely, I think that's best saved for another PR since it would affect some critical functionality (e.g. login). But I'll put up a ticket for it.

PR #3320, opened 2020-05-01
Add ability to configure subsample and max_input_fragment per project.

MarkAZhang:
# Description
Previously, we were using AppConfig to control custom `subsample` and `max_input_fragments`.
The downside of this approach is that you only get a single custom value that applies to all projects in a "whitelist".

At the time, we thought that maybe only a single project would need this custom value temporarily. However, at this time there are now multiple projects in the "subsample whitelist".

This changes the approach to be more intuitive by adding a `subsample_default` and `max_input_fragment_default` field on the project, allowing each project to be configured independently.

# Notes
- Admin options, if they are present, will still take precedence over the project defaults.
- The rollout plan is to move the AppConfig values over to the defaults for individual projects once this PR lands in production and then immediately deprecate and remove the AppConfig mechanism. 
- Under this approach, the user would first create the project they intend to upload to. An admin would then need to go in and change the project defaults on their behalf. Any samples uploaded to this project would then get the project defaults.

# Tests

* Wrote unit tests for the various scenarios.
* Also verified manually that samples get the expected defaults in various situations.

### DIFF
--- a/db/migrate/20200501044034_add_subsample_to_project.rb
+++ b/db/migrate/20200501044034_add_subsample_to_project.rb
@@ -1,0 +1,8 @@
+class AddSubsampleToProject < ActiveRecord::Migration[5.1]
+  def change
+    change_table :projects, bulk: true do |t|
+      t.integer :subsample_default, comment: "The default value of subsample for newly uploaded samples. Can be overridden by admin options."
+      t.integer :max_input_fragments_default, comment: "The default value of max_input_fragments for newly uploaded samples. Can be overridden by admin options."
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


db/migrate/20200501044034_add_subsample_to_project.rb
```diff
@@ -0,0 +1,8 @@
+class AddSubsampleToProject < ActiveRecord::Migration[5.1]
+  def change
+    change_table :projects, bulk: true do |t|
```
tfrcarvalho, line 3: We should discuss this pattern on future web meeting. I think we should start moving away from constantly extending these base tables and move these to extra columns into association tables. Alternatively, invest on JSON fields for these attributes (for which we should probably consider updating the DB to a version that supports it (although managed Aurora does not support MySQL 8).

PR #3321, opened 2020-05-01
Restyle coverage viz bottom sidebar error message for consistency.

MarkAZhang:
# Description

This is a follow-up to https://github.com/chanzuckerberg/idseq-web/pull/3315.
Jenn made a comment on the design of the error message. Since the issue was blocking the deploy, I moved the re-styling to this follow-up PR.

Before:
Screen Shot 2020-05-01 at 2 25 55 PM

After:
Screen Shot 2020-05-01 at 3 03 28 PM

# Notes

* There are enough differences between the other "no data message" and this one that I decided to just duplicate the html markup. The styling is shared between the two instances.

# Tests
* Verified that the message looks as expected and the styling of other links is unaffected.

### DIFF
--- a/app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
+++ b/app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
@@ -427,7 +427,7 @@ export default class CoverageVizBottomSidebar extends React.Component {
         <div className={cs.headerControls}>
           <div className={cs.vizLinkContainer}>
             <a
-              className={cs.vizLink}
+              className={cs.linkWithArrow}
               href={params.alignmentVizUrl}
               target="_blank"
               rel="noopener noreferrer"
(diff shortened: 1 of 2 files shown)


happyimadesignr:
Nice! thanks @MarkAZhang!

app/assets/src/components/common/CoverageVizBottomSidebar/CoverageVizBottomSidebar.jsx
```diff
@@ -467,13 +467,32 @@ export default class CoverageVizBottomSidebar extends React.Component {
...
+                    {
+                      accessionId: currentAccessionSummary.id,
+                      taxonId: params.taxonId,
+                      sampleId,
+                    }
+                  )
+                }
+              >
+                Contact us for help
+                <i className={cx("fa fa-chevron-right", cs.rightArrow)} />
```
lvreynoso, line 492: An `<i />` tag doesn't make any sense here - perhaps go with a `<span />`?

MarkAZhang: Great point!

The <i> tag has been used by Font Awesome for icons. Example: https://fontawesome.com/v4.7.0/icon/tag.
There doesn't seem to be a great reason for it. https://stackoverflow.com/questions/21559816/why-does-font-awesome-recommend-using-i-elements-instead-of-span-elements-i

However, since we use <i> tags elsewhere in our app for icons, I think we should keep the <i> tag for consistency.

PR #3322, opened 2020-05-02
Bump minimum idseq-cli version to 0.8.8

MarkAZhang:
# Description

Bump CLI version so users will have access to the bug fix here: https://github.com/chanzuckerberg/idseq-cli/pull/69.


MarkAZhang:
That's a good suggestion! Will do.

PR #3327, opened 2020-05-05
[IDSEQ-2221] Generate presigned s3 url for downloads

j-x-han:
# Description

See IDSEQ-2221. Replace sample download links with presigned S3 urls

# Notes

I found 3 download types (non-host reads, non-host contigs, unmapped reads) using the previous bad pattern of copying S3 files to rails prior to serving them. Let me know if there are any I missed!

image

# Tests

* Verify that the download links in the sample report sidebar download the expected file.
* Verified that attempting to download a file that doesn't exist doesn't result in the page crashing.


app/controllers/samples_controller.rb
```diff
@@ -991,8 +991,13 @@ def contigs_fasta
...
 
     if contigs_fasta_s3_path
-      @contigs_fasta = get_s3_file(contigs_fasta_s3_path)
-      send_data @contigs_fasta, filename: @sample.name + '_contigs.fasta'
+      filename = @sample.name + '_contigs.fasta'
+      @contigs_fasta_url = get_presigned_s3_url(contigs_fasta_s3_path, filename)
+      if @contigs_fasta_url
+        redirect_to @contigs_fasta_url
+      else
+        send_data nil, filename: filename
```
lvreynoso, line 999: If `get_presigned_s3_url()` returns `nil`, the user will get a 0 byte file download, is that correct? Would it be better to render an error instead?

gregdingle: at least we should log the error

app/helpers/pipeline_outputs_helper.rb
```diff
@@ -189,6 +189,26 @@ def get_s3_file(s3_path)
     end
   end
 
+  def get_presigned_s3_url(s3_path, filename)
+    s3 = Aws::S3::Resource.new(client: Client)
+    uri_parts = s3_path.split("/", 4)
```
cdebourcy, line 194: I saw the code that parses the URIs repeated in https://github.com/chanzuckerberg/idseq-web/blob/e70919eec874eecfdaa6d324959203279be7817d/app/controllers/samples_controller.rb#L1012 -- worth factoring it out into one of the helpers?

app/controllers/samples_controller.rb
j-x-han: Good call. Changed to render and log an error instead.

app/helpers/pipeline_outputs_helper.rb
j-x-han: Done! I found and replaced some other places where I saw this repeated.

app/controllers/samples_controller.rb
```diff
@@ -1010,9 +1019,7 @@ def contigs_summary
 
   # TODO(mark): Factor out into S3Helper file.
```
gregdingle, line 1020: might as well change this TODO

app/helpers/s3_helper.rb
```diff
@@ -0,0 +1,8 @@
+module S3Helper
```
gregdingle, line 1: I know you are responding to Mark's comment but I'd rather see this in `S3Util`. I think it is better to use helpers only when sharing instance state. Then the dependencies are clearer.

j-x-han: Sounds good, I've moved it to `S3Util`!

PR #3329, opened 2020-05-05
[Release fix] Update asset manifest to include application.css

lvreynoso:
# Description

The new version of the `sprockets` gem asks for an explicit declaration of `application.css` in `app/assets/config/manifest.js`. This PR does so.

# Notes

It's unclear if anything actually broke or what triggered the error, since the website continued to function fine.

# Tests

Since this is part of rendering actionview templates, testing endpoints that return html should test this. As noted above, currently rendering appears to continue to work.


lvreynoso:
Fixing this didn't fix the issue with importing font urls unfortunately :/

PR #3331, opened 2020-05-06
[Bug] Prevent upload of metadata csv with missing column names

j-x-han:
# Description

See IDSEQ-2679.

Users can currently upload metadata csvs with blank column headers; if a user then selects this blank option in the metadata dropdown on the heatmap, the page will crash.

This change prevents the user from uploading a metadata csv with missing column headers and also has the heatmap ignore blank metadata options rather than displaying them in the dropdown.

# Tests

* Verified that an error is displayed if a metadata CSV is uploaded with a missing column name.
* Verified that a heatmap will not display blank metadata options in the dropdown.

### DIFF
--- a/app/helpers/metadata_helper.rb
+++ b/app/helpers/metadata_helper.rb
@@ -235,10 +235,10 @@ module MetadataHelper
     warning_aggregator = ErrorAggregator.new
 
     # Require a header for every column.
-    # if (metadata["headers"].include? "")
-    #   errors.push(MetadataValidationErrors::MISSING_COLUMN_HEADER)
-    #   return { "errors" => errors, "warnings" => [] }
-    # end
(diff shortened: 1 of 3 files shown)


app/helpers/metadata_helper.rb
```diff
@@ -234,6 +234,12 @@ def validate_metadata_csv_for_samples(
     error_aggregator.set_metadata("num_cols", metadata["headers"].length)
     warning_aggregator = ErrorAggregator.new
 
+    # Require a header for every column.
+    if metadata["headers"].include? ""
```
lvreynoso, line 238: Could the headers include a nil value, or are they always strings?

j-x-han: I believe they're always converted into strings when the file is read.

PR #3333, opened 2020-05-07
[Release fix] remove sample accordion when necessary

cdebourcy:
# Description
The previous strategy for hiding "Samples in this Download" was not implemented properly so didn't have any effect. Now it is working properly.

# Tests
Downloads of type "Customer Support Request" don't have a "Samples in this Download" accordion in the Details Sidebar.

### DIFF
--- a/app/assets/src/components/common/DetailsSidebar/BulkDownloadDetailsMode/DetailsTab.jsx
+++ b/app/assets/src/components/common/DetailsSidebar/BulkDownloadDetailsMode/DetailsTab.jsx
@@ -55,7 +55,7 @@ export default class DetailsTab extends React.Component {
           )}
           <FieldList fields={fields} />
         </Accordion>
-        {bulkDownload.pipeline_runs && (
+        {bulkDownload.pipeline_runs.length > 0 && (
           <Accordion
             className={cs.accordion}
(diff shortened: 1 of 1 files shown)


app/assets/src/components/common/DetailsSidebar/BulkDownloadDetailsMode/DetailsTab.jsx
```diff
@@ -55,7 +55,7 @@ export default class DetailsTab extends React.Component {
           )}
           <FieldList fields={fields} />
         </Accordion>
-        {bulkDownload.pipeline_runs && (
+        {bulkDownload.pipeline_runs.length > 0 && (
```
tfrcarvalho, line 58: We are sure that pipeline_runs is never undefined, right?

cdebourcy: Yes, previously pipeline_runs was used without any checks: https://github.com/chanzuckerberg/idseq-web/blob/cdcde0dc35cded0cec3a7d51f9d944f7bbdf3454/app/assets/src/components/common/DetailsSidebar/BulkDownloadDetailsMode/DetailsTab.jsx#L59
So I assume this is fine

PR #3341, opened 2020-05-11
[IDSEQ-2509] Searching in DD, if the term is deleted there's no visual cue that it's still being applied

ovalenzuela19:
# Description
Searching in DD, if the term is deleted there's no visual cue that it's still being applied

Added to state* (not prop, sorry got it wrong in commit msg) lastSearchedTerm and onBlur method to fix bug.

# Notes 

# Tests


happyimadesignr:
Looks good to me! Thanks @ovalenzuela19!

tfrcarvalho:
@ovalenzuela19 Please write down the tests, as they also serve as instructions for the QA team. Or refer to the IDseq ticket if that has enough information.

PR #3343, opened 2020-05-12
[IDSEQ-2731] Help Center link is always blue

ovalenzuela19:
# Description

Added the color `$black !important` to `cs.option` to overwrite the blue `cs.link` color.

# Tests
Verified that the changes were successful by extending the user drop down menu on the top right of the screen and confirming that the Help Center link was black and not blue.


app/assets/src/components/common/Header.jsx
```diff
@@ -194,7 +194,8 @@ const AnnouncementBanner = ({ onClose }) => {
               }
             >
               Help Center
-            </ExternalLink>. Happy Holidays!
+            </ExternalLink>
+            . Happy Holidays!
```
ovalenzuela19, line 198: Looks like the linter doesn't want to take my changes :[

app/assets/src/components/common/Header.jsx
```diff
@@ -261,15 +262,15 @@ const UserMenuDropDown = ({
       <BareDropdown.Item
         key="help"
         text={
-          <ExternalLink
+          <a
```
tfrcarvalho, line 265: We still want to use ExternalLink... you should be able to use the classname `option` (passed below) to apply the style you need.

tfrcarvalho:
You can remove the sections you did not use and the italic sentence, which are meant as guidelines only.

app/assets/src/components/common/Header.jsx
ovalenzuela19: @tfrcarvalho I'm not sure what you mean by being able to use the classname option to apply the style I need. Do you mean adding in a `color` property to cs.option?

tfrcarvalho: yes, for color you might need to use the `!important` ( as `color: $black !important;`) due to a legacy css.

ovalenzuela19: I'm reading more about the Classnames package and see that `cs.link` in `Link.jsx` component is overwriting the color to be` $primary-light` (blue)

tfrcarvalho: Yes, that is the default we want to enforce for all links. In this case, it is an exception and you want to overwrite that.

ovalenzuela19: > We still want to use ExternalLink... you should be able to use the classname `option` (passed below) to apply the style you need.

Just wondering, why do we still want to use ExternalLink?

PR #3344, opened 2020-05-13
[IDSEQ-2650] Update spacing variables

j-x-han:
# Description

Update spacing variables according to IDSEQ-2650 (essentially renaming all spacing variables one size larger, e.g. $space-s => $space-m, but values themselves do not change). Visually, nothing should be affected.

# Notes

This change does not include work to replace pixel values with existing spacing values where possible; it's been agreed that we should replace those values incrementally on a component-by-component basis.

# Tests

* Navigated most of the main pages of the site (upload flow, data discovery, sample report, heatmap) and verified that spacing looked normal!


happyimadesignr:
Thanks @j-x-han, it looks good to me!!

I'll let @Janeece take a look also =]

Janeece:
@j-x-han , looks good to me too

PR #3348, opened 2020-05-14
[IDSEQ-2510] The metadata filtering and info buttons are confusing when you're on the visualization tab

ovalenzuela19:
# Description
The filter & info button on the DiscoveryHeader looks like they're active and clickable, but doesn't open when you're on the visualization tab.
If a user applied a filter on the samples tab and then goes to the visualizations tab it appears to have a filter but doesn't actually apply.

Goals: 
1) Remove the badge # from the filter icon 
2) Not having the icon appear clickable on hover. 
3) Repeat for the info icon

[UPDATE]
Create a tooltip that helps describes that Filter and Info buttons are unavailable. 
Check out the [ticket](https://jira.czi.team/secure/RapidBoard.jspa?rapidView=12&projectKey=IDSEQ&view=detail&selectedIssue=IDSEQ-2510) for more info

# Tests
To test that these changes fix the bug:
1. Navigate to _All Data_
2. Switch around between the _Projects, Samples, and Visualizations_ tabs
3. Pay attention to the filters icon and interaction.

changes available in sandbox!


happyimadesignr:
This is fantastic @ovalenzuela19! Thanks for the collab and being so detail-oriented!

app/assets/src/components/views/discovery/DiscoveryHeader.jsx
```diff
@@ -71,30 +71,52 @@ class DiscoveryHeader extends React.Component {
       tabs,
     } = this.props;
 
+    let disabledVisualizations = currentTab === "visualizations";
```
j-x-han, line 74: Would you mind renaming this variable to something like `disableSidebars` instead? Confused me at first glance 😅

ovalenzuela19: Of course! Good suggestion!

PR #3352, opened 2020-05-18
change total_runtime display to overall wait

cdebourcy:
# Description
This is a proposal. I'm curious to hear people's thoughts. 
Previously the runtime displayed for each sample only included time spent running Batch jobs and not the time spend waiting for workers. It would be more accurate to display the overall time from the pipeline creation until pipeline completion.

If people think this is a good change, I can invest the time to clean up the now unnecessary argument to the function.

# Tests
Numbers in the "Total Runtime" column on the project page may increase to account for wait time in addition to compute time.

### DIFF
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -289,12 +289,12 @@ module SamplesHelper
     samples.where(id: matching_sample_ids)
   end
 
-  def get_total_runtime(pipeline_run, run_stages)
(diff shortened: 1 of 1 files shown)


tfrcarvalho:
Technically, this is a bit risky. Any updates to pipeline run would have changed the `updated_at`. We never updated pipeline runs?

cdebourcy:
True. Closing for now!

tfrcarvalho:
The discussion might still make sense...

PR #3355, opened 2020-05-21
Async Elasticsearch

morsecodist:
# Description

This PR fixes issues that caused missing records in elasticsearch. It also makes our elasticsearch indexing asynchronous.

## Issues

These issues occurred because we add things to Elasticsearch with [elasticsearch-model[(https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-model), which relies on Active Record Callbacks. Any modifications that circumvent the Active Record Callback Flow result in missing data in Elasticsearch. 

### Taxon Lineages

These are only updated via the task `update_lineage_db`. About a year ago this task was modified to use raw SQL. We have been missing updates since then. To fix this I added a step to re-index after this job runs. It is a little slow but we can index the whole thing in <30 minutes and this task only runs once ever 28 days so I feel it isn't a huge deal.

### Metadata Bulk Import

To do bulk metadata importing we are using [https://github.com/zdennis/activerecord-import](https://github.com/zdennis/activerecord-import). This circumvents the Active Record Callback Flow. This is necessary to avoid blocking synchronous updates that would render the feature very slow. I created a version of this method that also indexes in Elasticsearch. However, this would still involve blocking synchronous updates without the async change I made.

## Async Elasticsearch

Currently we are blocking considering writes complete on updating the Elasticsearch Index. This slows down all of our database interactions with these tables. It also fails database operations when the re-index is not really required right away and could have been retried. In their docs elasticsearch-model [recommends async updates](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/README.md#asynchronous-callbacks). These callbacks also have logging and alerting so we can debug potential missingness in Elasticsearch in the future, and we can know when to initiate a re-index.

# Tests

Tested all of the relevant models, bulk imports, as well as the rake task.

### DIFF
--- a/app/helpers/elasticsearch_callbacks_helper.rb
+++ b/app/helpers/elasticsearch_callbacks_helper.rb
@@ -1,9 +1,5 @@
-require 'elasticsearch/model'
-
 # Provides Async Callbacks for Elasticsearch Syncing
 module ElasticsearchCallbacksHelper
-  include Elasticsearch::Model
-
   def self.included(base)
     return unless base.ancestors.include?(::ActiveRecord::Base)
     base.class_eval do
@@ -17,8 +13,8 @@ module ElasticsearchCallbacksHelper
     Resque.enqueue(
       ElasticsearchIndex,
       :index,
-      index_name,
-      document_type,
+      __elasticsearch__.index_name,
+      __elasticsearch__.document_type,
       id,
       __elasticsearch__.as_indexed_json
     )
@@ -27,9 +23,9 @@ module ElasticsearchCallbacksHelper
   def async_elasticsearch_delete
     Resque.enqueue(
       ElasticsearchIndex,
-      :index,
-      index_name,
-      document_type,
+      :delete,
+      __elasticsearch__.index_name,
+      __elasticsearch__.document_type,
       id,
       __elasticsearch__.as_indexed_json
     )
--- a/app/helpers/elasticsearch_callbacks_helper.rb
+++ b/app/helpers/elasticsearch_callbacks_helper.rb
@@ -1,5 +1,10 @@
 # Provides Async Callbacks for Elasticsearch Syncing
 module ElasticsearchCallbacksHelper
+  # WARNING: using this with a model means you must ensure activerecord
+  #  callbacks are called on all updates. This module updates elasticsearch
+  #  using these callbacks. If you must circumvent them somehow (eg. using raw
+  #  SQL or bulk_import) you must explicitly update elasticsearch appropriately.
+
   def self.included(base)
     return unless base.ancestors.include?(::ActiveRecord::Base)
     base.class_eval do
--- a/app/helpers/elasticsearch_callbacks_helper.rb
+++ b/app/helpers/elasticsearch_callbacks_helper.rb
@@ -1,0 +1,37 @@
+require 'elasticsearch/model'
+
+# Provides Async Callbacks for Elasticsearch Syncing
+module ElasticsearchCallbacksHelper
+  include Elasticsearch::Model
+
+  def self.included(base)
+    return unless base.ancestors.include?(::ActiveRecord::Base)
+    base.class_eval do
+      after_commit -> { async_elasticsearch_index }, on: :create
+      after_commit -> { async_elasticsearch_index }, on: :update
+      after_commit -> { async_elasticsearch_delete }, on: :destroy
+    end
+  end
+
+  def async_elasticsearch_index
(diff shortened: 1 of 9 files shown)


lib/tasks/update_lineagedb.rake
```diff
@@ -153,6 +153,18 @@ class LineageDatabaseImporter
     import_new_taxid_lineages!
     affected = build_new_taxon_lineages!
     upgrade_taxon_lineages!(affected, noverify)
+    # This is very slow, since this script is using raw SQL the
+    #  elasticsearch index will not be updated, since elasticsearch-model
+    #  relies on callbacks. The safest and easiest way to ensure the
+    #  index is in sync at the end is to refresh the whole thing.
+    #  Since this script is run rarely this shouldn't be a huge deal
+    #  but if it is slow you may want to look into only updating affected ids.
+    import_elasticsearch!
```
gregdingle, line 162: would it easy to update only the retired, updated and new records? those are separated out in this script, and they make up a small percent typically.

app/models/metadatum.rb
```diff
@@ -8,7 +8,7 @@ class Metadatum < ApplicationRecord
 
   if ELASTICSEARCH_ON
     include Elasticsearch::Model
-    include Elasticsearch::Model::Callbacks
+    include ElasticsearchCallbacksHelper
```
gregdingle, line 11: I want to see this declared loudly as a comment whenever `ElasticsearchCallbacksHelper` is included.

>  Any modifications that circumvent the Active Record Callback Flow result in missing data in Elasticsearch.

That should help spread the knowledge in the team.

app/jobs/elasticsearch_index.rb
```diff
@@ -0,0 +1,25 @@
...
+      Elasticsearch::Model.client.delete(index: index_name, type: index_type, id: record_id)
+      Rails.logger.debug("Elasticsearch: Successfully deleted record: #{record_id} source table: #{index_name}")
+    else raise ArgumentError, "Unknown operation '#{operation}'"
+    end
+  rescue Elasticsearch::Transport::Transport::Errors::NotFound
+    Rails.logger.debug("Elasticsearch: Attempted to delete document: #{record_id} from #{index_name} but it was not found")
+  rescue => e
+    Rails.logger.error(e)
+    LogUtil.log_backtrace(e)
+    LogUtil.log_err_and_airbrake("Elasticsearch failed to #{operation} record: #{record_id} source table: #{index_name}, with error: #{e.message}")
```
gregdingle, line 23: nice logging throughout here!

app/models/metadatum.rb
```diff
@@ -247,6 +253,21 @@ def self.bulk_load_import(to_create)
     errors
   end
 
+  # Bulk import safe to use with elasticsearch
+  def self.bulk_import_safe(to_create, opts = {})
```
gregdingle, line 257: why not just override `bulk_import`? it seems like that would be the safest way to ensure good behavior for idseq

app/helpers/elasticsearch_callbacks_helper.rb
```diff
@@ -0,0 +1,33 @@
+# Provides Async Callbacks for Elasticsearch Syncing
+module ElasticsearchCallbacksHelper
+  def self.included(base)
+    return unless base.ancestors.include?(::ActiveRecord::Base)
+    base.class_eval do
```
gregdingle, line 5: fancy ruby 🤩

gregdingle:
Also I'm wondering what is the plan to update this existing prod and staging DBs.

morsecodist:
> Also I'm wondering what is the plan to update the existing prod and staging DBs.

We can just run the import commands at any time after shipping this.

lib/tasks/update_lineagedb.rake
morsecodist: It wouldn't be too hard but it may not necessarily be faster. The total index import does more than just synchronously adding stuff one at a time, it has some batching logic for more efficient bulk updating. I could add a timestamp query maybe but I would need to do more research.

morsecodist: I tested with the date range query and that doesn't work because there are removed records so the only way to proceed would be a large in query or going ID by ID. Going one by one would most likely be slower and the large IN query is also dicey. I don't think it's worth spending more time on this because of how rarely this runs.

gregdingle: Ok sounds good.

PR #3356, opened 2020-05-21
[IDSEQ-2656] The filter for e-value says ">=" but right now it is only ">"

ovalenzuela19:
# Description
The threshold filter says "greater than or equal to" but things that are "equal" are being filtered out in the sample's report.
For more information, visit [IDSEQ-2656](https://jira.czi.team/browse/IDSEQ-2656) 

The difference between this PR and the PR that was reverted is that we are now comparing the threshold filter to the rendered data instead of changing what the rendered data looked like.

# Notes
The bug is caused by the function 
renderNtNrDecimalValues(val, decimalPlaces) in app/assets/src/components/views/SampleViewV2/ReportTable.jsx
 
The line Number(val).toFixed(decimalPlaces || 0)
has an unintended rounding effect that rounds up when given a float greater than 2 decimal places but specify to round to one decimal place. 

For example:
Mastadenovirus has an e-value of 88.27 – but since the value passes through renderNtNrDecimalValues(88.27, 1)
it gets rounded up to 88.3 and is displayed like that – even though the data is stored as 88.27.
 
This is why the threshold filter appears like it doesn't work properly. The same issue applies to every other column and filter so I am sure there are other bugs caused by this same issue. 

# Tests
Create a new threshold filter with any value and apply it.
The bug fix worked if the taxon still appears in the report after the threshold filter was applied.


app/assets/src/components/views/SampleViewV2/ReportTable.jsx
```diff
@@ -586,8 +586,8 @@ class ReportTable extends React.Component {
           rowData.genus
             ? getOr(nullValue, path, rowData)
             : sortDirection === "asc"
-              ? limits[0]
-              : limits[1],
+            ? limits[0]
```
ovalenzuela19, line 589: linter fixed this

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -1006,20 +1020,19 @@ export default class SampleViewV2 extends React.Component {
...
-                  metric={selectedOptions.metric}
-                  nameType={selectedOptions.nameType}
-                  onTaxonClick={this.handleTaxonClick}
-                  sample={sample}
-                  taxa={filteredReportData}
-                  useReportV2Format={true}
-                />
-              </div>
-            )}
+          {view === "tree" && filteredReportData.length > 0 && (
```
ovalenzuela19, line 1023: linter fixed this

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -1082,8 +1095,9 @@ export default class SampleViewV2 extends React.Component {
             </UserContext.Consumer>
           </div>
           {currentTab === "Report" && this.renderReport()}
-          {currentTab === "Antimicrobial Resistance" &&
-            amrData && <AMRView amr={amrData} />}
+          {currentTab === "Antimicrobial Resistance" && amrData && (
```
ovalenzuela19, line 1098: linter fixed this

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -344,8 +344,24 @@ export default class SampleViewV2 extends React.Component {
 
   getTaxonMetricValue = (row, metric) => {
     let parsedMetric = metric.split(":");
+    let metricDecimalPlaces = this.getDecimalPlacesFromMetric(parsedMetric[1]);
     let parsedValue = get(parsedMetric, row);
-    return parsedValue;
+    return parseFloat(parsedValue).toFixed(metricDecimalPlaces || 0);
+  };
+
+  getDecimalPlacesFromMetric = metric => {
```
tfrcarvalho, line 352: This should just be a constant defined on top.

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
tfrcarvalho: We could try to a full code lint update, to just do all at once. Do you want to try that on a follow up?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -344,8 +344,24 @@ export default class SampleViewV2 extends React.Component {
 
   getTaxonMetricValue = (row, metric) => {
     let parsedMetric = metric.split(":");
+    let metricDecimalPlaces = this.getDecimalPlacesFromMetric(parsedMetric[1]);
```
tfrcarvalho, line 347: I wonder if we should just do this on `fetchSampleReportData` just after we loaded the raw data. That way we would do it just once. wdyt?

app/assets/src/components/views/SampleViewV2/ReportTable.jsx
ovalenzuela19: Yeah I think that's a great idea!

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
ovalenzuela19: Yeah definitely I think it's a great idea!

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -348,6 +349,45 @@ export default class SampleViewV2 extends React.Component {
     return parsedValue;
   };
 
+  parseSpeciesMetricDecimalPlaces = species => {
```
tfrcarvalho, line 352: Lets call it `adjustMetricPrecision` ?
If this is only for species what happens with genus (i do not remember the full data structure...)?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -348,6 +349,45 @@ export default class SampleViewV2 extends React.Component {
...
+    let parsedSpeciesMetrics = Object.assign({}, species);
+    let metricDecimalPlaces = {
```
tfrcarvalho, line 354: Please define a constant on top. Eventually we might also have to move it out to keep consistent precision across the

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -348,6 +349,45 @@ export default class SampleViewV2 extends React.Component {
...
+      rpm: 1,
+      count: 0,
+      contigs: 0,
+      contig_r: 0,
+      percent_identity: 1,
+      alignment_length: 1,
+      e_value: 1,
+    };
+
+    Object.keys(species).forEach(outerMetric => {
```
tfrcarvalho, line 366: We use `count_type` to refer to `nt` and `nr` .

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -348,6 +349,45 @@ export default class SampleViewV2 extends React.Component {
     return parsedValue;
   };
 
+  parseSpeciesMetricDecimalPlaces = species => {
+    let parsedSpeciesMetrics = Object.assign({}, species);
```
tfrcarvalho, line 353: Any reason why yo can't edit species directly instead of a new object? would save space...

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -348,6 +349,45 @@ export default class SampleViewV2 extends React.Component {
...
+      count: 0,
+      contigs: 0,
+      contig_r: 0,
+      percent_identity: 1,
+      alignment_length: 1,
+      e_value: 1,
+    };
+
+    Object.keys(species).forEach(outerMetric => {
+      if (outerMetric in metricDecimalPlaces) {
```
tfrcarvalho, line 367: When is this condition true? We should always have either nt or nr...

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: If you us `Object.entries(species).foreach((count_type, metrics) => ` instead you get the `count_type` and the `metrics` immediately. Might make it slightly more readable?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -348,6 +349,45 @@ export default class SampleViewV2 extends React.Component {
...
+        parsedSpeciesMetrics[outerMetric] = parseFloat(
+          species[outerMetric]
+        ).toFixed(metricDecimalPlaces[outerMetric]);
+      } else if (outerMetric === "nt") {
```
tfrcarvalho, line 371: If you use Object.entries as above, this branch can be treated like the bottom one. Shorter code :)

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
ovalenzuela19: Sorry! This was because I was comparing the original species vs the species with adjusted decimal places and forgot to change back to species directly

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
ovalenzuela19: This is true if the metric is not in nr or nt and is in METRIC_DECIMAL_PLACES (which is a metric that we want to adjust the decimal places)

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
ovalenzuela19: a sorted genus just contains a genusTaxId and a children species just contains speciesTaxId. they're both similar in structure:
`sorted genus  (32) [570, 561, 10509, 547, 255204, 92793, 1386, 1827195, 63265, 1743, 590, 59732, 390250, 1705353, 71927, -200541000, 1649459, 5324, 1263, 544, 110932, 1822464, 1301, 265, 1298, 80865, 1912216, 2053, 32008, 469, 48736, 1827]`

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
ovalenzuela19: Here is the structure of a species:
`agg_score: "19195"
category: "bacteria"
common_name: ""
genus_tax_id: 1827
max_z_score: 6.475747921182336
name: "Rhodococcus erythropolis"
nr:
alignment_length: "49.0"
bg_mean: "40"
bg_stdev: "131"
count: "1"
e_value: "21.0"
percent_identity: "100.0"
rpm: "891.3"
z_score: "6.5"
__proto__: Object
nt:
alignment_length: "0.0"
bg_mean: "10"
bg_stdev: "31"
count: "0"
e_value: "0.0"
percent_identity: "0.0"
rpm: "0.0"
z_score: "-100.0"`

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
ovalenzuela19: countType seems a little misleading to me... would metricType be better? and instead of `metrics` would metricValue be better?
`
{
   agg_score: 19195.238841125654
   category: "bacteria"
   common_name: ""
   genus_tax_id: 1827
   max_z_score: 6.475747921182336
   name: "Rhodococcus erythropolis"
   NR: {
      alignment_length: 49
      bg_mean: 39.7436
      bg_stdev: 131.494
      count: 1
      e_value: 21
      percent_identity: 100
      rpm: 891.2655971479501
      z_score: 6.475747921182336
... (NT has same structure of NR)
}
`

ovalenzuela19: The formatting isn't so well here but NR and NT are nested within the species object which is why I think count_type is misleading since it also refers to `agg_score: 19195.238841125654
category: "bacteria"
common_name: ""
genus_tax_id: 1827
max_z_score: 6.475747921182336
name: "Rhodococcus erythropolis"`

ovalenzuela19: > If you us `Object.entries(species).foreach((count_type, metrics) => ` instead you get the `count_type` and the `metrics` immediately. Might make it slightly more readable?

This is really clean! I didn't know about Object.entries()!

tfrcarvalho: `countType` is what we call it across the entire app, so now is not a good time to change it. Also, NT and NR are not metrics types but just the database type we are counting against.`alignment_length`, `rpm` are the metrics..
So:
- `countType` is `nt`, `nr`
- `metrics` is the key `alignment_length`, `rpm`, etc. (and the object with key, values)
- `metricValue` is `6.4757479`, etc.

When you use the entries function it will match adequately.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: I see, forgot about the aggregate score.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -226,14 +238,16 @@ export default class SampleViewV2 extends React.Component {
...
-          return merge(
-            rawReportData.counts[SPECIES_LEVEL_INDEX][speciesTaxId],
-            {
-              highlighted: isHighlighted,
-              taxId: speciesTaxId,
-              taxLevel: "species",
-            }
+          const speciesInfo =
+            rawReportData.counts[SPECIES_LEVEL_INDEX][speciesTaxId];
+          let speciesWithAdjustedMetricPrecision = this.adjustMetricPrecision(
```
tfrcarvalho, line 243: This can be const.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -348,6 +362,33 @@ export default class SampleViewV2 extends React.Component {
     return parsedValue;
   };
 
+  adjustMetricPrecision = species => {
+    Object.entries(species).forEach(([countType, metricValue]) => {
+      if (countType in METRIC_DECIMAL_PLACES) {
+        species[countType] = parseFloat(metricValue).toFixed(
+          METRIC_DECIMAL_PLACES[countType]
+        );
+      } else if (countType === "nt") {
```
tfrcarvalho, line 371: This could almost be an else, but if you really want to do the sanity check do `if (countType === "nt" || countType === "nr")` and you can remove the bottom branch.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -348,6 +362,33 @@ export default class SampleViewV2 extends React.Component {
     return parsedValue;
   };
 
+  adjustMetricPrecision = species => {
+    Object.entries(species).forEach(([countType, metricValue]) => {
```
tfrcarvalho, line 366: Ok, I might have induced you in error. Here we don't yet know if it is a countType or a metricKey... so, lets call it key only.

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
```diff
@@ -348,6 +362,33 @@ export default class SampleViewV2 extends React.Component {
...
+      if (countType in METRIC_DECIMAL_PLACES) {
+        species[countType] = parseFloat(metricValue).toFixed(
+          METRIC_DECIMAL_PLACES[countType]
+        );
+      } else if (countType === "nt") {
+        Object.entries(species.nt).forEach(([countType, metricValue]) => {
```
tfrcarvalho, line 372: Here we know it is `metricKey`. We should also not reuse the variable we defined on the outer loop.

ovalenzuela19: Gotcha! So you mean replacing `countType` with `metricKey` right?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
ovalenzuela19: I am not sure if I understand the comments about `adjustMetricPrecision`. Here we talk about renaming `countType` to `key` - but then refer to replacing the two `else if` statements with one `if (countType === "nt" || countType === "nr")` in the other comment. Did you mean `if (key === "nt" || key === "nr")`?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
ovalenzuela19: Did you mean `if (key === "nt" || key === "nr")`?

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: Yes :)

app/assets/src/components/views/SampleViewV2/SampleViewV2.jsx
tfrcarvalho: Yes, I did not go back to fully fix old comments :)

ovalenzuela19:
Thank you for all of your feedback @tfrcarvalho!!

PR #3357, opened 2020-05-21
Update to Python 3 and Ruby 2.5 Buster image

lvreynoso:
# Description

Installs Python 3 on the Docker image for web. The base image is changed to `2.5-buster` to make Python 3.7 available.

Python 2 is long out of date and now officially unsupported. However the immediate motivation for this change is for the pipeline visualization work. With the pipeline's workflow described in WDL, parsing the workflow via miniwdl requires Python 3.6+, which `2.5-stretch` does not support.

# Notes

Part of pipeline visualization work for step function compatibility.

# Tests

* Server side tests pass, Travis YML updated
* No client side code changed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
         - redis
       before_install:
         # Install python and dependencies
-        - source ~/virtualenv/python2.7/bin/activate
+        - source ~/virtualenv/python3.7/bin/activate
         - pip install -r requirements.txt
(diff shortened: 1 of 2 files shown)


lvreynoso:
There is one Python script we have: `idd2wdl.py` for converting dag json files to wdl files. That script is for Python 3 anyways.

I saw the `mysql-client` package was removed and replaced it with the `default-mysql-client`, which is a MariaDB client. From my understanding, our version of MySQL is so old that we shouldn't have any issues (and I haven't had any so far).

PR #3358, opened 2020-05-21
[IDSEQ-2456] CURL Download option overwrites files with same name

ovalenzuela19:
# Description
Added the bulk_download ID to the tar.gz to prevent overwriting on conflicting files with same name – although overwriting of the same file with the same name + id will happen.

update: replaced the id with a timestamp

# Notes
I have searched through CURL's & tar's manuals and there's no functionality of incremental rename of files. One way to achieve the incremental rename of files would be to create a bash script (that each user would need to have).

# Tests

1. Go to downloads
2. Open details of file of these files
3. Select Advanced download and copy the URL
4. Open a terminal and paste the CURL pipeline
5. Perform steps 1-4 for a different download with the same file name that you just downloaded

### DIFF
--- a/app/assets/src/components/common/DetailsSidebar/BulkDownloadDetailsMode/AdvancedDownloadTab.jsx
+++ b/app/assets/src/components/common/DetailsSidebar/BulkDownloadDetailsMode/AdvancedDownloadTab.jsx
@@ -45,7 +45,7 @@ export default class AdvancedDownloadTab extends React.Component {
     }
 
     let currentTimestamp = moment().format("MM-D-YYYY hh-mm-ssa");
-    let bulkDownloadFileName = `${bulkDownload.download_name} - ${currentTimestamp}`;
+    let bulkDownloadFileName = `${bulkDownload.download_name}-${currentTimestamp}`;
 
     return `curl -L "${bulkDownload.presigned_output_url}" > "${bulkDownloadFileName}.tar.gz"\
         && mkdir -p "${bulkDownloadFileName}"\
--- a/app/assets/src/components/common/DetailsSidebar/BulkDownloadDetailsMode/AdvancedDownloadTab.jsx
+++ b/app/assets/src/components/common/DetailsSidebar/BulkDownloadDetailsMode/AdvancedDownloadTab.jsx
@@ -43,13 +43,11 @@ export default class AdvancedDownloadTab extends React.Component {
       return "Failed to generate command. Please contact us for help.";
     }
 
-    return `curl -L "${bulkDownload.presigned_output_url}" > "${
-      bulkDownload.download_name
-    }.tar.gz"\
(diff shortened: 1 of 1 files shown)


app/assets/src/components/common/DetailsSidebar/BulkDownloadDetailsMode/AdvancedDownloadTab.jsx
```diff
@@ -43,13 +43,11 @@ export default class AdvancedDownloadTab extends React.Component {
...
     }
 
-    return `curl -L "${bulkDownload.presigned_output_url}" > "${
-      bulkDownload.download_name
-    }.tar.gz"\
-        && mkdir -p "${bulkDownload.download_name}"\
-        && tar -zvxf "${bulkDownload.download_name}.tar.gz" -C "${
-      bulkDownload.download_name
-    }"\
+    let bulkDownloadFileName = `${bulkDownload.download_name} ID-${bulkDownload.id}`;
```
tfrcarvalho, line 46: What about adding a timestamp instead?
The id might be weird for the user, since it is an internal number that the user never sees.

ovalenzuela19: That’s a great idea! Olivia suggested the same thing earlier

ovalenzuela19: Here's an example of what the updated changes would output:
"Original Input Files - 05-26-2020 08-47-33am"

app/assets/src/components/common/DetailsSidebar/BulkDownloadDetailsMode/AdvancedDownloadTab.jsx
```diff
@@ -43,13 +44,12 @@ export default class AdvancedDownloadTab extends React.Component {
...
 
-    return `curl -L "${bulkDownload.presigned_output_url}" > "${
-      bulkDownload.download_name
-    }.tar.gz"\
-        && mkdir -p "${bulkDownload.download_name}"\
-        && tar -zvxf "${bulkDownload.download_name}.tar.gz" -C "${
-      bulkDownload.download_name
-    }"\
+    let currentTimestamp = moment().format("MM-D-YYYY hh-mm-ssa");
+    let bulkDownloadFileName = `${bulkDownload.download_name} - ${currentTimestamp}`;
```
tfrcarvalho, line 48: nit: We should actually avoid spaces... but i guess the download name itself can come with spaces?

ovalenzuela19: Sounds good! I'll make the change && yes the download name comes with spaces already

PR #3364, opened 2020-05-28
[IDSEQ-2724] Bad count badge locations in heatmap

ovalenzuela19:
# Description
The count badges for taxon categories and threshold filters were below where they should be.
To view the discussion thread, check out the JIRA ticket [here](https://jira.czi.team/browse/IDSEQ-2724)

# Notes
Decreased the space between filter dropdowns to the same as the report page
Changed "taxon categories" to "categories" as it is on the report page to match and to give it more space. (as a result of discussion thread)

# Tests
Create a heatmap with any # samples and add threshold filters and/or category filters

### DIFF
--- a/app/assets/src/components/ui/controls/dropdowns/common/DropdownTrigger.jsx
+++ b/app/assets/src/components/ui/controls/dropdowns/common/DropdownTrigger.jsx
@@ -17,6 +17,8 @@ class DropdownTrigger extends React.Component {
       placeholder,
     } = this.props;
 
+    const hasBadgeCount =
+      label === "Threshold Filters:" || label === "Categories:";
     return (
       <div
         className={cx(
@@ -29,7 +31,7 @@ class DropdownTrigger extends React.Component {
         )}
         onClick={onClick}
       >
-        <div className={cs.labelContainer}>
+        <div className={cx(cs.labelContainer, hasBadgeCount && cs.badgeCount)}>
           {label && <span className={cs.label}>{label}</span>}
           <span className={cx(value === null && cs.placeholder)}>
             {value || placeholder}
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
@@ -84,7 +84,7 @@ export default class SamplesHeatmapControls extends React.Component {
         onChange={this.onCategoryChange}
         selectedOptions={this.props.selectedOptions.categories}
         selectedSuboptions={this.props.selectedOptions.subcategories}
-        label="Taxon Categories"
+        label="Categories"
         disabled={this.props.loading || !this.props.data}
       />
     );
@@ -160,12 +160,12 @@ export default class SamplesHeatmapControls extends React.Component {
   };
 
   renderThresholdFilterSelect() {
+    const { options, selectedOptions } = this.props;
     return (
       <ThresholdFilterDropdown
-        options={this.props.options.thresholdFilters}
-        thresholds={this.props.selectedOptions.thresholdFilters}
+        options={options.thresholdFilters}
+        thresholds={selectedOptions.thresholdFilters}
         onApply={this.onThresholdFilterApply}
-        disabled={this.props.loading || !this.props.data}
       />
     );
   }
@@ -477,14 +477,13 @@ export default class SamplesHeatmapControls extends React.Component {
             <div className="col">{this.renderFilterTags()}</div>
           </div>
         )}
-        {!loading &&
-          displayFilterStats && (
-            <div className={cx(cs.filterRow, "row")}>
-              <div className={cx(cs.statsRow, "col")}>
-                {this.renderFilterStatsInfo()}
-              </div>
+        {!loading && displayFilterStats && (
+          <div className={cx(cs.filterRow, "row")}>
+            <div className={cx(cs.statsRow, "col")}>
+              {this.renderFilterStatsInfo()}
             </div>
-          )}
+          </div>
+        )}
         <Divider />
       </div>
     );
--- a/app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
+++ b/app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
@@ -22,13 +22,13 @@
 
     .filterRow {
       margin-top: $space-l;
-      margin-bottom: $space-s;
+      margin-bottom: $space-l;
     }
 
     .filterTagsList {
       margin-bottom: $space-l;
       .filterTag {
-        margin-right: 6px;
+        margin-right: $space-xs;
 
         &:last-child {
           margin-right: 0;
--- a/app/assets/src/components/ui/controls/dropdowns/common/dropdown_trigger.scss
+++ b/app/assets/src/components/ui/controls/dropdowns/common/dropdown_trigger.scss
@@ -32,6 +32,11 @@
     white-space: nowrap;
   }
 
+  .badgeCount {
+    display: flex;
+    justify-content: space-between;
+  }
+
   .label {
     font-weight: 600;
     margin-right: 5px;
--- a/app/assets/src/components/ui/controls/dropdowns/threshold_filter_dropdown.scss
+++ b/app/assets/src/components/ui/controls/dropdowns/threshold_filter_dropdown.scss
@@ -2,10 +2,6 @@
 @import "~styles/typography";
 
 :global(.ui):global(.dropdown).thresholdFilterDropdown {
-  .dropdownLabel {
-    margin-left: 6px;
-  }
-
   &:global(.active) {
     .dropdownTrigger {
       border: 1px solid $primary-light;


MarkAZhang:
Doing a quick review!

ovalenzuela19:
Update: I have to make some new changes to make the spacing between the filters similar to the sample report filters, create ellipses on threshold and category filters, and remove the badge count when it reaches the point where it can't be fully displayed

app/assets/src/components/ui/controls/dropdowns/common/DropdownTrigger.jsx
```diff
@@ -17,6 +17,8 @@ class DropdownTrigger extends React.Component {
       placeholder,
     } = this.props;
 
+    const hasBadgeCount =
+      label === "Threshold Filters:" || label === "Categories:";
```
MarkAZhang, line 21: A hard-coded conditional like this is error prone. If someone adds another dropdown that doesn't have label Threshold Filters or Categories, this code won't work for it and the bug will re-appear. The chance of that happening is quite high.

It looks like the cases we want to catch are ones where `<DropdownLabel>` is passed in as `this.props.value`. A better approach would be to check if `value` is a `DropdownLabel` component:
```
const hasDropdownLabel = get('type.name', this.props.value) === "DropdownLabel".
```
We use `get` from `lodash/fp` instead of doing `this.props.value.type.name` because if `value` is a string this second expression will fail.

I could add a comment explaining why we're taking this approach.
"When a DropdownLabel is passed in as the value, we need to add special styling so that the DropdownTrigger displays properly for narrow widths."

ovalenzuela19: This is great feedback! Thank you!! @MarkAZhang

happyimadesignr:
@ovalenzuela19 do you know if these changes affect any of the other filters in any unexpected way, say on the report page?

ovalenzuela19:
> @ovalenzuela19 do you know if these changes affect any of the other filters in any unexpected way, say on the report page?


> @ovalenzuela19 do you know if these changes affect any of the other filters in any unexpected way, say on the report page?

It did actually affect the samples report page! Thank's for bringing this up! I just added conditional hiding of the badge count

app/assets/src/components/ui/controls/dropdowns/common/DropdownTrigger.jsx
```diff
@@ -1,9 +1,53 @@
...
+  }
+
+  componentWillUnmount() {
+    if (this.resizeObserver) {
+      this.resizeObserver.disconnect();
+    }
+  }
+
+  handleFilterResize = labelContainers => {
+    const labelContainerWidth = labelContainers[0].contentRect.width - 3.7; // -3.7 removes padding
```
MarkAZhang: I would use `this.labelContainerRef.current.offsetWidth` here, just so that you're using the same method to fetch both widths. Generally better to use one approach instead of two if they're the same.

Also. it seems that the `3.7` and `1` values here are related to the padding. This seems to correspond to `margin-right: 5` for `.label` in the CSS. Maybe you can factor it out into a single variable with a comment (to make it a little cleaner):
```
// See dropdown_trigger.scss
const labelTextMarginRight = 5;
```

app/assets/src/components/ui/controls/dropdowns/common/DropdownTrigger.jsx
```diff
@@ -1,9 +1,53 @@
...
+    const labelTextWidth = this.labelRef.current.offsetWidth + 1;
+    const badgeCountWidth = 24;
```
MarkAZhang, line 37: I would add a comment `badge count width set by semantic-ui` so people understand where this value is ultimately coming from.

app/assets/src/components/ui/controls/dropdowns/common/DropdownTrigger.jsx
```diff
@@ -1,9 +1,53 @@
...
+
+    // If there's not enough space for the labelText + badgeCount hide, else show;
+    labelContainerWidth <= labelTextWidth + badgeCountWidth
```
MarkAZhang, line 40: Nit:
```
const hideDropdownLabel = labelContainerWidth <= labelTextWidth + badgeCountWidth;
this.setState({ hideDropdownLabel })
```

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -160,12 +160,12 @@ export default class SamplesHeatmapControls extends React.Component {
...
   renderThresholdFilterSelect() {
+    const { options, selectedOptions } = this.props;
     return (
       <ThresholdFilterDropdown
-        options={this.props.options.thresholdFilters}
-        thresholds={this.props.selectedOptions.thresholdFilters}
+        options={options.thresholdFilters}
+        thresholds={selectedOptions.thresholdFilters}
         onApply={this.onThresholdFilterApply}
-        disabled={this.props.loading || !this.props.data}
```
MarkAZhang, line 168: Did you mean to remove this disabled prop?

app/assets/src/components/ui/controls/dropdowns/common/DropdownTrigger.jsx
```diff
@@ -29,9 +77,27 @@ class DropdownTrigger extends React.Component {
...
-          <span className={cx(value === null && cs.placeholder)}>
+        <div
+          className={cs.labelContainer}
+          ref={hideBadgeIfInsufficientSpace && this.labelContainerRef}
+        >
+          {label && (
+            <span
+              className={cx(
+                cs.label,
+                hideBadgeIfInsufficientSpace && cs.disableMarginRight
```
MarkAZhang, line 88: Do you need disableMarginRight? It seems like the behavior is the same whether you disable it or not.

app/assets/src/components/ui/controls/dropdowns/MultipleNestedDropdown.jsx
```diff
@@ -239,6 +239,7 @@ class MultipleNestedDropdown extends React.Component {
         label={labelText}
         value={label}
         rounded={this.props.rounded}
+        hideBadgeIfInsufficientSpace={this.props.hideBadgeIfInsufficientSpace}
```
MarkAZhang, line 242: Make sure to add `hideBadgeIfInsufficientSpace` to the propTypes of all the relevant components.

app/assets/src/components/views/compare/SamplesHeatmapView/SamplesHeatmapControls.jsx
```diff
@@ -160,12 +161,13 @@ export default class SamplesHeatmapControls extends React.Component {
...
+        hideBadgeIfInsufficientSpace={true}
```
MarkAZhang, line 170: Nit: You can just do `hideBadgeIfInsufficientSpace` instead of `hideBadgeIfInsufficientSpace={true}`

app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
```diff
@@ -22,13 +22,26 @@
 
     .filterRow {
       margin-top: $space-l;
-      margin-bottom: $space-s;
+      margin-bottom: $space-l;
+      margin-right: 10px !important;
```
MarkAZhang, line 26: Not sure why you need `10px !important` here. It's not in the original styling.

app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
```diff
@@ -22,13 +22,26 @@
...
+    }
+
+    .filterElement {
+      display: inline-flex;
```
MarkAZhang, line 30: I don't think you need all of this styling. The following seems sufficient:
```
.filterRow {
  display: flex;
}

.filterElement {
  width: 16.66%;
  padding: 0 10px;
}
```

(`16.66%` looks a little better than the long decimal. Also, the previous styling seems to have `padding: 0 10px`)

app/assets/src/components/ui/controls/dropdowns/common/DropdownTrigger.jsx
ovalenzuela19: Yes I need disableMarginRight because otherwise the margin messes up `const labelTextWidth = this.labelRef.current.offsetWidth;` and I would have to add in `-3.7` again

app/assets/src/components/views/compare/SamplesHeatmapView/samples_heatmap_view.scss
ovalenzuela19: I just added "col s2" back & removed .filterElement

ovalenzuela19: I left `.filterRow` as is because I added back in `"col s2"` from Materialize.css. If I were to add in the `display.flex` and remove the margin-top and bot, it messes up the structure of filters

app/assets/src/components/ui/controls/dropdowns/common/DropdownTrigger.jsx
```diff
@@ -1,9 +1,56 @@
...
+    };
+  }
+
+  componentDidMount() {
+    const { value, hideBadgeIfInsufficientSpace } = this.props;
+
+    if (hideBadgeIfInsufficientSpace) {
+      const hasDropdownLabel = get("type.name", value) === "DropdownLabel";
+      if (
+        hasDropdownLabel &&
```
MarkAZhang, line 27: Looks like there's a bug here.

If the Dropdown starts out with no DropdownLabel (because no options are selected), but then you select an option, the resize code will not run (and you'll get the same bug as before). This is because you have a hasDropdownLabel guard around the resize observer here, and componentDidMount only runs when the component is initialized. (componentDidUpdate is the one that runs whenever the props update)

The least complex solution (but it is slightly less performant) is to take the hasDropdownLabel guard, and move it into `handleFilterResize`. So the resizeObserver always gets registered, but the resize code only runs when a DropdownLabel is passed in.

app/assets/src/components/ui/controls/dropdowns/common/DropdownTrigger.jsx
```diff
@@ -50,6 +116,13 @@ DropdownTrigger.propTypes = {
   disabled: PropTypes.bool,
   erred: PropTypes.bool,
   onClick: PropTypes.func,
+
+  /*
+  This prop is needed because we handle filter layout differently on different pages.
```
MarkAZhang, line 121: Having looked at the code a bit more, I would suggest adding "This prop should only be used when a DropdownLabel (a.k.a. the "badge") is passed in as the value." To make it more clear that this is meant for DropdownLabels, and the badge refers to the DropdownLabel.

ovalenzuela19:
> LGTM! Great work.

@MarkAZhang Thanks so much for your comments!!

PR #3365, opened 2020-05-28
Use border radius variables

tfrcarvalho:
# Description

* Define border radius variables and replace hard coded values

# Tests

* Browsed to the web to spot any visual anomalies.


tfrcarvalho:
> 1. app/assets/src/components/common/Heatmap/row_group_legend.scss

You have to sort the heatmap by genus and then hover the row labels.

> 1. &::-website-scrollbar (in "app/assets/src/components/ui/controls/dropdowns/bare_dropdown.scss")

It controls the border radius of the whole scrollbar (the background portion). You can see it in one of the dropdowns on the report page I believe. Look for class `innerMenu`.

> 1. .tooltip (in "app/assets/src/styles/_header.scss")

I cannot find usages of this class anymore.

> 1. app/assets/src/styles/_login.scss

Yes, I believe these are not used anymore. It was the old login page.

happyimadesignr:
@Janeece Heads up, I think we're just supposed to approve in the comments vs in the UI b/c approving in the UI allows it to merge I believe? @tfrcarvalho is this correct?

tfrcarvalho:
> @Janeece Heads up, I think we're just supposed to approve in the comments vs in the UI b/c approving in the UI allows it to merge I believe? @tfrcarvalho is this correct?

Technically yes, it bypasses approval by an engineer. But it is ok to approve anyway. :)

tfrcarvalho:
> @tfrcarvalho do you think we should remove the css that is no longer in use?

I can. I am trying to avoid deleting the code on the same PR since that can makes some of these changes harder. I will need to check if only that class is unused or if i should delete the whole file.
I will take a look and how obvious this case is and maybe delete it.

tfrcarvalho:
@Janeece, I was wrong. The login border radius is still in use in the forgot password page: https://idseq.net/users/password/new

Janeece:
> @Janeece, I was wrong. The login border radius is still in use in the forgot password page: https://idseq.net/users/password/new.
> These misses is why i am trying to avoid deleting right now. :)

Makes sense to me to remove things separately (and what a timely example!).  
I'm good with all of the updates, and approve. 

Sounds like usual approval procedure is to approve in a comment like this, so I'll just follow that. (but let me know if there's anything else needed from me).

PR #3366, opened 2020-05-29
Fix pipeline vix downloads for SFN runs

tfrcarvalho:
# Description

* Fix download paths for files exposed through the pipeline viz.

# Notes

Screen Shot 2020-05-28 at 10 53 31 PM

# Tests

* Open pipeline viz and verify that the links to the files are available and the user is able to download the file.


tfrcarvalho:
cc: @katrinakalantar

PR #3374, opened 2020-06-04
[IDSEQ-2769] Table headers overlap when the right side panel is opened or window size changes.

ovalenzuela19:
# Description
Table headers overlap when the right side panel is opened.

# Notes
Parent container contains the BaseTable with AutoSizer did not update width and height as expected. Setting the width & height of the parent container to 100% fixed the bug.

# Tests
Steps to Reproduce

1. Navigate to the All Data tab
2. Search for "aoeu"
3. Click the tab Samples** 
4. Shrink the window size by dragging the far right side of the browser OR expand the right side panel by clicking the blue info button. Also try expanding the window to full size and shrinking it.

### DIFF
--- a/app/assets/src/components/views/samples/samples_view.scss
+++ b/app/assets/src/components/views/samples/samples_view.scss
@@ -13,7 +13,6 @@
     flex: 1 0 auto;
     display: flex;
     flex-direction: column;
-    width: 100%;
     height: 100%;
   }
 
--- a/app/assets/src/components/views/samples/samples_view.scss
+++ b/app/assets/src/components/views/samples/samples_view.scss
@@ -13,6 +13,8 @@
     flex: 1 0 auto;
     display: flex;
     flex-direction: column;
+    width: 100%;
+    height: 100%;
   }
 
   .map {
--- a/app/assets/src/components/views/samples/samples_view.scss
+++ b/app/assets/src/components/views/samples/samples_view.scss
@@ -13,7 +13,6 @@
     flex: 1 0 auto;
     display: flex;
     flex-direction: column;
-    height: 100%;
   }
 
   .map {
--- a/app/assets/src/components/visualizations/table/BaseTable.jsx
+++ b/app/assets/src/components/visualizations/table/BaseTable.jsx
@@ -222,7 +222,10 @@ class BaseTable extends React.Component {
     const { activeColumns, columns } = this.state;
     const columnOrder = activeColumns || map("dataKey", columns);
     return (
-      <div className={cs.tableContainer}>
+      <div
+        className={cs.tableContainer}
+        style={{ flexBasis: defaultHeaderHeight }}
+      >
         <AutoSizer>
           {({ width, height }) => (
             <VirtualizedTable


app/assets/src/components/views/samples/samples_view.scss
```diff
@@ -13,6 +13,8 @@
     flex: 1 0 auto;
     display: flex;
     flex-direction: column;
+    width: 100%;
+    height: 100%;
```
tfrcarvalho, line 17: You should not need this... and can get tricky to mix flex grow with setting height this way. Why is height affecting this issue?

ovalenzuela19: The height is affecting this issue because it gets stuck once you expand the page to the entire window, just like the width does. If I just add in `width: 100%`, the bug persists

ovalenzuela19: I can't figure out why it works. I'm trying to find out why

ovalenzuela19: Found why adding the `height:100%` worked. At first the AutoSizer was setting a height of 0. After reading `usingAutoSizer.md` from reactVirtualized, I found that:

> AutoSizer expands to fill its parent but it will not stretch the parent. This is done to prevent problems with flexbox layouts. If AutoSizer is reporting a height (or width) of 0- then it's likely that the parent element (or one of its parents) has a height of 0. One easy way to test this is to add a style property (eg background-color: red;) to the parent to ensure that it is the correct size. (eg You may need to add height: 100% or flex: 1 to the parent.)

tfrcarvalho: i guess my question is why the parent was not stretching. It already has flex-grow of 1 - the first parameter of `flex: 1 0 auto`.

ovalenzuela19: I have no clue why it wasn't stretching. What's really weird is that `.table, .map` is under `.container` which contains `display: flex;
  flex-direction: column;
  height: 100%;
  flex: 1 0 auto;`

tfrcarvalho: Yeah, seems like a weird interaction with autosizer. Let me take a quick look later today, just to make sure there are no secondary effects.

ovalenzuela19: Sounds good! Thanks for your comments!

tfrcarvalho: This works a bit by chance... 
This only works because we are setting a minimal height for the `dataContainer` element outside. In the samples view, the toolbar takes over most of that space and the fact that the header of the table does not fit in the remainder of the space seem to affect how the table is rendered.
For instance if you inspect the banner you will see that the element overlaps with the header of the table.

Let's instead try changing the `div.tableContainer` in `BaseTable` to set `flex-basis` based on the 
header height like:

```
<div className={cs.tableContainer} style={{flexBasis: defaultHeaderHeight}}>
```

I think this will enforce that minimum size required for the table.
Let me know if it works

ovalenzuela19:
Ah this makes sense... That works! @tfrcarvalho thanks so much!

PR #3381, opened 2020-06-05
IDSEQ-2526 - Allow deploy script to work in idseq-prod account

davidrissato:
# Description

Change the deployment script to use the correct cluster name in idseq-prod.

# Tests

* Manual tests

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -13,7 +13,10 @@ end
 
 env = ARGV.shift
 image = ARGV.shift
-cluster = if env == "sandbox"
+
+aws_account_alias = JSON.parse(`aws iam list-account-aliases --output json --query "AccountAliases[0]"`)
+
+cluster = if env == "sandbox" or aws_account_alias == "idseq-prod"
             "idseq-#{env}-ecs"
           else
             # TODO(2020-01-23): Rename staging/prod clusters to new format (IDSEQ-2155).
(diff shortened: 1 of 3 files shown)


bin/deploy
```diff
@@ -13,7 +13,10 @@ end
...
 env = ARGV.shift
 image = ARGV.shift
-cluster = if env == "sandbox"
+
+aws_account_alias = JSON.parse(`aws iam list-account-aliases --output json --query "AccountAliases[0]"`)
+
+cluster = if env == "sandbox" or aws_account_alias == "idseq-prod"
             "idseq-#{env}-ecs"
           else
             # TODO(2020-01-23): Rename staging/prod clusters to new format (IDSEQ-2155).
```
tfrcarvalho, line 22: nit: should we update this comment, since prod is just going to be removed.

davidrissato: Maybe, but it is worth pointing out the comment still valid. `prod` still there, and `staging` is not following the new format either

davidrissato:
@tfrcarvalho  I realized there are 2 other commands that I need to update, I'll make changes to this PR in a few minutes.

bin/deploy
tfrcarvalho: What meant is that we should change it to mention staging only. Prod should simply be removed after (we should not spend time renaming it anymore)

PR #3384, opened 2020-06-08
[IDSEQ-2850] [Firefox] Scrolling creates large white gap in Heatmap between metadata and taxa rows

ovalenzuela19:
Signed-off-by: ovalenzuela19 <ovalenzuela@chanzuckerberg.com>

# Description
When using the heatmap on Firefox, trying to scroll creates a large white gap between metadata row and taxa rows.

# Notes
Firefox does not support event.wheelDeltaX or event.wheelDeltaY
We can use event.deltaY instead, as all browsers support that.
https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaX

# Tests
Open Firefox
Create a heatmap with any number of samples
Scroll in any direction

### DIFF
--- a/app/assets/src/components/visualizations/heatmap/Heatmap.js
+++ b/app/assets/src/components/visualizations/heatmap/Heatmap.js
@@ -478,8 +478,7 @@ export default class Heatmap {
     );
 
     // Set up space bar+click to pan behavior.
-    d3
-      .select("body")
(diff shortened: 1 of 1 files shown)


ovalenzuela19:
Just checked in with @happyimadesignr and this does not need review from her

PR #3386, opened 2020-06-08
[IDSEQ-1913][QA] Remove sortable style from columns that do not sort data

ovalenzuela19:
Signed-off-by: ovalenzuela19 <ovalenzuela@chanzuckerberg.com>

# Description
List headers showed inconsistent behavior (some with chevrons, text effects, or lack of both)

# Notes
Table headers "Project", "Created On", and "No. of Samples" appear as if they're clickable and did not have disableSort set to true. 
"Project" is a special case and required some added styling that .sortableHeader offers because it needs to be aligned left and not center.

# Tests
Steps to reproduce:

1. Go to projects
2. At Projects list (you are on the Projects tab, viewing a list of all Projects), mouseover column headers
3. Observe column header behavior

### DIFF
--- a/app/assets/src/components/visualizations/table/BaseTable.jsx
+++ b/app/assets/src/components/visualizations/table/BaseTable.jsx
@@ -113,10 +113,13 @@ class BaseTable extends React.Component {
             content={label}
           />
         )}
-        <SortIcon
-          sortDirection={sortDirection === "ASC" ? "ascending" : "descending"}
-          className={cx(cs.sortIcon, sortBy === dataKey && cs.active)}
-        />
+        {/* Commenting out because we don't have any sorting functionality right now.
+          TODO(omar): Uncomment when sorting functionality becomes live.
+          <SortIcon
+            sortDirection={sortDirection === "ASC" ? "ascending" : "descending"}
+            className={cx(cs.sortIcon, sortBy === dataKey && cs.active)}
+          />S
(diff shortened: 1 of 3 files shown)


app/assets/src/components/visualizations/table/BaseTable.jsx
```diff
@@ -65,9 +65,10 @@ class BaseTable extends React.Component {
     });
   }
 
-  basicHeaderRenderer({ columnData, label }) {
+  basicHeaderRenderer({ dataKey, columnData, label }) {
```
j-x-han, line 68: This actually isn't necessary to ensure the Project label is left-aligned, see the below comment

app/assets/src/components/visualizations/table/base_table.scss
```diff
@@ -40,6 +40,14 @@
   justify-content: center;
   text-align: center;
 
+  .projectHeader {
```
j-x-han, line 43: A more concise way to do this would be to edit the existing `.projectHeader` class in `projects_view.scss` and add `text-align: left`. That `.projectHeader` is already getting passed into the column's `headerClassName` in `ProjectsView.jsx` and will ensure the label is getting aligned to the left. You can see how this is done for the Samples tab! (See `.sampleHeader` in `SamplesView.jsx` and `samples_view.scss`)

